PR #3, opened 2015-06-22
Fix for TRAFODION-4

sureshsubbiah:
A prior checkin to address https://bugs.launchpad.net/trafodion/+bug/1460771 caused two tests to fail in the nightly build. The problem was twofold
a) VEG creation was disabled between old and new values in an the WHEN MATCHED clause of a MERGE statement. This caused several MERGE statements to fail.
b) Bulk load was incorrectly taking the merge code path.
Both these problem have been addressed by making the previous checkin apply only to UPSERT statements on tables with index. In addition a few other outstanding issues that could potentially cause inconsistency between table and index have been disabled. These include
a) MERGE statement on table with UNIQUE index
b) UPSERT statement on table with serialized columns.
Bulk load on tables with index will now cause the index to be rebuilt from scratch unless the table has a syskey, in which case the index will be loaded to at the same time as the table.

### DIFF
--- a/core/sql/optimizer/BindRelExpr.cpp
+++ b/core/sql/optimizer/BindRelExpr.cpp
@@ -4963,18 +4963,6 @@ RelExpr *RelRoot::bindNode(BindWA *bindWA)
       new (bindWA->wHeap()) AssignmentStHostVars(bindWA);
   }
 
-  if (isTrueRoot())
-    {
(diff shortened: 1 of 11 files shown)


sureshsubbiah:
jenkins ok

svarnau:
jenkins, add user

svarnau:
jenkins, ok to test

PR #4, opened 2015-06-23
Fix #1369853

ZhangXiuzhu:
T2 and T4 drivers return different errors if fetch/get data is issued
when cursor is after last row.

Change-Id: Ia91119b57c147e43ad5ea06fe1fd2c28ceca7c52


svarnau:
jenkins, ok to test

svarnau:
jenkins, retest please -- very strange build error

svarnau:
This fails test because of the merge conflict. The test tries to use the merged version of the pull request. Since it doesn't merge, nothing to test.  I'll try to make that more obvious for the future.

xwq:
I don't make sure who will review the changes about JDBC now . But I think -D_DEBUG option is very useful for developer. If the test passed, can anyone merged it. The changes only worked with -D_DEBUG option.

svarnau:
My previous comment was incorrect. There was no "merge" version to test, but it did have your branch version to test. The real problem was since you have not merged up to when the source tree was restructured. The test automation is expecting this code to be under the core directory.

xwq:
This pull request contain 2 commits. I think the commit 6f6017f76ce569ada36eecbc45de6f3e3633d6fb shouldn't be merged.

PR #5, opened 2015-06-23
[TRAFODION-6]JDBC T2 has compilation errors when use -D_DEBUG option

xwq:
With this check-in, JDBC T2 can use -D_DEBUG to make a debug version.
The debug version can output many debug informations and detect the
memory leak.
Change-Id: Ibf5dc2ae7e73b232bf0f3499fa3194d7f04c2e0e

### DIFF
--- a/core/conn/jdbc_type2/native/Debug.cpp
+++ b/core/conn/jdbc_type2/native/Debug.cpp
@@ -45,8 +45,8 @@
 #include <time.h>
 #include <sys/times.h>
 #include <sys/types.h>
+#include <sys/syscall.h>
 #include <unistd.h>
-#include <windows.h>
 #ifdef NSK_PLATFORM
 	#include <sqlWin.h>
 #else
@@ -56,13 +56,11 @@
 #include "SrvrCommon.h"
 #include "SrvrKds.h"
 #include "SqlInterface.h"
-#include "commondiags.h"
 //#include "spthread.h" commented by Venu for TSLX
 
 #include "jni.h"
 
 #include "Vproc.h"
-#include <tslxExt.h> //added by venu for TSLX
 
 // Maximum output line length
 #define DEBUG_MAX_LINE_LEN 1024UL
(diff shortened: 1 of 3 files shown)


svarnau:
jenkins, ok to test

xwq:
I don't know who will review the change about JDBC now. This change only affect -D_DEBUG option. And this option is very useful for developers. So can anyone review it and merge it .

core/conn/jdbc_type2/native/Debug.cpp
```diff
@@ -183,8 +184,9 @@ static const char *UCase(const char *text)
 
 static unsigned long GetThreadId(void)
 {
-	_TSLX_t thread_id = tslx_ext_pthread_self();
-	return((unsigned long) thread_id.field1);
+	//_TSLX_t thread_id = tslx_ext_pthread_self();
+	//return((unsigned long) thread_id.field1);
+        return pthread_self();
```
selvaganesang: Consider removing TSLX thread and pthread package references in T2 driver. T2 driver may not need to deal with threads at all. Also consider using syscall(SYS_gettid); to get the thread id to avoid pulling in pthread references.

You should feel free to drop all wanted cpp and .h files related to threads

xwq: I will use syscall(SYS_gettid).
 For removing TSLX thread references, I think it's better to file a new issue to do it.

selvaganesang: It should be ok to use the same JIRA to remove the unwanted files. If you prefer a different JIRA,  can you please submit a JIRA yourself using https://issues.apache.org/jira/browse/TRAFODION/
It is difficult to assign the JIRA to you if I file it.

core/conn/jdbc_type2/native/Debug.cpp
```diff
@@ -56,13 +56,11 @@
 #include "SrvrCommon.h"
 #include "SrvrKds.h"
 #include "SqlInterface.h"
-#include "commondiags.h"
 //#include "spthread.h" commented by Venu for TSLX
```
selvaganesang: Try to remove this comment too when possible

sureshsubbiah:
jenkins, retest

PR #6, opened 2015-06-23
JIRA TRAFODION-9

selvaganesang:
Non-blocking hbase operation to smoothen the data flow in trafodion engine

A new CQD HBASE_ASYNC_OPERATIONS is introduced to enable non-blocking
hbase operation for checkAndInsert operations. The default value is
'OFF'.

This allows the trafodion engine to insert rows into base table and
indexes at the same time. checkAndInsert operations always done under
a transaction ensuring the data consistency between tables and indexes
for a single row insert operation.

During this process, TM is refactored to enable passing a transaction
id created in a thread to another thread.

Change-Id: I0d09e5b0a3f988808c44bd83463df45fd21c67c0

### DIFF
--- a/core/sqf/src/tm/tmlib.cpp
+++ b/core/sqf/src/tm/tmlib.cpp
@@ -637,8 +637,8 @@ short REGISTERREGION(long transid, long startid, int pv_port, char *pa_hostname,
    TM_Transid lv_transid((TM_Native_Type) transid);
    TM_Transseq_Type lv_startid((TM_Transseq_Type) startid);
    // instantiate a gp_trans_thr object for this thread if needed.
-   TMlibTrace(("TMLIB_TRACE : REGISTERREGION ENTRY: txid: (%d,%d), startId: %ld, port: %d, hostname %s, length: %d, startcode: %ld, regionInfo: %s, length: %d.\n",
-            lv_transid.get_node(), lv_transid.get_seq_num(), startid, pv_port, pa_hostname, pv_hostname_length, pv_startcode, pa_regionInfo, pv_regionInfo_length), 2);
+   TMlibTrace(("TMLIB_TRACE : REGISTERREGION ENTRY: transid: %ld, txid: (%d,%d), startId: %ld, port: %d, hostname %s, length: %d, startcode: %ld, regionInfo: %s, length: %d.\n",
+            transid, lv_transid.get_node(), lv_transid.get_seq_num(), startid, pv_port, pa_hostname, pv_hostname_length, pv_startcode, pa_regionInfo, pv_regionInfo_length), 2);
 
    if (gp_trans_thr == NULL){
       TMlibTrace(("REGISTERREGION gp_trans_thr is null\n"), 2);
@@ -647,6 +647,21 @@ short REGISTERREGION(long transid, long startid, int pv_port, char *pa_hostname,
    }
 
    TM_Transaction *lp_currTrans = gp_trans_thr->get_current();
+   // Check if the passed-in transid is known to the thread
+   // if so, make that Transid current
+   if (lp_currTrans == NULL) {
+      lp_currTrans = gp_trans_thr->get_trans (lv_transid.get_native_type());
+      if (lp_currTrans != NULL) 
+         gp_trans_thr->set_current(lp_currTrans);
+   } 
+   // Check if the thread's current transid matches the passed-in transid
+   // If not, check if the passed-in transid is known to the thread
+   // if so, make that Transid current
+   else if (! lp_currTrans->equal(lv_transid)) {
+      lp_currTrans = gp_trans_thr->get_trans (lv_transid.get_native_type());
+      if (lp_currTrans != NULL) 
+         gp_trans_thr->set_current(lp_currTrans);
+   }
    TM_Transseq_Type lv_savedStartId = gp_trans_thr->get_startid();
 
    TMlibTrace(("REGISTERREGION lv_savedStartId is %ld.  Startid is %ld \n", (long) lv_savedStartId, startid), 2);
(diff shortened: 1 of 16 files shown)


sureshsubbiah:
jenkins, add user

core/sqf/src/tm/tmlib.cpp
```diff
@@ -647,6 +647,21 @@ short REGISTERREGION(long transid, long startid, int pv_port, char *pa_hostname,
...
+   } 
+   // Check if the thread's current transid matches the passed-in transid
+   // If not, check if the passed-in transid is known to the thread
+   // if so, make that Transid current
+   else if (! lp_currTrans->equal(lv_transid)) {
+      lp_currTrans = gp_trans_thr->get_trans (lv_transid.get_native_type());
+      if (lp_currTrans != NULL) 
+         gp_trans_thr->set_current(lp_currTrans);
+   }
    TM_Transseq_Type lv_savedStartId = gp_trans_thr->get_startid();
```
sbroeder: I think we need some code similar to this:
if (lv_savedStartId != lv_startid) {
   gp_trans_thr->set_startid(lv_startid);
}
Then we should pass lv_startid not lv_savedStartId into the call at line 677.

selvaganesang: Similar code exists few lines above and lv_savedStartId and lv_startId are one and the same at line 677. Also, can we take care of this if there are any issue in the next check-in

core/sql/executor/ExHbaseIUD.cpp
```diff
@@ -532,26 +534,64 @@ ExWorkProcRetcode ExHbaseAccessInsertSQTcb::work()
...
+            if (retcode == HBASE_RETRY_AGAIN) {
+               asyncCompleteRetryCount_++;
+               return WORK_CALL_AGAIN;
+            }
+            asyncCompleteRetryCount_ = 0;
+	    if (setupError(retcode, "ExpHbaseInterface::completeAsyncOperation")) {
+		step_ = HANDLE_ERROR;
+		break;
+            }
+            if (resultArray_[0] == FALSE) {
```
sureshsubbiah: Where is resultArray_ reset? Also can you please explain why this is an array of length 1?

core/sql/executor/ExHbaseAccess.h
```diff
@@ -1143,8 +1149,8 @@ class ExHbaseAccessSQRowsetTcb  : public ExHbaseAccessTcb
   queue_index nextRequest_;    // next request in down queue
 
   Lng32 numRetries_;
-
-  ExHbaseGetSQTaskTcb * getSQTaskTcb_;
```
sureshsubbiah: is getSQTaskTcb not used anywhere?

core/sql/executor/ExHbaseAccess.cpp
```diff
@@ -444,6 +444,9 @@ ExHbaseAccessTcb::ExHbaseAccessTcb(
   rows_.len = 0;
   colVal_.val = 0;
   colVal_.len = 0;
+  asyncCompleteRetryCount_ = 0;
+  asyncOperationTimeout_ = 2;
```
sureshsubbiah: A comment on how timeout increases will be helpful. You could tie this to Hbase client retry settings if you like that approach. Also maybe the max retry count could be configurable through cqd or hbase settings?

core/sql/executor/ExHbaseIUD.cpp
```diff
@@ -532,26 +534,64 @@ ExWorkProcRetcode ExHbaseAccessInsertSQTcb::work()
...
+		ComDiagsArea * diagsArea = NULL;
+		ExRaiseSqlError(getHeap(), &diagsArea, 
+				(ExeErrorCode)(8102));
```
sureshsubbiah: Could resultArray_[0] be FALSE for some reason other than uniqueness violation? Like say a some kind of RPC failure or HBase issue.

core/sql/executor/ExHbaseIUD.cpp
```diff
@@ -893,7 +943,48 @@ ExWorkProcRetcode ExHbaseAccessUpsertVsbbSQTcb::work()
               step_ = ALL_DONE;
 	  }
 	  break;
-
+        case ASYNC_INSERT_COMPLETE:
+          {
+            if (resultArray_  == NULL)
+                resultArray_ = new (getHeap()) NABoolean[hbaseAccessTdb().getHbaseRowsetVsbbSize()];
```
sureshsubbiah: Please ignore earlier comment about array of size 1.

core/sql/executor/ExHbaseIUD.cpp
```diff
@@ -3806,7 +3899,7 @@ ExWorkProcRetcode ExHbaseAccessSQRowsetTcb::work()
                      // another chance in case there is more input data.
                      // If parent doesn't input any more data on second (or
                      // later) chances, then process the request.
-                     if (numRetries_ == 2)    {
+                     if (numRetries_ == 3)    {
```
sureshsubbiah: Why is this being bumped up?

core/sql/executor/ExHbaseAccess.cpp
selvaganesang: Timeout increases helps to avoid cpu hogging when RS takes more time to respond. Yes. we will plan for a CQD later to provide a maximum timeout value. I would suggest avoiding coupling it with scanner timeout

core/sql/executor/ExHbaseIUD.cpp
```diff
@@ -4024,36 +4117,79 @@ ExWorkProcRetcode ExHbaseAccessSQRowsetTcb::work()
...
+            if (resultArray_  == NULL)
+                resultArray_ = new (getHeap()) NABoolean[hbaseAccessTdb().getHbaseRowsetVsbbSize()];
+            Int32 timeout;
+            if (asyncCompleteRetryCount_ < 10)
+               timeout = -1;
+            else {
+               asyncOperationTimeout_ = asyncOperationTimeout_ * 2;
+               timeout = asyncOperationTimeout_;
+            }
+            retcode = ehi_->completeAsyncOperation(timeout, resultArray_, numRowsInVsbbBuffer_);
```
sureshsubbiah: Later it might be good to refactor code from lines 4151 to 4170 (or maybe some other subset) into a separate method so that we don't have the same code block in multiple places (only if it is good design).

core/sql/executor/ExHbaseAccess.h
selvaganesang: Yes. RowsetTcb doesn't need getTaskTcb anymore

core/sql/executor/ExHbaseIUD.cpp
selvaganesang: resultArray_ is an array for two reasons. 
1) As an output parameter in java. To return the status of the Put operation
2) To enhance it to return an status array denoting the outcome of each Put operation.

This array is populated on the java side

core/sql/executor/ExHbaseIUD.cpp
selvaganesang: This should result in exception

core/sql/generator/GenRelUpdate.cpp
```diff
@@ -2747,6 +2747,10 @@ short HbaseInsert::codeGen(Generator *generator)
 
   generator->initTdbFields(hbasescan_tdb);
 
+  if (CmpCommon::getDefault(HBASE_ASYNC_OPERATIONS) == DF_ON
+           && t == ComTdbHbaseAccess::INSERT_)
```
sureshsubbiah: Does upsert pass this check also?

core/sql/executor/HTableClient.java
```diff
@@ -1112,7 +1173,7 @@ public boolean release(boolean cleanJniObject) throws IOException {
               try {
                  future.get(30, TimeUnit.SECONDS);
               } catch(TimeoutException | InterruptedException e) {
-		  logger.error("Pre-fetch Thread is Cancelled, " + e);
+		  logger.error("Asynchronos Thread is Cancelled, " + e);
```
sureshsubbiah: Typo in Asynchronous ?

core/sql/executor/ExHbaseIUD.cpp
selvaganesang: It was 3 earlier and I changed it 2 later. Now I bumped it 3, because the executor scheduler needs to cycle through the last list at least twice before this task could get a entry its down queue

core/sql/executor/ExHbaseIUD.cpp
selvaganesang: Thought about it. But decided against it because it could become convoluted with  the looping needed for timeout etc.

core/sql/executor/HTableClient.java
selvaganesang: Will fix it in my next checkin if there is no other critical issue to take care of

core/sql/generator/GenRelUpdate.cpp
selvaganesang: No. check few lines above

PR #9, opened 2015-06-25
Fix for Trafodion-5

selvaganesang:
System with large number of nodes, some created tables can not be used

The code that generates the unique ID that has been used as OBJECT_UID
is reworked to avoid getting negative numbers. Earlier, it was possible
to get negative number on the nodes at or above 64 nodes.

Please see the inline comments in CatSQLShare.cpp for the new
algorithm.

Change-Id: I7df665c16d9c156e110d9d2db3dcd82bd6dfd615


sureshsubbiah:
jenkins, retest

svarnau:
jenkins, retest please

PR #10, opened 2015-06-26
TRAFODION-13 Multiple MTD-related bug fixes, UDF and RI constraint fixes

zellerh:
Bug fixes for Multi-temperature Data aka MTD aka divisioning and a related bug for RI constraints. These will be converted to JIRAs soon.

Also added an Excel tool to draw query plans, has limited support only.

LaunchPad 1466209 Referential constraint column correspondence not recorded correctly in metadata

LaunchPad 1417739 Multi-temperature data: should not allow divisioning on float datatype.

LaunchPad 1417741 multi-temperature data: divisioning on (date_part('year', add_months(...)) fails with ERROR[3044], ERROR[1135]

LaunchPad 1417743 multi-temperature data: able to define a RI constraint on a system column.

LaunchPad 1427527 Multi-temperature data: create-table-like-with-constrains-without-division table should not be divisioned

LaunchPad 1442774 TMUDF: Compiling a TPCH query with TMUDF returns internal assertion in BaseTypes.cpp:118

Details:

LaunchPad 1466209 Referential constraint column correspondence not recorded correctly in metadata

For an RI constraint, we record a list of foreign key columns in the
KEYS metadata table, as well as the UID of the referenced uniqueness
constraint. The foreign key column list is matched with the unique key
columns. We needed to add logic to reorder the lists in case the
unique constraint was specified in a different order. For example, in
seabase/TEST020:

alter table test020t9 add constraint test020t9ri4
 foreign key (r1,r2) references test020t8 (c3,c2);

This will trigger the reorder logic to convert it to
 foreign key (r2,r1) references test020t8 (c2,c3)
to match the index on test020t8 (c2,c3).

LaunchPad 1417739 Multi-temperature data: should not allow divisioning on float datatype.

Added a check and a new error number 4257. Note that we do allow float
as key columns, and that could have similar issues that rounding
errors could cause us to find a row sometimes and not other times,
although those issues should be rare. Nevertheless, I prefer not to
take that chance and allow divisioning on floating point columns (it
does not make much sense anyway).

LaunchPad 1417741 multi-temperature data: divisioning on (date_part('year', add_months(...)) fails with ERROR[3044], ERROR[1135]

There are really two issues here. First, the CAST operator that is
internally used in ADD_MONTHS in the parser generates a nullable
result, since in the parser we don't know whether the result is
nullable or not. We need to fix that in the binder to avoid the error
1135.  Second, there were some cases where the interval leading
precision of a data type was not initialized correctly (error 3044).

LaunchPad 1417743 multi-temperature data: able to define a RI constraint on a system column.

This uncovered a bigger issue in the integration between MTD and RI
constraints. We need to ignore salt and divisioning columns in the
keys we consider, since these columns are redundant and not
user-visible. Added code to do that and while working on it, also
found the first bug in this list.

LaunchPad 1427527 Multi-temperature data: create-table-like-with-constrains-without-division table should not be divisioned

A typo in the parser code led us to ignore the WITHOUT DIVISION clause.

LaunchPad 1442774 TMUDF: Compiling a TPCH query with TMUDF returns internal assertion in BaseTypes.cpp:118

Used incorrect set of values in preCodeGen.


core/sqf/export/limited-support-tools/QPDrawer/PlanForest
```diff
@@ -0,0 +1,854 @@
...
+             "trafodion_vsbb_upser", _
+             "index_scan_unique", _
+             "insert", _
+             "insert_vsbb", _
+             "subset_delete", _
+             "subset_update", _
+             "trafodion_update", _
+             "trafodion_delete", _
+             "trafodion_insert", _
+             "trafodion_vsbb_updat", _
```
sureshsubbiah: First of all glad we got QPDrawer in.

trafodion_merge is missing? We need  a mental note to to change this file when a new leaf operator is added? i wonder why UDF, tuple, tuplelist are not here. I am assuming this is only for leaf nodes. This comment is minor and could be addressed if at all later.

selvaganesang: We also need a way to distinguish at the plan level if rowset operation is selected for update, delete and select.  It is available at the runtime operator level

zellerh:
Thanks for looking at this. Qifan made the changes for trafodion operators, I'm actually not very familiar with this code. Selva, this tool works off the EXPLAIN options 'f', so it cannot access any run-time information. Can we add the merge and UDF operators in a separate checkin, because I made a boo-boo in my local branch by doing a commit --amend, exactly what Steve said NOT to do :-( If there are other issues, we can maybe just abandon this request and I can make another one after fixing the issues.

selvaganesang:
Thanks Hans for clarification. At the run time, rowset operation is determined based on TDB attributes. So, it might be possible to massage the explain output similar to run time and then this tool can be modified. Currently, we need to decipher from the plan if it is single row get, multi-row get or scan from the full explain output. It would be good if can decipher this info easily from the explain options 'f' output and hence from this tool.

core/sqf/export/limited-support-tools/QPDrawer/PlanForest
zellerh: Actually, I don't think we need to handle UDF, tuple, tuplelist, because there is no table name associated with them, but if trafodion_merge is a leaf then we should probably add it. If missing, I think it will still display the operator, just not the table name it is operating on.

sureshsubbiah:
Yes Hans. Thank you for this change set. I will merge it in soon.

Thanks
Suresh

PR #15, opened 2015-06-30
Changes related to JIRA [TRAFODION-14]

selvaganesang:
upsert or merge into a table with indexes can result in
inconsistency between index and table

The change is to pass the recent row timestamp to the Delete
operator. However, this may not solve the above issue because
tombstone marker timestamp can be same or later than the
recently added row even with this change.

This change is pushed for the following reasons:
a) To enable Trafodion async operation concept for the merge
   operation
b) To optimize delete operation by removing unwanted memory
   allocation.


sureshsubbiah:
jenkins, test

PR #17, opened 2015-07-02
OSIM REWORK JIRA TRAFODION-16

HowardQin:


HowardQin:
How should I reply this question?

core/sql/optimizer/NAClusterInfo.cpp
```diff
@@ -296,13 +299,13 @@ NAClusterInfo::NAClusterInfo(CollHeap * heap)
...
       clusterToCPUMap_ = new(heap) NAHashDictionary<CollIndex,maps>
-          (&clusterNumHashFunc,17,TRUE, heap);
-      tableToClusterMap_ = new(heap) NAHashDictionary<CollIndex, maps> 
-          (&tableIdentHashFunc, 13, TRUE, heap);
-
+                                                          (&clusterNumHashFunc,17,TRUE, heap);
+                                                          
+      nodeIdToNodeNameMap_ = new(heap) NAHashDictionary<Int32, NAString>
+                                                          (&intHashFunc, 101,TRUE,heap);
+
```
zellerh: This is ok, but I'm wondering whether an array wouldn't be easier, since this translates integer node ids into node names?

core/sql/optimizer/OptimizerSimulator.cpp
```diff
@@ -278,9 +277,10 @@ OptimizerSimulator::OptimizerSimulator(CollHeap *heap)
     sysCallLogFilePath_[sc]=NULL;
     writeSysCallStream_[sc]=NULL;
   }
+  // Init log4cxx logging
+  CommonLogger::instance().initLog4cxx("log4cxx.trafodion.osim.config");
```
zellerh: Isn't this running in the Trafodion process, master executor or compiler? If so, won't that interfere with the logger already established for the process? I'm wondering whether we need a separate logger for OSIM or whether we could just reuse the existing one?

core/sql/export/NAStringDef.cpp
```diff
@@ -347,6 +347,143 @@ NAString::resize(size_t N)
...
+
+//This static function calculates and allocates buffer for the printf-like formatted string,
+//and the formatted string is copied to buffer.
+char* NAString::buildBuffer(const char* formatTemplate, va_list args)
+{
+  // calculate needed bytes to allocate memory from system heap.
+  int bufferSize = 20049;
+  int msgSize = 0;
+  //buffer is managed by this static function
+  static __thread char *buffer = NULL;
```
zellerh: Should use the THREAD_P macro defined in Platform.h here instead of __thread.

zellerh:
Thank you, looks good to me.

sureshsubbiah:
jenkins, add user

core/sql/optimizer/OptimizerSimulator.cpp
HowardQin: Yes, OSIM is running in compiler.
The initialization process is not necessary, but we do need a separate logger to print OSIM message to console.

core/sql/export/NAStringDef.cpp
HowardQin: reworked see last commit.

core/sql/optimizer/OptimizerSimulator.cpp
zellerh: Hi Howard, thanks for the explanation. Can I ask two more questions? First, if we do a CONTROL OSIM command, then we will go through this code. That probably means that all subsequent log messages (after the OSIM command completes) for this process will appear on the screen and won't be logged to files anymore? Also, if people do the OSIM command from trafci or from a JDBC client, they would not see the console log?

Assuming I understand this correctly, here is one possible solution: Don't use a logger. Instead, use printf (or similar) to log to a file, along with the other information. Then issue a warning in the OSIM command that looks like this: **** WARNING 1234: Log output has been written to /somedir/osim_log_20150707_1234.log. For debugging OSIM code, you can just use the regular logger and examine the log files. Also, you can use an #ifdef OSIM_DEBUG or similar if you want to print debugging info to the screen.

Again, I hope I understood the problem well enough that what I say above makes sense.

HowardQin: Hi Hans, I know your concern, in this latest commit, I will only use fprintf for OSIM logging, actually, fprintf was prefered in the first place, and messages weren't shown in trafci client, OSIM logger is not going to impact logging of other modules.

nonstop-qfchen:
Hi Howard,

Thank you. Will review the new set. 

I have made a change in HQC area and would like you to review it. Can you send an email to my esgyn account?   My review message to your HP email bounced back.

core/sql/optimizer/OptimizerSimulator.cpp
zellerh: Great, thanks for removing it.

svarnau:
Jenkins message did not posted.  Should have been:

Refer to this link for build results (access rights to CI server needed): 
https://jenkins02.trafodion.org/job/Check-PR/37/
Test FAILURE

nonstop-qfchen:
The code in set "rework on logger, NAString, DDL" looks good!

PR #19, opened 2015-07-06
Fixes for 1438896 and 1465356

robertamarton:
1438896: Internal error during create or replace view

The objectUID check when getting privilege information was not correct.

1465356: Revoke privilege from role returns dependent object error

There is a check in mainline revoke code to determine if the object type is a
view and if the SELECT privilege is no longer applicable. If so, then the
dependent error is returned. However, this code is incorrect and actually the
correct code exists in the gatherViewPrivileges method. The view check has been
removed.

Change-Id: Ia1a00f18bad301dd1439c2ceb079d69808de7bdd

### DIFF
--- a/sql/sqlcomp/PrivMgrPrivileges.cpp
+++ b/sql/sqlcomp/PrivMgrPrivileges.cpp
@@ -431,6 +431,22 @@ PrivMgrPrivileges::PrivMgrPrivileges (
 }
 
 // ----------------------------------------------------------------------------
+// Construct a PrivMgrPrivileges object for on objectUID
+// ----------------------------------------------------------------------------
+PrivMgrPrivileges::PrivMgrPrivileges (
+   const int64_t objectUID,
+   const std::string &metadataLocation,
+   ComDiagsArea *pDiags)
+: PrivMgr(metadataLocation, pDiags),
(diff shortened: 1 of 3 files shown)


robertamarton:
Changes to fix a privilege manager issue with create or replace view and a change to remove a semantic check that is no longer needed and incorrect.

sureshsubbiah:
jenkins, add user

DaveBirdsall:
Code changes look good to me. One question though: Do we still need the old constructor that doesn't take an object ID?

sql/sqlcomp/PrivMgrPrivileges.cpp
```diff
@@ -431,6 +431,22 @@ PrivMgrPrivileges::PrivMgrPrivileges (
 }
 
 // ----------------------------------------------------------------------------
+// Construct a PrivMgrPrivileges object for on objectUID
```
arvind-narain: Looks good - Comment could be updated "for an objectUID"

robertamarton:
This code was committed from an old repository that is no longer around.  There were many missing files that indicates issues with the original commit.  Since the original repository is no longer available, decided to close out this pull request and start over.

PR #31, opened 2015-07-18
TRAFODION-38 First working TMUDFs in Java

zellerh:


svarnau:
Sorry, still debugging why jenkins module is not posting the link correctly.
https://jenkins02.trafodion.org/job/Check-PR/53/

core/Makefile
```diff
@@ -135,3 +135,28 @@ find-absolute-dlls:
 
 check-copyrights:
 	python ./updateCopyrightCheck.py
+
+eclipse: rest/.project sql/.project sqf/src/seatrans/hbase-trx/.project .project
+	@echo
+	@echo
+	@echo "  Now you can import 5 projects in Eclipse:"
```
DaveBirdsall: The echo says 5 projects, but I only see make file steps for 4?

core/sql/executor/JavaObjectInterface.cpp
```diff
@@ -147,19 +147,29 @@ int JavaObjectInterface::createJVM()
...
+       debugTimeout_ = atoi(debugTimeoutStr);
+     const char *suspendOnDebug = getenv("JVM_SUSPEND_ON_DEBUG");
+     char debugOptions[300];
+
+     debugPort_ = debugPort;
+     // to allow debugging multiple processes at the same time,
+     // specify a port that is a multiple of 1000 and the code will
+     // add pid mod 1000 to the port number to use
+     if (debugPort_ % 1000 == 0)
+       debugPort_ += (GetCliGlobals()->myPin() % 1000);
```
DaveBirdsall: What's the probability of port collisions? How does one identify them? I suppose a way to work around port collisions is to stop and restart the affected processes?

core/Makefile
zellerh: The HBase-trx make step actually creates two projects, hbase-trx and hbase-protocol.

core/sql/executor/JavaObjectInterface.cpp
zellerh: Yes, this is not perfect. If the user makes sure they have 1000 unused ports then we can guarantee that there are no port conflicts. One error I got from the debugger was "connection refused", which I think might have been caused by a port conflict.

core/sql/langman/LmJavaExceptionReporter.cpp
```diff
@@ -792,3 +792,89 @@ LmJavaExceptionReporter::reportInternalSPJException(LmHandle jt,
...
+        }
+      else
+        {
+          // Report the return status as an internal error, since
+          // we didn't get a UDRException. This should be rare, since
+          // Java exceptions are caught above.
+          char buf1[4];
+          snprintf(buf1, sizeof(buf1), "%d", callPhase);
+          char buf2[80];
+          snprintf(buf2, sizeof(buf2), "Return status %d from JNI call", returnStatus);
```
DaveBirdsall: I've noticed that in the last year or so we are putting more message text in our code (to avoid multiplying error messages in SqlciErrors.txt probably). If there is ever a desire to translate our error messages into another language, this will prove to be an obstacle.

core/sql/langman/LmLangManagerJava.cpp
```diff
@@ -694,6 +712,52 @@ void LmLanguageManagerJava::initialize(LmResult &result,
...
+         "I");
+
+    if (jni->ExceptionOccurred() ||
+        udrQueueStateField_ == NULL)
+      {
+        exceptionReporter_->insertDiags(diagsArea_,
+                                        -LME_JVM_SYS_CLASS_ERROR,
+                                        "org/trafodion/sql/udr/UDR$QueueStateInfo");
+        result = LM_ERR;
+        return;
```
DaveBirdsall: Just curious: Why no deletelocalref in this code path?

core/sql/regress/udr/EXPECTED001
```diff
@@ -195,6 +244,33 @@ CREATE TABLE_MAPPING FUNCTION TRAFODION.SCH.SESSIONIZE_DYNAMIC
 
 --- SQL operation complete.
 >>
+>>get table_mapping functions for library TEST001_Java ;
```
DaveBirdsall: "table mapping" without the underscore would seem to be more natural.

core/sql/src/main/java/org/trafodion/sql/udr/TupleInfo.java
```diff
@@ -1309,100 +1277,27 @@ public void setString(int colNum, String val) throws UDRException {
...
-                    row_.putInt(t.getVcLenIndOffset(),stringLen);
-                else
-                    row_.putShort(t.getVcLenIndOffset(),(short)stringLen);
-            }
-            break ;
+            // fall through to next case
+
         case CHAR:
         case VARCHAR:
+        case BLOB:
```
DaveBirdsall: A question: Is it appropriate to blank-pad a BLOB? (It might be binary data, for example.)

DaveBirdsall:
Code changes look good to me. Nice work on a very useful feature. It would be good if someone else could review these changes also.

core/sql/langman/LmJavaExceptionReporter.cpp
zellerh: Yes, sorry, there is some English text here, but I think this particular error message goes through significant trouble to allow translation into other languages. The message text is this: Internal error in communicating with UDR $0~String0 through the object interface in the $1~String1 phase ($2~String2). Details: $3~String3.

core/sql/langman/LmLangManagerJava.cpp
zellerh: Thanks, I'll fix that.

core/sql/src/main/java/org/trafodion/sql/udr/TupleInfo.java
zellerh: Would we blank-pad a BLOB? As far as I can tell, the putEncodedStringIntoBuffer() method only blank-pads fixed-length fields, and a BLOB should be variable length. We probably should add a setBytes() or setBinary() method for BLOBs. BLOB and CLOB support for TMUDFs is not fully designed and supported yet, I've been talking to Sandhya about the best ways to do this.

sureshsubbiah:
This a beautiful change. I see no issues. Thank you fixing some of the bugs I have introduced in *.java files from a previous delivery. Please let us know if we should pull this change in now or wait for some rework agreed to above (in response to Dave's comments). I particularly like the changes to udr/TEST001 that show how the new UDF can be used.

core/sql/regress/udr/EXPECTED001
sureshsubbiah: I found that difficult to do in our parser without increasing conflicts. This is not to say that we should not try again.

zellerh:
Thank you to both of you, Dave and Suresh, for reviewing this big change! Sorry for the delay, I didn't know how to interpret the Jenkins output that indicated a release build failure. I just committed  the fix to the bug Dave found and the one found by Jenkins.

PR #32, opened 2015-07-21
Fix for [TRAFODION-14]

selvaganesang:
upsert or merge into a table with indexes can result
in inconsistency between index and table

This change avoids the HBase issue of rows getting lost when the
 delete and put operations are done in the same milliseconds
timestamp dimension for the given rowid.

In the trafodion_delete operation execution, the row is skipped
for deletion if the new and old rowid evaluates to the same value.

This needs change from the compiler sub-task in the JIRA to complete
for it to be effective.

In addition removed yet another memory allocation for the delete
operation.

### DIFF
--- a/core/sql/comexe/ComTdbHbaseAccess.cpp
+++ b/core/sql/comexe/ComTdbHbaseAccess.cpp
@@ -127,6 +127,7 @@ ComTdbHbaseAccess::ComTdbHbaseAccess(
   returnMergeInsertExpr_(returnMergeInsertExpr),
   encodedKeyExpr_(encodedKeyExpr),
   keyColValExpr_(keyColValExpr),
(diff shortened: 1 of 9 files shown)


zellerh:
Thanks, looks good to me. Now the ball is in my court with the corresponding compiler fix, TRAFODION-40 (https://issues.apache.org/jira/browse/TRAFODION-40).

sureshsubbiah:
Looks good to me too. Thanks for the change.

svarnau:
Thanks, Suresh for being on top of merging PR's that are ready.
In future, I think committers can provide a better commit message than the default one that git merge provides. I'll update instructions to help at https://cwiki.apache.org/confluence/display/TRAFODION/Committer+Workflow

PR #34, opened 2015-07-29
TRAFODION-40 Make index-maintenance delete conditional

zellerh:
This fixes the issue where we update an indexed column to itself, and
the index maintenance tree both deletes and re-inserts the index row
within the same millisecond. This makes the index row disappear, as the
delete takes precedence over the insert, even though it happened earlier.

As a side-effect, index maintenance may get faster, because we are
also replacing the CheckAndPut of the index insert with a Put.
We could further improve this by processing multiple Puts together.


DaveBirdsall:
Code changes look good to me.

sureshsubbiah:
Change looks good to me too. Merging it now.

PR #37, opened 2015-07-31
[TRAFODION-1423] Indexes on trafodion should be created in aligned ro…

selvaganesang:
…w format by default

A new CQD TRAF_INDEX_ALIGNED_ROW_FORMAT is set to ON by default.
The indexes are created in the aligned row format when this
CQD is set to ON and the table has no serialized column.


selvaganesang:
Please don’t merge this change to master yet.  I need to push in some more change removing the commented out lines and resolve conflicts. Also, there is one failure in qat test, that I am working on. Most likely I will deliver a known diff file for now.

selvaganesang:
All changes are here. Can one of the committers request for full test run with this commit please?

selvaganesang:
I have already started the full test, so committers don't bother to start the full test.

core/sql/generator/GenRelScan.cpp
```diff
@@ -2203,10 +2203,46 @@ short HbaseAccess::codeGen(Generator * generator)
...
+         tablename = space->AllocateAndCopyToAlignedSpace(GenGetQualifiedName(getIndexDesc()->getNAFileSet()->getFileSetName()), 0);
+         if (getIndexDesc()->isClusteringIndex())
+         {
+            //base table
+            snapshotName = (char*)getTableDesc()->getNATable()->getSnapshotName() ;
+           if (snapshotName == NULL)
+             latestSnpSupport = latest_snp_no_snapshot_available;
+          }
+          else
+            latestSnpSupport = latest_snp_index_table;
```
sureshsubbiah: Do we support the case where the base table has no snapshot defined on it but an index on the table does have a snapshot?

selvaganesang: This part of the code is moved here from few lines below in the same function just to get the tablename. The snapshot functionality is not affected by this change. Yes. We need to figure out  the consequences if the snapshot for the index is not created, when table has snapshot and vice versa. But, it is outside the scope of this change

DaveBirdsall: It's outside of this change, but I'm curious: How in general do we guarantee consistency of a table snapshot with index snapshots?

core/sql/optimizer/NATable.cpp
```diff
@@ -7596,6 +7617,19 @@ NABoolean  NATable::getHbaseTableInfo(Int32& hbtIndexLevels, Int32& hbtBlockSize
...
+{
+  if (qualifiedName_.getText().compareTo(objectName, NAString::exact) == 0)
+     return getSQLMXAlignedTable();
+  else {
+     for (CollIndex i = 0; i < indexes_.entries(); i++) {
+         if (indexes_[i]->getFileSetName().getQualifiedNameAsAnsiString().compareTo(objectName, NAString::exact) == 0)
+            return indexes_[i]->isSqlmxAlignedRowFormat();
+     }  
+  }
+  return getSQLMXAlignedTable();
```
DaveBirdsall: Should we ever get here? (If we do, is this an internal error?)

DaveBirdsall:
Code changes look OK to me.

core/sql/optimizer/NAFileSet.h
```diff
@@ -72,6 +72,13 @@ enum FileOrganizationEnum
   KEY_SEQUENCED_FILE,
   HASH_FILE
 };
+
+enum RowFormatEnum {
+  SQLMX_UNKNOWN_FORMAT,
+  SQLMX_ROW_FORMAT,
+  SQLMX_ALIGNED_ROW_FORMAT,
+  SQLMX_HBASE_FORMAT
```
zellerh: SQLMX_ALIGNED_ROW_FORMAT is already used in NATable.h for a flag. Should we avoid this conflict (renaming either one). Also, I'm wondering whether this should go into ComSmallDefs, since it is fairly fundamental.

core/sql/optimizer/NATable.h
```diff
@@ -576,7 +576,17 @@ class NATable : public NABasicObject
   }
 
   NABoolean isSQLMXAlignedTable() const
-  {  return (flags_ & SQLMX_ALIGNED_ROW_FORMAT) != 0; }
+  {
```
zellerh: Now that we have file formats per index, we should move this method from NATable to NAFileSet, for example as NAFileSet::isSQLMXAlignedFormat(). Instead of calling this method on an NATable *t, the caller should then call t->getClusteringIndex()->isSQLMXAlignedFormat(). I would also have some comments on the implementation below, but they are irrelevant if you replace this method.

core/sql/optimizer/NATable.h
```diff
@@ -808,6 +824,9 @@ class NATable : public NABasicObject
   NABoolean getRegionsNodeName(Int32 partns, ARRAY(const char *)& nodeNames) const;
 
 private:
+  NABoolean getSQLMXAlignedTable() const
```
zellerh: Maybe remove this method and the flag and put it into NAFileSet?

core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
```diff
@@ -8358,6 +8358,19 @@ desc_struct * CmpSeabaseDDL::getSeabaseUserTableDesc(const NAString &catName,
       Lng32 keyColCount = *(Lng32*)vi->get(6);
       Lng32 nonKeyColCount = *(Lng32*)vi->get(7);
       Lng32 idxNumSaltPartns = *(Lng32*)vi->get(8);
+      char * format = vi->get(9);
+      Lng32 idxRowFormat;
+
+      if (memcmp(format, COM_ALIGNED_FORMAT_LIT, 2) == 0)
```
zellerh: Not sure I understand this: This gets the row format from the TABLES table, so the index row format is the same for all the indexes, is that correct? Wouldn't we need to do that per index? For example, older tables could have an aligned base table and unaligned indexes, while newer tables and indexes could both be aligned?

zellerh:
See some line-based comments. I can feel your pain, as I was dealing with a similar issue where things were organized on a per-table basis that I needed on a per-index basis...

DaveBirdsall:
Selva, I see you've submitted a new pull request for this JIRA? Does that mean you're abandoning this one?

selvaganesang:
Will be abandoning this PR because the changes are moved to PR #42

PR #38, opened 2015-07-31
[TRAFODION-28] Part 3 of changes to update license text

DaveBirdsall:
This check-in updates the license text in the remaining
subdirectories of the core directory. Still to do are a
few files in the core directory itself (I forgot these),
and the other non-core diretories in incubator-trafodion.


sureshsubbiah:
Thanks! The change has been merged

PR #43, opened 2015-08-05
[TRAFODION-1438] Fixing Win ODBC crashes with long cert file name length(over 30 bytes)

ryzuo:
Change the driver to determine the length of buffer for certificate
file name at runtime, and fill the name with secured strcpy version(strcpy_s)


svarnau:
jenkins, add user

arvind-narain:
Thanks for the change. Could you please check on the git settings so that the eols are maintained for windows ?

hegdean:
Thanks for the change.  We will run into similar issue when using Linux driver. Can you  please make similar change for linux driver as well.

ryzuo:
For there's the pwd. file under core/sqf/src/tm/tools/ which Windows does not recognize, it makes it's always in the changed list and not able to remove from cache, thus I can't commit from Windows. So I did this via Linux.
Although these are Windows codes, LF is default configuration of git, and conventionally people uses LF when submitting codes into git even for Windows platform code for compliance purposes, and Visual Studio supports LF anyway. I'm thinking we will format all those to LF in the future.

PR #42, opened 2015-08-05
[TRAFODION-1423] Index improvements- Indexes in aligned Row format

selvaganesang:

PR #44, opened 2015-08-05
[Trafodion-1424] Enable CIF for Trafodion scan operator

sureshsubbiah:
When a row is read is read by the scan operator from native HBase or Trafodion tables (either aligned or regular format) a convert expression is applied to convert the row into a form usable by rest of query tree. Output of convert expression is now in aligned format, with varchars not being blank padded. Change can be disabled with cqd COMPRESSED_INTERNAL_FORMAT 'OFF'. Majority of change is due to Khaled Bouaziz.


core/sql/generator/GenRelScan.cpp
```diff
@@ -2183,6 +2182,17 @@ short HbaseAccess::codeGen(Generator * generator)
   if (getTableDesc()->getNATable()->hasAddedColumn())
     hasAddedColumns = TRUE;
 
+  NABoolean alignedFormatTable = 
+    getTableDesc()->getNATable()->isSQLMXAlignedTable();
```
DaveBirdsall: Will this line conflict with Selva's change? (See https://github.com/apache/incubator-trafodion/pull/42). I'm wondering about ths isSQLMXAlignedTable() call in particular.

DaveBirdsall:
Change looks good to me though I'm wondering if it will conflict with Selva's change (https://github.com/apache/incubator-trafodion/pull/42).

DaveBirdsall:
Yes, just merged pull request #42 and I see that this one now has a conflict to be resolved.

core/sql/executor/ExHbaseAccess.cpp
```diff
@@ -379,7 +379,8 @@ ExHbaseAccessTcb::ExHbaseAccessTcb(
       latestVersionNumForCols_ = new(glob->getDefaultHeap()) 
 	long[hbaseAccessTdb.workCriDesc_->getTupleDescriptor(hbaseAccessTdb.asciiTuppIndex_)->numAttrs()] ;
     }
-
+  
+  convertRowLen_ = hbaseAccessTdb.convertRowLen_;
```
selvaganesang: Use the getter method instead

core/sql/executor/ExHbaseAccess.cpp
```diff
@@ -1645,12 +1650,18 @@ Lng32 ExHbaseAccessTcb::createSQRowFromAlignedFormat(Int64 *latestRowTimestamp)
...
-	convertExpr()->eval(pentry_down->getAtp(), workAtp_, NULL, asciiRowLen);
+	convertExpr()->eval(pentry_down->getAtp(), workAtp_, NULL, 
+			    asciiRowLen, &rowLen);
       if (evalRetCode == ex_expr::EXPR_ERROR)
 	{
 	  return -1;
 	}
+      if (hbaseAccessTdb().getUseCif() &&
+      rowLen < convertRowLen_)
+        convertRowLen_=rowLen;
```
selvaganesang: Will the rowLen be less than convertRowLen from the compiler. I believe convertRowLen_ might have been ujsed to pre-allocate row buffer. There could be buffer overflow. Can you please confirm that is not the case.

sureshsubbiah: Yes rowLen can be hbaseAccessTdb().convertRowLen(). Row buffer has been pre-allocated with hbaseAccessTdb().convertRowLen(). Now that we have set convertRowLen_ to be actual size of output row for this particular row, I think this ensure that we copy only the required bytes for this output row. Please see change ExHbaseSelect.cpp

sureshsubbiah: Merge conflict and comment to to accesor method has been addressed in second commit. Thanks for the review.

PR #48, opened 2015-08-07
[TRAFODION-1419] Support for multiple column families in traf table

anoopsharma00:
details in jira


DaveBirdsall:
Code changes look good to me.

sandhyasun:
Looks fine to me too. Will merge this after giving folks a few more hours to take  look and comment.

sureshsubbiah:
Looks good to me too. Should be a wonderful addition to Trafodion. One minor comment is that max allowed col family value of 32 may be too high given what we know about HBase behaviour with more than say 10 column families.

PR #49, opened 2015-08-07
[TRAFODION-1426] null values in RI constraint

anoopsharma00:
null values in columns of referencing table incorrectly cause referential constraint to fail


anoopsharma00:
details and testcase in jira. Also part of regress/seabase/test020.

DaveBirdsall:
Changes look good to me. Will commit these if no-one beats me to it.

PR #52, opened 2015-08-11
[TRAFODION-1444] Reduced pathlength for IUD

selvaganesang:
Reduced the number of JNI to java transitions and
memory allocation by streamlining the Trafodion-HBase
client layer for IUD commands

Also removed unreferenced code and variables in this layer.

### DIFF
--- a/core/sql/executor/HBaseClient.java
+++ b/core/sql/executor/HBaseClient.java
@@ -77,6 +77,7 @@ import org.apache.hadoop.hbase.HRegionInfo;
 import org.apache.hadoop.hbase.ClusterStatus;
 import org.apache.hadoop.hbase.ServerName;
 
+import java.util.concurrent.ExecutionException;
 import java.util.Set;
 import java.util.TreeSet;
 
@@ -1419,6 +1420,80 @@ public class HBaseClient {
                         throws IOException {
       HTableClient htc = getHTableClient(jniObject, tblName, useTRex);
       return htc.startGet(transID, rowIDs, columns, timestamp);
+  }
+
+  public boolean insertRow(long jniObject, String tblName, boolean useTRex, long transID, byte[] rowID,
+                         Object row,
+                         long timestamp,
+                         boolean checkAndPut,
+                         boolean asyncOperation) throws IOException, InterruptedException, ExecutionException {
+
+      HTableClient htc = getHTableClient(jniObject, tblName, useTRex);
+      boolean ret = htc.putRow(transID, rowID, row, null, null,
+                                checkAndPut, asyncOperation);
+      if (asyncOperation == false)
+         releaseHTableClient(htc);
+      return ret;
+  }
+
+  public boolean checkAndUpdateRow(long jniObject, String tblName, boolean useTRex, long transID, byte[] rowID,
+                         Object columnsToUpdate,
+                         byte[] columnToCheck, byte[] columnValToCheck,
+                         long timestamp,
+                         boolean asyncOperation) throws IOException, InterruptedException, ExecutionException {
+      boolean checkAndPut = true;
+      HTableClient htc = getHTableClient(jniObject, tblName, useTRex);
+      boolean ret = htc.putRow(transID, rowID, columnsToUpdate, columnToCheck, columnValToCheck,
+                                checkAndPut, asyncOperation);
+      if (asyncOperation == false)
+         releaseHTableClient(htc);
+      return ret;
+  }
+
+  public boolean insertRows(long jniObject, String tblName, boolean useTRex, long transID, 
+			 short rowIDLen,
+                         Object rowIDs,
+                         Object rows,
+                         long timestamp,
+                         boolean autoFlush,
+                         boolean asyncOperation) throws IOException, InterruptedException, ExecutionException {
+      HTableClient htc = getHTableClient(jniObject, tblName, useTRex);
+      boolean ret = htc.putRows(transID, rowIDLen, rowIDs, rows, timestamp, autoFlush, asyncOperation);
+      if (asyncOperation == false)
+         releaseHTableClient(htc);
+      return ret;
+  }
+
+  public boolean deleteRow(long jniObject, String tblName, boolean useTRex, long transID, 
+                                 byte[] rowID,
+                                 Object[] columns,
+                                 long timestamp, boolean asyncOperation) throws IOException {
+      HTableClient htc = getHTableClient(jniObject, tblName, useTRex);
+      boolean ret = htc.deleteRow(transID, rowID, columns, timestamp);
+      if (asyncOperation == false)
+         releaseHTableClient(htc);
+      return ret;
+  }
+
+  public boolean deleteRows(long jniObject, String tblName, boolean useTRex, long transID, short rowIDLen, Object rowIDs,
+                      long timestamp, 
+                      boolean asyncOperation) throws IOException, InterruptedException, ExecutionException {
+      HTableClient htc = getHTableClient(jniObject, tblName, useTRex);
+      boolean ret = htc.deleteRows(transID, rowIDLen, rowIDs, timestamp);
+      if (asyncOperation == false)
+         releaseHTableClient(htc);
+      return ret;
+  }
+
+  public boolean checkAndDeleteRow(long jniObject, String tblName, boolean useTRex, long transID, 
+                                 byte[] rowID,
+                                 byte[] columnToCheck, byte[] colValToCheck,
+                                 long timestamp, boolean asyncOperation) throws IOException {
+      HTableClient htc = getHTableClient(jniObject, tblName, useTRex);
+      boolean ret = htc.checkAndDeleteRow(transID, rowID, columnToCheck, colValToCheck, timestamp);
+      if (asyncOperation == false)
+         releaseHTableClient(htc);
+      return ret;
   }
 
   public boolean  createCounterTable(String tabName,  String famName) throws IOException, MasterNotRunningException
--- a/core/sql/executor/HBaseClient.java
+++ b/core/sql/executor/HBaseClient.java
@@ -1431,7 +1431,9 @@ public class HBaseClient {
       HTableClient htc = getHTableClient(jniObject, tblName, useTRex);
       boolean ret = htc.putRow(transID, rowID, row, null, null,
                                 checkAndPut, asyncOperation);
-      if (asyncOperation == false)
+      if (asyncOperation == true)
+         htc.setJavaObject(jniObject);
+      else
          releaseHTableClient(htc);
       return ret;
   }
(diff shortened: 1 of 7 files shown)


selvaganesang:
jenkins, retest

selvaganesang:
Here is the explanation for this change. Earlier, IUD operations involve the following 3 steps
1. HBaseClient_JNI::getHTableClient
2  HTableClient_JNI::insertRow , insertRows, deleteRow, deleteRows ...
3. HBaseClient_JNI::releaseHTableClient

There are 3 JNI to java transistions and a creation of HTableClient_JNI object. Now with this change there is only one step

HBaseClient_JNI::insertRow, insertRows, deleteRow ...

This method takes in all the parameters needed to accomplish the task and makes only one transition from JNI to java. Also, there is no need to create HTableClient_JNI object if the asynchronous operation is not attempted.

core/sql/executor/HBaseClient.java
```diff
@@ -1421,6 +1422,80 @@ public int startGet(long jniObject, String tblName, boolean useTRex, long transI
...
+      if (asyncOperation == false)
+         releaseHTableClient(htc);
+      return ret;
+  }
+
+  public boolean deleteRow(long jniObject, String tblName, boolean useTRex, long transID, 
+                                 byte[] rowID,
+                                 Object[] columns,
+                                 long timestamp, boolean asyncOperation) throws IOException {
+      HTableClient htc = getHTableClient(jniObject, tblName, useTRex);
```
sureshsubbiah: How does error handling work now? Suppose tdblname refers to a non-existent table, will we get an exception that will get carried all the way back?

selvaganesang: Yes. It should return IOException. And the IOException is translated into an error in the JNI layer. It is no different from calling these methods separately from JNI.

core/sql/executor/HBaseClient_JNI.cpp
```diff
@@ -2844,819 +2871,980 @@ HBC_RetCode HBaseClient_JNI::getHbaseTableInfo(const char* tblName,
...
-//////////////////////////////////////////////////////////////////////////////
-HTableClient_JNI::~HTableClient_JNI()
-{
-  //QRLogger::log(CAT_JNI_TOP, LL_DEBUG, "HTableClient destructor called.");
-  cleanupResultInfo();
-  if (tableName_ != NULL)
-  {
-     NADELETEBASIC(tableName_, heap_);
+  if (asyncOperation) {
+     HTableClient_JNI *htc = new (heap) HTableClient_JNI(heap, (jobject)-1);
```
sureshsubbiah: Should this just be htc = ... ?
In line 2883 we have already declared the htc variable.

core/sql/executor/HBaseClient_JNI.cpp
```diff
@@ -2844,819 +2871,980 @@ HBC_RetCode HBaseClient_JNI::getHbaseTableInfo(const char* tblName,
...
+  }
+  if (hbs)
+    hbs->incBytesRead(rowID.len + row.len);
+  if (jresult == false) {
+     jenv_->PopLocalFrame(NULL);
+    if (htc != NULL)
+        NADELETE(htc, HTableClient_JNI, heap);
+     return HBC_ERROR_INSERTROW_DUP_ROWID;
+  }
+  *outHtc = htc;
```
sureshsubbiah: Just so that I understand, htc can be NULL at this point (for sync operations)?

core/sql/executor/HBaseClient_JNI.cpp
```diff
@@ -2844,819 +2871,980 @@ HBC_RetCode HBaseClient_JNI::getHbaseTableInfo(const char* tblName,
...
+      ExHbaseAccessStats *hbs, bool useTRex, Int64 transID, short rowIDLen, HbaseStr rowIDs,
+      HbaseStr rows, Int64 timestamp, bool autoFlush, bool asyncOperation,
+      HTableClient_JNI **outHtc)
 {
-  QRLogger::log(CAT_SQL_HBASE, LL_DEBUG, "HTableClient_JNI::startScan() called.");
+  
+  HTableClient_JNI *htc = NULL;
 
+  if (asyncOperation) {
+     HTableClient_JNI *htc = new (heap) HTableClient_JNI(heap, (jobject)-1);
```
sureshsubbiah: same comment as before regarding declaration. I am assuming other methods will have this also. Please check.

core/sql/exp/ExpHbaseInterface.cpp
```diff
@@ -1315,32 +1192,30 @@ Lng32 ExpHbaseInterface_JNI::checkAndInsertRow(
...
+  NABoolean checkAndPut = TRUE;
+
   if (noXn)
-    transID = 0;
-  else
+    transID = 0; 
+  else 
     transID = getTransactionIDFromContext();
- 
+  retCode_ = client_->insertRow((NAHeap *)heap_, tblName.val, hbs_,
```
zellerh: Since the method is called checkAndInsertRow, should we not call checkAndInsertRow here as well? Sorry, haven't looked yet at the surround what really happens, but this could either be wrong or confusing - unless I'm wrong or confused :-)

core/sql/executor/HBaseClient_JNI.cpp
```diff
@@ -2844,819 +2871,980 @@ HBC_RetCode HBaseClient_JNI::getHbaseTableInfo(const char* tblName,
...
+  jenv_->SetByteArrayRegion(jba_columnValToCheck, 0, columnValToCheck.len, (const jbyte*)columnValToCheck.val);
+  jlong j_htc = (long)htc;
+  jboolean j_useTRex = useTRex;
+  jlong j_tid = transID;  
+  jlong j_ts = timestamp;
+  jboolean j_asyncOperation = asyncOperation;
+ 
+  if (hbs)
+    hbs->getTimer().start();
+  tsRecentJMFromJNI = JavaMethods_[JM_HBC_DIRECT_CHECKANDUPDATE_ROW].jm_full_name;
```
sureshsubbiah: What is line 3138 for?

core/sql/executor/HBaseClient_JNI.cpp
```diff
@@ -2844,819 +2871,980 @@ HBC_RetCode HBaseClient_JNI::getHbaseTableInfo(const char* tblName,
...
+     htc->setHbaseStats(hbs);
+  }
   if (jenv_->PushLocalFrame(jniHandleCapacity_) != 0) {
     getExceptionDetails();
-    return HTC_ERROR_GETROWSOPEN_EXCEPTION;
+    return HBC_ERROR_DELETEROW_EXCEPTION;
+  }
+  jstring js_tblName = jenv_->NewStringUTF(tableName);
+  if (js_tblName == NULL) {
+    GetCliGlobals()->setJniErrorStr(getErrorText(HBC_ERROR_INSERTROWS_PARAM));
```
sureshsubbiah: Is the exception correct here?

core/sql/executor/HBaseClient_JNI.cpp
```diff
@@ -2844,819 +2871,980 @@ HBC_RetCode HBaseClient_JNI::getHbaseTableInfo(const char* tblName,
...
   }
-  jbyteArray jba_rowID = jenv_->NewByteArray(rowID.len);
-  if (jba_rowID == NULL) 
-  {
-     GetCliGlobals()->setJniErrorStr(getErrorText(HTC_ERROR_DELETEROW_PARAM));
-     jenv_->PopLocalFrame(NULL);
-     return HTC_ERROR_DELETEROW_PARAM;
+  jstring js_tblName = jenv_->NewStringUTF(tableName);
+  if (js_tblName == NULL) {
+    GetCliGlobals()->setJniErrorStr(getErrorText(HBC_ERROR_INSERTROWS_PARAM));
```
sureshsubbiah: Exception seems incorrect.

zellerh:
Looks good to me, I'm glad Suresh went into more detail than I did.

core/sql/executor/HBaseClient_JNI.cpp
selvaganesang: Yes. Thanks for catching it. Let me unit test asyncOperation

core/sql/executor/HBaseClient_JNI.cpp
selvaganesang: Yes. htc is null for sync operations. This asyncHtc_ in ExpHbaseInterface_JNI will be set to null. Hence, there is no need for release the asyncHtc_

core/sql/executor/HBaseClient_JNI.cpp
selvaganesang: Yes. You are correct. I have fixed it all places

core/sql/executor/HBaseClient_JNI.cpp
selvaganesang: This is useful when there is a crash in JVM. You will be able to figure out which java method was called from JNI at the time of crash from this thread specific variable.

core/sql/executor/HBaseClient_JNI.cpp
selvaganesang: Thanks for finding it. Fixed it.

core/sql/executor/HBaseClient_JNI.cpp
selvaganesang: Thanks. Fixed it

core/sql/exp/ExpHbaseInterface.cpp
selvaganesang: checkAndInsertRow and insertRow is implemented using the same method HBaseClient_JNI::insertRow. There is a parameter checkAndPut is used to determine if checkAndPut is needed

core/sql/executor/HBaseClient_JNI.h
```diff
@@ -576,6 +550,30 @@ class HBaseClient_JNI : public JavaObjectInterface
   HBC_RetCode incrCounter( const char * tabName, const char * rowId, const char * famName,
```
sureshsubbiah: Could incrCounter method benefit from the new change? Or maybe it is already getting is own htc and deleting it at the end?

core/sql/exp/ExpHbaseInterface.cpp
```diff
@@ -797,178 +734,149 @@ Lng32 ExpHbaseInterface_JNI::getRowsOpen(
...
           NABoolean asyncOperation)
 {
-  HTableClient_JNI *htc = client_->getHTableClient((NAHeap *)heap_, tblName.val, useTRex_, hbs_);
-  if (htc == NULL) {
-    retCode_ = HBC_ERROR_GET_HTC_EXCEPTION;
-    return HBASE_OPEN_ERROR;
-  }
-  
-  Int64 transID;
+  HTableClient_JNI *htc;
```
sureshsubbiah: Should we initialize to NULL here? If comment is valid please check similar places in other methods too.

core/sql/executor/HBaseClient_JNI.h
selvaganesang: incrCounter uses the regular hbase table. It doesn't use trafodion table. But, I think this code is already optimized for JNI to java transitions.

core/sql/exp/ExpHbaseInterface.cpp
selvaganesang: Not needed. htc used as a output parameter and the insertRow method below initializes it null or htc for asnyc operation when it is success.

sureshsubbiah:
Looks good. Thanks. Merging it.

selvaganesang:
Also fixed the failure with compGeneral/TEST042

PR #54, opened 2015-08-12
[TRAFODION-1447]Windows ODBC Driver not able to process "UPSERT" statement

ryzuo:
Add "UPSERT" statement type supports.


arvind-narain:
Thanks for the quick change. Solves the issue seen if client uses row-wise binding with UPSERT. It's difficult to see the actual change though since  > 4k lines got touched.

+        else if ((strcmp(token, "INSERT") == 0) || (strcmp(token, "INS") == 0) ||
+                (strcmp(token, "UPSERT") == 0))  // Add for "UPSERT" support
+            m_StmtType = TYPE_INSERT;

sureshsubbiah:
This change has been tested now for both windows and linux. Merging it. Thanks.

PR #55, opened 2015-08-13
[TRAFODION-1448] Leak in HTableClient objects

selvaganesang:
Refactored the code to ensure that HTableClient object in java
is released for every batch of gets.

In the process removed unused code in HBaseClient_JNI.cpp and
ExpHbaseInterface.cpp


sureshsubbiah:
Looks good. merging.

PR #56, opened 2015-08-14
TRAFODION-1455. Renamed pwd. to pwd.cpp

venkat1m:
This JIRA addresses that issue where pwd. file gets deleted on windows


sureshsubbiah:
looks good. merging.

PR #57, opened 2015-08-14
[TRAFODION - 1456] Fix for rowset upsert into table with index assert. 

sureshsubbiah:
Change is in the binder, on how characteristic inputs are assigned.


sandhyasun:
Looks good to me. Will wait for a couple of  hours and then merge this one.

PR #58, opened 2015-08-14
[[Trafodion 1452]] Cloudera parcel support and EPEL fix

amandamoran:
Trafodion can now be installed on cloudera parcels. 

traf_setup was trying to use pdsh and pdcp for the epel repo 
when it had not been installed.


DaveBirdsall:
Looks OK to me but I don't fully understand it.

amandamoran:
Cloudera parcels installs in a different directory path than Cloudera packages. 
Example with HBase: /opt/cloudera/parcels/lib/hbase vs. /usr/lib/hbase

We needed to modify traf_cloudera_mods98 since it needs to copy the hbase-trx*.jar file to the correct <path>/hbase/lib directory. We also needed to modify sqenvcom.sh to set the correct classpaths with the correct locations.

PR #59, opened 2015-08-14
[TRAFODION-1454] Unable to detect the query id via RMS while running …

selvaganesang:
…the tools like ODB

Active query concept is now introduced to list queries which either have finished executing within the given number of seconds or still being executed. The negative number in LAST_ACTIVITY_SECS means that the query has completed that many seconds ago.

cd $MY_SQROOT/export/limited-support-tools/LSO
./offender -s active

CURRENT_TIMESTAMP           LAST_ACTIVITY_SECS    QUERY_ID                                                        EXEC_STATE                SOURCE_TEXT
--------------------------  --------------------  -------------------------------------------------------------   ------------------------  ---------------------------------------------------------------------------
2015-08-14 01:25:46.040274                   -18  MXID11000008455212306275447582585000000000206U3333300_175_S1    CLOSE                     "upsert into t12 values (1,1,1) ;

Also fixed a bug where TDBID_DETAIL stats(instance level stats) is not being displayed for HBase access and HDFS scan operators when the query is not yet finished.


DaveBirdsall:
Code changes look OK to me. Will commit later today if no objections appear.

core/sqf/export/limited-support-tools/LSO/README
```diff
@@ -152,40 +153,52 @@ URRENT_TIMESTAMP           NODE         PID          EXE_MEM_HIGH_WM_MB  EXE_MEM
...
+013-02-22 16:29:30.430629                  1772  MXID11000030466212228308797313024000000000106U6553500_6_SQL_ENV_Q6            FETCH           "select cast(COUNT(*) as int) from HP_SYSTEM_CATALOG.MXCS_SCHEMA.ENVIRONMENTVALUES where env_id = cast(? as integer) FOR READ
+ 
+
+ACTIVE QUERIES
+================
+
+This statement lists queries which either have finished executing within the
+given number of seconds or still being executed. The negative number in 
+LAST_ACTIVITY_SECS means that the query has completed that many seconds ago. 
+
```
sureshsubbiah: what will LAST_ACTIVITY_SECS be for a query that is still executing. NULL ?

selvaganesang: If the query is still executing, LAST_ACTIVITY_SECS will be the number of seconds(a positive number) the query has been running till then. The queries are listed in the descending order of LAST_ACTIVITY_SECS to help the customer to focus on these queries first

core/sql/executor/ExStats.cpp
```diff
@@ -10666,6 +10688,27 @@ NABoolean ExMasterStats::filterForCpuStats(short subReqType,
...
+         if (exeStartTime_ != -1) {
+            if (exeEndTime_ != -1)
+               tsToCompare = exeEndTime_;
+            else
+               tsToCompare = exeStartTime_;
+            lastActivity_ = (Int32)((currTimestamp-tsToCompare) / (Int64)1000000);
+            if (exeEndTime_ == -1)
+               return TRUE;
+            else
+            if (lastActivity_ <= etTimeInSecs) {
```
sureshsubbiah: So active queries with a default of 30 means all currently active queries (regardless of how long they have been active for) + queries that completed in last 30 seconds. Please ignore my previous comment.

sureshsubbiah:
Looks good to me Selva. Did we test with ODB to check if it helped? I can merge it now since Dave has reviewed already.

core/sql/executor/ExStats.cpp
selvaganesang: Correct. Yes. 30 seconds applies to the query that have finished.

selvaganesang:
I tested it using Trafci to confirm the behavior when ODBC Server is involved. I am expecting the ODB to behave in the same way

PR #60, opened 2015-08-17
TRAFODION-1457 Issues related to decoding HBase region keys and TEXT …

zellerh:
Fixed a few issues relating to decoding binary encoded key values of UTF8 and interval data type.
Also moved the logic of double-quoting strings into the methods that update the TEXT table.
Renamed some regression files that used obsolete suffixes like .LINUX.


core/sql/optimizer/BindItemExpr.cpp
```diff
@@ -10298,23 +10298,23 @@ NABoolean ZZZBinderFunction::isPadWithSpace (ExprValueId& padExpr, CharInfo::Cha
...
+       ((ConstValue *)padExpr->castToItemExpr())->getConstStr(FALSE));
      Int32 i = 0;
      NABoolean foundSingleQuote = FALSE;
-     for (const NAWchar *s = padString.data(); *s; s++)
+     for (const char *s = padString.data(); *s; s++)
      {
-       i++;  // 0x0027 is the single quote character '
-       if ((!foundSingleQuote)&&(*s != 0x0027)) // loop through 
+       i++;  // 0x27 is the single quote character '
+       if ((!foundSingleQuote)&&(*s != 0x27)) // loop through
```
DaveBirdsall: Slightly more portable to code *s != '\''  (it would work in EBCDIC, for example, while 0x27 would not).

core/sql/optimizer/BindItemExpr.cpp
```diff
@@ -10298,23 +10298,23 @@ NABoolean ZZZBinderFunction::isPadWithSpace (ExprValueId& padExpr, CharInfo::Cha
...
        else if ((!foundSingleQuote))
        {
         foundSingleQuote = TRUE; // found the leading single quote.
         continue;
        }
-       if ((i == (padString.length())) && (*s == 0x0027)) // trailing single quote
+       if ((i == (padString.length())) && (*s == 0x27)) // trailing single quote
          continue;
-       if (*s != 0x0020) // 0x0020 is the single space character ' '
+       if (*s != 0x20) // 0x20 is the single space character ' '
```
DaveBirdsall: Again, slightly more portable to code *s != ' '

core/sql/optimizer/NATable.cpp
```diff
@@ -1472,58 +1478,161 @@ static ItemExpr * getRangePartitionBoundaryValuesFromEncodedKeys(
               // value. Note that the buffer has a prefix of some bytes with actual key
               // values, followed by bytes that are zeroed out. 
 
+              // the number of bytes actually provided in the key (not filled in)
+              int numBytesInProvidedVal = encodedKeyLen-valOffset;
+
+              if (nullHdrSize && numBytesInProvidedVal <= 0)
```
DaveBirdsall: The condition numBytesInProvidedVal <= 0 should always be true by definition of columnIsPartiallyProvided. Doesn't hurt anything though, just makes reading more complicated.

core/sql/optimizer/BindItemExpr.cpp
zellerh: Done.

core/sql/optimizer/BindItemExpr.cpp
zellerh: Done.

core/sql/optimizer/NATable.cpp
zellerh: Couldn't this happen, for example?

currOffset = 10
encodedKeyLen = 11
nullHdrSize = 2
valOffset = 12

In other words, the partially provided key covers only the NULL indicator or part of it.

core/sql/sqlcomp/CmpSeabaseDDLcommon.cpp
```diff
@@ -6137,53 +6138,82 @@ short CmpSeabaseDDL::validateDivisionByExprForDDL(ItemExpr *divExpr)
...
+  ElemDDLPartitionList * pPartitionList = NULL;
+  ElemDDLPartitionRange *pPartitionRange = NULL;
+
+      numSplits = numSaltSplits;
+
+      // for salt splits, only the first key column value
+      // is variable, the rest use the default values
+      for (int k=1; k<numKeys; k++)
+        splitValuesAsText[k] = defaultSplits[k];
+
```
DaveBirdsall: Indentation is inconsistent (if you happen to fix something else perhaps you could fix this too)

DaveBirdsall:
I like the refactoring done in CharType.cpp. Seems like much cleaner code. Similarly the handling of doubled quotes in the TEXT table is cleaner now (this bedeviled me when I did the ALTER TABLE HBASE_OPTIONS stuff).

PR #61, opened 2015-08-19
[TRAFODION-1460] Enable EXPLAIN IN RMS feature by default

selvaganesang:
EXPLAIN_IN_RMS is now enabled by default. The explain plan
can now be obtained from a different session using

explain for qid <qid>
explain options 'f' for qid <qid>

The explain fragment in packed format is retained in RMS shared
segment till the query is deallocated.

### DIFF
--- a/core/sql/cli/SessionDefaults.cpp
+++ b/core/sql/cli/SessionDefaults.cpp
@@ -139,7 +139,6 @@ SessionDefaults::SessionDefaults(CollHeap * heap)
        cancelLogging_(TRUE),
        suspendLogging_(TRUE),
        aqrWarn_(0),
-       explainInRMS_(FALSE),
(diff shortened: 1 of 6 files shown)


selvaganesang:
jenkins, retest

selvaganesang:
I measured the amount of memory allocated in RMS shared segment using GET STATISTICS FOR RMS ALL command before and after this change.
Before:
Node 0
Stats Heap High WM 2,551,176
Node 1
Stats Heap High WM 1,824,432
After EXPLAIN_IN_RMS is changed to 'ON'
Node 0
Stats Heap High WM 2,811,504
Node 1
Stats Heap High WM 1,839,480

sureshsubbiah:
Looks great. Thanks for explaining.

PR #62, opened 2015-08-19
Fixes for Jira's trafodion-19 & trafodion-29

robertamarton:
See Jira's for a description.
Also removed obsolete code from the parser


core/sql/common/ComMisc.cpp
```diff
@@ -99,3 +99,17 @@ NABoolean ComIsTrafodionReservedSchemaName(
   return FALSE;
 }
 
+// schema names of pattern "_HV ... _" and "_HB_ ... _" are reserved to store
```
DaveBirdsall: Nit: Comment is not quite right... it's _HV_ and _HB_ (the trailing underscore after _HV is missing).

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -1571,6 +1569,21 @@ NATable *BindWA::getNATable(CorrName& corrName,
       ((QualifiedName&)(table->getTableName())).setIsVolatile(TRUE);
     }
       
+  // For now, do not allow access through the Trafodion external name created for
+  // the HIVE object unless the inDDL flag is set.  inDDL is set for drop 
+  // table and SHOWDDL statements.  
+  // TDB - may want to merge the Trafodion version with the HIVE version.
+  // TDB - similar operation may be needed for external HBase tables
+  if ((table) && (table->isExternalTable() && (! bindWA->inDDL())))
```
DaveBirdsall: Does this give correct behavior, for example, if an external name is referenced in the SELECT part of a "CREATE TABLE LIKE ... SELECT ..." statement?

core/sql/optimizer/ObjectNames.h
```diff
@@ -692,6 +689,13 @@ class CorrName : public NABasicObject
     defaultMatchCount_(-1),
     flagbits_(0)
   {
+    setIsExternal(ComIsTrafodionExternalSchemaName(schemaName));
+    //Int32 len (schemaName.length());
```
DaveBirdsall: Do you want to delete the commented code?

core/sql/optimizer/RelExeUtil.cpp
```diff
@@ -3733,6 +3734,23 @@ RelExpr * DDLExpr::bindNode(BindWA *bindWA)
...
       }
 
+      // Hive tables can only be specified as external and must be created
+      // with the FOR clause
+      if (qualObjName_.isHive()) 
+      {
+        if (createTableNode->isExternal())
+        {
+          externalTable = TRUE;
+          isHbase_ = TRUE;
```
DaveBirdsall: Just checking: We set isHbase_ to TRUE for Hive tables? Is there an isHive_ member in the class?

DaveBirdsall:
The changes look good to me. The only comments I had were fairly minor.

core/sql/bin/SqlciErrors.txt
```diff
@@ -179,8 +179,8 @@
...
 1178 ZZZZZ 99999 ADVANCED MAJOR DIALOUT Internal error: unable to find catalog associated with uid $0~string0.  This catalog is being accessed by role $1~string1.
 1179 ZZZZZ 99999 ADVANCED MAJOR DBADMIN Internal error: inconsistent object definition for $0~TableName found for object in name space $0~string0. Object owned by $1~string1. Unable to find object details in $2~string2;
-1180 ZZZZZ 99999 BEGINNER MAJOR DBADMIN The required subvolume name for locations in schema $0~SchemaName is $1~String0.
-1181 ZZZZZ 99999 BEGINNER MAJOR DBADMIN Label $0~String0 could not be dropped (file error $1~NSKCode). 
+1180 ZZZZZ 99999 BEGINNER MAJOR DBADMIN Trying to create an external table $0~TableName with a different schema or table name than the source table $1~TableName.  The external schema and table name must be the same as the source. 
+1181 ZZZZZ 99999 BEGINNER MAJOR DBADMIN Trying to create a schema with name $0~TableName to store the definition of a native HIVE or HBASE table and the name is too long.  Maximum length supported is $0~Int0.
```
nonstop-qfchen: Can we supply the external source as an argument (HIVE or HBASE) here?

core/sql/common/ComMisc.cpp
```diff
@@ -99,3 +99,17 @@ NABoolean ComIsTrafodionReservedSchemaName(
...
+// external hive and hbase tables
+NABoolean ComIsTrafodionExternalSchemaName (
+                                    const NAString &schName)
+{
+  Int32 len (schName.length());
+  Int32 prefixLen = sizeof(HIVE_EXT_SCHEMA_PREFIX);
```
nonstop-qfchen: Should we use another string as HBASE_EXT_SCHEMA_PREFIX?

core/sql/optimizer/BindRelExpr.cpp
nonstop-qfchen: Update stats utility also needs to access the external table.  Can we change the inDDL() method name to allowExternalHBaseTable() and allowExternalHiveTable()?

core/sql/sqlcomp/CmpSeabaseDDLcommon.cpp
```diff
@@ -3422,7 +3440,7 @@ Int64 CmpSeabaseDDL::getObjectUIDandOwners(
     strcpy(cfvd, " and valid_def = 'Y' ");
 
   char buf[4000];
```
nonstop-qfchen: May consider moving this array to heap, because of the size of the array. Stack overflow prevention.

core/sql/sqlcomp/CmpSeabaseDDLmd.h
```diff
@@ -238,6 +238,16 @@ static const QString seabaseObjectsUniqIdxDDL[] =
   {" ; "}
 };
 
+enum SeabaseObjectsFlags {
+  SEABASE_OBJECT_IS_EXTERNAL_HIVE  = 0x0000000000000001, // set if this object
```
nonstop-qfchen: May consider dropping the prefix SEABASE_.

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -348,20 +348,32 @@ void CmpSeabaseDDL::createSeabaseLibrary(
       processReturn();
       return;
     }
+
+  ComTdbVirtTableTableInfo * tableInfo = new(STMTHEAP) ComTdbVirtTableTableInfo[1];
```
nonstop-qfchen: these assignments should be part of the arguments to a constructor to ComTdbVirtTableTableInfo.

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -1009,18 +1022,30 @@ void CmpSeabaseDDL::createSeabaseRoutine(
       return;
     }
 
+  ComTdbVirtTableTableInfo * tableInfo = new(STMTHEAP) ComTdbVirtTableTableInfo[1];
```
nonstop-qfchen: Same comment as above. Should be part of a new constructor taking arguments.

core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
```diff
@@ -378,6 +378,244 @@ void CmpSeabaseDDL::createSeabaseTableLike(
...
+  // table's schema name as the default schema
+  ComObjectName tgtTableName(createTableNode->getTableName(), COM_TABLE_NAME);
+  ComAnsiNamePart currCatAnsiName(getSystemCatalog());
+  ComAnsiNamePart currSchAnsiName(srcTableName.getSchemaNamePart().getExternalName());
+  tgtTableName.applyDefaults(currCatName, currSchAnsiName);
+
+  // For now the object and schema name for the target table must match the
+  // object and schema name for the source table
+  if (tgtTableName.getObjectNamePart().getExternalName() != 
+      srcTableName.getObjectNamePart().getExternalName())
```
nonstop-qfchen: In HIVE (since 0.13), a column name can be quoted with backtick characters (not the same as double quoted SQL identifiers), see http://docs.hortonworks.com/HDPDocuments/HDP2/HDP-2.1.5/bk_dataintegration/content/ch_using-hive-quotedidentifiers.html. 

Just wonder if we are going to support that. 

For the table names, we are going to only allow those with alphanumeric plus underscore characters, right?

core/sql/common/ComMisc.cpp
robertamarton: Don't see this???

core/sql/optimizer/ObjectNames.h
robertamarton: Yes, meant to do it but forgot

core/sql/optimizer/RelExeUtil.cpp
robertamarton: This is kind of tricky.  there is an isHive method in qualified name but I also needed to set isHbase_ in order for executeSeabaseDDL to work.  It this field is not set, then executeSeabaseDDL returns an unsupported error.

core/sql/bin/SqlciErrors.txt
robertamarton: Can do, good suggestion.

core/sql/common/ComMisc.cpp
robertamarton: the prefixes are the same but should probably fix this in case they change or we add something different.

core/sql/sqlcomp/CmpSeabaseDDLcommon.cpp
robertamarton: All the SQL requests generated by the DDL component use this method to create strings on the fly.  It is a good suggestion but we may want to consider some generic method and change all the string creations to use this method.  Do we really have this much of a concern for stack overflow?

core/sql/sqlcomp/CmpSeabaseDDLmd.h
robertamarton: I generated this enum based off a similar FLAGS field for the COLUMNS table.  
num SeabaseColumnsFlags {
  SEABASE_COLUMN_IS_SALT       = 0x0000000000000001, // don't rely on this quite yet,                                                   
  SEABASE_COLUMN_IS_DIVISION   = 0x0000000000000002

I think we should be consistent in the naming.

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
robertamarton: I just copied this from similar code in the DDL component.  If we decide to change this, then we should change all occurances.  Something to consider.

core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
robertamarton: In the JIRA I have a list of tasks left to be done.  One of them was testing with the latest version of HIVE which supports special characters. HBase, as you know, has a much more limited set of special characters.  So this will require some work and we may need to support a trafodion name different than the hive name.  Supporting a different name is also on the list of things left to do.

core/sql/common/ComMisc.cpp
robertamarton: The ... was supposed to signify the schema name, perhaps "_HV_schema_" and "_HB_schema_" would look better.  The ... to indicate the schema name may be confusing.

core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
```diff
@@ -1181,6 +1421,30 @@ short CmpSeabaseDDL::createSeabaseTable2(
       return -1;
     }
 
+  // If creating an external table, go perform operation
+  if (createTableNode->isExternal())
+    {
+      retcode = createSeabaseTableExternalHive(cliInterface, createTableNode, currCatName, currSchName);
```
nonstop-qfchen: if this function returns -2, should we perform a cleanup operation of the meta-data tables here?

core/sql/optimizer/BindRelExpr.cpp
robertamarton: inDDL was added for compound create schema commands a while back and does not seem to be used today.  I added it as a stop gate since it was available.  It does make sense to add it as separate fields but I didn't do it because I was not sure if we were going to keep this rule going forward.  We may want to allow all accesses through the external name. I will add this to my list of things to do.

robertamarton: Create table like <external table> does create a regular table correctly - a table with the same signature as the external table and the new table is not external.  However, we do allow it to be created in the _HV_ schema.   So a check is needed in create code to not allow objects created in schema beginning with "_HV_" and "_HB_". 

create external table like external table does not work because the source and target tables have to be the same names.  The error returned is a unique constraint violation so a better error would be nice.   If we ever allow the external name to be different than the hive table, then we should allow this option.

I also notice an incorrect error returned when creating an external table with the "select as" clause.  This should be fixed.

create external table promotion as select * from hive.hive.promotion;
*** ERROR[4258] Trying to access external table TRAFODION."_HV_HIVE_".PROMOTION through its external name format. Please use the native table name.

core/sql/common/ComMisc.cpp
robertamarton: sorry, I did not catch the "underscore" comment - I will fix this in a subsequent delivery.

core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
robertamarton: There is definitively an issue here that was cut and pasted from create table like further down. 

First off, the createSeabaseExternalHive should only be returning 0 or -1.  0 is successful and -1 is failure.  A -2 is only needed if the associated HBase table needs to be removed (hence a cleanup).  Since there is no external table, a -2 should not be returned.  

This particular code code should return the retcode instead of -1:

  if (createTableNode->isExternal())
    {
      retcode = createSeabaseTableExternalHive(cliInterface, createTableNode, currCatName, currSchName);
      . . .
      return retcode;
    }

PR #63, opened 2015-08-20
TRAFODION-1459 SHOWDDL shows incorrect DIVISION BY + regressions fix

zellerh:
Fixing 1459 and also checking in a new expected file that should solve
the compGeneral/TEST042 failures we have seen over the past two days,
caused by my earlier commit. The expected file + known diffs worked on
a debug build, but it is not sufficient for a release build.


DaveBirdsall:
Looks good to me.

PR #66, opened 2015-08-27
fix #JIRA 1449: missing end key.

nonstop-qfchen:
Please pull the request.  Thanks!

### DIFF
--- a/core/sql/generator/GenPreCode.cpp
+++ b/core/sql/generator/GenPreCode.cpp
@@ -154,15 +154,19 @@ static NABoolean processConstHBaseKeys(Generator * generator,
   // convert built-in search key to entries with constants, if possible
   if (skey->areAllKeysConstants(TRUE))
     {
-      ValueIdSet exePreds;
       ValueIdSet nonKeyColumnSet;
       idesc->getNonKeyColumnSet(nonKeyColumnSet);
 
-      // seed exePreds with all predicates, key and non-key
-      // they will be reduced by the method below
-      skey->getKeyPredicates(exePreds);
+      // seed keyPreds with only the full key predicate from skey
+      ValueIdSet keyPreds = skey->getFullKeyPredicates();
+
+      // include executorPreds and selection predicates 
+      // but exclude the full key predicates.
+      ValueIdSet exePreds;
+
       exePreds += executorPreds;
       exePreds += relExpr->getSelectionPred();
+      exePreds.subtractSet(keyPreds);
 
       HbaseSearchKey::makeHBaseSearchKeys(
            skey,
@@ -170,11 +174,15 @@ static NABoolean processConstHBaseKeys(Generator * generator,
            skey->getIndexDesc()->getOrderOfKeyValues(),
            relExpr->getGroupAttr()->getCharacteristicInputs(),
            TRUE, /* forward scan */
-           exePreds,
+           keyPreds,
            nonKeyColumnSet,
            idesc,
            relExpr->getGroupAttr()->getCharacteristicOutputs(),
            mySearchKeys);
+
+      // Include any remaining key predicates that have not been 
+      // picked up (to be used as the HBase search keys).
+      exePreds += keyPreds;
 
       TableDesc *tdesc = NULL;
       if (mySearchKeys.entries()>0)
--- a/core/sql/optimizer/SearchKey.cpp
+++ b/core/sql/optimizer/SearchKey.cpp
@@ -201,6 +201,11 @@ void SearchKey::init(const ValueIdList & keyColumns,
 					      );
 
 
+  // Find any full key predicate in setOfPredicates that refers 
+  // to a key column and save the result in fullKeyPredicates_.
+  computeFullKeyPredicates(setOfPredicates);
+  
+
   // We do not want to remove the predicates which have two constants
   // in their VEG. These should be evaluated like any other predicate.
 
@@ -2708,6 +2713,65 @@ void SearchKey::computeCoveredLeadingKeys()
        if ( !found )
          break;
     }
+  }
+}
+
+void SearchKey::computeFullKeyPredicates(ValueIdSet& predicates)
+{
+  // If the key column is an INDEX COLUMN, convert it into a base column
+  ValueIdList baseKeyCols;
+
+  ValueId vid;
+  const ValueIdList& keyColumns = getKeyColumns();
+  for (CollIndex i=0; i<keyColumns.entries(); i++ ) {
+     if ( keyColumns[i].getItemExpr()->getOperatorType() == ITM_INDEXCOLUMN )
+       vid = ((IndexColumn*)(keyColumns[i].getItemExpr()))->getDefinition();
+     else
+       vid = keyColumns[i];
+
+     baseKeyCols.insertAt(baseKeyCols.entries(), vid);
+  }
+
+  // compute the # of leading key columns covered by the key columns 
+  // in the disjunct.
+   ValueIdSet columnsCoveredByKey;
+
+   vid = predicates.init();
+   for (; predicates.next(vid); predicates.advance(vid)) {
+
+      columnsCoveredByKey.clear();
+      vid.getItemExpr()->findAll(ITM_BASECOLUMN, columnsCoveredByKey, TRUE, TRUE);
+
+      // for each vid (or predicate), check the base column list
+      for (CollIndex i=0; i<baseKeyCols.entries(); i++ ) {
+
+         NABoolean found = FALSE;
+
+         // the predicate directly refers the base column 
+         if ( columnsCoveredByKey.containsTheGivenValue(baseKeyCols[i]) ) {
+            fullKeyPredicates_ += vid;
+            found = TRUE;
+         } else {
+            // do a search by column position
+           BaseColumn* keyCol = (BaseColumn*)(baseKeyCols[i].getItemExpr());
+
+
+           ValueId cvid = columnsCoveredByKey.init();
+           for (; columnsCoveredByKey.next(cvid); columnsCoveredByKey.advance(cvid)) {
+              BaseColumn* colInKeyPred =  (BaseColumn*)(cvid.getItemExpr());
+ 
+              // the key column in the predicate is the same as the base column 
+              if ( keyCol->getColNumber() == colInKeyPred->getColNumber() ) {
+                 fullKeyPredicates_ += vid;
+                 found = TRUE;
+                 break;
+              }
+          }
+        }
+
+        if ( found )
+         break;
+     }
   }
 }
 
--- a/core/sql/optimizer/SimpleScanOptimizer.cpp
+++ b/core/sql/optimizer/SimpleScanOptimizer.cpp
@@ -218,25 +218,14 @@ SimpleFileScanOptimizer::isLogicalSubPartitioned() const
 SearchKey *
 SimpleFileScanOptimizer::constructSearchKey()
 {
+  // We do not need to incude the flatten version of RANGE SPEC predicates 
+  // to the exePreds because SearchKey is capable of handling such predicates.
+  // In addition, we do not want to confuse the SearchKey::ini() method 
+  // about what are the original predicates and what are derived predicates, 
+  // since combing these two forms will confuse SearchKey::makeHBaseSearchKey() 
+  // and cause the missing of the end keys (JIRA1449).
   ValueIdSet exePreds(getRelExpr().getSelectionPred());
 
-  ValueIdList selectionPredList(exePreds);
-  ItemExpr *inputItemExprTree;
-  if((CmpCommon::getDefault(RANGESPEC_TRANSFORMATION) == DF_ON )
-	  && ( exePreds.entries())){
-   ValueIdSet resultSet; 
-   inputItemExprTree =  selectionPredList.rebuildExprTree(ITM_AND,FALSE,FALSE);
-   ItemExpr* resultOld = revertBackToOldTree(CmpCommon::statementHeap(), 
-   		  inputItemExprTree);
-//	revertBackToOldTreeUsingValueIdSet(exePreds, resultSet);
-//	exePreds.clear();	
-//	ItemExpr* resultOld =  resultSet.rebuildExprTree(ITM_AND,FALSE,FALSE);
-//	exePreds += resultSet;
-//	doNotReplaceAnItemExpressionForLikePredicates(resultOld,exePreds,resultOld);
-	resultOld->convertToValueIdSet(exePreds, NULL, ITM_AND, FALSE);
-    doNotReplaceAnItemExpressionForLikePredicates(resultOld,exePreds,resultOld);
-   }
- 
   const Disjuncts *curDisjuncts = &(getDisjuncts());
 
   if(isLogicalSubPartitioned()) {
--- a/core/sql/optimizer/mdam.cpp
+++ b/core/sql/optimizer/mdam.cpp
@@ -2007,39 +2007,39 @@ void Disjuncts::computeCommonPredicates()
       get(disjunct,i);
       if(CmpCommon::getDefault(RANGESPEC_TRANSFORMATION) == DF_ON )
       {
-	    ValueIdSet inVidset = disjunct.getAsValueIdSet();
-	    ValueIdSet outVidset,parsedVs;
-        for (ValueId predId = inVidset.init();
-	     inVidset.next(predId);
-	     inVidset.advance(predId) )
-	    {
-         if(predId.getItemExpr()->getOperatorType() == ITM_RANGE_SPEC_FUNC )
-         {
-		  if(predId.getItemExpr()->child(1)->getOperatorType() == ITM_AND ){
-	        predId.getItemExpr()->child(1)->convertToValueIdSet(parsedVs, NULL, ITM_AND, FALSE);
+	 ValueIdSet inVidset = disjunct.getAsValueIdSet();
+	 ValueIdSet outVidset,parsedVs;
+         for (ValueId predId = inVidset.init();
+	      inVidset.next(predId);
+	      inVidset.advance(predId) )
+	 {
+            if(predId.getItemExpr()->getOperatorType() == ITM_RANGE_SPEC_FUNC )
+            {
+	       if(predId.getItemExpr()->child(1)->getOperatorType() == ITM_AND ){
+	          predId.getItemExpr()->child(1)->convertToValueIdSet(parsedVs, NULL, ITM_AND, FALSE);
 		    outVidset +=parsedVs;
-		 }
-	     else if(predId.getItemExpr()->child(1)->getOperatorType() != ITM_AND 
+            }
+	    else if(predId.getItemExpr()->child(1)->getOperatorType() != ITM_AND 
 			 && predId.getItemExpr()->child(1)->getOperatorType() != ITM_OR)
-	      outVidset += predId.getItemExpr()->child(1)->castToItemExpr()->getValueId();	    
-	   }
+	       outVidset += predId.getItemExpr()->child(1)->castToItemExpr()->getValueId();	    
+           }
 	   else
-	    outVidset +=predId;
-	    parsedVs.clear();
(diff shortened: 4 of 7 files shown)


core/sql/generator/GenPreCode.cpp
```diff
@@ -170,12 +174,16 @@ static NABoolean processConstHBaseKeys(Generator * generator,
...
-           exePreds,
+           keyPreds,
            nonKeyColumnSet,
            idesc,
            relExpr->getGroupAttr()->getCharacteristicOutputs(),
            mySearchKeys);
 
+      // Include any remaining key predicates that have not been 
+      // picked up (to be used as the HBase search keys).
+      exePreds += keyPreds;
```
DaveBirdsall: Not sure I understand the point of subtracting the keyPreds at line 169 then just adding them back here. The exePreds set is not passed to HbaseSearchKey::makeHBaseSearchKeys.

core/sql/optimizer/SearchKey.cpp
```diff
@@ -2711,6 +2716,65 @@ void SearchKey::computeCoveredLeadingKeys()
...
+  ValueId vid;
+  const ValueIdList& keyColumns = getKeyColumns();
+  for (CollIndex i=0; i<keyColumns.entries(); i++ ) {
+     if ( keyColumns[i].getItemExpr()->getOperatorType() == ITM_INDEXCOLUMN )
+       vid = ((IndexColumn*)(keyColumns[i].getItemExpr()))->getDefinition();
+     else
+       vid = keyColumns[i];
+
+     baseKeyCols.insertAt(baseKeyCols.entries(), vid);
+  }
```
DaveBirdsall: So this code works for index key columns and base table key columns (because if it is a base table key column, we go to line 2730 and don't attempt any mapping).

core/sql/optimizer/SearchKey.cpp
```diff
@@ -2711,6 +2716,65 @@ void SearchKey::computeCoveredLeadingKeys()
...
+
+  // compute the # of leading key columns covered by the key columns 
+  // in the disjunct.
```
DaveBirdsall: Not sure I understand this comment since I don't see a counter being incremented anywhere. I do see the fullKeyPredicates_ member being populated. I'm guessing the intent is to get the covered leading prefix of the key. The correctness of the code I'm guessing depends on "predicates" being ordered by key position but "predicates" is a ValueIdSet, so does it have an order?

core/sql/generator/GenPreCode.cpp
nonstop-qfchen: The key predicates are computed by the SearchKey already. So we take it out
from executor+selection predicates.

Later on, if makeHBaseSearchKey() decides any part of the key predicates
that can not be included in the Hbase search key, we add those back.

On Thu, Aug 27, 2015 at 10:03 AM, DaveBirdsall <notifications@github.com>
wrote:

> In core/sql/generator/GenPreCode.cpp
> <https://github.com/apache/incubator-trafodion/pull/66#discussion_r38105680>
> :
>
> >             nonKeyColumnSet,
> >             idesc,
> >             relExpr->getGroupAttr()->getCharacteristicOutputs(),
> >             mySearchKeys);
> >
> > +      // Include any remaining key predicates that have not been
> > +      // picked up (to be used as the HBase search keys).
> > +      exePreds += keyPreds;
>
> Not sure I understand the point of subtracting the keyPreds at line 169
> then just adding them back here. The exePreds set is not passed to
> HbaseSearchKey::makeHBaseSearchKeys.
>
> —
> <https://github.com/apache/incubator-trafodion/pull/66/files#r38105680>.
>



-- 
Regards, --Qifan

core/sql/optimizer/SearchKey.cpp
nonstop-qfchen: OK. I'll take that comment out in next check-in.  Thanks

On Thu, Aug 27, 2015 at 10:25 AM, DaveBirdsall <notifications@github.com>
wrote:

> In core/sql/optimizer/SearchKey.cpp
> <https://github.com/apache/incubator-trafodion/pull/66#discussion_r38108486>
> :
>
> > +  // If the key column is an INDEX COLUMN, convert it into a base column
> > +  ValueIdList baseKeyCols;
> > +
> > +  ValueId vid;
> > +  const ValueIdList& keyColumns = getKeyColumns();
> > +  for (CollIndex i=0; i<keyColumns.entries(); i++ ) {
> > +     if ( keyColumns[i].getItemExpr()->getOperatorType() == ITM_INDEXCOLUMN )
> > +       vid = ((IndexColumn*)(keyColumns[i].getItemExpr()))->getDefinition();
> > +     else
> > +       vid = keyColumns[i];
> > +
> > +     baseKeyCols.insertAt(baseKeyCols.entries(), vid);
> > +  }
> > +
> > +  // compute the # of leading key columns covered by the key columns
> > +  // in the disjunct.
>
> Not sure I understand this comment since I don't see a counter being
> incremented anywhere. I do see the fullKeyPredicates_ member being
> populated. I'm guessing the intent is to get the covered leading prefix of
> the key. The correctness of the code I'm guessing depends on "predicates"
> being ordered by key position but "predicates" is a ValueIdSet, so does it
> have an order?
>
> —
> <https://github.com/apache/incubator-trafodion/pull/66/files#r38108486>.
>



-- 
Regards, --Qifan

core/sql/generator/GenPreCode.cpp
DaveBirdsall: So, is keyPreds side-effected by the HbaseSearchKey::makeHBaseSearchKeys call? Then it might be different when added back.

nonstop-qfchen: Yes. Please refer to the comment of the method. It contains the
"left-over".

On Thu, Aug 27, 2015 at 10:31 AM, DaveBirdsall <notifications@github.com>
wrote:

> In core/sql/generator/GenPreCode.cpp
> <https://github.com/apache/incubator-trafodion/pull/66#discussion_r38109234>
> :
>
> >             nonKeyColumnSet,
> >             idesc,
> >             relExpr->getGroupAttr()->getCharacteristicOutputs(),
> >             mySearchKeys);
> >
> > +      // Include any remaining key predicates that have not been
> > +      // picked up (to be used as the HBase search keys).
> > +      exePreds += keyPreds;
>
> So, is keyPreds side-effected by the HbaseSearchKey::makeHBaseSearchKeys
> call? Then it might be different when added back.
>
> —
> <https://github.com/apache/incubator-trafodion/pull/66/files#r38109234>.
>



-- 
Regards, --Qifan

core/sql/generator/GenPreCode.cpp
```diff
@@ -154,15 +154,19 @@ static NABoolean processConstHBaseKeys(Generator * generator,
...
+      // seed keyPreds with only the full key predicate from skey
+      ValueIdSet keyPreds = skey->getFullKeyPredicates();
+
+      // include executorPreds and selection predicates 
+      // but exclude the full key predicates.
+      ValueIdSet exePreds;
+
       exePreds += executorPreds;
       exePreds += relExpr->getSelectionPred();
+      exePreds.subtractSet(keyPreds);
```
sureshsubbiah: I am wondering if this is a good thing to do. There are instances where we want a predicate to be both an executor predicate as well as a key predicate. In certain MDAM disjuncts the key predicate guarantees that a minimal set of rows is scanned but an exe pred has to be evaluated on the same pred again to guarantee correct results. I realize that the comment is vague without an example or two. I can  try to find one if necessary. Also since this PreCodeGen a predicate can be physically implemented differently based on what is available. That may also require us to keep things in the exepred?

core/sql/optimizer/SearchKey.cpp
```diff
@@ -2711,6 +2716,65 @@ void SearchKey::computeCoveredLeadingKeys()
...
+      // for each vid (or predicate), check the base column list
+      for (CollIndex i=0; i<baseKeyCols.entries(); i++ ) {
+
+         NABoolean found = FALSE;
+
+         // the predicate directly refers the base column 
+         if ( columnsCoveredByKey.containsTheGivenValue(baseKeyCols[i]) ) {
+            fullKeyPredicates_ += vid;
+            found = TRUE;
+         } else {
```
sureshsubbiah: I don't understand why we need this else branch? Is it because a particular basecolumn may have different valueids and we may not have found a match the "contains" check in the if branch.Can you please give an example where the else branch was actually taken?
Also could comparing by position lead to a false hit in the case where the same table is scanned twice? If we have an OR pred then we can get columns from a the second scan in the preds for the first table scan (I think). If we just compare by position we may think we have a match but the column may refer to the second scan and not the first. Not sure if this will happen as something else could prevent this from happening.

core/sql/optimizer/SimpleScanOptimizer.cpp
```diff
@@ -218,25 +218,14 @@ SimpleFileScanOptimizer::isLogicalSubPartitioned() const
 SearchKey *
 SimpleFileScanOptimizer::constructSearchKey()
 {
+  // We do not need to incude the flatten version of RANGE SPEC predicates 
+  // to the exePreds because SearchKey is capable of handling such predicates.
```
sureshsubbiah: I remember in the MDAM document there are examples where a predicate had to be in mdam disjunct as well as in exe pred to give correct results. This was because mdam disjunct could only guarantee minimal set of rows were returned (in such cases) instead of something that exactly matched the predicate. I can look for the example if it will help.

core/sql/optimizer/mdam.cpp
```diff
@@ -2007,39 +2007,39 @@ void Disjuncts::computeCommonPredicates()
...
-      {
+     }
+     else
+     {
 	if (i==0)
 	  commonPredicates_.insert(disjunct.getAsValueIdSet());
 	else
 	  commonPredicates_.intersectSet(disjunct.getAsValueIdSet());
-      }
+     }
```
sureshsubbiah: just for my understanding, there is no change in this file and we the diffs are all whitespaces. I am sorry I do not know how to ignore whitespace in this diff tool.

core/sql/regress/seabase/EXPECTED010
```diff
@@ -6113,6 +6114,7 @@ DESCRIPTION
   begin_keys(incl) ....... 4,b,-2147483648
   end_keys(excl) ......... 4,b,2
   new_rec_expr ........... (D assign 'upd')
+  predicate .............. (C < 2)
```
sureshsubbiah: It looks like we have added an exe pred that would clearly be covered by the begin/end key. I was expecting the changes I saw to have the opposite effect (i.e. miss adding necessary exe preds), so clearly my understanding is incomplete. However this is not that desirable also. Any ideas why this happened?

core/sql/regress/seabase/EXPECTED010
```diff
@@ -10316,12 +10319,12 @@ CREATE TABLE TRAFODION.SCH."delim_tab"
...
-Schema(Active  ): VOLATILE_SCHEMA_MXID110000212952122994710660444710000000002
+Schema(Active  ): VOLATILE_SCHEMA_MXID110000212132123073588113805260000000002
 
 --- SQL operation complete.
 >>get all volatile tables;
 
-Schema(Active  ): VOLATILE_SCHEMA_MXID110000212952122994710660444710000000002
+Schema(Active  ): VOLATILE_SCHEMA_MXID110000212132123073588113805260000000002
   Table: VTAB1
   Table: VTAB2
```
sureshsubbiah: i am guilty of not doing this many times. It will be good to add a query or two to regression tests that will show the end-key being added due to this change. This is an important change and also one that could be unintentionally broken by future work.

core/sql/optimizer/SimpleScanOptimizer.cpp
zellerh: Good question, and it makes me think of two things. One is that the example where MDAM can't retrieve the exact set is probably one that involves complex predicates on two or more columns, while RangeSpec itself is probably simple enough. But RangeSpec AND <some other predicate> might fall into this category. The second question is whether we really handle RangeSpec predicates correctly in MDAM code generation.

sureshsubbiah:
This is a wonderful change.

core/sql/optimizer/SearchKey.cpp
nonstop-qfchen: The new method is modeled from SearchKey::computeCoveredLeadingKeys().

The case of dealing the same table twice is interesting. But I think all
the predicates we are dealing are local, right?

On Thu, Aug 27, 2015 at 10:53 AM, Suresh Subbiah <notifications@github.com>
wrote:

> In core/sql/optimizer/SearchKey.cpp
> <https://github.com/apache/incubator-trafodion/pull/66#discussion_r38111897>
> :
>
> > +   vid = predicates.init();
> > +   for (; predicates.next(vid); predicates.advance(vid)) {
> > +
> > +      columnsCoveredByKey.clear();
> > +      vid.getItemExpr()->findAll(ITM_BASECOLUMN, columnsCoveredByKey, TRUE, TRUE);
> > +
> > +      // for each vid (or predicate), check the base column list
> > +      for (CollIndex i=0; i<baseKeyCols.entries(); i++ ) {
> > +
> > +         NABoolean found = FALSE;
> > +
> > +         // the predicate directly refers the base column
> > +         if ( columnsCoveredByKey.containsTheGivenValue(baseKeyCols[i]) ) {
> > +            fullKeyPredicates_ += vid;
> > +            found = TRUE;
> > +         } else {
>
> I don't understand why we need this else branch? Is it because a
> particular basecolumn may have different valueids and we may not have found
> a match the "contains" check in the if branch.Can you please give an
> example where the else branch was actually taken?
> Also could comparing by position lead to a false hit in the case where the
> same table is scanned twice? If we have an OR pred then we can get columns
> from a the second scan in the preds for the first table scan (I think). If
> we just compare by position we may think we have a match but the column may
> refer to the second scan and not the first. Not sure if this will happen as
> something else could prevent this from happening.
>
> —
> <https://github.com/apache/incubator-trafodion/pull/66/files#r38111897>.
>



-- 
Regards, --Qifan

core/sql/optimizer/mdam.cpp
nonstop-qfchen: Yes. there is no real change in this file. I aligned the blocks.

On Thu, Aug 27, 2015 at 11:00 AM, Suresh Subbiah <notifications@github.com>
wrote:

> In core/sql/optimizer/mdam.cpp
> <https://github.com/apache/incubator-trafodion/pull/66#discussion_r38112704>
> :
>
> >  	if (i==0)
> >  	  commonPredicates_.insert(disjunct.getAsValueIdSet());
> >  	else
> >  	  commonPredicates_.intersectSet(disjunct.getAsValueIdSet());
> > -      }
> > +     }
>
> just for my understanding, there is no change in this file and we the
> diffs are all whitespaces. I am sorry I do not know how to ignore
> whitespace in this diff tool.
>
> —
> <https://github.com/apache/incubator-trafodion/pull/66/files#r38112704>.
>



-- 
Regards, --Qifan

core/sql/regress/seabase/EXPECTED010
nonstop-qfchen: That sounds a good idea.

How about I added some next time?

On Thu, Aug 27, 2015 at 11:07 AM, Suresh Subbiah <notifications@github.com>
wrote:

> In core/sql/regress/seabase/EXPECTED010
> <https://github.com/apache/incubator-trafodion/pull/66#discussion_r38113428>
> :
>
> >
> >  --- SQL operation complete.
> >  >>get all volatile tables;
> >
> > -Schema(Active  ): VOLATILE_SCHEMA_MXID110000212952122994710660444710000000002
> > +Schema(Active  ): VOLATILE_SCHEMA_MXID110000212132123073588113805260000000002
> >    Table: VTAB1
> >    Table: VTAB2
> >
>
> i am guilty of not doing this many times. It will be good to add a query
> or two to regression tests that will show the end-key being added due to
> this change. This is an important change and also one that could be
> unintentionally broken by future work.
>
> —
> <https://github.com/apache/incubator-trafodion/pull/66/files#r38113428>.
>



-- 
Regards, --Qifan

core/sql/optimizer/SimpleScanOptimizer.cpp
nonstop-qfchen: My intention on the comment is that SearchKey is capable of processing
RANGE SPEC, and therefore there is no need to feed the flatten version of
it.

That complex predicates and RANGE SPEC may be partially represented in the
SeachKey probably is a known fact.  The fix tries to address it in one
aspect.

There maybe other areas that require some work, such as
cardinality/costing.

On Thu, Aug 27, 2015 at 11:08 AM, Hans Zeller <notifications@github.com>
wrote:

> In core/sql/optimizer/SimpleScanOptimizer.cpp
> <https://github.com/apache/incubator-trafodion/pull/66#discussion_r38113574>
> :
>
> > @@ -218,25 +218,14 @@ SimpleFileScanOptimizer::isLogicalSubPartitioned() const
> >  SearchKey *
> >  SimpleFileScanOptimizer::constructSearchKey()
> >  {
> > +  // We do not need to incude the flatten version of RANGE SPEC predicates
> > +  // to the exePreds because SearchKey is capable of handling such predicates.
>
> Good question, and it makes me think of two things. One is that the
> example where MDAM can't retrieve the exact set is probably one that
> involves complex predicates on two or more columns, while RangeSpec itself
> is probably simple enough. But RangeSpec AND might fall into this category.
> The second question is whether we really handle RangeSpec predicates
> correctly in MDAM code generation.
>
> —
> <https://github.com/apache/incubator-trafodion/pull/66/files#r38113574>.
>



-- 
Regards, --Qifan

core/sql/regress/seabase/EXPECTED010
sureshsubbiah: Yes, that will be perfect. Thank you.

core/sql/optimizer/SearchKey.cpp
sureshsubbiah: I think we can get a nonlocal predicate if there is a OR (or a CASE)

For example select * from t1 a, t1 b where a.col1 = 1 OR b.col2 = 2 ;
I think this predicate will get to pushed to both scans. It may end up being a IS NOT NULL after codegen, precodegen should see the predicate I think.

nonstop-qfchen: Can you confirm it by set a breaking point in SearchKey::init()?

Thanks

On Thu, Aug 27, 2015 at 12:04 PM, Suresh Subbiah <notifications@github.com>
wrote:

> In core/sql/optimizer/SearchKey.cpp
> <https://github.com/apache/incubator-trafodion/pull/66#discussion_r38119659>
> :
>
> > +   vid = predicates.init();
> > +   for (; predicates.next(vid); predicates.advance(vid)) {
> > +
> > +      columnsCoveredByKey.clear();
> > +      vid.getItemExpr()->findAll(ITM_BASECOLUMN, columnsCoveredByKey, TRUE, TRUE);
> > +
> > +      // for each vid (or predicate), check the base column list
> > +      for (CollIndex i=0; i<baseKeyCols.entries(); i++ ) {
> > +
> > +         NABoolean found = FALSE;
> > +
> > +         // the predicate directly refers the base column
> > +         if ( columnsCoveredByKey.containsTheGivenValue(baseKeyCols[i]) ) {
> > +            fullKeyPredicates_ += vid;
> > +            found = TRUE;
> > +         } else {
>
> I think we can get a nonlocal predicate if there is a OR (or a CASE)
>
> For example select * from t1 a, t1 b where a.col1 = 1 OR b.col2 = 2 ;
> I think this predicate will get to pushed to both scans. It may end up
> being a IS NOT NULL after codegen, precodegen should see the predicate I
> think.
>
> —
> <https://github.com/apache/incubator-trafodion/pull/66/files#r38119659>.
>



-- 
Regards, --Qifan

DaveBirdsall:
I'm unsure of the status of this change. Is there consensus that it should be committed as-is, or are there additional changes needed? (I'm neutral on the matter due to my relative ignorance.)

PR #67, opened 2015-08-27
[TRAFODION-1453] Improvements in Hive Scan

selvaganesang:
The null byte is now skipped over to look for the column and record delimiter while scanning the hdfs row buffer. The null byte will be part of the column value. The behavior of null byte in the column value is unknown. Skipping over null byte is now in sync with other tool's behavior.

Also, improved the code to scan the buffer for both record and column delimiter once. Earlier, the buffer was scanned twice; once for the record delimiter and another for column delimiter.

No errors are returned when null byte is found.

A sql error 7002 is now reported when a unsupported row format type is accessed from Trafodion.

*** ERROR[7002] The input row format of the hive table is not yet supported.

### DIFF
--- a/core/sql/executor/ExHdfsScan.cpp
+++ b/core/sql/executor/ExHdfsScan.cpp
@@ -45,6 +45,7 @@
 
 #include "ExpORCinterface.h"
 
+
 ex_tcb * ExHdfsScanTdb::build(ex_globals * glob)
 {
   ExExeStmtGlobals * exe_glob = glob->castToExExeStmtGlobals();
@@ -716,8 +717,8 @@ ExWorkProcRetcode ExHdfsScanTcb::work()
 	      {
 		// Position in the hdfsScanBuffer_ to the
 		// first record delimiter.  
-		hdfsBufNextRow_ = strchr(hdfsScanBuffer_,
-                                         hdfsScanTdb().recordDelimiter_);
+		hdfsBufNextRow_ = hdfs_strchr(hdfsScanBuffer_,
+                                         hdfsScanTdb().recordDelimiter_, hdfsScanBuffer_+trailingPrevRead_+ bytesRead_);
 		// May be that the record is too long? Or data isn't ascii?
 		// Or delimiter is incorrect.
 		if (! hdfsBufNextRow_)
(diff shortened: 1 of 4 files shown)


core/sql/executor/ExHdfsScan.cpp
```diff
@@ -1427,24 +1424,28 @@ char * ExHdfsScanTcb::extractAndTransformAsciiSourceToSqlRow(int &err,
...
+               return sourceRowEnd+1;
+         }
+         short len = 0;
+	 len = sourceColEnd - sourceData;
+         if (rdSeen) {
+            sourceRowEnd = sourceColEnd; 
+            hdfsLoggingRowEnd_  = sourceRowEnd;
+            if ((endOfRequestedRange_) && 
+                (sourceRowEnd >= endOfRequestedRange_))
+                   *(sourceRowEnd +1)= RANGE_DELIMITER;
```
sureshsubbiah: I am nervous about adding to table data. What will happen  if is '\002' in table data? If we are assuming that '\002' will not be present, can you please add a cqd or something that will allow us to change the RANGE_DELIMITER to something else if necessary. If you agree, subsequent checkin will be good.

core/sql/executor/ExHdfsScan.cpp
```diff
@@ -1427,24 +1424,28 @@ char * ExHdfsScanTcb::extractAndTransformAsciiSourceToSqlRow(int &err,
...
-            { 
-              len = sourceDataEnd - sourceData;
-              if (i != hdfsScanTdb().convertSkipListSize_ - 1)
-                isTrailingMissingColumn = TRUE;
-            }
-
-          if (attr) // this is a needed column. We need to convert
-          {
+      if (!isTrailingMissingColumn) {
+         sourceColEnd = hdfs_strchr(sourceData, rd, cd, sourceDataEnd, &rdSeen);
```
sureshsubbiah: What about NULL? We used to treat \N as NULL before I think. Can we still do that. For non-string columns will a missing value be treated as NULL?

sureshsubbiah:
Thank you. This is a much needed change.

core/sql/executor/ExHdfsScan.cpp
selvaganesang: Ok, I can fix this without the need for cqd by enabling when to look for range delimiter

core/sql/executor/ExHdfsScan.cpp
selvaganesang: This behavior is not changed with the modified code. See few lines below wherein the code determines the null value based on the content and the length

DaveBirdsall:
I will commit it

PR #73, opened 2015-09-03
TRAFODION-1480 Alternative to lscpu to get cores and cpus using info from /proc/cpuinfo

radumarias:
- Try to get cores and processors from /proc/cpuinfo when we cannot get from lscpu.


sureshsubbiah:
jenkins, add user

sureshsubbiah:
Thank you for this change. It looks great.

DaveBirdsall:
Thanks also for this change. It looks good to me also. Amanda, do you want to take a look and do the committing honors if you agree?

amandamoran:
Looks great! Thanks so much for finding this issue and fixing it!

PR #74, opened 2015-09-04
Support for features described in JIRA TRAFODION-19 && TRAFODION-1483

robertamarton:


core/sql/common/ComMisc.cpp
```diff
@@ -99,17 +100,74 @@ NABoolean ComIsTrafodionReservedSchemaName(
...
+      schName(len-1) == '_' ))
+    return TRUE;
+
   return FALSE;
 }
+
+// ----------------------------------------------------------------------------
+// function: ComConvertNativeNameToTrafName
+//
+// this fuction converts the native HIVE or HBASE object name into its
```
DaveBirdsall: function? (typo)

core/sql/common/ComMisc.cpp
```diff
@@ -99,17 +100,74 @@ NABoolean ComIsTrafodionReservedSchemaName(
...
+
+  // Catalog name is "TRAFODION"
+  NAString convertedName (CmpSeabaseDDL::getSystemCatalogStatic());
+  convertedName += ".";
+
+  // append transformed schema name, convert internal name to external format
+  ComAnsiNamePart internalAnsiName(tempSchemaName, ComAnsiNamePart::INTERNAL_FORMAT);
+  convertedName += internalAnsiName.getExternalName();
+
+  // object  name is appended without change
```
DaveBirdsall: A question: Do we know that object names always conform to regular identifier syntax? Or if they don't, are they already in delimited form when we enter this method?

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -1569,12 +1569,11 @@ NATable *BindWA::getNATable(CorrName& corrName,
...
-  // the HIVE object unless the inDDL flag is set.  inDDL is set for drop 
-  // table and SHOWDDL statements.  
-  // TDB - may want to merge the Trafodion version with the HIVE version.
-  // TDB - similar operation may be needed for external HBase tables
-  if ((table) && (table->isExternalTable() && (! bindWA->inDDL())))
+  // For now, don't allow access through the Trafodion external name created for
+  // native HIVE or HBASE objects unless the allowExternalTables flag is set.  
+  // allowExternalTables is set for drop table and SHOWDDL statements.  
+  // TDB - may want to merge the Trafodion version with the native version.
+  if ((table) && (table->isExternalTable() && (! bindWA->allowExternalTables())))
```
sureshsubbiah: I have forgotten why we thought this restriction is necessary.

A user registers a Hive/HBase table with us but is not able to use the table with the registered name in DML. Will UPDATE STATs will be on the original name too (and update stats code will find out registered name and use it for Table_uid?). This was the registration concept is almost like an internal implementation idea, at some point could even be done by some utility program for all existing Hive/hBase tables, and the user need not be aware of it?  

I am struggling with a user would want to register a table, if they cannot use it by that name.

DaveBirdsall:
Code changes look good to me.

core/sql/optimizer/RelExeUtil.cpp
```diff
@@ -3814,23 +3814,21 @@ RelExpr * DDLExpr::bindNode(BindWA *bindWA)
...
-          externalTable = TRUE;
-        }
-        else
-        {
-          *CmpCommon::diags() << DgSqlCode(-4222) << DgString0("DDL");
-          bindWA->setErrStatus();
-          return NULL;
+          if (dropTableNode->isExternal())
+            {
+              isHbase_ = TRUE;
```
sureshsubbiah: What does this flag mean? should it be set for a Hive external table too?

core/sql/parser/StmtDDLDrop.cpp
```diff
@@ -561,7 +561,8 @@ StmtDDLDropSchema::StmtDDLDropSchema(//const SchemaName & schemaName,
   // If the schema name specified is reserved name, users cannot drop them.
   // They can only be dropped internally.
   if ((! Get_SqlParser_Flags(INTERNAL_QUERY_FROM_EXEUTIL)) &&
-      (ComIsTrafodionReservedSchemaName(schemaQualName_.getSchemaName())))
+      (ComIsTrafodionReservedSchemaName(schemaQualName_.getSchemaName())) &&
+      (!ComIsTrafodionExternalSchemaName(schemaQualName_.getSchemaName())))
```
sureshsubbiah: Should this be a ! on line 565? I would have expected the condition to be similar to line 564. If I am wrong please excuse my mistake.

core/sql/regress/hive/EXPECTED009
```diff
@@ -266,10 +276,301 @@ A            B            C
...
+--- SQL operation failed with errors.
+>>create table t009t2 as select * from "_HV_SCH_T009_".t009t2;
+
+*** ERROR[4258] Trying to access external table TRAFODION."_HV_SCH_T009_".T009T2 through its external name format. Please use the native table name.
+
+*** ERROR[8822] The statement was not prepared.
+
+>>
+>>-- this create succeeds
+>>create table t009t1 as select * from hive.sch_t009.t009t1;
```
sureshsubbiah: Is t009t1 a Trafodion table or a hive external table (with an internal already created in a different script)?

core/sql/regress/hive/EXPECTED009
```diff
@@ -266,10 +276,301 @@ A            B            C
...
+>>-- this causes the external table to be invalid
+>>
+>>-- cleanup data from the old table, and create/load data with additional column
+>>sh regrhadoop.ksh fs -rm   /user/hive/exttables/t009t1/*;
+>>sh regrhive.ksh -v -f $REGRTSTDIR/TEST009_b.hive.sql;
+>>
+>>-- should fail - column mismatch
+>>select count(*) from hive.sch_t009.t009t1;
+
+*** ERROR[3078] The column list for HIVE.SCH_T009.T009T1 does not match its external table representation defined as TRAFODION."_HV_SCH_T009_".T009T1.
```
sureshsubbiah: Should this be a warning instead of an error? The exernal table is not being directly used here. Should the user be stopped and forced to make sure that the two definitions are consistent?

core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
```diff
@@ -379,84 +387,40 @@ void CmpSeabaseDDL::createSeabaseTableLike(
 }
 
 // ----------------------------------------------------------------------------
-// Method: createSeabaseTableExternalHive
+// Method: createSeabaseTableExternal
```
sureshsubbiah: Thanks. This will be a very useful methods during PoCs. Maybe the bulk loader could take a create option and use this method to create an empty Traf table before loading into it.

sureshsubbiah:
Changes look great.  They are ready to be merged in my opinion. Thank you.  
Are the externals fully described in the JIRA? I will read it as I find that I don't have the big picture in  mind anymore.

core/sql/common/ComMisc.cpp
robertamarton: Yep, will fix on my next delivery.

core/sql/common/ComMisc.cpp
robertamarton: We require the names to be in external format before entering this method.  I am using the ComAnsiNamePart class to convert from external to internal and back.  The constructor calls a method called "copyExternalName" which checks for a valid name.  However,  the constructor ignores the errors.  It looks like the method "isValid" should be called after constructing the object.  In this case, the name will be empty.  

I did run tests with both delimited and non-delimited names so the code is working correctly today.  In the future, we should probably change this code.  We should also look at other places where ComAnsiNamePart is called to make sure a similar issue does not exist.

core/sql/optimizer/BindRelExpr.cpp
robertamarton: We could open this up and allow the customer to use either name.  The same results should be returned independent of the name.  They are sort of like synonyms. But the way it is coded is that the customer will always use the hive.sch.table or hbase.sch.table name.  They really don't need to worry about the registered (external) name.  

We could open it up for all requests but there will be additional code and testing.  I could create a JIRA to change this or wait and see how people use external tables to see if this is a desired feature.

core/sql/optimizer/RelExeUtil.cpp
robertamarton: isExternal is set when the "external" keyword is used in the drop table syntax.  It is set for both hbase and hive tables.

core/sql/parser/StmtDDLDrop.cpp
robertamarton: Sorry, I should have written this cleaner.  

The first check - if the parserflag is set, then continue and don't return an error.

The second check - if the schema name does not begin with an underscore, then continue and don't return an error.

The final check - if the schema name is not an external name, that is begins with "_HV_" or "_HB_", then this is an error.  We don't want to drop tables beginning with the "_".

I added this check so people who have the privilege can drop externally generated schemas.  If this check did not exist, then you have to be DB__ROOT to remove these schemas.  I thought this was a little extreme.

core/sql/regress/hive/EXPECTED009
robertamarton: I actually toyed with this and could not decide whether this was an error or warning.

When I was testing, I discovered that the information I stored in the metadata did not match the information returned in the hive_tbl_desc.  This actually pointed to a couple of errors in the code.  There is also an issue when a mismatch does occur, that an error is returned and you have no idea what the issue is.  I had to add debug code to figure it out.

So this area can be handled better.  I can go either way - error or warning.

robertamarton:
I have not updated the external document for a while.  Let me work on it this week so it matches what is actually coded.

Thank you all for reviewing this code.

PR #76, opened 2015-09-09
[[Trafodion 1133]] [[Trafodion 1088]] [[Trafodion 1481]]

amandamoran:
Trafodion 1133 and Trafodion 1088 allow installer to add a node.

Trafodion 1481 seperates rpm installs into its own script so that it can be
run during an upgrade (to make sure all new packages needed are added). Bug still needs to be fixed is that it runs too many times (is ran once per node) and doesn't need to be. Will create a new jira for that small bug. Would like to check this in to prevent issue when upgrading.


install/installer/traf_config_setup
```diff
@@ -550,10 +562,10 @@ else
    exit -1
 fi
 
-hadoopVersion=$(curl -su $username:$password http://$URL/api/v1/clusters | grep version | grep -c CDH)
+hadoopVersion=$(curl -su admin:admin http://$URL/api/v1/clusters | grep version | grep -c CDH)
```
DaveBirdsall: Did you mean to hard-code admin:admin here (and below)? Seems there was a recent change to move away from this?

install/installer/trafodion_config_default
```diff
@@ -125,6 +125,9 @@ export PASSWORD="admin"
 # hadoop cluster name
 export CLUSTER_NAME=""
 
+#Trafodion Cluster name, normally the same as Cloudera cluster name
+export CLUSTERNAME=""
```
DaveBirdsall: Glad I saw this comment otherwise I'd have no clue about the difference between CLUSTER_NAME and CLUSTERNAME. Maybe these should be named differently? (Perhaps TRAFODION_CLUSTER_NAME for one of them?)

install/installer/traf_package_setup
```diff
@@ -0,0 +1,224 @@
...
+#========================================
+#Setting up pdsh variables
+if [ $node_count -eq 1 ]; then
+    TRAF_PDSH=""
+    TRAF_PDCP=""
+else
+    # use the -S option to cause pdsh to return largest of
+    # the remote command return values so we can tell if one
+    # or more of the remote commands failed
+    TRAF_PDSH="pdsh -R exec $MY_NODES ssh -q -n %h"
```
DaveBirdsall: The comment says, "use the -S option" but the commands don't seem to use it.

install/installer/traf_package_setup
```diff
@@ -0,0 +1,224 @@
...
+            if [ $node_count -ne 1 ]; then
+               for node in $NODE_LIST
+               do
+                  scp -q $LOCAL_WORKDIR/$epel_rpm $(whoami)@$node:$HOME
+               done
+            else
+               cp -rf $epel_rpm $HOME
+            fi
+
+            if [ $? != 0 ]; then
```
DaveBirdsall: What happens if scp fails on the first node but succeeds on the remaining ones? What would the value of $? be in that case?

install/installer/traf_config_setup
amandamoran: Sorry, I guess the merge didn't happened as I thought it would. Let me fix and check in.

install/installer/trafodion_config_default
amandamoran: CLUSTERNAME is used through out Trafodion and it can not be changed. It was deleted awhile back, since it was believed to not being used and had some bad consequences.

install/installer/traf_package_setup
amandamoran: Comment is incorrect, I will delete it and check in.

install/installer/traf_package_setup
amandamoran: Ah! Good thinking! Let me put in error checking! Thanks!

amandamoran:
jenkins, retest

arvind-narain:
Test failure seem to be related to generation or usage of certificate file, currently suspect some permission issues - checking.

amandamoran:
Great, thanks!

On Fri, Sep 11, 2015 at 11:12 AM, arvind-narain <notifications@github.com>
wrote:

> Test failure seem to be related to generation or usage of certificate
> file, currently suspect some permission issues - checking.
>
> —
> <https://github.com/apache/incubator-trafodion/pull/76#issuecomment-139616625>
> .
>



-- 
Thanks,

Amanda Moran

arvind-narain:
seems like CLUSTERNAME is getting set, hence mxosrvrs look for the certificate file in $HOME/sqcert rather than $MY_SQROOT/sqcert. I think we should not set the default value in trafodion_config_default.

trafodion_config_default:export CLUSTERNAME=""

Also do please look into sqenvcom.sh for CLUSTERNAME check.

PR #77, opened 2015-09-10
[TRAFODION-9] Non-blocking hbase operation to smoothen the data flow …

selvaganesang:
…in trafodion engine

The non-blocking Hbase operation for the index maintenance operators is now enabled by
default for IUD, merge and update commands.

The Hbase operations put, checkAndPut and delete operations from the trafodion engine
for the index maintenance is done using Executor Services to make it non-blocking and the
data flow in the trafodion operators can continue up to the point till it needs to pause for completion.

checkAndDelete is never used for index maintenance now.

Also, ensured that index maintenance uses rowset or single row operators when it should.

### DIFF
--- a/core/sql/executor/ExHbaseIUD.cpp
+++ b/core/sql/executor/ExHbaseIUD.cpp
@@ -417,6 +417,7 @@ ExWorkProcRetcode ExHbaseAccessInsertSQTcb::work()
 	    matches_ = 0;
             asyncCompleteRetryCount_ = 0;
             asyncOperationTimeout_ = 1;
+            asyncOperation_ = hbaseAccessTdb().asyncOperations() && getTransactionIDFromContext();
 	    step_ = INSERT_INIT;
 	  }
 	  break;
@@ -525,7 +526,7 @@ ExWorkProcRetcode ExHbaseAccessInsertSQTcb::work()
 	                                      row_,
                                               hbaseAccessTdb().useHbaseXn(),
                                               insColTSval_,
-                                              hbaseAccessTdb().asyncOperations());
+                                              asyncOperation_);
 
 	    if (retcode == HBASE_DUP_ROW_ERROR) // row exists, return error
 	      {
@@ -549,7 +550,7 @@ ExWorkProcRetcode ExHbaseAccessInsertSQTcb::work()
                 break;
             }
 	    matches_++;
-	    if (hbaseAccessTdb().asyncOperations()) {
+	    if (asyncOperation_) {
                 step_ = ASYNC_INSERT_COMPLETE;
                 return WORK_CALL_AGAIN;
             }
@@ -605,7 +606,7 @@ ExWorkProcRetcode ExHbaseAccessInsertSQTcb::work()
 				      row_,
                                       hbaseAccessTdb().useHbaseXn(),
 				      insColTSval_,
-                                      hbaseAccessTdb().asyncOperations());
+                                      asyncOperation_);
 
 	    if (setupError(retcode, "ExpHbaseInterface::insertRow")) {
 		step_ = HANDLE_ERROR;
@@ -616,7 +617,7 @@ ExWorkProcRetcode ExHbaseAccessInsertSQTcb::work()
 		break;
 	    }
 	    matches_++;
-	    if (hbaseAccessTdb().asyncOperations()) {
+	    if (asyncOperation_) {
                 step_ = ASYNC_INSERT_COMPLETE;
                 return WORK_CALL_AGAIN;
             }
@@ -668,7 +669,7 @@ ExWorkProcRetcode ExHbaseAccessInsertSQTcb::work()
 				     &rc, FALSE))
 		  return 1;
 	      }
-	    if (hbaseAccessTdb().asyncOperations()) {
+	    if (asyncOperation_) {
                step_ = ASYNC_INSERT_COMPLETE;
                return WORK_CALL_AGAIN;
             }
@@ -779,6 +780,7 @@ ExWorkProcRetcode ExHbaseAccessUpsertVsbbSQTcb::work()
 	    rowsInserted_ = 0;
             asyncCompleteRetryCount_ = 0;
             asyncOperationTimeout_ = 1;
+            asyncOperation_ = hbaseAccessTdb().asyncOperations() && getTransactionIDFromContext();
             numRowsInVsbbBuffer_ = 0;
 	    step_ = INSERT_INIT;
 	  }
@@ -921,14 +923,14 @@ ExWorkProcRetcode ExHbaseAccessUpsertVsbbSQTcb::work()
 				       hbaseAccessTdb().useHbaseXn(),
                                        insColTSval_,
                                        hbaseAccessTdb().getIsTrafLoadAutoFlush(), 
-				       hbaseAccessTdb().asyncOperations());
+				       asyncOperation_);
 
 	    if (setupError(retcode, "ExpHbaseInterface::insertRows"))
 	      {
 		step_ = HANDLE_ERROR;
 		break;
 	      }
-            if (hbaseAccessTdb().asyncOperations()) {
+            if (asyncOperation_) {
 		lastHandledStep_ = step_;
                 step_ = ASYNC_INSERT_COMPLETE;
                 break;
@@ -2065,7 +2067,7 @@ ExWorkProcRetcode ExHbaseUMDtrafUniqueTaskTcb::work(short &rc)
 	                                     tcb_->row_,
 					     (tcb_->hbaseAccessTdb().useHbaseXn() ? TRUE : FALSE),
 					     -1, //colTS_
-                                             FALSE); //tcb_->hbaseAccessTdb().asyncOperations());
+                                             tcb_->asyncOperation_);
 	    if ( tcb_->setupError(retcode, "ExpHbaseInterface::insertRow"))
 	      {
 		step_ = HANDLE_ERROR;
@@ -2099,7 +2101,7 @@ ExWorkProcRetcode ExHbaseUMDtrafUniqueTaskTcb::work(short &rc)
 						     colValToCheck_,
                                                      tcb_->hbaseAccessTdb().useHbaseXn(),
 						     -1, //colTS_
-                                                     FALSE); //tcb_->hbaseAccessTdb().asyncOperations());
+                                                     tcb_->asyncOperation_);
 
 	    if (retcode == HBASE_ROW_NOTFOUND_ERROR)
 	      {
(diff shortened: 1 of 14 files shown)


core/sql/generator/GenRelUpdate.cpp
```diff
@@ -2652,7 +2656,8 @@ short HbaseInsert::codeGen(Generator *generator)
   queue_index downqueuelength = (queue_index)getDefault(GEN_DP2I_SIZE_DOWN);
   Int32 numBuffers = getDefault(GEN_DP2I_NUM_BUFFERS);
 
-  if (getInsertType() == Insert::VSBB_INSERT_USER)
+  if (getInsertType() == Insert::VSBB_INSERT_USER &&
+              generator->oltOptInfo()->multipleRowsReturned())
     downqueuelength = 400;
```
sureshsubbiah: If at some future point we could change this 400 to a cqd setting or a cardinality based setting or even a constant that will read better. No change is needed now.

sureshsubbiah:
Change looks great to me Selva. I am not sure how we can merge when PR tests has failed? Should we rerun tests? Or maybe do something to testware (known diff etc) as part of this pull request?

DaveBirdsall:
Looks good to me too.

core/sql/executor/ExHbaseIUD.cpp
```diff
@@ -4193,7 +4188,7 @@ ExWorkProcRetcode ExHbaseAccessSQRowsetTcb::work()
 	      step_ = DONE;
 	  }
 	  break;
-      case ASYNC_INSERT_COMPLETE:
+      case ASYNC_OPERATION_COMPLETE:
```
zellerh: Would it be better to call this step "COMPLETE_ASYNC_OPERATION"? ASYNC_OPERATION_COMPLETE sounds as if the async operation is already complete, but what it really means is that we need to complete the operation.

Don't want to hold up this change, if you agree then maybe you can rename it in some future checkin.

core/sql/generator/GenRelUpdate.cpp
zellerh: Also, don't queue lengths have to be a power of 2? So, maybe use 512 when we make this change?

zellerh:
Looks good to me, too.

I had some trouble convincing myself that async operations won't cause false duplicate  row errors. Here is how I convinced myself, let me know if that's wrong:

Non-unique indexes: Each IM row contains the clustering key, so there are no conflict between rows. For conflicts between the IM delete and the IM update, we now use a pre-condition to suppress conflicting deletes and inserts. So, we should be fine with async operations.

Unique indexes: They could conflict between rows. In some cases the compiler will add a blocking operator (a sort) to avoid those conflicts. Async should be ok for those. In other cases we might not have the blocking operator. For those, there is already a risk of false duplicate errors, so async is not fundamentally altering the picture. Also, the compiler uses a "robust delete", which also checks the clustering key, so that the IM operation does not delete an index row for another base table row.

sureshsubbiah:
Hans, I think the union between the delete and insert for index maintenance is a blocked union. Will that not be enough to prevent false uniquness violations for unique indexes (with and without async) when blocking sort is absent?

core/sql/executor/ExHbaseIUD.cpp
selvaganesang: I can change it Hans. Also, I wanted to confirm if the stats returned is correct. So, I might do another code push

selvaganesang:
Currently, we support insert command only with unique indexes. So, async operation is not an issue. When we support upsert command, we need to ensure that async operation can work with unique indexes too. 

 With the following table definition for t12 and unique index t12i3, let's say the following rows already exist

(1,2,3,4)
(11,12,13,14)

I am not sure about the semantics of the upsert command with unique indexes. Which of the following commands should succeed?

Upsert (1,2,3,4) 
Upsert(1,2,3,5)  
Upsert (1,11,3,4) 

-- Definition of Trafodion table TRAFODION.SELVA.T12
-- Definition current  Thu Sep 10 20:41:39 2015

  (
    A                                LARGEINT NO DEFAULT NOT NULL NOT DROPPABLE
      SERIALIZED
  , B                                INT NO DEFAULT NOT NULL NOT DROPPABLE
      SERIALIZED
  , C                                INT DEFAULT NULL SERIALIZED
  , D                                INT DEFAULT NULL SERIALIZED
  , "_SALT_"                         INT UNSIGNED NO DEFAULT NOT NULL NOT
      DROPPABLE SERIALIZED
  )
  PRIMARY KEY ("_SALT_" ASC, A ASC, B ASC)

-- Definition of Trafodion unique index T12I3
-- Definition current  Thu Sep 10 20:41:39 2015

  (
    D                                INT DEFAULT NULL SERIALIZED
  , "_SALT_"                         INT UNSIGNED NO DEFAULT NOT NULL NOT
      DROPPABLE SERIALIZED
  , A                                LARGEINT NO DEFAULT NOT NULL NOT DROPPABLE
      SERIALIZED
  , B                                INT NO DEFAULT NOT NULL NOT DROPPABLE
      SERIALIZED
  )
  PRIMARY KEY
  (
    D ASC
  )

selvaganesang:
How about upsert (1,2,3,14) also

zellerh:
Suresh makes a good point that I hadn't understood well before. I guess I didn't assume that we wait for the async delete to be complete before we send the up queue entry, so we still ensure that the delete happens before the insert.

About Selva's question:

upsert (1,2,3,4) (set d=4) ==> Success, pre-condition skips index maintenance?
upsert(1,2,3,5) (set d=5) ==> Success, delete 4 and insert 5 into the unique index.
upsert 1,11,3,4) (insert (1,11,3,4) ==> Success if we did second upsert, failure without second upsert
upsert (1,2,3,14) ==> Failure, duplicate 14

It might be better to use showddl and real SQL statements to make sure the scenario is not ambiguous.

sureshsubbiah:
I am about to enable upsert for table with unique indexes. Will checkout these scenarios and report back.

selvaganesang:
Async operation doesn't wait for the individual operation in the IM to complete before we send the up queue entry. In fact, the up query entry is sent so that other operators can do their part and push their operation to hbase. The async operation is forced to wait for completion after processing either a single down queue entry or multiple down queue entries depending upon the type of operator involved(Single row operation or rowset operation). 

If the order of operations is important for unique indexes, then the async operation should be disabled. For the above use cases with the unique indexes, the async operation shouldn't be problem at all. I just went through these cases with Suresh and we think the order of operation is not needed due to pre-condition.

For the update command where-in the rows needs to be deleted before the new values can be inserted ( similar to updating a column value in a multi-column primary key), the plan is complex. The plan operates in async fashion for the old rows first and then operates in async fashion for the new rows.

Thanks to Suresh, we concluded that it should be ok to have async operation enabled for this case too.

selvaganesang:
Also, just wanted to update the PR with this info. The insert command is non-blocking with base table operation meaning insertion into base table and the indexes can happen simultaneously. In case of upsert command, insertion into base table happens first in blocking fashion and then the insertion into all the indexes can happen at the same time. The merge operator in upsert command is still blocking. We can make this operation also non-blocking.  I will attempt to do it later.

DaveBirdsall:
Just wanted to check: Are we ready to commit this change? Suresh, Hans, any outstanding concerns? Selva, is there anything you wish to update before commit?

selvaganesang:
Dave, I am done with the code changes unless there are more review
comments. I just need to correct my last comment about the differences
between upsert and insert.



Selva



*Sent:* Friday, September 11, 2015 9:14 AM
*Cc:* selvaganesang <selva.govindarajan@esgyn.com>
*Subject:* Re: [incubator-trafodion] [TRAFODION-9] Non-blocking hbase
operation to smoothen the data flow … (#77)



Just wanted to check: Are we ready to commit this change? Suresh, Hans, any
outstanding concerns? Selva, is there anything you wish to update before
commit?

—
<https://github.com/apache/incubator-trafodion/pull/77#issuecomment-139588479>
.

sureshsubbiah:
Dave, thanks for checking. I think this change is ready to be merged.

DaveBirdsall:
OK, I can commit it.

sureshsubbiah:
Happy to report that with the fix for Trafodion 1456 all the 4 upsert test cases Selva proposed above work as Hans expects them to.

Results below. Thank you for such a good test case Selva, Hans.

>>create table TRAFODION.JIRA.T12
(
A LARGEINT NO DEFAULT NOT NULL NOT DROPPABLE
SERIALIZED
, B INT NO DEFAULT NOT NULL NOT DROPPABLE
SERIALIZED
, C INT DEFAULT NULL SERIALIZED
, D INT DEFAULT NULL SERIALIZED
, PRIMARY KEY (A ASC, B ASC)
)
SALT USING 2 PARTITIONS;
--- SQL operation complete.
>>create unique index t12i3 on t12(d);
--- SQL operation complete.

>>upsert into t12 values (1,2,3,4), (11,12,13,14) ;
--- 2 row(s) inserted.
>>select * from t12 ;
A                     B            C            D          
--------------------  -----------  -----------  -----------

                   1            2            3            4
                  11           12           13           14

--- 2 row(s) selected.
>>set parserflags 1 ;

--- SQL operation complete.
>>select * from table(index_table t12i3);

D@           _SALT_      A                     B          
-----------  ----------  --------------------  -----------

          4           0                     1            2
         14           1                    11           12

--- 2 row(s) selected.
>>upsert into t12 values (1,2,3,4) ;

--- 1 row(s) inserted.
>>select * from t12 ;

A                     B            C            D          
--------------------  -----------  -----------  -----------

                   1            2            3            4
                  11           12           13           14

--- 2 row(s) selected.
>>select * from table(index_table t12i3);
D@           _SALT_      A                     B          
-----------  ----------  --------------------  -----------

          4           0                     1            2
         14           1                    11           12

--- 2 row(s) selected.
>>upsert into t12 values (1,2,3,5) ;

--- 1 row(s) inserted.
>>select * from t12 ;

A                     B            C            D          
--------------------  -----------  -----------  -----------

                   1            2            3            5
                  11           12           13           14

--- 2 row(s) selected.
>>select * from table(index_table t12i3);
D@           _SALT_      A                     B          
-----------  ----------  --------------------  -----------

          5           0                     1            2
         14           1                    11           12

--- 2 row(s) selected.
>>upsert into t12 values (1,11,3,4) ;

--- 1 row(s) inserted.
>>select * from t12 ;
A                     B            C            D          
--------------------  -----------  -----------  -----------

                   1            2            3            5
                   1           11            3            4
                  11           12           13           14

--- 3 row(s) selected.
>>select * from table(index_table t12i3);
D@           _SALT_      A                     B          
-----------  ----------  --------------------  -----------

          4           1                     1           11
          5           0                     1            2
         14           1                    11           12

--- 3 row(s) selected.
>>delete from t12 ;

--- 3 row(s) deleted.
>>upsert into t12 values (1,2,3,4), (11,12,13,14) ;
--- 2 row(s) inserted.
>>select * from t12 ;
A                     B            C            D          
--------------------  -----------  -----------  -----------

                   1            2            3            4
                  11           12           13           14

--- 2 row(s) selected.
>>select * from table(index_table t12i3);
D@           _SALT_      A                     B          
-----------  ----------  --------------------  -----------

          4           0                     1            2
         14           1                    11           12

--- 2 row(s) selected.
>>upsert into t12 values (1,11,3,4) ;

*** ERROR[8102] The operation is prevented by a unique constraint.

--- 0 row(s) inserted.
>>upsert into t12 values (1,2,3,14) ;

*** ERROR[8102] The operation is prevented by a unique constraint.

--- 0 row(s) inserted.
>>

PR #78, opened 2015-09-10
Changes to sqenvcom.sh file to make it easier to run on Apache HBase w/o distro

zellerh:
The script now checks for common HBase and Hadoop setup scenarios, for
example with environment variables HADOOP_PREFIX, HBASE_HOME and
HIVE_HOME set.

This work is not complete yet, we need to document the entire process
and probably need to make a few small code changes to fix interfaces
that went away or are not yet added in Apache Hadoop/HBase.

### DIFF
--- a/core/sqf/sqenvcom.sh
+++ b/core/sqf/sqenvcom.sh
@@ -433,6 +433,8 @@ elif [[ -d /opt/mapr ]]; then
 
   # We tried this with MapR 3.1, which has hadoop-0.20.2, hbase-0.94.13, hive-0.12
 
+  [[ $SQ_VERBOSE == 1 ]] && echo "Found /opt/mapr, this is a MapR distro"
+
   # Note that hadoopversion and hiveversion are not officially
   # supported by MapR, only hbaseversion is. We recommend creating
   # these files to guide Trafodion to the right version, if necessary.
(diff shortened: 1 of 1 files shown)


sureshsubbiah:
Excited about this change. I did not see anything that could be improved. Can merge after a few other look at it.

core/sqf/sqenvcom.sh
```diff
@@ -737,7 +863,7 @@ SQ_CLASSPATH=${SQ_CLASSPATH#:}
 
 # add Hadoop and HBase config dirs to classpath, if they exist
 if [[ -n "$HADOOP_CNF_DIR" ]]; then SQ_CLASSPATH="$SQ_CLASSPATH:$HADOOP_CNF_DIR"; fi
-if [[ -n "$HIVE_CNF_DIR"   ]]; then SQ_CLASSPATH="$SQ_CLASSPATH:$HBASE_CNF_DIR";  fi
+if [[ -n "$HBASE_CNF_DIR"  ]]; then SQ_CLASSPATH="$SQ_CLASSPATH:$HBASE_CNF_DIR";  fi
```
DaveBirdsall: Nice catch on finding this bug.

DaveBirdsall:
Looks good to me.

selvaganesang:
Looks good to me. Do you think that we should include only the client side hbase jar files alone in the CLASSPATH. Is it feasible to do it?

zellerh:
There are several opportunities to shorten the classpath: One is including only the client jar files, another is avoiding the duplication of the Hadoop jars in HBase and the Hadoop jars in Hadoop. We do call more interfaces than just what's in the client, so as far as I remember, reducing the jars included is a tedious and not very effective thing to do.

PR #83, opened 2015-09-17
[TRAFODION-1456] Merge statement returns error 3241 when there is unique index in the table [TRAFODION-1493] Logs TMDUF does not scan udr logs Also enabled merge with views and merge with RI constraint.

sureshsubbiah:


core/sql/executor/ExHbaseIUD.cpp
```diff
@@ -472,7 +472,15 @@ ExWorkProcRetcode ExHbaseAccessInsertSQTcb::work()
 	  {
 	    rc = applyPred(scanExpr());
 	    if (rc == 1) // expr is true or no expr
-	      step_ = CREATE_MUTATIONS;
+	      {
+		 rc = evalInsDelPreCondExpr();
```
selvaganesang: Should we consider move this expression evaluation to SETUP_INSERT step since it evaluates from queue entry.

sureshsubbiah: I thought of the preConditionExpr as something similar to the scanExpr. Agree that it is on upqueue though. Do not see any benefit from changing except if we are thinking that some steps are avoided? I can try it out, but would rather do that in a separate checkin if we want to go that route.

About VsbbUpsert, right now the preCondition expression is evaluated only for unique index for which we need CheckAndPut. I am assuming this is always in the SQInsertTcb. Merge on native HBase tables is also not supported yet.

DaveBirdsall:
Changes look good to me. I take it the insert expression (generalization of the delete expression to insDel expression) is there for views with check option?

sureshsubbiah:
Issues found by Selva and Dave have been addressed. View with check option was not being processed correctly by merge, this is now disabled. Thank you very much forthe review. I will be merging this change in in the next 30 mins, unless there are objections.

PR #84, opened 2015-09-17
Small bug fix to allow Cloudera to install from Local Repo

amandamoran:
Adding small change to allow Cloudera to install from a local repo


DaveBirdsall:
Looks good to me.

PR #89, opened 2015-09-22
[TRAFODION-1473] External file support for LOB input and Extract

sandhyasun:
•	Support filetolob  function to input file data into a lob .
•	Support for Extract syntax to extract  length of a given lob.
•	Support for Extract syntax to extract data given a lob handle. This includes support for lobtoFile function to extract the lobdata into a file.
Various syntaxes :
Extract lobtofile (lob <lob handle> , <filename>) ; if file exists error, if file does not exist create.
Extract lobtofile (lob <lob handle> , <filename>, truncate ) ; if file exists truncater, if file does not exist create.
Extract lobtofile (lob <lob handle> , <filename>, create, append ) ; if file exists append, if file does not exist error
Extract lobtofile (lob <lob handle> , <filename>, create, append ) ; if file exists append, if file does not exist create.

•	Remove the 2G limitation for lob data files.
•	Support for extract from a lob column to an external file(on platform linux or hdfs file) in chunks .
•	Support for input from an external file (on platform linux or hdfs file) in chunks into lob column.
•	Changed semantics of lob  extract to file to match  hdfs’s APIs (copyToLocalFile)
•	Lob Max size is now a CQD and configurable
•	Lob Max Chunk Size limits how much data is used to allocate and write in chunks. Earlier the chunk size was sixed and now it is configurable with a CQD.
•	Earlier the lob extract size was limited to LOB_OUTPUT_SIZE which was 32K
•	Now it’s limited by LOB_MAX_CHUNK_MEM_SIZE but can also be specified by the caller in syntax.  (there is a check to ensure it is less than LOB_MX_CHUNK_MEM_SIZE).

•	Added Regression test to:

1) insert into lob column using :
a)	local text file.
b)	hdfs text file
c)	local binary file (jpg file)
d)	hdfs binary format file (jpg file)

2) Regression test to extract from lob column  into :
a)	Local text file
b)	Local hdfs text file
c)	Local binary file (jpg file)
d)	Hdfs binary file (jpg file)

There are still items TBD for this JIRA. See JIRA for details.


svarnau:
How big are these jpeg files?
Do we have legal right to use them (even as tests data)?
Could the test scripts download or generate data rather than including them in the source code?

sandhyasun:
The files need to be on platform. They don't have any copyright issues, they are used to test chunking and inserting into lobs in parts. If I make them smaller, the code won't get exercised. 
-rw-rw-r--. 1 sandhyasun sandhyasun 431839 Sep 17 06:06 anoush.jpg
-rw-rw-r--. 1 sandhyasun sandhyasun 211849 Sep 17 06:06 deep.jpg 
Looks like there are expected and test files that are larger than this too.
Does anyone have a concern with these size files being checked in ?

svarnau:
I'll go ahead and raise the size limit that the static checks checking for, so the tests can proceed.

svarnau:
jenkins, retest please

core/sql/exp/ExpLOBaccess.cpp
```diff
@@ -1599,203 +1627,290 @@ Ex_Lob_Error ExLob::readDataToMem(char *memAddr,
...
-      return LOB_DEST_FILE_OPEN_ERROR;
-    }
-
-    if (pwrite(fdDestFile, data, desc.getSize(), 0) == -1) {
-      free(data);
-      close(fdDestFile);
-      return LOB_DATA_WRITE_ERROR;
-    }
+	  
+    if (fdDestFile >0 )
```
DaveBirdsall: Should this be >= 0? (open could return 0 on success; highly unlikely here because file descriptor 0 is likely already in use, but conceivable.)

core/sql/exp/ExpLOBaccess.cpp
```diff
@@ -1599,203 +1627,290 @@ Ex_Lob_Error ExLob::readDataToMem(char *memAddr,
...
+      {
+	if ((LobTgtFileFlags)fileflags == Lob_Error_Or_Create)
+	    return LOB_TARGET_FILE_EXISTS_ERROR;	  
+      }
+    if (fdDestFile == -1) 
+      {
+	if (((LobTgtFileFlags)fileflags == Lob_Append_Or_Error) ||
+	    ((LobTgtFileFlags)fileflags == Lob_Truncate_Or_Error))
+	  return LOB_TARGET_FILE_OPEN_ERROR;
```
DaveBirdsall: Should we capture errno on these open failures? errno would indicate why the open failed.

core/sql/exp/ExpLOBaccess.cpp
```diff
@@ -1599,203 +1627,290 @@ Ex_Lob_Error ExLob::readDataToMem(char *memAddr,
...
-    getLobGlobalHeap()->deallocateMemory(lobData);
-    fdTgtFile = NULL;
-    /*
-#ifdef SQ_USE_HDFS
-    // tbd
-#else
-    if (pread(fdData_, data, desc.getSize(), desc.getOffset()) == -1) {
-      free(data);
-      return LOB_DATA_READ_ERROR;
+Ex_Lob_Error ExLob::readDataToHdfsFile(char *tgtFileName,  Int64 offset, Int64 size, Int64 &writeOperLen, Int64 lobMaxChunkMemLen, Int32 fileflags)
```
DaveBirdsall: Would it make sense to strongly type the fileflags parameter as LobTgtFileFlags instead of Int32?

core/sql/parser/sqlparser.y
```diff
@@ -12682,15 +12682,15 @@ update_obj_to_lob_function :
 			        {
 				  $$ = new (PARSERHEAP()) LOBupdate( $3, NULL, LOBoper::STRING_, TRUE);
 				}
-                          | TOK_FILETOLOB '(' TOK_LOCATION character_literal_sbyte ',' TOK_FILE character_literal_sbyte ')'
+                          | TOK_FILETOLOB '('character_literal_sbyte ')'
```
DaveBirdsall: I like the new syntax better. Is there any ANSI syntax for this?

DaveBirdsall:
Code changes look good. I have the same question as Steve on the pictures though: Are they in the public domain? If not, perhaps substitute some other pictures? (Could just snap a picture on your phone perhaps.)

core/sql/exp/ExpLOBaccess.cpp
sandhyasun: You're right. Will change the check.

core/sql/exp/ExpLOBaccess.cpp
sandhyasun: WiIll consider doing that  in a nother checkin. The error message and interfaces will need to be enhanced to provide more error detail.

core/sql/exp/ExpLOBaccess.cpp
sandhyasun: I have piggybacked/overloaded an Int32 param in the ExLobsOper method which is used widely for many lob operations. FileFlags is relevant only for the read interface (for extract).  So left it as an Int32.

core/sql/parser/sqlparser.y
sandhyasun: Nothing mentioned in standard. So this is homegrown syntax .

sandhyasun:
Yes pictures are public domain - no copyright notices.

PR #91, opened 2015-09-24
[TRAFODION-1497] Fix expected results for compGeneral/TEST042

DaveBirdsall:
The fix to JIRA TRAFODION-1494 unfortunately changed the results for compGeneral/TEST042. This check-in updates the expected results for that test.

### DIFF
--- a/core/sql/regress/compGeneral/EXPECTED042
+++ b/core/sql/regress/compGeneral/EXPECTED042
@@ -1,9 +1,9 @@
 >>showstats for table t042_orderline on ol_o_id detail;
 
 Detailed Histogram data for Table TRAFODION.ORDERENTRY.T042_ORDERLINE
-Table ID: 6902403919319426105
+Table ID: 2761910211648663657
 
-Hist ID:    1306243874
+Hist ID:    1309682661
 Column(s):  OL_O_ID
 Total Rows: 10
 Total UEC:  4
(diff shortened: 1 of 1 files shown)


core/sql/regress/compGeneral/EXPECTED042
```diff
@@ -1871,8 +1871,8 @@ NUM_ENTRIES  TEXT_ENTRIES  NUM_PLANS
 NUM_ENTRIES  TEXT_ENTRIES  NUM_PLANS 
 -----------  ------------  ----------
 
-         17             0          17
-          1             0           1
+         22             0          22
```
sureshsubbiah: Does this mean more metadata queries are being cached now? Dow we know what the new entries are? There is no doubt that the fix for 1494 is necessary and required. I am not able to understand fully the changes in this expected file. If it will take too much effort to find out, we could simply accept the result. I am good with that too.

core/sql/regress/compGeneral/EXPECTED042
```diff
@@ -1977,25 +1977,31 @@ NUM_HITS    NUM_PARAMS
...
+         0           3
          0           3
+         0           4
          0           5
          1           1
          1           1
          1           2
          1           2
          1           3
+         1           3
```
sureshsubbiah: This looks more serious. We have more entries in the query cache, but are getting fewer hits. I wonder what happened.

DaveBirdsall: Same number of queries as before. What's happening is that some of the cache entries are not matching up due to differences in CQDs. That's not surprising as the bug in 1494 prevented the toggling of CQDs in some circumstances. In this particular test suite, the metadata queries get cached very early on and the increase in cache usage stops. (I found this out by peppering a copy of the test script with selects against querycache.) So it is up-front growth. Of course if we wish, we could investigate further and perhaps add or delete some of the CQDs in the metadata code to improve the entry sharing.

DaveBirdsall:
jenkins, retest please

sureshsubbiah:
Thank you for explaining.

PR #96, opened 2015-09-30
[TRAFODION-1514] Stop using custom suffix (Nv44) of ICU libs.

ryzuo:
We keep using our customized suffix from Neoview time, Nv4,
for ICU lib files, this causes confusion to developers who wants
to build Linux ODBC driver. Remove this suffix from the name of
lib files, to get rid of the confusion.

### DIFF
--- a/core/conn/unixodbc/odbc/odbcclient/unixcli/makefile.lnx
+++ b/core/conn/unixodbc/odbc/odbcclient/unixcli/makefile.lnx
@@ -29,7 +29,7 @@ VPATH = ./DrvrManager:./trace:./cli:./TCPIPV4:./Interface:./common:./platform:./
 
 C_FLAGS = -m64 -Dunixcli -DVERSION3 -DMXLINUX -DDISABLE_TRACE -DSIZEOF_LONG=8 -DSIZEOF_LONG_INT=8 -Wall -fPIC -w -DASYNCIO -DTRACE_COMPRESSION -I$(ICU)/linux64/include -I/usr/include/openssl
 
-LIBS = -lpthread -lrt -ldl -lz -L$(ICU)/linux64/lib -licuucNv44 -Wl,--hash-style=both /usr/lib64/libssl.a /usr/lib64/libcrypto.a -Wl,--version-script=linux.exports
+LIBS = -lpthread -lrt -ldl -lz -L$(ICU)/linux64/lib -licuuc -Wl,--hash-style=both /usr/lib64/libssl.a /usr/lib64/libcrypto.a -Wl,--version-script=linux.exports
 
 ifeq ($(MAKECMDGOALS),linux64_release)
 	OBJDIR = ./obj/linux64/release/
--- a/core/conn/unixodbc/odbc/odbcclient/unixcli/package/linux64_install.sh
+++ b/core/conn/unixodbc/odbc/odbcclient/unixcli/package/linux64_install.sh
@@ -40,7 +40,7 @@ ETC_DIR=
 INCLUDE_DIR=
 SAMPLE_DIR=
 #package files, excluding this file
-pkgfiles=(libhpodbc_l64.so.?.*.? MXODSN MD5SUM LICENSE hpsqlext.h connect_test.cpp license.txt libicuucNv44.so.44 libicudataNv44.so.44)
+pkgfiles=(libhpodbc_l64.so.?.*.? MXODSN MD5SUM LICENSE hpsqlext.h connect_test.cpp license.txt libicuuc.so.44 libicudata.so.44)
 
 #check whether the package directory has all files (or this script is invoked form "PkgTmp" dir)
 function check_package {
@@ -151,9 +151,9 @@ function install_lib {
       err=0
       mv -f $lib $lib.SAV
       (( err += $? ))
-	  mv -f libicuucNv44.so.44 libicuucNv44.so.44.SAV     
+	  mv -f libicuuc.so.44 libicuuc.so.44.SAV     
       (( err += $? ))   
-	  mv -f libicudataNv44.so.44 libicudataNv44.so.44.SAV     
+	  mv -f libicudata.so.44 libicudata.so.44.SAV     
       (( err += $? ))    
       #remove link
       rm -f libhpodbc_l64.so.$ver
@@ -168,21 +168,21 @@ function install_lib {
    fi
    #copy library
    err=0
-   cp -f $PKG_DIR/libicuucNv44.so.44  .
+   cp -f $PKG_DIR/libicuuc.so.44  .
    (( err += $? ))
-   chmod 555 libicuucNv44.so.44
+   chmod 555 libicuuc.so.44
    (( err += $? ))
-   rm -f libicuucNv44.so
+   rm -f libicuuc.so
    (( err += $? ))
-   ln -s libicuucNv44.so.44 libicuucNv44.so
+   ln -s libicuuc.so.44 libicuuc.so
    (( err += $? ))
-   cp -f $PKG_DIR/libicudataNv44.so.44  .
+   cp -f $PKG_DIR/libicudata.so.44  .
    (( err += $? ))
-   chmod 555 libicudataNv44.so.44
+   chmod 555 libicudata.so.44
    (( err += $? ))       
-   rm -f libicudataNv44.so
+   rm -f libicudata.so
    (( err += $? ))       
-   ln -s libicudataNv44.so.44 libicudataNv44.so
+   ln -s libicudata.so.44 libicudata.so
    (( err += $? ))       
    cp -f $PKG_DIR/libhpodbc_l64.so.?.*.? .
    (( err += $? ))
--- a/core/conn/unixodbc/odbc/odbcclient/unixcli/package/mklnxpkg.sh
+++ b/core/conn/unixodbc/odbc/odbcclient/unixcli/package/mklnxpkg.sh
@@ -95,9 +95,9 @@ then
     let err=0
     cp ../../../libtrafodbc_l64.so ${libname##*/}
     ((err += $?))
-    cp ${ICU}/linux64/lib/libicuucNv44.so.44.0 libicuucNv44.so.44
+    cp ${ICU}/linux64/lib/libicuuc.so.44.0 libicuuc.so.44
     ((err += $?))
-    cp ${ICU}/linux64/lib/libicudataNv44.so.44.0 libicudataNv44.so.44
+    cp ${ICU}/linux64/lib/libicudata.so.44.0 libicudata.so.44
     ((err += $?))
     md5sum ./${libname##*/} >  MD5SUM
     ((err += $?))
--- a/core/conn/unixodbc/odbc/odbcclient/unixcli/package/trafodbclnx_install.sh
+++ b/core/conn/unixodbc/odbc/odbcclient/unixcli/package/trafodbclnx_install.sh
@@ -40,7 +40,7 @@ ETC_DIR=
 #INCLUDE_DIR=
 SAMPLE_DIR=
 #package files, excluding this file
-pkgfiles=(libtrafodbc_l64.so TRAFDSN MD5SUM LICENSE connect_test.cpp license.txt libicuucNv44.so.44 libicudataNv44.so.44)
+pkgfiles=(libtrafodbc_l64.so TRAFDSN MD5SUM LICENSE connect_test.cpp license.txt libicuuc.so.44 libicudata.so.44)
 
 #check whether the package directory has all files (or this script is invoked form "PkgTmp" dir)
 function check_package {
@@ -146,9 +146,9 @@ function install_lib {
       err=0
       mv -f $lib $lib.SAV
       (( err += $? ))
-	  mv -f libicuucNv44.so.44 libicuucNv44.so.44.SAV     
+	  mv -f libicuuc.so.44 libicuuc.so.44.SAV     
       (( err += $? ))   
-	  mv -f libicudataNv44.so.44 libicudataNv44.so.44.SAV     
+	  mv -f libicudata.so.44 libicudata.so.44.SAV     
       (( err += $? ))    
       #remove link
       rm -f libtrafodbc_l64.so.1
@@ -163,21 +163,21 @@ function install_lib {
    fi
    #copy library
    err=0
-   cp -f $PKG_DIR/libicuucNv44.so.44  .
+   cp -f $PKG_DIR/libicuuc.so.44  .
    (( err += $? ))
-   chmod 555 libicuucNv44.so.44
+   chmod 555 libicuuc.so.44
    (( err += $? ))
-   rm -f libicuucNv44.so
+   rm -f libicuuc.so
    (( err += $? ))
-   ln -s libicuucNv44.so.44 libicuucNv44.so
+   ln -s libicuuc.so.44 libicuuc.so
    (( err += $? ))
-   cp -f $PKG_DIR/libicudataNv44.so.44  .
+   cp -f $PKG_DIR/libicudata.so.44  .
    (( err += $? ))
-   chmod 555 libicudataNv44.so.44
+   chmod 555 libicudata.so.44
    (( err += $? ))
-   rm -f libicudataNv44.so
+   rm -f libicudata.so
    (( err += $? ))
-   ln -s libicudataNv44.so.44 libicudataNv44.so
+   ln -s libicudata.so.44 libicudata.so
    (( err += $? ))
    cp -f $PKG_DIR/libtrafodbc_l64.so .
    (( err += $? ))
--- a/install/traf_tools_setup.sh
+++ b/install/traf_tools_setup.sh
@@ -229,7 +229,7 @@ cd $BASEDIR
 wget http://download.icu-project.org/files/icu4c/4.4/icu4c-4_4-src.tgz
 tar -xzf icu4c-4_4-src.tgz
 cd icu/source
-./configure --with-library-suffix=Nv44 --prefix=$TOOLSDIR/icu4.4/linux64
+./configure --prefix=$TOOLSDIR/icu4.4/linux64
 make
 make check
 make install


arvind-narain:
Do please check on the dependency site for icu's library suffix. Also will need an update to TOOLSDIR /opt/home/tools accordingly for dev and build systems.

https://cwiki.apache.org/confluence/display/TRAFODION/Additional+Build+Tools

ryzuo:
The apache wiki site still refers Nv44 (while the old wiki.trafodion.org site refers Trf), this must be updated.
Look over all the 3rd party open source libs, ICU is the only one we adds suffix, I don't think this is necessary. And contributors from community might have ICU libs already on their system, and make them have another copy of the libs is not necessary. We better not to use any suffix anymore, although there might be works to update all our dev systems.

hegdean:
With the change in place do u have new instructions for the ICU dependency on build tools page? 
https://cwiki.apache.org/confluence/display/TRAFODION/Additional+Build+Tools

ryzuo:
Sure, but I don't think I have access to update the wiki page...

arvind-narain:
I think that can be updated. But before this change goes in we need to get the development and build environment setup for the change. Also you would need to look at the new scripts that were checked in recently to install dependencies (traf_tools_setup.sh - TRAFODION-1513) and any other installer related scripts. Email to the development community for the upcoming change with instructions would be good - the same can then be used to update the wiki page.

ryzuo:
Update the traf_tools_setup.sh accordingly

svarnau:
Disregard latest test. Inadvertent test was cancelled.

arvind-narain:
Hi Steve, 

Could you please help in setting up icu4.4 so that both prefixes work ( with and without Nv44 ) - thanks for your help. This should help in getting the build and packaging step to succeed.
Steps :
1.      cd $TOOLSDIR/icu4.4/linux64/lib
2.      ln -s libicudataNv44.so.44 libicudata.so.44
3.      ln -s  libicuucNv44.so.44 libicuuc.so.44

svarnau:
Yes, I'm working on this today.  I need to make this change forward compatible and get it to all our test and dev machines.  Please be sure the build instructions get updated on the wiki: https://cwiki.apache.org/confluence/display/TRAFODION/Additional+Build+Tools

svarnau:
jenkins, retest -- icu library naming updated in build/test environment

arvind-narain:
Sorry about that Steve. 

The packaging script expects filename ending with .0. Could you also do the following ? 

Sorry for the trouble and thanks.

1. cd $TOOLSDIR/icu4.4/linux64/lib
2. ln -s libicudataNv44.so.44.0     libicudata.so.44.0
3. ln -s libicuucNv44.so.44.0        libicuuc.so.44.0

svarnau:
Updated. Should propagate fully in about a half hour or so

svarnau:
jenkins, retest -- additional links for .0

arvind-narain:
Thanks Steve - that seems to have worked.

arvind-narain:
Hi RuoYu, 

Could you please work with Steve and check which icu library file the driver is not able to find during runtime ? The location of the files for that particular test is /home/jenkins/workspace/pyodbc_test-cm5.3/trafodion/dcs/src/test/pytests/odbc_driver.

Thanks

arvind-narain:
Hi RuoYu, 

Most probably the workaround of linking to keep both versions may not work, since the soname remains with suffix Nv44. Suggest to generate the icu lib directory utilizing the steps in traf_tools_setup.sh (modified) and move those to $TOOLSDIR/icu4.4/linux64/lib (or it might be possible to generate in existing directory). Please test this out so that both libraries exist and provide steps for Steve to execute.

I looked at the following:
http://userguide.icu-project.org/packaging

Thanks

ryzuo:
The driver links libicudata.so and libicuuc.so, without any .44 nor .0 ending, have you checked that these 2 still have Nv44 suffix?
The symbolic links should be created like below:
ln -s libicudata.so.44.0 libicudata.so
ln -s libicuuc.so.44.0 libicuuc.so

Could you please check that?

arvind-narain:
You might be right RuoYu. Could you please test this out in your environment with and without the suffix and let Steve know all the steps he should be doing. For example:

cd $TOOLSDIR/icu4.4/linux64/lib
rm -f libicudata.so.44.0
rm -f libicudata.so.44
rm -f libicudata.so
ln -s libicudataNv44.so.44.0 libicudata.so.44.0
ln -s libicudata.so.44.0 libicudata.so.44
ln -s libicudata.so.44.0 libicudata.so

rm -f libicuuc.so.44.0
rm -f libicuuc.so.44
rm -f libicuuc.so
ln -s libicuucNv44.so.44.0 libicuuc.so.44.0
ln -s libicuuc.so.44.0 libicuuc.so.44
ln -s libicuuc.so.44.0 libicuuc.so

I think if this is to be done on one system and then propagated, then just doing the configure/make/make check/make install without suffix and getting it installed in the existing target directory may be a safer way. 

Do please test in your environment, suggest steps to Steve and followup on the test failures/success.

Thanks

ryzuo:
Hi Arvind,
We tested this updated traf_tools_setup.sh script to setup a new machine that day I update this PR, it just ran perfectly fine. And Weiqing used the symbolic link to setup and run make on a new machine again, succeeded.  The links were created just like listed above.

The traf_tools_setup.sh has the steps of configure/make/make check/make install already, if you take a look you will the steps. And I just ran it again without the suffix, everything runs good.

PS: the traf_tools_setup.sh needs sudo privilege to run if the TOOLSDIR is set to /opt/home/tools, otherwise make install of all additional tools will fail. It's out of the concern of this discussion anyway, but we better document that in wiki.

Regards

svarnau:
@ryzuo, I can't just run the script and install the result. The test machines must be backward compatible to work with current versions of code.  So the lib directory must have both the old and new library names.  

Currently they have this set up:
lrwxrwxrwx. 1 svarnau svarnau       20 Oct 15 00:25 libicuucNv44.so -> libicuucNv44.so.44.0
lrwxrwxrwx. 1 svarnau svarnau       20 Oct 15 00:25 libicuucNv44.so.44 -> libicuucNv44.so.44.0
-rwxr-xr-x. 1 svarnau svarnau  5968571 Dec  7  2013 libicuucNv44.so.44.0
lrwxrwxrwx. 1 svarnau svarnau       20 Oct 15 00:25 libicuuc.so -> libicuucNv44.so.44.0
lrwxrwxrwx. 1 svarnau svarnau       20 Oct 15 00:25 libicuuc.so.44 -> libicuucNv44.so.44.0
lrwxrwxrwx. 1 svarnau svarnau       20 Oct 15 00:25 libicuuc.so.44.0 -> libicuucNv44.so.44.0

Why won't this work?

xwq:
@svarnau Do you create the link libicudata.so too ? If you give the error message , we can find the reason. We can't just guess the reason.
Bye the way RuoYu's solution works for me.

svarnau:
The latest test got further than previous. The test log shows:
('01000', "[01000] [unixODBC][Driver Manager]Can't open lib '/home/jenkins/workspace/pyodbc_test-cm5.3/trafodion/dcs/src/test/pytests/odbc_driver/libtrafodbc_drvr64.so' : file not found (0) (SQLDriverConnect)")

It looks to me like something is looking for the shared lib in the location were it was built, and not the location where odbc driver was installed.

svarnau:
No, I don't have that quite right. it is not built in that location, but the pyodbc tests set it up there. Not sure exactly why it gives error, perhaps there is something in that test script that needs to change as well.  Give the pyodbc tests a try to see if you can reproduce it.

arvind-narain:
@svarnau We could do an ldd on libtradodbc_drvr64.so file - that will show if it is looking for *Nv44 suffix still.

@svarnau Could you try out the set of commands I gave earlier ? It does maintain backward compatibility. Previous version just looks for the following and that is not being touched. We are adding another link to *Nv44.so.0 files so that we have the same setup for *icuuc.so and *icudata.so. This setup was tested by @ryzuo and @xwq as per the discussion earlier. 

lrwxrwxrwx. 1 svarnau svarnau       22 Oct 15 00:19 libicudataNv44.so -> libicudataNv44.so.44.0
lrwxrwxrwx. 1 svarnau svarnau       22 Oct 15 00:19 libicudataNv44.so.44 -> libicudataNv44.so.44.0
-rw-r--r--. 1 svarnau svarnau 14924889 Dec  7  2013 libicudataNv44.so.44.0
lrwxrwxrwx. 1 svarnau svarnau      20 Oct 15 00:19 libicuucNv44.so -> libicuucNv44.so.44.0
lrwxrwxrwx. 1 svarnau svarnau      20 Oct 15 00:19 libicuucNv44.so.44 -> libicuucNv44.so.44.0
-rwxr-xr-x. 1 svarnau svarnau 5968571 Dec  7  2013 libicuucNv44.so.44.0

@ryzuo and @xwq  you can look at the log files via - test name->Console Output -> click on detailed logs link - that basically has all the logs and relevant information saved. Under that you'll have logs where this particular test stores the files used and output.

http://traf-logs.esgyn.com/PullReq/96/23/pyodbc_test-cm5.3/
http://traf-logs.esgyn.com/PullReq/96/23/pyodbc_test-cm5.3/logs/

arvind-narain:
Basically need something similar to the following (4th and 5th line edited):
Similar for libicudata.
lrwxrwxrwx. 1 svarnau svarnau 20 Oct 15 00:25 libicuucNv44.so -> libicuucNv44.so.44.0
lrwxrwxrwx. 1 svarnau svarnau 20 Oct 15 00:25 libicuucNv44.so.44 -> libicuucNv44.so.44.0
-rwxr-xr-x. 1 svarnau svarnau 5968571 Dec 7 2013 libicuucNv44.so.44.0
lrwxrwxrwx. 1 svarnau svarnau 20 Oct 15 00:25 libicuuc.so -> libicuuc.so.44.0
lrwxrwxrwx. 1 svarnau svarnau 20 Oct 15 00:25 libicuuc.so.44 -> libicuuc.so.44.0
lrwxrwxrwx. 1 svarnau svarnau 20 Oct 15 00:25 libicuuc.so.44.0 -> libicuucNv44.so.44.0

ryzuo:
Hi, Arvind and Steve

If you look into the make.log(close to bottom of the entire log), you will see the odbc driver has already been made succesfully:
http://traf-logs.esgyn.com/PullReq/96/23/build-master-debug/Make.log

So this failure is not related to this change.

I went to take a look at the test install log:
http://traf-logs.esgyn.com/PullReq/96/23/pyodbc_test-cm5.3/console.html

And I noticed the driver is install at /home/jenkins/workspace/pyodbc_test-cm5.3/clients/odbc", but if you look into the odbcinst.ini configuration, you will see the driver dir is specified at "/home/jenkins/workspace/pyodbc_test-cm5.3/trafodion/dcs/src/test/pytests/odbc_driver/". The setup of  the test environment is incorrect.

Could you please double check the test setup scripts and the machine? I don't see this is in git repo, this is what I can tell just from the log.

Thanks& Regards
RuoYu

arvind-narain:
Hi RuoYu,

This test is working fine in daily builds. Do check the console output where a link is established. File not found error seen in the test logs I believe is due to some missing .so file - in this case *Nv44* - my guess - we could do a ldd to confirm. Should be easy to simulate if you use the current files on /opt/home/tools.

+ cd /home/jenkins/workspace/pyodbc_test-cm5.3/trafodion/dcs/src/test/pytests
+ mkdir /home/jenkins/workspace/pyodbc_test-cm5.3/trafodion/dcs/src/test/pytests/logs
+ rm -rf /home/jenkins/workspace/pyodbc_test-cm5.3/trafodion/dcs/src/test/pytests/odbc_driver
+ ln -s /home/jenkins/workspace/pyodbc_test-cm5.3/clients/odbc /home/jenkins/workspace/pyodbc_test-cm5.3/trafodion/dcs/src/test/pytests/odbc_driver

ryzuo:
Hi Arvind and Steve

I use unixODBC, which is used by pyodbc, to simulated this on my box, it just runs fine, I don't see the error. Could you please check is there's only one copy of ICU libs in the Jenkins box?

ryzuo:
Sorry... mouse jumping......

Could you please make sure the driver is installed successfully? And also please double check is the ICU libs are in the LD_LIBRARY_PATH, and also, please make sure there's no other copy messes it up.

I ldd checked the lib file, just everything it needs are all in:
[zuor@zuor-centos-v64 install]$ ldd libtrafodbc_drvr64.so
        linux-vdso.so.1 =>  (0x00007ffcb9065000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f6420a4d000)
        librt.so.1 => /lib64/librt.so.1 (0x00007f6420844000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007f6420640000)
        libz.so.1 => /lib64/libz.so.1 (0x00007f642042a000)
        libicuuc.so.44 => ./libicuuc.so.44 (0x00007f64200cd000)
        libstdc++.so.6 => /usr/lib64/libstdc++.so.6 (0x00007f641fdc7000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f641fb43000)
        libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007f641f92c000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f641f598000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f64211b2000)
        libicudata.so.44 => ./libicudata.so.44 (0x00007f641e55c000)

I don't have pyodbc on hand by now, so I compiled the connect_test.cpp sample file with unixODBC Driver Manager, which is used by pyodbc:
g++ -g -L./ -lodbc -o test connect_test.cpp
and ran a test, anyway, everything runs fine on my system... I have no idea what went wrong on Jenkins... It still might be some env setup issue.

hegdean:
pyodbc tests are part of the incubator-trafodion tree, You can find the pyodbc tests in incubator-trafodion/dcs/src/test/pytests. It is very simple to run.  There is a README that provide instructions on how to run it else if you have updated your branch to pick the latest codebase helper scripts are provided to run the test in your environment(swpyodbc)

arvind-narain:
I think your environment for $TOOLSDIR files is a bit different than others,possibly because of the setup script run earlier. Do look at the current env posted by Steve and the link I posted for icu packaging - what .so file information is used at compile and link time. Current env we are still getting Nv44 when -licuuc is done. Either for build and runtime env we need to do the steps I mentioned earlier for linking or we need to install without suffix so that both versions are available in the target directory linux64/lib

-----Original Message-----

svarnau:
jenkins, retest -- I've changed the symlinks per Arvind's suggestion.

svarnau:
jenkins, retest -- yet another attempt to make icu4.4 forward/backward compatible

arvind-narain:
last attempt was with generation of icu library files without suffix. This one failed because of incompatible glibc between runtime system (glibc 2.12) and where icu was compiled (glibc 2.15). Steve will be recompiling and installing icu. Thanks a lot Steve for your help and patience.

svarnau:
jenkins, retest -- icu rebuild done and propagated again

arvind-narain:
@ryzuo - Do please see how we can maintain backward compatibility with the change submitted. Currently because we have the include files from Nv44 suffix and the lib from the default icu the generated driver is giving error while compiling ( connect_test.cpp ) or running ( via python ). Do check if introducing some env variable will help or some other way we can maintain this backward compatibility.

svarnau:
Perhaps a different path instead of TOOLSDIR/icu4.4 would work.  If the new non-suffix libs were in different location?

arvind-narain:
@ryzuo Could you please work on Steve's suggestion and make the corresponding change?  Let Steve know the new location under TOOLSDIR so that he can install icu without prefix. The location of icu installation is specified in sqenvcom.sh and LocalSettingsTemplate.sh via ICU - it's current value is $TOOLSDIR/icu4.4.

Do also check on the document source Trafodion_Client_Installation_Guide.xml for removing the suffix.

arvind-narain:
@ryzuo do please comment on any progress made on this pull request - thanks.

ryzuo:
@arvind-narain We realized, that might because the functions of ICU have suffix Nv44 as well while the libs were built with the suffix option... Do please double check the .so files with nm, they might have Nv44 suffix. Thus, soft link does not work, you might need to build the ICU libs without suffix option to make it work. My system has 2 editions coexist when I tested that.
capture

svarnau:
@ryzuo We did that build without the suffix on the last go-around, but did have to put them in the same directory ($TOOLSDIR/icu4.4/linux64/lib).  It did not work, which is why I suggested using a directory path different than icu4.4, so that both can co-exist completely separately.

sureshsubbiah:
@ryzuo could you please comment on Steve's suggestion? Even if it worked on our individual test environments, when a change does not work in Steve's official build environment we may not be able to merge it. You could get temporary access to the build machine from Steve and look there if that will help. Thank you for contributing this change and your patient responses to all questions.

arvind-narain:
@ryzuo Do please update this pull request with any latest changes that you might have done/tested to address/follow Steve's suggestions. Thanks.

ryzuo:
Hi,

I don't have any updates for this PR. Actually I don't think there's necessary that we keep them coexist. The purpose of this change is eliminate the old libs with Nv44 suffix. Although they coexist on my dev machine, but I never used the old ones since I changed this. To make them coexist, we need two edition of the make file and packaging script? Or introduce some env var allow developers to switch between the new and the old? That will be more confusing... So my opinion is we must update our dev environment. Since the traf_tools_setup.sh script is updated as well, I think this won't be a difficult thing to do. Some contributors I know in Shanghai have changed this and keep using it for a while, and run smoothly.

If you really have difficult to update your environments, I think I will have no choice but close this PR...

svarnau:
Hi @ryzuo, I think there is a mis-communication on the point of co-existence of the new/old libraries. As you say there is no need for the new version of code to be able to switch back and forth. There is need, however, for old versions of code to be able to be built/tested on the same machines. 

This is necessary for the transition period so that every workspace on a development machine does not have to update at the same time. More importantly, it is also needed to be able to support an old release of code.

We went through several rounds of trying to have both sets of libraries in the same directory, but a cleaner way will be to just have the new libraries at a different path. Then when the new change comes in it changes where we look for ICU libraries, such as TOOLSDIR/icu44_2.  That enables the old code to coexist on the same machine.

sandhyasun:
Need to move this one along to the next step. Looks like some rework is needed here from what Steve suggests. Is it work in progress ?

PR #98, opened 2015-10-01
TRAFODION-1502 Build remaining Java files in SQL executor and ustat through Maven

zellerh:
Most of the Trafodion Java source files are built through Maven, using
projects DCS, REST, HBase-trx and SQL. A few files remain in the
core/sql/executor and core/sql/ustat directories that are built
through javac commands in the Makefile and helper files. Also, these
files don't conform to common naming standards and therefore are a bit
harder to debug with tools like Eclipse.

This change does the following:

1. Move these remaining Java source files under core/sql/src,
   where the UDR-related files already are.
2. Include the class files in trafodion-sql-1.2.0.jar, the
   deliverable of the core/sql Maven project.
3. Change the package name to org.trafodion.sql for all these files.
4. Remove the trafodion-HBaseAccess-1.2.0.jar file (content
   will be included in trafodion-sql-1.2.0.jar). This jar file
   used to be built with a Makefile.
5. Make some source code changes to reflect the new package names
   and to avoid calling private method of class
   org.apache.hadoop.hive.ql.io.orc.OrcStruct.
6. Remove Makefile rules for compiling Java files.

Also, this change includes a change of the component name for the
event logger. We use SQL.COMP now, to make analyzing log messages
for Trafodion easier. Venkat asked for this change.


DaveBirdsall:
Looks good to me. Would be nice to get another pair eyes on this one.

zellerh:
Thank and I'm hoping Roberta can take a look, we discussed this change.

robertamarton:
Looks good to me also.

PR #99, opened 2015-10-02
Test130 modifications to get it to work with installer setup of hdfs …

sandhyasun:
…directories. Added some extra cleanup that caused test to fail if run multiple times.


robertamarton:
Looks good to me

sureshsubbiah:
Looks good. Thank you. Merging the change.

PR #100, opened 2015-10-02
Fix JIRA 1503

nonstop-qfchen:
This is to fix JIRA 1503. 

The code change involves rewinding the ComDiags area including the maxDiagsId_ data member. Without doing so, maxDiagsId_ becomes out of sync with # of errors and # of warnings and lead to the test "da->getNumber() != daMark" to fail. With the fix, da->mark() call returns a value of 0, a good starting point to detect any potential errors. 

GenExpGenerator.cpp:

  daMark = da->mark();
  ex_expr::exp_return_type evalReturnCode = defaultValueIdList.evalAtCompileTime
    (0,
     ExpTupleDesc::PACKED_FORMAT,
     defaultValue, attr->getDefaultValueStorageLength() + extraFltLen,
     &length, &offset, da);

  if ((evalReturnCode != ex_expr::EXPR_OK) || (da->getNumber() != daMark))


robertamarton:
Thanks for changing this, it looks fine.  I use this same mechanism in privilege manager code so I will check to see if a similar change is required.

PR #101, opened 2015-10-04
In regression /executorTEST130, swhadoop needs to be replaced with regrhadoop.ksh script. If not it f…

sandhyasun:
…ails on official nightly build machines that do not have swhadoop helper scripts.


sureshsubbiah:
Change looks good. Merging.

PR #104, opened 2015-10-05
TRAFODION-1504 check for environment in sqgen

zellerh:
Avoid the common mistake of doing an sqgen in an environment that
is not clean, where changes to CLASSPATH or other environment variables
have been done after sourcing sqenv.sh.

This commit also fixes TRAFODION-1505, delete the sw* convenience scripts
created by install_local_hadoop when we do swuninstall_local_hadoop.


sureshsubbiah:
Looks good. Thank you Hans.

DaveBirdsall:
Bless you for making this change. You'll save many wrong turns in the future.

PR #105, opened 2015-10-06
Copyright changes and cleanup of obsolete files in dcs to get prepared for a clean RAT report (sandhya copyright2)

sandhyasun:
These changes consist of  adding new headers to some files that RAT flagged as unlicensed  headers , lots of cleanup in odbc and dcs directories by Arvind.
One this is merged, will add a script and list of exclusion files so we cna run RAT on our source tree either automated or on a regular basis .


DaveBirdsall:
Looks good to me. I take it a lot of the removed ODBC stuff is old NSK-isms that are no longer used?

PR #107, opened 2015-10-06
[TRAFODION-1482] [TRAFODION-1474] [TRAFODION-1279]

sureshsubbiah:
Please see JIRA issues for information on each fix.
The files are
1482: HTableClient.java
1474: GenUdr.cpp
1279:NaColumn.h, NAColumn.cpp, CmpSeabaseDDLtable.cpp

### DIFF
--- a/core/sql/optimizer/NAColumn.cpp
+++ b/core/sql/optimizer/NAColumn.cpp
@@ -749,6 +749,27 @@ Lng32 NAColumnArray::getOffset(Lng32 position) const
 
   return result;
 }
+
+Lng32 NAColumnArray::getMaxTrafHbaseColQualifier() const
+{
+  NAColumn *column;
+  char * colQualPtr;
+  Lng32 colQualLen;
+  Int64 colQVal;
+  Lng32 maxVal = 0;
+
+  for (CollIndex i = 0; i < entries(); i++)
+    {
+      column = (*this)[i];
+      colQualPtr = (char*)column->getHbaseColQual().data();
+      colQualLen = column->getHbaseColQual().length();
+      colQVal = str_atoi(colQualPtr, colQualLen);
+      if (colQVal > maxVal)
+	maxVal = colQVal ;
+    }
+  return colQVal;
+}
+
 // LCOV_EXCL_STOP
 
 //method to reset an NAColumn object after a statement
--- a/core/sql/optimizer/NAColumn.cpp
+++ b/core/sql/optimizer/NAColumn.cpp
@@ -767,7 +767,7 @@ Lng32 NAColumnArray::getMaxTrafHbaseColQualifier() const
       if (colQVal > maxVal)
 	maxVal = colQVal ;
     }
-  return colQVal;
+  return maxVal;
 }
 
 // LCOV_EXCL_STOP
--- a/core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
+++ b/core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
@@ -4596,6 +4596,8 @@ void CmpSeabaseDDL::alterSeabaseTableAddColumn(
   char * col_name = new(STMTHEAP) char[colName.length() + 1];
   strcpy(col_name, (char*)colName.data());
 
+  Lng32 maxColQual = nacolArr.getMaxTrafHbaseColQualifier();
+
   NAString quotedHeading;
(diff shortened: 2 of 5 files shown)


core/sql/optimizer/NAColumn.cpp
```diff
@@ -749,6 +749,27 @@ Lng32 NAColumnArray::getOffset(Lng32 position) const
...
+  for (CollIndex i = 0; i < entries(); i++)
+    {
+      column = (*this)[i];
+      colQualPtr = (char*)column->getHbaseColQual().data();
+      colQualLen = column->getHbaseColQual().length();
+      colQVal = str_atoi(colQualPtr, colQualLen);
+      if (colQVal > maxVal)
+	maxVal = colQVal ;
+    }
+  return colQVal;
```
zellerh: Shouldn't this return maxVal? It probably works with colQVal, since the column qualifiers are in ascending order. Also, what about initializing maxVal with -1? We don't support tables with no columns, so this hair-splitting and a theoretical  argument.

core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
```diff
@@ -4609,7 +4611,7 @@ void CmpSeabaseDDL::alterSeabaseTableAddColumn(
     }
 
   Int64 objUID = naTable->objectUid().castToInt64();
-  str_sprintf(query, "insert into %s.\"%s\".%s values (%Ld, '%s', %d, '%s', %d, '%s', %d, %d, %d, %d, %d, '%s', %d, %d, '%s', %d, '%s', '%s', '%s', '%d', '%s', '%s', %Ld )",
+  str_sprintf(query, "insert into %s.\"%s\".%s values (%Ld, '%s', %d, '%s', %d, '%s', %d, %d, %d, %d, %d, '%s', %d, %d, '%s', %d, '%s', '%s', '%s', '%u', '%s', '%s', %Ld )",
```
zellerh: Shouldn't this be %d?

zellerh:
Thanks for fixing these, and I have the feeling that I probably broke the EXPLAIN for UDRs, sorry about that.

core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
sureshsubbiah: This was intentional. We are using this value as an unsigned int in the generator. I athought we might get into some kind of overflow situation if the number of columns is in between signed_int_max and unsigned_int_max.

Code snippet from HbaseAccess:createHbaseColId()
 Int64 colQval = str_atoi(colQualPtr, colQualLen);
      
      if (colQval <= UCHAR_MAX)
	{
	  unsigned char c = (unsigned char)colQval;
	  cid.append((char*)&c, 1);
	}
      else if (colQval <= USHRT_MAX)
	{
	  unsigned short s = (unsigned short)colQval;
	  cid.append((char*)&s, 2);
	}
      else if (colQval <= ULONG_MAX)
	{
	  Lng32 l = (Lng32)colQval;
	  cid.append((char*)&l, 4);
	}

zellerh: It's just that maxColQual is defined as a Lng32 on line 4599 and that is an int, not an unsigned int.

core/sql/optimizer/NAColumn.cpp
sureshsubbiah: Thanks for catching this. I will fix the return value in a minute. The smallest colQualifier we add is 1 for a Traf table. We don't seem to add colQualifier with value 0. Returning a negative value will mean that it will have to be checked by the caller.

sureshsubbiah:
Thank you for catching my basic mistake Hans.

core/sql/optimizer/NAColumn.cpp
zellerh: Thanks, sounds great.

DaveBirdsall:
Changes look good to me.

core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
sureshsubbiah: You are right Hans. I misread the line in Platform.h. Let me fix this in my next checkin.

sureshsubbiah:
Thank you for the review Hans, Dave.

PR #109, opened 2015-10-08
[TRAFODION-1513] Trafodion build from source tar failing

robertamarton:


DaveBirdsall:
Looks good to me.

sureshsubbiah:
Looks good to me too. Will wait some more to see if others would like to review.

core/sqf/sqenvcom.sh
```diff
@@ -95,17 +95,20 @@ cpucnt=$(grep processor /proc/cpuinfo | wc -l)
...
 export MAKEFLAGS="-j$cpucnt"
 
+# For now set up the TOOLSDIR, it may be overwritten later when the
+# .trafodion configuration file is loaded.
 if [[ -n "$CLUSTERNAME" ]]; then
-  export MY_ROOT=/opt/hp
-  export TOOLSDIR=${TOOLSDIR:-/home/tools}
+  if [ -z "$TOOLSDIR" ]; then
+    export TOOLSDIR=${TOOLSDIR:-/home/tools}
+  fi
```
svarnau: Setting TOOLSDIR here seems to be doing same thing as before, but more complicated (3 lines instead of one).  If we are explicitly checking for zero-length, then we might as well use the simple assignment:
   export TOOLSDIR=/home/tools

core/sqf/sqenvcom.sh
```diff
@@ -795,13 +799,9 @@ fi
 
 # Non-standard or newer version tools
 export BISON="${TOOLSDIR}/bison_3_linux/bin/bison"     # Need 1.3 version or later
-
-
 export LLVM="${TOOLSDIR}/dest-llvm-3.2"
 export UDIS86="${TOOLSDIR}/udis86-1.7.2"
 export ICU="${TOOLSDIR}/icu4.4"
-export QT_TOOLKIT="${TOOLSDIR}/Qt-4.8.5-64"
-
```
svarnau: Not setting QT_TOOLKIT here means compiler debugger it is no longer included in the standard builds, including jenkins builds?

core/sqf/sql/scripts/install_hadoop_regr_test_env
```diff
@@ -134,12 +134,12 @@ fi
...
-  echo "environment variable. Then retry installing. Sorry, this is due"
-  echo "to the TPC wanting your email address."
+  # Please download the tpcds_kit.zip file from"
+  # "http://www.tpc.org/tpcds/dsgen-download-request.asp"
+  # "and place it into a directory pointed to by the MY_LOCAL_SW_DIST"
+  # "environment variable. Then retry installing. Sorry, this is due"
+  # "to the TPC wanting your email address."
+  echo "The testware tpcds_kit.zip does not exist and will not be installed"
   exit 99
 fi
```
svarnau: Might want to give message about the implications. The hive tests (and one core test) will not work without this.

install/traf_tools_setup.sh
```diff
@@ -0,0 +1,249 @@
...
+cd $BASEDIR
+wget http://archive.apache.org/dist/thrift/0.9.0/thrift-0.9.0.tar.gz
+tar -xzf thrift-0.9.0.tar.gz
+cd thrift-0.9.0
+./configure --prefix=$TOOLSDIR/thrift-0.9.0 -without-qt
+make
+make install
+if [ $VERBOSE -eq 1 ]; then
+  echo "STEP completed:  installation of THRIFT"
+fi
```
svarnau: Really good to have this script. I hope we keep it up to date as we migrate to new versions!

svarnau:
All minor comments, except I'm concerned whether the QT variable change removes compiler debugger from check,daily, and release builds.

core/sqf/sql/scripts/install_hadoop_regr_test_env
zellerh: Also, unfortunately, the TPC rewrote their test kit and we would need to update our build procedure to use the new files. So, following these instructions will actually no longer work. We are all using an old, saved copy of tpcds_kit.zip.

core/sqf/sql/scripts/install_local_hadoop
```diff
@@ -925,8 +933,10 @@ else
 
   if [ -f $MY_LOCAL_SW_DIST/${HADOOP_TAR} ]; then
     cp $MY_LOCAL_SW_DIST/${HADOOP_TAR} .
+    echo "Downloaded Hadoop tar file: $MY_LOCAL_SW_DIST/${HADOOP_TAR}"
```
zellerh: Given that we didn't download anything, should this message say "copied" instead of "downloaded"? Same for the following ones.

core/sqf/sql/scripts/install_local_hadoop
```diff
@@ -565,8 +572,9 @@ fi
 
 # check for missing tpcds_kit.zip file
 install_hadoop_regr_test_env --check
+INSTALL_TPCDS=0
 if [ $? -ne 0 ]; then
-  exit 1
+  INSTALL_TPCDS=1
```
zellerh: I would have expected that we set INSTALL_TPCDS to 0 here, since we do not want to attempt the install in this case.

install/traf_tools_setup.sh
```diff
@@ -0,0 +1,249 @@
...
+if [ "$TOOLSDIR" == "" ]; then
+  echo
+  echo "ERROR: install directory (-t)  is not specified"
+  Usage;
+  exit 1;
+fi
+
+if [ ! -d "$TOOLSDIR" ]; then                                                    
+  echo
+  echo "ERROR: install directory ($TOOLSDIR) does not exist"
```
zellerh: Since this sets up $TOOLSDIR, should we create it here?

core/sqf/sqenvcom.sh
robertamarton: No, this prevents the TOOLSDIR from being overwritten to either /opt/home/tools or /home/tools.  On a VM or a generic system, the tools directory does not exist. If you invoke sqenvcom.sh after setting your own TOOLSDIR envvar and don't create the .trafodion file, your tools directory variable is overwritten and a subsequent build fails.

zellerh: I think what Steve is saying that the old syntax

export TOOLSDIR=${TOOLSDIR:-/home/tools}

already does that, it does nothing if TOOLSDIR is already set.

core/sqf/sqenvcom.sh
zellerh: Agree, could we leave this in? If we skip the GUI for the official Apache build script that's fine, but Jenkins and developers should continue to build it.

robertamarton: Is there a requirement to build the QT_TOOLKIT?  I asked around and it seemed to be an optional product. This would be one less dependency that the user has to provide.  Given that I was unable to get QT to build on a VM, (it kept wanting more and more dependencies which I could not find).   I assumed that if someone wants this tool, then we can provide instructions on how to build it.  I can add a section on the build page to discuss how to build and use this tool.

zellerh: That's fine, but Jenkins and most developers routinely build the GUI on our machines today, and that should continue to work. We use the DISPLAY command (invoking the GUI) all the time to debug queries.

core/sqf/sql/scripts/install_hadoop_regr_test_env
robertamarton: Most of the contributors that build Trafodion code are expecting things to work without change including the running of all regressions.  To do this, a repository has been created that contains everything these contributors need.  However, not all contributors have this luxury.  As Hans pointed out, other contributors, cannot even follow the directions to download this software.  That said, our regression tests require this to be installed - at least our Hive tests.  Let me write a JIRA to on this.  In the meantime, I can display something to the effect that some of the developer regressions will not pass.

core/sqf/sql/scripts/install_local_hadoop
robertamarton: Good point!

install/traf_tools_setup.sh
robertamarton: I did remove QT, Protocol buffers, and Maven from this list.  I was thinking about adding it back in. If "protoc --version" does not produce the correct version then install it. If "mvn --version" does not produce the correct version than install it.  However, I am not sure it is worth the effort.  I assume that most system have these packages installed.  What do you think?

zellerh: It it's not too much work it would be good to add it, but I would not say that it's required. Given how widespread the use of Maven is in Hadoop, we can probably expect people to have it.

Protoc is a real trouble maker (at least for me). I'm going to check in a script that downloads Hadoop sources and builds libhdfs. This script also builds a local copy of protoc 2.5.0, since the default for CentOS (2.3.0) is too old. Maybe I can remove that if this script sets up protoc 2.5.0.

Qt would be very good to have in this setup, since it's mostly meant for developers, who will need the compiler GUI. But, again, it's not required. Maybe Howard can help us set this up.

svarnau: If a "good enough" version can be installed from standard linux distro, then I think we should prefer that. This script should concentrate on what is not in the standard distro.

core/sqf/sqenvcom.sh
robertamarton: Okay, I see what Steve is saying, I can simplify the code

core/sqf/sqenvcom.sh
robertamarton: Let me play with it.  It will need additional script changes.

install/traf_tools_setup.sh
robertamarton: I am able to build with protoc version 2.3.0.  Apparently you need 2.5.0 to build libhdfs...  I will write a JIRA and make these changes in a future delivery -  In build instructions describe how to install later versions of protocol buffers and QT.  I can also update the script to install version 2.5.0 of protocol buffers.

core/sqf/sql/scripts/install_local_hadoop
robertamarton: You are correct.  I thought I tested this version of the script but apparently not. Will fix it.

install/traf_tools_setup.sh
robertamarton: Probably.  I can create it if it doesn't exit.

robertamarton:
I will push some more changes that will:
   Simplify the TOOLSDIR check in sqenvcom.sh
   See if there is a simple solution to fixing the QT issue
   Mention that HIVE regression fail if TPC is not installed
   Fix the INSTALL_TPC check in install_local_hadoop and rework the output text
   Add a mkdir in the script if the tools directory does not exist

Create 2 JIRAs
- fix the dependency for TPC in order to run HIVE tests
- Add installation of QT and protobufs in the traf_tools_setup.sh script

install/traf_tools_setup.sh
zellerh: Good to know, but for now we are probably fine. I can handle the protobuf requirement inside my script for now - or we can move the code to your script (and also update all of our VM images).

core/sqf/sqenvcom.sh
zellerh: In my next checkin, I am planning to remove the checks for QT_TOOLKIT from the $MY_SQROOT/../bldenvchk.sh script. I'll also remove the check for the Hadoop lib and include directories.

If you want you can take out the QT_TOOLKIT from sqenvcom.sh and I can put it back in once I updated bldenvchk.sh.

robertamarton:
I just pushed the requested.  I did not get any e-mail message for this push but it does show up in the pull request as a separate commit ...

robertamarton:
I reworked changes for building, can this request now be merged?



    Roberta



*From:* Trafodion Automated Test [mailto:notifications@github.com]
*Sent:* Thursday, October 8, 2015 8:19 PM
*Subject:* Re: [incubator-trafodion] [TRAFODION-1513] Trafodion build from
source tar failing (#109)



Test Passed.
https://jenkins.esgyn.com/job/Check-PR/200/

—
<https://github.com/apache/incubator-trafodion/pull/109#issuecomment-146745263>
.

svarnau:
Looks good to me.

DaveBirdsall:
Looks good to me too. Should we wait for Hans to take another look?

zellerh:
Looks good to me as well. Good to see the build process getting simpler!

DaveBirdsall:
I will merge it.

PR #112, opened 2015-10-09
[TRAFODION-1516] A trafodion infrastructure program mxssmp dumps cor…

selvaganesang:
…e at times

SSMP was dumping core sometimes when the DCS MASTER tries to cancel a query using pid.
When the pid is more than 32767,  the executor was not handling it correctly resulting
in negative number for pid. This resulted in partial junk message being sent to SSMP.
The junk message causes ssmp to dump core sometimes.


DaveBirdsall:
Looks good to me. I'm happy to commit it.

sureshsubbiah:
Please proceed Dave. I see nothing that can be improved.

PR #118, opened 2015-10-13
TRAFODION-1521 Build Trafodion without having HBase installed

zellerh:
Adding a new script, core/sqf/sql/scripts/get_libhdfs_files. The new
script will download Google Protocol Buffers 2.5.0 and a source tar
file for Hadoop and build protobuf-2.5.0 and the Hadoop native
libraries. It will then copy the libraries (libhadoop.so and
libhdfs.so) to $MY_SQROOT/export/lib$ {SQ_MBTYPE}. It will also copy
the include file hdfs.h to $MY_SQROOT/include. Since this step is very
time-consuming, it takes a shortcut if Hadoop is installed
(e.g. with install_local_hadoop), in that case it will copy the
existing native libraries.  To address the remaining needs for jar
files during the build, the hbasetmlib2 build is converted to
Maven.  We also remove the case in sqenvcom.sh where a
configuration does not have install_local_hadoop, Cloudera,
Hortonworks or MapR installed, but has a TOOLSDIR. Such cases should
use the recently added case for a native Apache Hadoop installation or
the case where no HBase installation is found. That last case will
print out a reminder that can be ignored.

Note: Users will need to do "sqgen" to update their classpath. This is
because we need a new jar in the classpath: trafodion-dtm-1.2.0.jar.


core/sqf/sql/scripts/get_libhdfs_files
```diff
@@ -0,0 +1,186 @@
...
+       if [[ $# < 1 ]]; then
+         echo "Expecting argument after -d or --tempDir"
+         exit 1
+       fi
+       LIBHDFS_TEMP_DIR="$1"
+       ;;
+
+    *)
+       echo "Unknown command line option: $arg"
+       echo "Usage $0 [ -f | --force ]"
```
DaveBirdsall: Did you mean to echo $USAGE here?

core/sqf/sql/scripts/get_libhdfs_files
```diff
@@ -0,0 +1,186 @@
...
+  fi
+
+  if [[ $FORCE_BUILD == true ]]; then
+    rm -rf ${LIBHDFS_TEMP_DIR}/${PROTOBUF_ID}
+    rm -rf ${LIBHDFS_TEMP_DIR}/${PROTOBUF_TGT_ID}
+  fi
+
+  if [[ ! -d ${PROTOBUF_ID} ]]; then
+    echo "Unpacking Google Protocol Buffer tar file..." | tee -a ${LOGFILE}
+    rm -rf ${LIBHDFS_TEMP_DIR}/${PROTOBUF_TGT_ID}
```
DaveBirdsall: Might get an error on this if you went through the code path above that also does this "rm".

core/sqf/sql/scripts/get_libhdfs_files
```diff
@@ -0,0 +1,186 @@
...
+      exit 1
+    fi
+    make 2>&1 >>${LOGFILE}
+    if [[ $? != 0 ]]; then
+      echo "Error during make step, exiting" | tee -a ${LOGFILE}
+      exit 1
+    fi
+    # skip the tests
+    # make check 2>&1 >>${LOGFILE}
+    if [[ $? != 0 ]]; then
```
DaveBirdsall: Probably should comment out this "if" block too since the "make check" is commented out. Though it is harmless (the "if" should always be false).

core/sqf/sqenvcom.sh
```diff
@@ -580,6 +547,13 @@ EOF
...
   fi
 
+  if [[ -d $TOOLSDIR/thrift-0.9.0 ]]; then
+    # this is mostly for a build environment, where we need
+    # thrift from TOOLSDIR
+    export THRIFT_LIB_DIR=$TOOLSDIR/thrift-0.9.0/lib
+    export THRIFT_INC_DIR=$TOOLSDIR/thrift-0.9.0/include
+  fi
+
   if [ -n "$HBASE_CNF_DIR" -a -n "$HADOOP_CNF_DIR" -a \
```
robertamarton: You removed Thrift from the required list in bldenvchk and set the variable only if Thrift is installed in the TOOLSDIR.  What happens if Thrift is not installed?

core/sql/nskgmake/Makerules.linux
```diff
@@ -80,20 +80,14 @@ OBJSUFFIX := o
...
 #######################################################################
-NSK_SQ    := $(MY_SQROOT)
-NSK       := $(MY_SQROOT)/sql/nsk
-XMPIROOT        = $(NSK_SQ)/export/lib$(SQ_MBTYPE)
-NSKBIN		= $(NSK_SQ)/export/bin$(SQ_MBTYPE)
-NSKROOT         = $(NSK_SQ)/sql/nsk
+LIBROOT         = $(MY_SQROOT)/export/lib$(SQ_MBTYPE)
+TRAF_BIN	= $(MY_SQROOT)/export/bin$(SQ_MBTYPE)
+XMPIROOT        = $(LIBROOT)
 ifeq ($(SQ_BUILD_TYPE),release)
```
DaveBirdsall: Glad to see the NSK stuff cleaned up.

DaveBirdsall:
Looks good to me. All of my comments are minor.

core/sqf/sql/scripts/get_libhdfs_files
```diff
@@ -0,0 +1,186 @@
+#!/bin/sh
+
+# This script downloads and/or makes the required libhdfs files
```
robertamarton: Do we need the Apache header?

core/sqf/sql/scripts/get_libhdfs_files
```diff
@@ -0,0 +1,186 @@
...
+  LIBHDFS_TEMP_DIR=${MY_SQROOT}/sql/libhdfs_files
+fi
+LOGFILE=${LIBHDFS_TEMP_DIR}/build.log
+
+# Hadoop source tar file to build libhdfs from
+HADOOP_SRC_MIRROR_URL=https://archive.apache.org/dist/hadoop/common/hadoop-2.6.0
+HADOOP_ID=hadoop-2.6.0
+HADOOP_SRC_ID=${HADOOP_ID}-src
+HADOOP_SRC_TAR=${HADOOP_SRC_ID}.tar.gz
+
```
robertamarton: We have been using 2.5.? versions.  Is 2.6.0 compatible with HBase .98?

core/sqf/sql/scripts/get_libhdfs_files
```diff
@@ -0,0 +1,186 @@
...
+# files to build required version of Google Protocol Buffers
+PROTOBUF_MIRROR_URL=https://github.com/google/protobuf/releases/download/v2.5.0
+PROTOBUF_ID=protobuf-2.5.0
+PROTOBUF_TAR=${PROTOBUF_ID}.tar.gz
+
+# result of protobuf build
+PROTOBUF_TGT_ID=protobuf-tgt
+
```
robertamarton: Would it be useful to do :  protoc --version and if the default version is 2.5.0 or greater, than remove this download step and use the installed version?

core/sqf/sql/scripts/get_libhdfs_files
zellerh: Thanks, will fix that.

core/sqf/sql/scripts/get_libhdfs_files
zellerh: The "-rf" option will suppress errors if the files don't exist.

core/sqf/sql/scripts/get_libhdfs_files
zellerh: Sure, will do.

core/sqf/sql/scripts/get_libhdfs_files
```diff
@@ -0,0 +1,186 @@
...
+       ;;
+
+    -d|--tempDir)
+       shift
+       if [[ $# < 1 ]]; then
+         echo "Expecting argument after -d or --tempDir"
+         exit 1
+       fi
+       LIBHDFS_TEMP_DIR="$1"
+       ;;
```
robertamarton: It looks like someone can set LIBHDFS_TEMP_DIR before calling this script but if they specify the -d option, tempDir is overwritten - I assume this is the desired action.  Also, LOGFILE is set earlier based on LIBHDFS_TEMP_DIR, if the user changes this value here, should the LOGFILE also be changed to match?

core/sqf/sql/scripts/get_libhdfs_files
```diff
@@ -0,0 +1,186 @@
...
+done
+
+
+if [[ $FORCE_BUILD == true || \
+      ! -e ${TGT_INC_DIR}/hdfs.h || \
+      ! -e ${TGT_LIB_DIR}/libhdfs.so ]]; then
+
+  if [ ! -d $LIBHDFS_TEMP_DIR ]; then
+    mkdir $LIBHDFS_TEMP_DIR
+  fi
```
robertamarton: What happens if the directory has not been created.  For example, you may need sudo access to create the temp directory or the user specified some invalid location.

core/sqf/sql/scripts/get_libhdfs_files
```diff
@@ -0,0 +1,186 @@
...
+  esac
+
+  shift
+done
+
+
+if [[ $FORCE_BUILD == true || \
+      ! -e ${TGT_INC_DIR}/hdfs.h || \
+      ! -e ${TGT_LIB_DIR}/libhdfs.so ]]; then
+
```
robertamarton: There is another library - libhadoop.so, so the assumption is that if libhdfs.so exists so does libhadoop.so.

core/sqf/sqenvcom.sh
zellerh: Good point, I should not have removed it from bldenvchk (did that when I tried to eliminate Thrift altogether - unsuccessfully). I'll put the check back into bldenvchk, because the build would fail.

core/sqf/sql/scripts/get_libhdfs_files
zellerh: Yes, thanks.

core/sqf/sql/scripts/get_libhdfs_files
```diff
@@ -0,0 +1,186 @@
...
+  if [[ ! -d ${PROTOBUF_ID} ]]; then
+    echo "Unpacking Google Protocol Buffer tar file..." | tee -a ${LOGFILE}
+    rm -rf ${LIBHDFS_TEMP_DIR}/${PROTOBUF_TGT_ID}
+    tar -xzf ${PROTOBUF_TAR} >>${LOGFILE}
+  fi
+
+  if [[ ! -d $PROTOBUF_TGT_ID ]]; then
+    cd ${PROTOBUF_ID}
+    echo "Building Google Protocol Buffers, this could take a while..." | tee -a ${LOGFILE}
+    if [[ $VERBOSE == true ]]; then
```
robertamarton: Why can't we just download binaries for protocol buffers?

core/sqf/sql/scripts/get_libhdfs_files
zellerh: Sorry, I don't remember anymore why I picked 2.6.0, but it does work. We use 2.5.0-cdh5.3.0 elsewhere, which is different from Apache 2.5.0, and I might have run into some interface issues with the Apache 2.5.0 version.

core/sqf/sql/scripts/get_libhdfs_files
```diff
@@ -0,0 +1,186 @@
...
+
+  # Final check whether all the needed files are there
+  if [[ ! -r ${TGT_INC_DIR}/hdfs.h || \
+        ! -r ${TGT_LIB_DIR}/libhdfs.so ]]; then
+    echo "Error, not all files were created" | tee -a ${LOGFILE}
+    ls -l ${TGT_INC_DIR}/hdfs.h
+    ls -l ${TGT_LIB_DIR}/libhdfs.so
+    exit 1
+  fi
+fi
```
robertamarton: This is great - thanks for adding this make step.

core/sqf/sql/scripts/get_libhdfs_files
zellerh: It might also have been because at some point Apache started to distribute the 64 bit native library in the binary distribution - that would not be a strong reason of course, since we are building the library here anyway.

robertamarton:
Should get someone from the foundation team to look through these changes.

core/sqf/sql/scripts/get_libhdfs_files
zellerh: I left out libhadoop.so on purpose. Some of the tar files don't seem to include it and it is not needed to build, only at runtime. There might be a problem if it is missing both at compile time and at runtime.

core/sqf/sql/scripts/get_libhdfs_files
zellerh: Thanks, will fix that.

core/sqf/sql/scripts/get_libhdfs_files
zellerh: I didn't see a place where you can download binaries. The RPMs that are available for protoc on CentOS don't include version 2.5.0. Let me know if you find a good place to download this as a binary.

core/sqf/sqenvcom.sh
sandhyasun: Is CURL_INC_DIR and CURL_LIB_DIR removed ? We need it for LOB support. The code is not yet active due to a few issues hence no regression test yet for this case.

core/sqf/sql/scripts/get_libhdfs_files
zellerh: Yes, good idea, I'll add that.

zellerh:
There is one more dependency we need to take care of: The newget_libhdfs_files script to build the native Hadoop libraries needs cmake, so we need to do the following on all of our machines:

sudo yum -y install cmake

I'll work with Steve to get that done before we can merge the change. This is the reason why the Jenkins test above failed.

robertamarton:
I changed the build wiki page to include this dependency.



    Roberta



*Sent:* Tuesday, October 13, 2015 9:58 AM
*Subject:* Re: [incubator-trafodion] TRAFODION-1521 Build Trafodion without
having HBase installed (#118)



There is one more dependency we need to take care of: The
newget_libhdfs_files script to build the native Hadoop libraries needs
cmake, so we need to do the following on all of our machines:

sudo yum -y install cmake

I'll work with Steve to get that done before we can merge the change. This
is the reason why the Jenkins test above failed.

—
<https://github.com/apache/incubator-trafodion/pull/118#issuecomment-147777536>
.

core/sqf/sqenvcom.sh
zellerh: No, these are still there. Good to know, I was wondering what those are used for. Thanks for checking.

core/sqf/sql/scripts/get_libhdfs_files
zellerh: Yes, the command line options should override environment variables, if specified (mixing the two is probably not a good idea). I will fix the log file issue and update the LOGFILE variable here. Thanks.

svarnau:
Ahh, so it is indirect dependency, needed to build libhdfs.  I will get it
added to the jenkins build machine recipe.



--Steve



*From:* Roberta Marton [mailto:notifications@github.com]
*Sent:* Tuesday, October 13, 2015 10:02 AM
*Subject:* Re: [incubator-trafodion] TRAFODION-1521 Build Trafodion without
having HBase installed (#118)



I changed the build wiki page to include this dependency.



Roberta



*From:* Hans Zeller [mailto:notifications@github.com
<notifications@github.com>]
*Sent:* Tuesday, October 13, 2015 9:58 AM
*Subject:* Re: [incubator-trafodion] TRAFODION-1521 Build Trafodion without
having HBase installed (#118)



There is one more dependency we need to take care of: The
newget_libhdfs_files script to build the native Hadoop libraries needs
cmake, so we need to do the following on all of our machines:

sudo yum -y install cmake

I'll work with Steve to get that done before we can merge the change. This
is the reason why the Jenkins test above failed.

—
<
https://github.com/apache/incubator-trafodion/pull/118#issuecomment-147777536
>
.

—
<https://github.com/apache/incubator-trafodion/pull/118#issuecomment-147778501>
.

zellerh:
Thanks, let us know, Steve, when Jenkins is ready to build this version then. It did go down the right path in sqenvcom.sh and it built much of protobuf and Hadoop, so besides cmake I'm cautiously optimistic this change will work with your current build setup.

svarnau:
Ready now.



--Steve



*Sent:* Tuesday, October 13, 2015 10:20 AM
*Cc:* Steve Varnau <steve.varnau@esgyn.com>
*Subject:* Re: [incubator-trafodion] TRAFODION-1521 Build Trafodion without
having HBase installed (#118)



Thanks, let us know, Steve, when Jenkins is ready to build this version
then. It did go down the right path in sqenvcom.sh and it built much of
protobuf and Hadoop, so besides cmake I'm cautiously optimistic this change
will work with your current build setup.

—
<https://github.com/apache/incubator-trafodion/pull/118#issuecomment-147782705>
.

zellerh:
Thanks, I asked Sean and Shang-Sheng to take a look as well, as Roberta suggested, once they have commented and once you have reviewed the rework I hope this change will be ready.

sbroeder:
Looks good to me.

zellerh:
jenkins, retest

bucaojit:
Reviewed the TM changes: hbasetmlib2/ directory.  Looks good.  Nice work converting the build to Maven!

One thing that we will keep in mind is whether to add pom/Makefile changes to support other distros (besides CDH) similar to the build in the sqf/src/seatrans/hbase-trx directory.

DaveBirdsall:
Looks good to me. Just waiting for a passing test.

robertamarton:
Rework looks good, thanks for making the changes.

DaveBirdsall:
So, are we waiting for Steve to resolve the cmake dependency on our build machines? Hoping Steve will trigger a retest when that is ready.

svarnau:
Also need to address this failure:
2015-10-13 12:59:03 Error: Build created 25175 untracked file(s)
2015-10-13 12:59:03        Update .gitignore files

An impressive list of new build objects.

zellerh:
This PR may go for two awards, for the most untracked objects and the most commits :-)

Thanks for pointing that out. I'll fix with my next commit (waiting for test results to see whether there are any other issues).

DaveBirdsall:
I think you need one or two more commits to make the second record. But another record this PR is achieving is scrutiny by the most reviewers, I think. Community development in action!

DaveBirdsall:
Are we good to go on this one now?

zellerh:
Hi Dave,

Yes, thanks for your great patience, I think it is ready now.

Hans

PR #120, opened 2015-10-14
[TRAFODION-1525]Locale for JDBCT4 doesn't work

kevinxu021:
Change the base name of the bundler.
Fixed an NullPointerException for messageArguments as well.


DaveBirdsall:
Looks good to me.

svarnau:
jenkins, add user

core/conn/jdbc_type4/src/org/trafodion/jdbc/t4/HPT4Messages.java
```diff
@@ -295,7 +292,7 @@ static HPT4Exception createSQLException(T4Properties t4props, Locale msgLocale,
 			if (messageArguments != null) {
 				message = message.concat(" With parameters: ");
 				while (true) {
-					message = message.concat(messageArguments[i++].toString());
+					message = message.concat(messageArguments[i++] + "");
```
arvind-narain: This is minor - do we need to make the same change on line 83 also ?

core/conn/jdbc_type4/src/org/trafodion/jdbc/t4/HPT4Messages.java
```diff
@@ -251,18 +253,13 @@ static HPT4Exception createSQLException(T4Properties t4props, Locale msgLocale,
 			Object p[] = T4LoggingUtilities.makeParams(t4props, messageId, messageArguments);
 			log.logp(Level.SEVERE, "HPT4Messages", "createSQLException", "", p);
 		}
-
-		Locale currentLocale;
+		
+		Locale currentLocale = t4props == null ? null : t4props.getLocale();
+		currentLocale = currentLocale == null ? Locale.getDefault(): currentLocale;
+		
 		int sqlcode;
```
arvind-narain: Do please check that ignoring msgLocale is ok for all callers.

arvind-narain:
Starting the merge since its needed for https://github.com/apache/incubator-trafodion/pull/121. Do please open another pull request if any of the above comments needs further changes.

core/conn/jdbc_type4/src/org/trafodion/jdbc/t4/HPT4Messages.java
kevinxu021: Arvind, Dont worry about it. t4props is the only way to get locale and should be the only way. As i said, that is meaningless that some messages are Chinese but the others are English. If we do that, it will involve more misunderstandings.

PR #128, opened 2015-10-16
[TRAFODION-1532]TEST042 allowing longer compile time and collect system stats

HowardQin:


sureshsubbiah:
Looks good. Thanks for the change. I can merge it in if no issues are found in the next few hours.

PR #129, opened 2015-10-16
[TRAFODION-1535] Explain and RMS statistics output are not in sync

selvaganesang:
when the vsbb is chosen for Insert/Upsert command.

See JIRA for details


core/sql/optimizer/RelUpdate.h
```diff
@@ -1830,7 +1830,8 @@ class HbaseInsert: public Insert
                CollHeap *oHeap = CmpCommon::statementHeap(),
                InsertType insertType = SIMPLE_INSERT)
        : Insert(name,tabId,otype,child,NULL,NULL,oHeap, insertType),
-         returnRow_(FALSE)
+         returnRow_(FALSE),
+         vsbbInsert_(FALSE)
```
sureshsubbiah: CopyTopNode is not carrying over this new flag. I suppose it does not matter since the flag is being set in the generator. Please consider adding that in a subsequent checkin, if you think it will be a good thing to do.

sureshsubbiah:
Changes look great. Thanks for making explain more reliable.

core/sql/optimizer/RelUpdate.h
selvaganesang: Will do just to make it consistent.

PR #130, opened 2015-10-16
Jira 1531

bucaojit:


narendragoyal:
Looks good! Nice refactoring that will help with future code maintenance!

sbroeder:
Looks correct, but do we need similar checks in put/delete<list of>?

bucaojit:
Hi Sean.  Yes, I added the checkRow for both the list of puts/deletes and single put/delete.  The put(List) actually calls the put(single) in the code so the checkRow on line 3935 covers both cases.  As for delete, single delete is line 3641 and delete(List) is line 3661.

sbroeder:
Got it.  This looks fine.  Thanks Oliver.

sureshsubbiah:
Merging. Thanks for change and the review all.

PR #131, opened 2015-10-16
[TRAFODION 1539] HADOOP_PATH will now be set

amandamoran:
HADOOP_PATH was not being set in config file.


sureshsubbiah:
Looks good. Thanks for the change.

PR #134, opened 2015-10-19
TRAFODION-1541 Fix make errors when building outside of git

zellerh:
Removed the Koraktor Maven plugin from SQL and hbasetmlib2 pom files,
since it fails when we build a source tree that is not managed by git.

Additional changes:

- Added "clean" steps to remove the built jar files
- Suppressed "svn not found" error when building outside of git
- Added maven to traf_tools_setup.sh

### DIFF
--- a/install/traf_tools_setup.sh
+++ b/install/traf_tools_setup.sh
@@ -263,11 +263,11 @@ fi
 
 # install maven, if not available already
 MAVEN_VERSION=`mvn --version 2>/dev/null | grep 'Apache Maven' | cut -f 3 -d ' '`
-if [[ "$MAVEN_VERSION" != "3.0.5" ]]; then
+if [[ ! "$MAVEN_VERSION" =~ "3." ]]; then
   # install maven
   cd $BASEDIR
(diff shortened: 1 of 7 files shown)


svarnau:
looks good to me.

install/traf_tools_setup.sh
```diff
@@ -260,3 +260,14 @@ make install
 if [ $VERBOSE -eq 1 ]; then
   echo "STEP completed:  installation of THRIFT"
 fi
+
+# install maven, if not available already
+MAVEN_VERSION=`mvn --version 2>/dev/null | grep 'Apache Maven' | cut -f 3 -d ' '`
+if [[ "$MAVEN_VERSION" != "3.0.5" ]]; then
```
robertamarton: The current default is maven 3.3.3.  I have been using this version without issue.  Do we really need to check for an older version?

zellerh:
No, I can change it so we accept any 3.x Maven version and install 3.3.3 if we don't find any. I just picked 3.0.5 because that's what's installed on my machine.

robertamarton:
Any objections to me committing this fix?

PR #137, opened 2015-10-20
[TRAFODION-34] Support region splitting and re-balancing with transactions active

bucaojit:
Functionality being checked in disabled.

To enable the flush/read functionality, enter the following property into hbase-site.xml:
  <property>
    <name>hbase.transaction.split.delay.noflush</name>
    <value>false</value>
  </property>


DaveBirdsall:
Code looks OK to me but it is beyond my area of expertise. Is there someone with transaction background who can review and give thumbs up?

shangshengtung:
Code looks OK.

PR #143, opened 2015-10-22
[TRAFODION-1553] Missing License causing AL static check failure

bucaojit:


svarnau:
Thanks for the quick response on making the RAT license check happy.

sandhyasun:
Looks good.

PR #145, opened 2015-10-23
[TRAFODION-1560] Improve script that sets up tools needed for Trafodion build

robertamarton:
Thanks for reviewing


core/sqf/sqenvcom.sh
```diff
@@ -469,24 +469,24 @@ else
...
 
-    See http://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/ClusterSetup.html
+       Execute the install_local_hadoop script which performs a single node
+       install using a popular Hadoop distribution 
 
-    Yet another option is to use the install_local_hadoop script on a
-    single node for evaluation or development.
+          cd $MY_SQROOT/sql/scripts
+          install_local_hadoop [-p <port option>]
+          source ./install_traf_components
```
zellerh: install_traf_components is a shell script and should be executed, not sourced. It has "exit" commands in it, so if it is sourced, it will kill the invoker's shell when something goes wrong and they won't be able to see what happened.

svarnau:
Looks like you changed <version> in rest/pom.xml and dcs/pom.xml, but not <package.version>.
I think both values need to change in both pom files.

core/sqf/sqenvcom.sh
robertamarton: You are correct, I will fix this.

robertamarton:
Looks like I missed a place for version updates.  Will fix this also.

install/traf_tools_setup.sh
```diff
@@ -62,9 +62,43 @@ function Usage {
...
+function downloadSource
+{
+  # currently only tar files ending in "tar.gz" and "tgz" are recognized
+  TARSUFFIX="tar.gz"
+  if [[ ! $1 == *$"$TARSUFFIX" ]]; then
+    TARSUFFIX="tgz"
+  fi
+
+  if [ ! -e $BASEDIR/$2.$TARSUFFIX ]; then
+    wget $1  >>$LOGFILE 2>&1
```
zellerh: The function takes two arguments, the URL and the name of the extracted directory. It would misbehave, however, if the name of the extracted directory would not match what's in the file. Wouldn't it be easier to just use a single argument $1 and to figure out the name of the file to extract with `basename $1`. I think that would be consistent with the way wget behaves. If the function takes a second argument, then maybe that second argument should be the name of the tar file to extract, including the suffix?

install/traf_tools_setup.sh
```diff
@@ -43,6 +43,7 @@
 #   common services such as naming, configuration management, synchronization, 
 #   and group services.
 # Thrift: Communications and data serialization tool
+# Maven: Build tool that is only installed if compatible version does not exist
```
zellerh: Actually, that version check is commented out below, maybe change this comment as well?

zellerh:
Looks good to me, with a few small comments. Thanks for making the message printed out by sqenvcom.sh better and more useful!

install/traf_tools_setup.sh
robertamarton: That is a good suggestion.  I would like to make other improvements to this script and will make a note of this issue.

install/traf_tools_setup.sh
robertamarton: This was a mistake.  I was testing the code where Maven was already installed.  Removed the commented lines.

sandhyasun:
Looks good. Merging.

PR #146, opened 2015-10-26
[TRAFODION-1561]List all configurations

kevinxu021:
The way to know what kinds of configurations it has on DCS. More people ask for it. Enjoy.

### DIFF
--- a/dcs/src/main/resources/dcs-webapps/master/servers.jsp
+++ b/dcs/src/main/resources/dcs-webapps/master/servers.jsp
@@ -194,5 +194,22 @@ entries
     <%
     }  
     %>
+<br>
+<h2> List all configurations</h2>
+<table>
+<tr><th>Properties Name</th><th>Value</th></tr>
+<%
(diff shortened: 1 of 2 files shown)


svarnau:
jenkins, add user

DaveBirdsall:
This is out of my area of expertise, but it looks good to me. Will commit it later today if there are no other comments.

hegdean:
This change is of no harm but helpful during troubleshooting as it displays not only DCS specific configuration but much more than that.  It would have been nice if this was displayed as a link on the main page  which would then show list the entire configuration on a new page rather than on the main page

kevinxu021:
Hegdean, Thanks for your comments. Ref to Apache Storm-UI Topology page. All-in-one instead of switching pages may be helpful here since you may get wanted info by CTRL+F on browser. It may support fold/unfold and something more will be added. But all the other things will be done in another pull-request. By the way, all configurations are from dcs-default.xml and dcs-site.xml. Don't worry, its style and backend might have significant changes in the next milestone.

PR #147, opened 2015-10-27
TRAFODION-1549 Extends event_log_reader TMUDF to read DCS logs and overflow files from all components

hegdean:
Fix in installer so DCS_INSTALL_DIR is set before executing sqgen so that UDFs can read this environment variable


sureshsubbiah:
Looks perfect. Since this change needed to get nightly tests to pass, I will merge by tonight if there are no issues found in the next 3 hrs.

PR #148, opened 2015-10-27
[TRAFODION-1564]Split/balance is not being delayed

bucaojit:
Region split/balance is not being delayed if Trx Scanner obj is present
Fix was to delay on scanners list clear OR pending list clear instead
of AND.


sureshsubbiah:
Change looks good to me. Merging.

PR #149, opened 2015-10-28
[[TRAFODION 1567]] HDFS commands can not be done on nodes without HDFS

amandamoran:


amandamoran:
jenkins, retest

DaveBirdsall:
Just making sure I understand the fix. It looks like the old code looked around for a node running Cloudera Manager, then ran a bunch of things on that node. The new code just runs things on the local node. So, I take it that the tacit assumption in the old code was that HDFS would be running on any node where Cloudera Manager was running but that assumption turns out not to be true? The new tacit assumption is that HDFS is running on the local node. Why is this a better assumption? 

Given that the new assumption is good, the code changes look great.

amandamoran:
Great questions.
So, I take it that the tacit assumption in the old code was that HDFS would
be running on any node where Cloudera Manager was running but that
assumption turns out not to be true? That is correct. In all our previous
testing I thought they would always be together. That is not the case at
all. Actually, from further study they are almost never together that is
not the recommended configuration.

The new tacit assumption is that HDFS is running on the local node. Why is
this a better assumption? Trafodion must be installed on nodes with HBase
and HBase needs HDFS. Trafodion does a check (not in this file... earlier
in install) that makes sure every node being installed on has HBase (so by
proxy HDFS).

An improvement in the future will be for the installer to be ran from
anywhere. Example: Hadoop Nodes n1 n2 n3 n4  Trafodion nodes n3 n4 and
install from n0 (a node not in cluster) n1-n4.

On Thu, Oct 29, 2015 at 8:47 AM, DaveBirdsall <notifications@github.com>
wrote:

> Just making sure I understand the fix. It looks like the old code looked
> around for a node running Cloudera Manager, then ran a bunch of things on
> that node. The new code just runs things on the local node. So, I take it
> that the tacit assumption in the old code was that HDFS would be running on
> any node where Cloudera Manager was running but that assumption turns out
> not to be true? The new tacit assumption is that HDFS is running on the
> local node. Why is this a better assumption?
>
> Given that the new assumption is good, the code changes look great.
>
> —
> <https://github.com/apache/incubator-trafodion/pull/149#issuecomment-152221325>
> .
>



-- 
Thanks,

Amanda Moran

PR #154, opened 2015-10-30
[[ TRAFODION 1563 ]] Fix of INSTALL_LOG file with current date

amandamoran:
On an upgrade, INSTALL_LOG was getting the old installer name (because
the config file had already been sourced). Added new variable to "remember"
the new installer log.


sureshsubbiah:
Looks good. Merging.

PR #155, opened 2015-11-01
update expected file for HIVE/TEST018

nonstop-qfchen:
I think the diff is due to some other reason, since my changes are about update stats on hive tables which is not done in TEST018.  

Here are some interesting findings. 
1. If I prepare the query separately, I do not see the warnings. 
2. If I ran the test with explain, the explain shows reasonable cardinality for the query in question, which indicates the stats is present.
3. Earlier on, I noticed the disappearance of the same warning from core/test126. 

A JIRA will be filed to track the unnecessary warning.

### DIFF
--- a/core/sql/regress/hive/EXPECTED018
+++ b/core/sql/regress/hive/EXPECTED018
@@ -140,9 +140,9 @@ Task:  CLEANUP         Status: Started    Object: TRAFODION.HBASE.CUSTOMER_ADDRE
 Task:  CLEANUP         Status: Ended      Object: TRAFODION.HBASE.CUSTOMER_ADDRESS
 Task:  PREPARATION     Status: Started    Object: TRAFODION.HBASE.CUSTOMER_ADDRESS
        Rows Processed: 50000 
-Task:  PREPARATION     Status: Ended      ET: 00:00:09.002
+Task:  PREPARATION     Status: Ended      ET: 00:00:09.186
 Task:  COMPLETION      Status: Started    Object: TRAFODION.HBASE.CUSTOMER_ADDRESS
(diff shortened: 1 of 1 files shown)


sureshsubbiah:
Given that the warning appears intermittently, will a known diff file or a filter to consume the warning (if produced) be a better solution? Thank you for taking care of this test failure.

nonstop-qfchen:
Yes, once we fix the bug, we will regenerate the expected file.

On Mon, Nov 2, 2015 at 9:20 AM, Suresh Subbiah <notifications@github.com>
wrote:

> Given that the warning appears intermittently, will a known diff file or a
> filter to consume the warning (if produced) be a better solution? Thank you
> for taking care of this test failure.
>
> —
> <https://github.com/apache/incubator-trafodion/pull/155#issuecomment-153050777>
> .
>



-- 
Regards, --Qifan

DaveBirdsall:
Looks good to me.

sureshsubbiah:
While we fix the bug will the test fail intermittently, even if we merge this change? I hope I am not understanding the situation incorrectly. Thanks for explaining.

nonstop-qfchen:
Actually, it was determined that the fix has nothing with the appearance
and disappearance of the warning after one more around of the review of the
change, and the fact that update stats was not called during the test.


On Mon, Nov 2, 2015 at 5:20 PM, Suresh Subbiah <notifications@github.com>
wrote:

> While we fix the bug will the test fail intermittently, even if we merge
> this change? I hope I am not understanding the situation incorrectly.
> Thanks for explaining.
>
> —
> <https://github.com/apache/incubator-trafodion/pull/155#issuecomment-153190127>
> .
>



-- 
Regards, --Qifan

PR #158, opened 2015-11-02
[TRAFODION-1576] Performance improvement for backup-restore and reducing offline interval for backup.

sureshsubbiah:


DaveBirdsall:
Looks good to me. Though the new Java utility module is beyond my expertise.

core/sqf/hbase_utilities/src/main/java/org/trafodion/utility/backuprestore/TrafExportSnapshot.java
```diff
@@ -0,0 +1,1076 @@
...
+    	    runCopyJob(inputRoot, outputRoot, snapshotName, snapshotDir, verifyChecksum,
+    		       filesUser, filesGroup, filesMode, mappers, bandwidthMB);
+    	}
+    	else {
+    	     	LOG.info("Copy Snapshot Data Files");
+    	     	for (Pair<SnapshotFileInfo, Long> fileInfo: snapshotFiles) {
+    	     		inputInfo = fileInfo.getFirst();
+    	     		Path outputPath = getCopyOutputPath(conf, inputInfo);
+    	     		Path inputPath = getCopyInputPath(conf, inputInfo, inputFs);
+    	     		FileUtil.copy(inputFs, inputPath, outputFs, outputPath, false, false, conf);
```
selvaganesang: Should we consider setting the filesuser, filesGroup and filesMode on the target cluster similar to runCopyJob. This may be needed if we need to Export  a single table using this utility in future.

selvaganesang:
Please consider testing with HDP and apache hbase to ensure that this utility works with these distributions also

DaveBirdsall:
Just checking: Where are we on this change?

sureshsubbiah:
Sorry for delayed response reviewers.
I was looking into addressing Selva's comments but would like to defer them now
a) To set filesUser, filesGroup and filesMode on target cluster when FileUtil.copy is used, I could use the preserveAttributes method that is in Hbase' ExportSnapshot class. The method would have to be made more wide widely available than its current private visibility. I would like to look into how refactoring can be done better and also think of a way to test this change. Currently I don't have a cluster where hbase is installed under a different id.
b) Testing on apache and hdp will be good. I will try to add a small regression test to the backup script. Also need more time to get access to hdp and vanilla apache clusters.

I hope the change can be merged for now and I will address these two issue in my next delivery in this area.

DaveBirdsall:
Selva, are you good with that?

selvaganesang:
Yes. No problem.

DaveBirdsall:
I will commit it now.

PR #161, opened 2015-11-05
[TRAFODION-1590] UPSERT into <table> VALUES (<subquery>, ...) fails

sureshsubbiah:


core/sqf/hbase_utilities/backup_and_restore/run_full_trafodion_backup.sh
```diff
@@ -229,17 +238,29 @@ if  [[ $? -eq 0 ]]; then
     exit 1;
 fi
 
+## Restart Trafodion
+if [[ ${stay_offline} -ne 1 ]]; then
+ start_trafodion $trafodion_user 
+else
+ echo "Trafodion can now be restarted at your convinience. Backup script does not need for it to be offline anymore."
```
DaveBirdsall: Typo, should be "convenience"

DaveBirdsall:
Not sure I understand this pull request. It appears to contain much of the same material in a previous pull request, PR 158?

sureshsubbiah:
Yes, it is the same Dave. Sorry, I did not intend for the B/R commit to be included in this Pull Request. I used the same workspace, but a different branch. Apparently that is not enough. Let me find out what it will take to remove the B/R commit from here. It should not matter as it is the same change, but I would prefer to not commit it a second time.

sureshsubbiah:
Following Steve's advice I merged with master again and now the B/R commit is no longer visible in the PR. There is just a big merge commit. Thank you Steve.
PS Steve also indicated previous situation would have been harmless.

DaveBirdsall:
Looks good. Just wanted to check: Were the changes to runregr_executor.ksh intended? Or is that debug code?

sureshsubbiah:
Intended Dave. When I ran executor test015 I saw an error flash by on the screen where it was looking for the begin_odbc.ksh script and not finding it. I thought it will be good to not execute that line. There are a few other cqds and settings that are a leftover from old days. It will be a bigger task to clean them up too. Should have explained this in the commit message. Thanks for catching it.

DaveBirdsall:
Given that tests pass, I think this is ready to commit.

PR #162, opened 2015-11-05
[TRAFODION-1591] Script to find ports not-in-use

DaveBirdsall:
This check-in adds a new script in the sql/scripts directory, findPort.py. It is useful in finding a range of ports that are not in use before using install_local_hadoop.

### DIFF
--- a/core/sqf/sql/scripts/findPort.py
+++ b/core/sqf/sql/scripts/findPort.py
@@ -1,0 +1,106 @@
+# @@@ START COPYRIGHT @@@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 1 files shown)


selvaganesang:
Will netstat print ports in use from other users?

DaveBirdsall:
Yes.

sandhyasun:
This is a very useful script. I wonder if the 300 range can also be parameterized ? Even so , this version of script can go in. Merging this.

sandhyasun:
In the abive comment, I meant the port range that is now hardcoded to 200 (not 300)

DaveBirdsall:
Hi Sandhya: Did you mean, check for ranges of 300 ports rather than 200?

PR #165, opened 2015-11-09
Resolving the following JIRAs related to LOB feature :

sandhyasun:
TRAFODION-1604 - Fixed one place in CharType.h where 100 was the max limit for lob handle length
TRAFODION-1596 - Added checks in alter code to prevent altering adding LOB columns.
TRAFODION-1598 - Added several syntax filxes in parser and ExpLOBaccess.cpp to address these problems.
TRAFODION-1599 - Added checks in binder to prevent sample columns from being LOB columns.
TRAFODION-1602 - Added checks in DDL layer to prevent LOB columns as unique constraints or store by.

### DIFF
--- a/core/sql/exp/ExpLOBaccess.cpp
+++ b/core/sql/exp/ExpLOBaccess.cpp
@@ -1693,7 +1693,9 @@ Ex_Lob_Error ExLob::readDataToLocalFile(char *fileName,  Int64 offset, Int64 siz
     // open the targte file for writing 
     int filePerms = S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH;
     int openFlags = O_RDWR ; // O_DIRECT needs mem alignment
-    if ((LobTgtFileFlags)fileflags == Lob_Append_Or_Error )
+    if (((LobTgtFileFlags)fileflags == Lob_Append_Or_Error ) ||
+	((LobTgtFileFlags)fileflags == Lob_Error_Or_Create ) ||
+	((LobTgtFileFlags)fileflags == Lob_Append_Or_Create))
       openFlags |= O_APPEND;
     else
       openFlags |= O_TRUNC;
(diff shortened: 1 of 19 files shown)


core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -7469,7 +7469,14 @@ RelExpr *Scan::bindNode(BindWA *bindWA)
      bindWA->setErrStatus();
      return NULL;
   }
-
+  if (naTable->hasLobColumn() && isSampleScan())
```
sureshsubbiah: Does this mean we cannot update statistics with sampling on a table with LOB columns?

sureshsubbiah:
Changes look good. Thank you. There appears to be some problems with Jenkins though.

core/sql/optimizer/BindRelExpr.cpp
sandhyasun: Update statistics right now returns an error when a table has lob columns. Will open a JIRA for that. But your question on this also showed that this if check is a bit too restrictive. Will submit a change to move the check from binder phase to the generator phase where it's easier to detect sample columns alone.

core/sql/exp/ExpLOBaccess.cpp
```diff
@@ -1693,7 +1693,9 @@ Ex_Lob_Error ExLob::readDataToLocalFile(char *fileName,  Int64 offset, Int64 siz
     // open the targte file for writing 
     int filePerms = S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH;
     int openFlags = O_RDWR ; // O_DIRECT needs mem alignment
-    if ((LobTgtFileFlags)fileflags == Lob_Append_Or_Error )
+    if (((LobTgtFileFlags)fileflags == Lob_Append_Or_Error ) ||
```
DaveBirdsall: Declaring fileflags to be of type LobTgtFileFlags would eliminate the need for all this casting. And would be safer. (Not a show-stopper though.)

DaveBirdsall:
The changes look good to me.

PR #166, opened 2015-11-10
[TRAFODION-1530]deprecated put.getFamilyMap

mashengchen:


sureshsubbiah:
jenkins, add user

sureshsubbiah:
Change has been reviewed and approved by Trina Krug. Thank you. I will merge once tests pass, unless there are other comments.

sureshsubbiah:
jenkins, retest

DaveBirdsall:
Looks good to me. Is Trina on JIRA yet? Would be good if she can give her input directly.

selvaganesang:
Is the deprecated API getFamilyMap removed from some distro? It looks like we are creating more objects just to take care of the deprecated API. Won't it be good to use Cell concepts all the way to server instead of just to take care of this API alone on the client side.

mashengchen:
This method is missing in some HBase 0.98.x version. It seems to be there in 0.98.1 and gone in 0.98.6, at least in the CDH versions.
See HBASE-10339. The method also seems to be gone in HBase 1 and 2, so we can hopefully replace it with the new method getFamilyCellMap().

sureshsubbiah:
Selva, do you have any objections to merging this change?

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/SsccTransactionalTable.java
```diff
@@ -692,22 +693,23 @@ public SsccPutMultipleTransactionalResponse call(SsccRegionService instance) thr
...
-				}
-			}
-		}
-	}
+    // validate for well-formedness
+    public void validatePut(final Put put) throws IllegalArgumentException {
+        if (put.isEmpty()) {
+            throw new IllegalArgumentException("No columns to insert");
+        }
+        if (maxKeyValueSize > 0) {
```
selvaganesang: This looks like the code is never executed because maxKeyValueSize is not set at all. Should we remove it?  Also, it is not clear why should the keyValue size should be restricted in Trafodion.

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionalTable.java
```diff
@@ -765,23 +765,24 @@ else if (result.hasException())
...
-			}
-		}
-	}
-	
+    // validate for well-formedness
+    public void validatePut(final Put put) throws IllegalArgumentException {
+        if (put.isEmpty()) {
+            throw new IllegalArgumentException("No columns to insert");
+        }
+        if (maxKeyValueSize > 0) {
```
selvaganesang: Same comment as above

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/regionserver/transactional/SingleVersionDeleteNotSupported.java
```diff
@@ -53,13 +54,15 @@ public SingleVersionDeleteNotSupported() {
...
      */
     public static void validateDelete(final Delete delete) throws SingleVersionDeleteNotSupported {
-        Collection<List<KeyValue>> values = delete.getFamilyMap().values();
-        for (List<KeyValue> value : values) {
-            for (KeyValue kv : value) {
-                if (Type.Delete.getCode() == kv.getType()) {
+        Collection<List<Cell>> values = delete.getFamilyCellMap().values();
+        for (List<Cell> value : values) {
+            for (Cell cell : value) {
+                KeyValue kv = new KeyValue(cell);
```
selvaganesang: Consider using cell.getTypeByte instead of constructing a new object kv

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/SsccTransactionalTable.java
mashengchen: yes, the code is never executed, also in KeyValue source code, it restricted the size itself.
but i guess the person who code here will use a constant to restricted it future, and he(she) may be forget to add the constant.

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/regionserver/transactional/SingleVersionDeleteNotSupported.java
mashengchen: got it

prashanth-vasudev:
maxKeyValueSize is not set. So the piece of code is not executed , looks like.  if there is plan to retain the code for later, I would change  "kv.getLength()" to "KeyValueUtil.length(cell)" where ever it is used. 

Rest of the code changes look good to me.

selvaganesang:
Are we ready to merge this change in?  I would recommend filling a JIRA to remove references to KeyValue and use Cell concepts in the trx-jar part of the code. This code is good to go in.

DaveBirdsall:
Selva, you should now have committer privileges. Would you like to test that out by committing this change?

selvaganesang:
Yes Dave. I will commit this change.

PR #167, opened 2015-11-10
[TRAFODION-1575] Avoid transforming update into delete and insert

selvaganesang:
commands to improve performance of update statements. When a table has
check constraint the update command is transformed into insert and delete.

Added code to evaluate constraint expressions in all TCBs so that
constraint checking is done correctly. But, the subtask [TRAFODION-1610]
needs to implemented before this change is exercised.

### DIFF
--- a/core/sql/optimizer/BindRelExpr.cpp
+++ b/core/sql/optimizer/BindRelExpr.cpp
@@ -10535,7 +10535,7 @@ RelExpr *Update::bindNode(BindWA *bindWA)
   NABoolean transformUpdateKey = updatesClusteringKeyOrUniqueIndexKey(bindWA);
   if (bindWA->errStatus()) // error occurred in updatesCKOrUniqueIndexKey()
     return this;
-
+  // To be removed when TRAFODION-1610 is implemented.
   NABoolean xnsfrmHbaseUpdate = FALSE;
   if ((hbaseOper()) && (NOT isMerge()))
     {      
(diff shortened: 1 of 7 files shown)


DaveBirdsall:
Looks good to me. An interesting test case will be UPSERT on a column with a check constraint, for both the case where the row exists and where the row does not. Similarly for MERGE with both a WHEN MATCHED clause and a WHEN NOT MATCHED clause.

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -10543,26 +10543,19 @@ RelExpr *Update::bindNode(BindWA *bindWA)
...
-	}
-      else if (avoidHalloween())
-	{
-	  xnsfrmHbaseUpdate = TRUE;
-	}
       else if (getCheckConstraints().entries())
-	{
-	  xnsfrmHbaseUpdate = TRUE;
-	}
+       {
```
sureshsubbiah: I suppose the default attribute can be removed from nadefaults.cpp and DefaultConstants.h also?

sureshsubbiah:
Changes look good to me. Thank you for doing this. I second Dave in asking for some testcases to be added to regressions. Maybe that can be done as part of JIRA 1610. If we do do decide to remove an attribute from nadefaults.cpp, that can be done later too. I will wait for a few hours and merge if there are no other comments.

selvaganesang:
Thanks Dave and Suresh. I will revisit the requested change when Hans is done with TRAFODION-1610

PR #169, opened 2015-11-11
Couple of fixes for JIRA issues. Details below.

anoopsharma00:
1) JIRA 1606. Create of a table LIKE a metadata table
              now succeeds.
2) JIRA 1611. An error was incorrectly being returned if the
              number of columns in a user specified column family
              exceeded 32. That has been fixed.


sureshsubbiah:
Change looks good. Please consider adding JIRA id to title of Pull request next time "[TRAFODION-1610]". That will cause this pull request information to be automatically placed in the JIRA for 1610, giving a record of the checkin. This seems to work on for the first JIRA in the title though.

robertamarton:
looks good to me

PR #170, opened 2015-11-11
Trafodion-1612 and Trafodion-1613

robertamarton:
1612: Stop building HBase-trx for MAPR, due to instable Maven repository

- removed mapr from the build_all step in the hbase_trx Makefile, we now create
  6 jar files instead of 7

1613: build failing due to mismatched Maven downloads

When building Trafodion from source objects, Maven is required.  It is not
possible to download Maven from yum, instead it needs to be installed. There
are a few problems related to how Maven is managed with Trafodion builds.

- In many environments, Maven is already installed.  The check to see if Maven
  is already installed is commented out in traf_tools_setup.sh. This has been
  fixed.

- Today, sqenvcom.sh checks to see if mvn is installed, if not, it sets it to
  version 3.0.5 but we install 3.3.3. Changed the code to return an error:
  "ERROR unable to find mvn (MAVEN) in path" to inform the user that mvn needs
  to be added to the path before building. An error is also returned during
  the build (make) if mvn is not found in the path.

- As part of the above change, it was noted that many of the tools were being
  set to the value of $TOOLSDIR when $TOOLSDIR may not be set.  Checks were
  added to see if the environment variable was already set. This allows
  someone to specify a different location for these tools as described in the
  additional build tools wiki page.

- Changed the output returned in bldenvchk.sh.  BISON does not have to be in
  the path but needs to be specified correctly in the environment variable.

### DIFF
--- a/core/bldenvchk.sh
+++ b/core/bldenvchk.sh
@@ -50,7 +50,7 @@ for AVAR in $VARFILELIST; do
       printf '%s =\t%s\n' $AVAR $AVALUE
    fi
    if [[ $RESULT != 0 ]]; then
-      echo "*** Error: $AVAR file not found in PATH."
+      echo "*** Error: $AVAR file not found in PATH or file not fully qualified."
       RETVAL=1
    fi
 done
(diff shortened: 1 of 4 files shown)


DaveBirdsall:
Looks good to me. Thank you for making these changes.

PR #171, opened 2015-11-13
Fix for TRAFODION-1623

robertamarton:
Removed mvn in path check from sqenvcom.sh

### DIFF
--- a/core/sqf/sqenvcom.sh
+++ b/core/sqf/sqenvcom.sh
@@ -774,13 +774,7 @@ export ANT=ant
 export AR=ar
 export FLEX=flex
 export CXX=g++
-
-# Maven may not yet be in the path. Check and report an error
 export MAVEN=mvn
-if [[ -z "$(which $MAVEN 2> /dev/null)" ]]; then
-  echo " *** ERROR unable to find mvn (MAVEN) in path"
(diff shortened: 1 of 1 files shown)


svarnau:
if you want to retain the error check, then it should go into core/bldenvchk.sh script that is run only during build time, not runtime.

zellerh:
Looks good to me. Putting it into bldenvchk.sh is good, but if the build fails, the error message should be fairly clear and easy to diagnose without a special check.

robertamarton:
The check is already in bldenvchk, I just also erroneously added it to
sqenvcom.sh.



     Roberta



*From:* Steve Varnau [mailto:notifications@github.com]
*Sent:* Thursday, November 12, 2015 5:23 PM
*Subject:* Re: [incubator-trafodion] Fix for TRAFODION-1623 (#171)



if you want to retain the error check, then it should go into
core/bldenvchk.sh script that is run only during build time, not runtime.

—
<https://github.com/apache/incubator-trafodion/pull/171#issuecomment-156288465>
.

svarnau:
You're right, it is already in the bldenvchk.sh!  Looks good to me.

svarnau:
jenkins, retest -- infrastructure change was pre-mature for 1.3 release; jar file fix not back-ported to this branch.

PR #173, opened 2015-11-13
[TRAFODION-1417] Reduce locations to change release version number

svarnau:
To change the trafodion release version major/minor number, there were about a dozen locations in four different files that needed to be modified.

This set of changes eliminates all those places except for the single place in core/sqf/sqenvcom.sh.

It does add maven warnings in DCS and REST build logs, same as in other components that already use maven with environment variable version number.

jdbc_test.py script now takes additional options to pass in correct version numbers. If you use swjdbc script, then it is taken care of. Other environments that call jdbc_test.py directly (e.g., jenkins test automation) should also be changed. But the default values match the current release, so the change is not urgent.

### DIFF
--- a/core/rest/pom.xml
+++ b/core/rest/pom.xml
@@ -401,7 +401,7 @@
   	<jython-standalone.version>2.5.3</jython-standalone.version>
   	<jettison.version>1.3.5</jettison.version>
   	<displaytag.version>1.1.1</displaytag.version>
-  	<jdbct4.version>2.0.0</jdbct4.version>
+  	<jdbct4.version>${project.version}</jdbct4.version>
    	
     <!-- Plugin Dependencies -->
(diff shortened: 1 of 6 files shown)


DaveBirdsall:
Wonderful change! Reduces the opportunity for human error when we change versions.

hegdean:
Finally we  fixed it. Thanks.  Changes  look good to me.

arvind-narain:
Thanks Steve for this change - looks good

zellerh:
See also TRAFODION-1522. To get rid of the warnings eventually, we can try to make a Maven project hierarchy and then change the version in the top-level pom.xml file. That would add one more place to update the version, but I hope that would be acceptable. Sorry, didn't get to work on this project hierarchy yet.

PR #174, opened 2015-11-13
[TRAFODION-1585] Allow MDAM on inner child of NJ in certain circumsta…

DaveBirdsall:
…nces. Details are given in the JIRA.

### DIFF
--- a/core/sql/optimizer/OptPhysRelExpr.cpp
+++ b/core/sql/optimizer/OptPhysRelExpr.cpp
@@ -3917,7 +3917,7 @@ InputPhysicalProperty* NestedJoin::generateIpp(
   Cardinality maxRows;
   CardConstraint * outerCardinalityConstraint = NULL;
 
-  if (getGroupAttr()->hasCardConstraint(minRows, maxRows))
+  if (child(0).getGroupAttr()->hasCardConstraint(minRows, maxRows))
     {
       outerCardinalityConstraint = new(CmpCommon::statementHeap())
                    CardConstraint(minRows,maxRows);
--- a/core/sql/optimizer/OptPhysRelExpr.cpp
+++ b/core/sql/optimizer/OptPhysRelExpr.cpp
@@ -3910,6 +3910,20 @@ InputPhysicalProperty* NestedJoin::generateIpp(
   InputPhysicalProperty* ipp = NULL;
 
   // ----------------------------------------------------------------
+  // If there is a cardinality constraint on the outer child, we'll
+  // pass that information along to the inner child.
+  // ----------------------------------------------------------------
+  Cardinality minRows;
+  Cardinality maxRows;
+  CardConstraint * outerCardinalityConstraint = NULL;
+
+  if (getGroupAttr()->hasCardConstraint(minRows, maxRows))
+    {
+      outerCardinalityConstraint = new(CmpCommon::statementHeap())
+                   CardConstraint(minRows,maxRows);
+    }
+
+  // ----------------------------------------------------------------
   // Get the sort order of my left child and pass this information to
   // the context for optimizing my right child.  (Costing the inner
   // table access has a dependency on the order from the outer table).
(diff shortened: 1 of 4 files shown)


core/sql/optimizer/OptPhysRelExpr.cpp
```diff
@@ -3910,6 +3910,20 @@ InputPhysicalProperty* NestedJoin::generateIpp(
...
 
   // ----------------------------------------------------------------
+  // If there is a cardinality constraint on the outer child, we'll
+  // pass that information along to the inner child.
+  // ----------------------------------------------------------------
+  Cardinality minRows;
+  Cardinality maxRows;
+  CardConstraint * outerCardinalityConstraint = NULL;
+
+  if (getGroupAttr()->hasCardConstraint(minRows, maxRows))
```
zellerh: I would have expected that this would examine the join's left child's group attributes, not those of the join itself?

core/sql/optimizer/ScanOptimizer.cpp
```diff
@@ -3364,7 +3364,25 @@ ScanOptimizer::isMdamEnabled() const
...
+    const InputPhysicalProperty* ippForMe =
+      getContext().getInputPhysicalProperty();
+    if (ippForMe)
+      {
+        const CardConstraint * outerCardConstraint =
+          ippForMe->getOuterCardConstraint();
+        if (outerCardConstraint)
+          {
+            Cardinality upperBound = outerCardConstraint->getUpperBound();
+            if (upperBound <= (Cardinality)1)
```
zellerh: Sounds good for now, maybe we can extend this to small cardinalities > 1 in the future, e.g. by using an integer-valued CQD here?

core/sql/optimizer/OptPhysRelExpr.cpp
DaveBirdsall: Good catch... I think you are right. I will fix this. I wondered why my unit test succeeded with this mistake. It succeeded because both legs of the join had cardinality constraint of one so the join itself had this constraint also.

core/sql/optimizer/ScanOptimizer.cpp
DaveBirdsall: There is already a CQD just a few lines down in the code that will allow this for estimates. I wanted to avoid introducing yet another CQD. That was the point of the code a few lines down that used the cardinality constraint to sharpen the row estimate.

DaveBirdsall:
Please do not merge until I have addressed the bug Hans found above.

core/sql/optimizer/OptPhysRelExpr.cpp
sureshsubbiah: Maybe in the case described in TRAFODION-1585 the CardConstraint on the outer child, resulted in a CardConstraint being synthesized for the Join too? (since the Join is on key columns?), hence this change produced the desired plan.

That raises the question on whether the CardConstraint that is sent down to the inner child states that
a) The number of probes or unique values for input from outer is limited by the CradConstraint
or b) whether the output of the inner child satisfies the CardConstraint. This is a basic question, but my knowledge of how Ipp works is limited.

Also do we need to consider inputs or probes that could come from the top. I am thinking that the right child of an NJ can see several requests, even though the left child send only one probe for each request that it receives.

All the questions I have are minor and for my personal knowledge, I think the change is good.

DaveBirdsall: Yes, the intent of the change is to send any cardinality constraint on the result of the left child down to the right child scan as a constraint on the number of probes. Before this change, no such constraint was passed down; we relied on estimates only. The difference is a matter of certainty. MDAM was considered dangerous to do on the inner child of a nested join due to unreliability of estimates in some plans.

sureshsubbiah:
Looks good to me Dave.

zellerh:
Hi,

Did some more thinking about this, and had the following thought:

We send input logical properties to the right child of the nested join (Context::getInputLogProp()). Those contain a cardinality and a max. cardinality (Context::getInputLogProp()->getMaxCardEst()). Both of these cardinalities should be in the range of the min. and max. cardinality established by constraints.

IMHO it would be better if we would check that max cardinality instead of passing another object down to the right child of the join.

There are multiple other questions I have with the current solution, but first I want to find out whether you are ok with this solution.

Ideally we should guarantee that estimated and max. cardinality should always be within the cardinality boundaries established by constraints (GroupAttr::hasCardConstraint()), but I couldn't find any code that ensures that - maybe I just didn't look carefully enough. Anyway, that could be handled through a separate JIRA, if necessary. For now I would assume that this is the case.

My apologies for being late, I should have thought about it and commented earlier on!

DaveBirdsall:
Please do not merge this change while I investigate Hans' suggestion. Thanks.

DaveBirdsall:
Closing this pull request, as it has been replaced by https://github.com/apache/incubator-trafodion/pull/519.

PR #175, opened 2015-11-15
Fix for the executor/TEST106 failure caused due to JIRA TRAFODION-1625

sandhyasun:
A delete query on a 200K large table was marked as a unique delete. This was found when the long operation was not being cancelled when a cancel query was issued. Unique queries are typcially not cancelable since they are  short running.   The issue happens when rowset deletes are chosen . This  plan change happened recently due to a costing change. The rowset operation needs to be marked as cancelable.

### DIFF
--- a/core/sql/generator/GenPreCode.cpp
+++ b/core/sql/generator/GenPreCode.cpp
@@ -4934,8 +4934,9 @@ RelExpr * HbaseDelete::preCodeGen(Generator * generator,
   else
   if (isUnique)
     {
-      // do not cancel unique queries.
-      generator->setMayNotCancel(TRUE);
+      if (!generator->oltOptInfo()->multipleRowsReturned())
+	// do not cancel unique queries.
+	generator->setMayNotCancel(TRUE);
       uniqueHbaseOper() = TRUE;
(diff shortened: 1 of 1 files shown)


selvaganesang:
Please consider taking care of this issue in HbaseUpdate and yet another place where this flag is being set.

core/sql/generator/GenPreCode.cpp
```diff
@@ -4934,8 +4934,9 @@ RelExpr * HbaseDelete::preCodeGen(Generator * generator,
   else
   if (isUnique)
     {
-      // do not cancel unique queries.
-      generator->setMayNotCancel(TRUE);
+      if (!generator->oltOptInfo()->multipleRowsReturned())
+	// do not cancel unique queries.
+	generator->setMayNotCancel(TRUE);
```
selvaganesang: Revise the comment please.

sandhyasun: Done updating the comment. Also added a similar check to HbaseUpdate . Hbase Access already had the check in place and added just a comment. I did not find any other place that needed this check.

sureshsubbiah:
Looks good to me. Thanks.

PR #176, opened 2015-11-16
[TRAFODION-1627] 'sqgen' cannot copy files to all nodes.

mkby:


svarnau:
jenkins, add user

selvaganesang:
Looks good to me

selvaganesang:
Requested Narendra Goyal and Prashanth Vasudev to review this change.

prashanth-vasudev:
Looks good to me. I see mon.env and shell.env only in src/seabed/test  only in work station environment. May not be needed for normal working environment.

selvaganesang:
I will be merging this change tomorrow because the files mon.env and shell.env are not generated by sqgen. But these files are sourced in by monitor and shell respectively in the code to set up its own properties.

amandamoran:
Question: Would these files need to be copied over if sqgen is needed to be ran after the first install? 

Question 2: Would sqgen ever be needed to run after the first install?

PR #177, opened 2015-11-16
[TRAFODION-1628] Implement T2 Driver's Rowsets ability to enhance the batch insert performance

ryzuo:
This is the implementation of T2 Driver rowsets ability of batch insert and fetch as well. Includes Interface, functionality and method changes as below:
1. Add the interface code for communication between JAVA
and native via protocol buffer.
2. Overload the PREPARE/EXUECUTE functions of native code to
implement rowsets functionality.
3. Rewrite/overload the PREPARE/EXECUTE/FETCH code of JAVA
PreparedStatement class.


core/conn/jdbc_type2/native/CSrvrStmt.cpp
```diff
@@ -180,12 +241,88 @@ SQLRETURN SRVR_STMT_HDL::Prepare(const SQLValue_def *inSqlString, short inStmtTy
...
     stmtType = inStmtType;
     holdability = inHoldability;
 
     CLI_DEBUG_RETURN_SQL(PREPARE(this));
 }
 
+SQLRETURN SRVR_STMT_HDL::Execute(const char *inCursorName, IDL_long inInputRowCnt, IDL_short inSqlStmtType, 
+        const SQLValueList_def *inValueList, 
+        short inSqlAsyncEnable, Int32 inQueryTimeout)
+{
```
selvaganesang: The method signature of this method and the one below looks same. If so, it is not clear how the compiler decides to resolve this method.

arvind-narain:
Thanks for this change. This will surely help in usage of T2 via multithreaded DCS.

While the review is in progress, could you also plan to implement some T2 specific unit test cases for this change ? Thank you.

core/conn/jdbc_type2/native/CSrvrStmt.cpp
ryzuo: Hi Selva,
They are different actually, if you look at it carefully, the one below has one more argument SQLValueList_def *outValueList, they are overloaded functions.

ryzuo:
Hi Arvind,

Actually the T2 Phoenix tests basically covered all we need for this (and this does not change any API of JDBC supplied for user). However, I have some tests for counting the performance of batch insert which I used it to compare it against the old driver and T4 driver. I can push it to somewhere in the repo.

core/conn/jdbc_type2/native/CSrvrStmt.cpp
selvaganesang: Thanks. It is my bad. I forgot to expand the hidden lines.

selvaganesang:
For this commit to be merged, you just need to ensure that the check-PR tests passes. That is a minimum requirement.

Also, this change involves many files. Do you think a review meeting would be helpful to get this change reviewed and merged. Please feel free to invite me and others who have exposure to this area of code.

arvind-narain:
jenkins, retest -- Build error for extracting hadoop-2.6.0-src.tar.gz: Cannot open: No such file or directory

selvaganesang:
It is good to see this merge passed the minimum requirement. I just glanced through the code and saw there is a lot of rework being done in memory management of T2 driver. Earlier, we had ensured that there is no memory growth when we ran YCSB select and update benchmarks with T2 driver. Can you please work to get these benchmarks done and ensure that there is no memory growth in T2 driver.

hegdean:
Can you please include me in the review meeting as well so we can get rowset functionality reviewed and merged. Regarding tests, please include them under dcs/src/test folders so they are run regularly along with the other tests

core/conn/jdbc_type2/native/CSrvrStmt.cpp
```diff
@@ -341,7 +479,7 @@ SQLRETURN SRVR_STMT_HDL::Close(unsigned short inFreeResourceOpt)
     SQLRETURN rc;
 
     if (stmtType == INTERNAL_STMT) CLI_DEBUG_RETURN_SQL(SQL_SUCCESS);
-    cleanupSQLMessage();
+    //cleanupSQLMessage();
```
ryzuo: Sorry, this is not on purpose, I will uncomment this within an updates along with some other changes of the memory allocation/release management code.

DaveBirdsall:
Is this ready to commit?

selvaganesang:
I have requested for YCSB performance regressions to be run and also a meeting to review this change.

ryzuo:
I've never did YCSB tests before, whom should I seek help for this? I failed to find instructions and script which is used before on wiki and github.

For meeting to review this, this is a big change, I do not rather to explain every line of the codes. Thus I'd like to gather topics and questions about what you want to know via the meeting first.

sandhyasun:
Trying to move this one along. Did you get the needed help to run YCSB ?  For meeting, perhaps a Skype meeting with a link here would be good and let the reviewer drive it by asking you questions.

ryzuo:
Got some instructions of YCSB from Dennis already. I'm working on it now, but still waiting for physical machine to make it run, now I only have virtual machine which may cannot afford such kind of performance tests. Will send out the online meeting link after some Christmas, I will have a few days off   next week.

ryzuo:
I booked a skype meeting at 1pm -3pm this Thursday,  if it's not convenient for all of you, please let me know.
https://meet.partner.lync.cn/esgyncom-esgyn/ruoyu.zuo/VST5VC96

ryzuo:
sorry, comment above is Shanghai time, for PST, it must be Wednesday 9pm - 11pm, 13th.

ryzuo:
I have run several times of YCSB tests, observed memory leak, it took me long time to track it down, and I found one serious problem in YCSB's code, look into all the test cases in YCSBWorkload2.java, all PreparedStatement instances it creates are never closed even when the statement execution is done, this causes serious memory leak, until the connection is closed or the process exits.
Previously, the PreparedStatement does not actually have batch ability, so the memory cost is very slow, but if you run statements without close again and again, as long as the time last long enough, along with more and more un-closed statements stay in, I can also observe the memory leak, just the growing rate is slightly enough cause no serious problem. I believe if it runs long enough this way, memory will be eventually exhausted.
With the new driver, because of the rowsets functions have much higher memory cost, the growing rate is much higher as well. But if we close the PreparedStatement every time once it's execution is done, I don't observe leak then.

I think the YCSB codes must be updated, make sure the PreparedStatement instances are closed once they are done with their job. Driver does not manage to close the PreparedStatement, because  we never know when the user will use this statement again, it's one time prepared, N times execution statement.

The attached screenshot is one example of the un-closed PreparedStatenent code in YCSBWorkload2.java
31e7 tm

selvaganesang:
.Is the Singletonselect_stmt set to NULL everytime after execution. If not, isn't the statement is prepared once and executed many times. I believe YCSB test is designed to prepare once and execute many times. if the preparedStatement is a IUD statement, all the memory allocated at the time of execution should be re-used everytime. In case of Select statement, the application can choose to call ResultSet.close().  If you close a prepareStatement, it is expected that the statement is prepared again.  

Yes. PreparedStatement is kept around till the connection is closed or till the user application explicitly closes the statement. YCSB wants to keep the statement around till the connection is closed. YCSB tests are not designed to prepare the statement everytime, because the execution time is very short. It is prepare once and execute many times kind of test.

A well written driver, should protect itself even if the application doesn't close the Resultset explicitly. It should trigger resource allocation when the ResultSet close object goes out of scope and gets garbage collected. This holds good for Statement or PreparedStatement also. Trafodion driver is supposed to create weak references and poll the associated reference queue for deallocating the resources when the object has no other reference. However, that may not be relevant here, because YCSB tests are designed to prepare once and execute many times.

You can test it with T4 driver and check if there is growth in mxosrvr memory similar to T2.

ryzuo:
Jenkins, retest

DaveBirdsall:
It looks like just one regression test is failing, seabase/TEST011. Looking at recent versions of it, I'm wondering if merging the latest version of that test into your change would cure the problem?

DaveBirdsall:
Ignore my comment; the issue with seabase/TEST011 is a merge issue that is being addressed separately. Once that is done, try another retest.

DaveBirdsall:
jenkins, retest

DaveBirdsall:
Is this ready to merge now? @ryzuo, would you like to commit it?

selvaganesang:
It would be good to update this PR with the findings from YCSB testing before this PR is merged

selvaganesang:
@ryzuo You can update if virtual memory held steady with the YCSB testing.

ryzuo:
@selvaganesang I had another minor update of the code this time. As I override the new/delete operation to trace memory usage(actually not exactly override new/delete operation, as we have macro wrapping the new/delete operations, I managed simply add tracing code in those wrappers, as you can see them in Debug.h/.cpp), I do found about 3 leaking places, but not significant, I fixed them already in the recent commits, left only one global object(srvrGlobal), which should be living during the entire lifetime until process exit, thus we cannot free it. Also, I've gone through the codes very carefully(almost line by line), and I did not find any other leaking memories.

I managed to run a group of YCSB tests with scale 10/stream 5 and scale 10/stream 10. I have some problem with uploading the log files here earlier this morning((maybe network or the zip file is not supported by github? no idea...), I will have another try later, or just put it on OneDriver or something else, and paste the URL here.

As T2 actually residents on server local, so the tests actually only run on single node as well as all streams are on the one same node, and seems rowsets requires ESPs cost plenty memories. According to the machine resource limitation, that's all I can ran for now. But still, I observed the memory growing during the tests run, but now I have no idea where the leaking spot is, as the tracing does not show anything from several rounds of tests, neither as I read the code. My suspicion is there might be some leaking in CLI, or may some references requested through CLI were not closed properly through, the modern CLI rowset APIs are quite different from the old fashion calls which T2 used to use. Anyway, I'm not quite familiar with CLI calls, so they are kind of complicated to me. And as we don't have any documents of CLI APIs(or it's just I never found, if you know there's any, please tel me), so this still needs time to investigate, and some help to understand the CLI would be better.

My opinion is we cannot have a pie with one bite, so we can merge this part first, in the meanwhile we keep investigate on the mem growing and CLI calls, and we will have another PR for this.

sandhyasun:
I think it would be reasonable to have a JIRA describing the results and growth of memory during these tests. Studying the CLI API alone will not be enough since a whole lot of additional processing happen during execution. It could be any of the executor TCBs executing the sql statement as well. So some careful experimenting and repeated analysis needs to be done and some joint effort with executor may be needed to achieve this.  So it would be good to merge the functionality in and analyze the growth separately as part of the new JIRA. Will wait to see if  atleast one  other committer agrees.

ryzuo:
[ycsb-log.zip](https://github.com/apache/incubator-trafodion/files/343674/ycsb-log.zip)

selvaganesang:
When the project was initially being open sourced from HP,  the memory leak was observed with T4  driver.  @arvind-narain  and I worked together to identify the root cause of the issue with the mxosrvr and fixed it. I don't recollect the memory leak was observed with T2 driver. @arvind-narain can comment on this too. It might be good to check if there is memory growth with T2 driver without these changes by running the same tests before merging this.

sandhyasun:
Agreed with previous comment. @ryzuo  perhaps you have already done this. But it would help resolve the concern  if you ran YCSB without your changes and see if there is similar memory growth or 0 growth as @selvaganesang is saying and attaching those logs too.   If you still do see issues only with your changes we can definitely come up with a recourse to hep you analyze the growth or else address them separately in a different JIRA.

ryzuo:
I understand your concerns. Indeed there's no growing with T2 observed without these changes, but there's growing with MXOSRVR. However, like I said, I have mem tracing code, of course includes all these new changes, shows no leaking already (about 3 leaking before, fixed already in the latest commit). So my suspicious is, probably there's inappropriate calls of CLI(maybe some descriptor or statement is not close, or anything else, I have no idea, there's no reference for CLI calls...), and it might cause something in executor during the statements's execution time like @sandhyasun said. T2 rowsets functionality uses a group of CLI calls like MXOSRVR does, slightly different(so looks like T2 is growing faster, or maybe it's just because it's multi-threaded, and all "connections" resident in one single process on the same machine). That's the only way that I can think of now, if you have anyother way out of this, surely  that would be great.

ryzuo:
[mem_N_stmt_logs.zip](https://github.com/apache/incubator-trafodion/files/348988/mem_N_stmt_logs.zip)

These are the mem trace log and a statement trace log files last time I ran. The only object has never been deleted is created in SQLMXDriver.cpp line 114 (each row for each thread), and that's for srvrGlobal, a global object should be living in entire connection life cycle, thus we can't and don't have place to release it. And as well as the COMMIT and ROLLBACK statement.

selvaganesang:
I had used google memory profiler earlier. See http://goog-perftools.sourceforge.net/doc/heap_profiler.html .  With this tool, I was able to narrow down the memory leak outside of Trafodion SQL engine. This includes any client component that issues Trafodion SQL CLI calls.

You can also monitor the memory usage in trafodion SQL using get process statistics for <pid>( pid is of the form <node_id>, <pin> )command after the application reached a steady state. In case of YCSB, the application can be assume to have reached the steady state after first interval of information is displayed. Trafodion SQL uses its own heap management to manage the memory within the SQL engine. Please see the manuals for description of output of get process statistics command.

anoopsharma00:
This request has been opened for a while now.
What will it take to come to a decision on this?
What actions need to be done or demonstrated to come
to a conclusion regarding it?

PR #178, opened 2015-11-16
[TRAFODION-1423] indexes are now created in aligned format by default

selvaganesang:
The indexes row format is now independent of the table. The table
can be in format that aids in pushing down the predicates to
hbase, but the index is now created in aligned format by default.

SHOWDDL <table_name> now shows the row format of the index when it
is in aligned row format.

### DIFF
--- a/core/sql/executor/HBaseClient_JNI.cpp
+++ b/core/sql/executor/HBaseClient_JNI.cpp
@@ -2875,14 +2875,7 @@ HBC_RetCode HBaseClient_JNI::getHbaseTableInfo(const char* tblName,
   blockSize = arrayElems[1];
   if (isCopy == JNI_TRUE)
      jenv_->ReleaseIntArrayElements(jHtabInfo, arrayElems, JNI_ABORT);
-
-  if (jresult == false)
-  {
-    logError(CAT_SQL_HBASE, "HBaseClient_JNI::getHbaseTableInfo()", getLastError());
-    jenv_->PopLocalFrame(NULL);
-    return HBC_ERROR_GET_HBTI_EXCEPTION;
-  }
-
+  jenv_->PopLocalFrame(NULL);
   return HBC_OK;  // Table exists.
 }
 
--- a/core/sql/executor/HBaseClient_JNI.cpp
+++ b/core/sql/executor/HBaseClient_JNI.cpp
@@ -2845,10 +2845,15 @@ HBC_RetCode HBaseClient_JNI::getHbaseTableInfo(const char* tblName,
   if (jenv_ == NULL)
      if (initJVM() != JOI_OK)
          return HBC_ERROR_INIT_PARAM;
+  if (jenv_->PushLocalFrame(jniHandleCapacity_) != 0) {
+     getExceptionDetails();
+     return HBC_ERROR_GET_LATEST_SNP_EXCEPTION;
+  }
   jstring js_tblName = jenv_->NewStringUTF(tblName);
   if (js_tblName == NULL)
   {
     GetCliGlobals()->setJniErrorStr(getErrorText(HBC_ERROR_GET_HBTI_PARAM));
+    jenv_->PopLocalFrame(NULL);
     return HBC_ERROR_GET_HBTI_PARAM;
   }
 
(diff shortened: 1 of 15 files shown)


core/sql/generator/GenExpGenerator.cpp
```diff
@@ -2892,7 +2892,8 @@ short ExpGenerator::generateKeyEncodeExpr(const IndexDesc * indexDesc,
       if (handleSerialization)
 	{
 	  NAColumn * nac = indexDesc->getNAFileSet()->getIndexKeyColumns()[i];
-	  if (CmpSeabaseDDL::isEncodingNeededForSerialization(nac))
+          NABoolean isAlignedRowFormat = indexDesc->getNAFileSet()->isSqlmxAlignedRowFormat();
+	  if ((!isAlignedRowFormat) && CmpSeabaseDDL::isEncodingNeededForSerialization(nac))
```
sureshsubbiah: Since the NAColumn object has a data member describing its serialization state, why do we ned the additional check of !isAlignedFormat here?

For example t1.b is serialized bit the correspond column in the index t1_ix1.b is part of an aligned row and therefor not serailized. I would have expected CmpSeabaseDDL::isEncodingNeededForSerialization(nac) to return FALSE for t1_ix1.b since you havetaken care to set up two different NAColumn objects for the table and index.

core/sql/generator/GenRelScan.cpp
```diff
@@ -302,7 +302,7 @@ int HbaseAccess::createAsciiColAndCastExpr2(Generator * generator,
   asciiValue = new (h) NATypeToItem(newGivenType->newCopy(h));
   castValue = new(h) Cast(asciiValue, newGivenType); 
 
-  if (HbaseAccess::isEncodingNeededForSerialization(colNode))
+  if ((!alignedFormat) && HbaseAccess::isEncodingNeededForSerialization(colNode))
```
sureshsubbiah: This comment is about line 264, the NABoolean HbaseAccess::isEncodingNeededForSerialization(ItemExpr * colNode) method. I could not add a comment  there.

Is this method expected to return the correct answer now if it is called on an Index object which is aligned, but whose parent table is not aligned? I would expect the answer to be  YES, and therefore the additional change in line 305 is not necessary. If that is not the case, then some kind of comment on line 296 starting that it cannot be used directly on aligned indexes will be good.

sureshsubbiah:
Change looks good. It will improve performance with indexes. I am slightly concerned with how binder stage changes on resetting the serialization attribute of NAColumn in an index are not really used in the generator. Most likely I have missed something in the flow of logic.

selvaganesang:
Thanks Suresh for your comments. The serialization concept is not supported for aligned format. There is no need to check if encoding is needed for the serialized column for aligned row format. Ideally, I could have encapsulated this concept in both the methods of IsEncodingNeededForSerialization, but chose to do it this way.

selvaganesang:
The test seabase/TEST022 failed possibly because of change in plan when the rows in vanilla hbase table  are deleted. The plan change might have been triggered by the fix for TRAFODION-1633. The delete command works if cqd hbase_delete_costing is turned off. I am running the full regressions in my environment to check if any other tests are failing due to this fix. I might turn off the cqd hbase_delete_costing by default.

selvaganesang:
All tests passed except seabase/TEST022. Instead of changing the default value of hbase_delete_costing cqd, I will turn it off in the test script. I think the issue might be that the rowset plan is chosen for deletes against native hbase tables.

core/sql/executor/HBaseClient_JNI.cpp
```diff
@@ -2856,26 +2861,25 @@ HBC_RetCode HBaseClient_JNI::getHbaseTableInfo(const char* tblName,
...
+  blockSize = arrayElems[1];
+  if (isCopy == JNI_TRUE)
+     jenv_->ReleaseIntArrayElements(jHtabInfo, arrayElems, JNI_ABORT);
 
   if (jresult == false)
   {
     logError(CAT_SQL_HBASE, "HBaseClient_JNI::getHbaseTableInfo()", getLastError());
+    jenv_->PopLocalFrame(NULL);
     return HBC_ERROR_GET_HBTI_EXCEPTION;
   }
```
DaveBirdsall: Why don't we PopLocalFrame in the OK return case?

core/sql/regress/seabase/TEST022
```diff
@@ -20,7 +20,7 @@
 -- under the License.
 --
 -- @@@ END COPYRIGHT @@@
-
+cqd hbase_delete_costing 'off';
```
DaveBirdsall: Just curious why this is needed.

DaveBirdsall:
Looks good to me. One question, though; see in-line comments.

core/sql/executor/HBaseClient_JNI.cpp
selvaganesang: Should do it. Will submit a change because it can cause memory leak.

DaveBirdsall:
Looks good to me also. When the new tests pass I think it is ready to commit.

DaveBirdsall:
I will commit this now.

PR #180, opened 2015-11-18
[TRAFODION-1631]Using Zookeeper node creation time for time check

kevinxu021:
Fix the issue that if Master node system time is later than the rest of nodes.

### DIFF
--- a/dcs/src/main/java/org/trafodion/dcs/master/DcsMaster.java
+++ b/dcs/src/main/java/org/trafodion/dcs/master/DcsMaster.java
@@ -22,22 +22,11 @@ under the License.
  */
 package org.trafodion.dcs.master;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.net.InetAddress;
-import java.net.InetSocketAddress;
-import java.net.ServerSocket;
-import java.net.NetworkInterface;
-import java.nio.charset.Charset;
-import java.util.Enumeration;
-import java.util.*;
 import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.Executors;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
-import java.util.concurrent.ExecutionException;
 
-import org.apache.commons.io.IOUtils;
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.GnuParser;
 import org.apache.commons.cli.Options;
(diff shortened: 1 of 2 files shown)


svarnau:
jenkins, add user

DaveBirdsall:
It is a bit outside of my area of expertise but the code looks fine to me.

selvaganesang:
Thanks for removing the imported classes that are not used. Can you please provide some more details on the issue that this change resolves, it would be helpful to review this change.

kevinxu021:
Selva, there is a JIRA for more detailed info. https://issues.apache.org/jira/browse/TRAFODION-1631.
In a word, assume it has a 3-nodes cluster, node1 ST(system time) 00:00:03, node2 ST 00:00:01 and node3 ST 00:00:02. Now if the DCSMaster start on node1, it will show nothing on DCS UI because DCSServers ST is ahead of DCSMaster. You may know what will happen if DCSMaster starts on node2. Yes, it'll show one DCSServer on DCSUI.

arvind-narain:
Does Zookeeper sync the times for quorum configuration? If it doesn't and picks up the ctime from the node where the physical open is done, then we may face the same issue. Could you check if ntpd service is running on all nodes ? Maybe coding for tolerating some delta difference might be good.

selvaganesang:
Thanks Kevin. I should have known to look at the JIRA also.  Such a variation in the time of the nodes forming the cluster can cause all kinds of problems within our software stack and hbase. So, ntpd service is the way to go.

kevinxu021:
Arvind, After testing, yes, it'll not impact ctime sequence in Zookeeper even no NTPD service.
Selva, On production, yes, NTPD service is required. But for this obvious issue, i think it should be fixed since it will confuse users that the number of DCS processes doesn't match the number DCS UI without NTPD. It's a no-harm fix and only impacts DCS UI.

selvaganesang:
Agreed Kevin that it is a no-harm fix. But, the variation between the time of the nodes can cause lot more problems in hbase. See https://blog.cloudera.com/blog/2011/04/hbase-dos-and-donts/. With Trafodion, we let hbase to add the timestamp  when a row is inserted/deleted/updated. The variation in time impacts in general and more so when the query involves ESP processes running on different nodes. This can cause the rows to disappear or reappear unintentionally. Hence, I feel fixing this issue at DCS level might be no-harm for DCS, but it won't help the overall state of the cluster with the variation in time of the nodes. I would give +0 for this fix tending more towards -1. However I will let other reviewers to decide.

arvind-narain:
Agree with Selva. Better would be to fix the underlying issue by seeing if ntp/ntpd service is running. Fixing just DCS may not indicate that cluster is actually able to tolerate the time difference.

 It might also be helpful to write a utility that checks on $MY_NODES for some tolerance difference between the stat time reported by zookeeper during a dummy node creation and the actual time on the node. This could then flag the nodes that are beyond the tolerance limit and help administrator in fixing the problem.

kevinxu021:
Thank you for comments. Still have some questions. As you said, it seems NTP is required for Trafodion cluster. SQCHECK/DCSCHECK should check it and if the user dont have one on each node, dont allow to start anything. I got confused. It should not be allowed to start an abnormal cluster. By the way, Cloudera Manager only prints a WARNING message if it doesnt have NTP and doesnt have any abvious mismatching. Sure, if you confirm with it, i'll get back the fixes.

arvind-narain:
Thanks Kevin.

Following will verify NTP status:
pdsh $MY_NODES "ntpstat" | grep unsynchronised

Installer script should be modified to check if times are sync'd and/or if ntp is installed/started. Including Amanda @diericka  for change in installer.

amandamoran:
Hi there All-

Please make a jira for this and assign to me.

https://github.com/amandamoran

Thanks!

On Mon, Nov 23, 2015 at 1:57 PM, arvind-narain <notifications@github.com>
wrote:

> Thanks Kevin.
>
> Following will verify NTP status:
> pdsh $MY_NODES "ntpstat" | grep unsynchronised
>
> Installer script should be modified to check if times are sync'd and/or if
> ntp is installed/started. Including Amanda @diericka
> <https://github.com/diericka> for change in installer.
>
> —
> <https://github.com/apache/incubator-trafodion/pull/180#issuecomment-159078458>
> .
>



-- 
Thanks,

Amanda Moran

DaveBirdsall:
Where does this issue stand? Is there a rework coming?

kevinxu021:
it should get rid of time check. this pull request is going to close.

PR #181, opened 2015-11-18
TRAFODION-1635. Changes to a few script related sql message file and to lob startup.

sandhyasun:
Includes change from Narendra for starting lob process when a node is intergrated.(gensq.pl)

### DIFF
--- a/core/sqf/sql/scripts/lobstart
+++ b/core/sqf/sql/scripts/lobstart
@@ -30,8 +30,8 @@ let totalNodes=`grep 'Up' $tmp_node_status | wc -l`
 while [ ${count} -lt $totalNodes ]
 do
 sqshell -a > /dev/null << eof
-set {process \zlobsrv${count} } PERSIST_RETRIES=10,60
-set {process \zlobsrv${count} } PERSIST_ZONES=0,1
+set {process \$zlobsrv${count} } PERSIST_RETRIES=10,60
+set {process \$zlobsrv${count} } PERSIST_ZONES=${count}
 exec {name \$zlobsrv${count}, nowait, nid ${count}, out stdout_\$ZLOBSRV${count} } mxlobsrvr
 eof
 echo Successfully started \$zlobsrv${count}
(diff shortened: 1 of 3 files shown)


sandhyasun:
The change to makemsg.ksh was made since the check for pdcp was not a relaiable one. To detect that we are on a cluster the MY_NODES additional check is more reliable.

DaveBirdsall:
Looks OK to me.

core/sqf/sql/scripts/lobstart
```diff
@@ -30,8 +30,8 @@ let totalNodes=`grep 'Up' $tmp_node_status | wc -l`
 while [ ${count} -lt $totalNodes ]
 do
 sqshell -a > /dev/null << eof
-set {process \zlobsrv${count} } PERSIST_RETRIES=10,60
-set {process \zlobsrv${count} } PERSIST_ZONES=0,1
+set {process \$zlobsrv${count} } PERSIST_RETRIES=10,60
+set {process \$zlobsrv${count} } PERSIST_ZONES=${count}
 exec {name \$zlobsrv${count}, nowait, nid ${count}, out stdout_\$ZLOBSRV${count} } mxlobsrvr
```
selvaganesang: It looks like this lobstart script attempts to configure and start the lob servers in the nodes which are Up, while the change in sqgen configures it on all nodes. Let's say if we bring the cluster with one node down, the lobstart won't even configure the lob process on that node. When the node is reintegrated, will the lob processes come up. Also, my recommendation would be to configure the lobprocess in sqgen and let sqgen generate the lobstart script. That way, when the cluster is started with node down, the lob process would also be configured for the downed node and flagged as not available.

sandhyasun: Yes the lob processes will come up when a down node is reintegrated thanks to the change made in gensq.pl. Stylistically it could be better to have the lobstart and the lob config db scripts together. But functionally this setup works just fine. So  would prefer not to make further changes to rework this at this point.

DaveBirdsall:
Looks like this is ready to commit?

sandhyasun:
Yes it is ready - just waiting.

DaveBirdsall:
Waiting for what?

sandhyasun:
To be committed :-) I will go ahead and commit this one .

PR #182, opened 2015-11-19
[TRAFODION-1100] Creator of view in private schema unable to select from view

robertamarton:


core/sql/regress/catman1/TEST141
```diff
@@ -0,0 +1,394 @@
...
+-- specific language governing permissions and limitations
+-- under the License.
+--
+-- @@@ END COPYRIGHT @@@
+--
+-- ============================================================================
+-- Functionality: Extended support for views for all levels:
+--    column, object, component
+--
+-- Added in response to JIRA 1100
```
DaveBirdsall: Nit, might be more clear to say, JIRA TRAFODION-1100.

core/sql/sqlcomp/CmpSeabaseDDLview.cpp
```diff
@@ -1404,17 +1456,22 @@ short CmpSeabaseDDL::dropMetadataViews(ExeCliInterface * cliInterface)
 static bool checkAccessPrivileges(
    const ParTableUsageList & vtul,
    const ParViewColTableColsUsageList & vctcul,
+   NABoolean viewCreator,
    PrivMgrBitmap & privilegesBitmap,
    PrivMgrBitmap & grantableBitmap)
    
 {
+  BindWA bindWA(ActiveSchemaDB(),CmpCommon::context(),FALSE/*inDDL*/);
```
DaveBirdsall: Thanks for changing the indentation. This is much easier to read. (At least for me.)

core/sql/sqlcomp/CmpSeabaseDDLview.cpp
```diff
@@ -1498,32 +1577,50 @@ PrivColumnBitmap colGrantableBitmap;
...
+                                                           naTable->getObjectType(),
+                                                           naTable->getOwner(),
+                                                           privs);
+
+        if (retcode == STATUS_ERROR)
+        {
+           *CmpCommon::diags() << DgSqlCode(-CAT_UNABLE_TO_RETRIEVE_PRIVS);
+           return false;
+        }
+        pPrivInfo = &privs;
```
DaveBirdsall: So, in one code path pPrivInfo points to a heap-allocated structure while in this code path it points to a stack variable. Is there any issue with memory leaks in the first case?

DaveBirdsall:
Aside from one possible memory leak, it looks good to me.

core/sql/sqlcomp/CmpSeabaseDDLview.cpp
robertamarton: As part of creating the natable, the list of privileges the current user has is allocated and stored in the class.  This information is retained until the entry is removed.  There should not be any memory leak.

DaveBirdsall: I see... so in the one code path I alluded to, pPrivInfo is not pointing to a heap-allocated object but is pointing to a member of an object (possibly heap-allocated), and other code will take care of destroying that larger object.

DaveBirdsall:
Roberta answered my concern. Looks good to me.

sandhyasun:
Looks good to me too.

PR #183, opened 2015-11-20
[TRAFODION-1557]connot connect using Tableau

xwq:
[TRAOFDION-1557] DataSource cannot connect using Tableau
using LONG_PTR instead of long to cast a point in some MFC function.
in Windows 10 x64 , Visual Stduio 2013, long is 4 bytes and point is 8 bytes.


arvind-narain:
jenkins, retest -- failure in test162, not related - will check if it fails again.

arvind-narain:
Looks good to me. Checked on other calls to SetWindowLongPtr()  - these were the ones that needed the change. Will merge once the test passes and others have had a chance to review.

PR #184, opened 2015-11-21
[TRAFODION-1634] Update stats on Hive tables is slow

sureshsubbiah:


core/sql/ustat/hs_globals.cpp
```diff
@@ -3857,7 +3759,8 @@ Lng32 HSSample::make(NABoolean rowCountIsEstimate, // input
     // need to use a vanilla INSERT statement. Otherwise, we can use SIDETREE
     // INSERTS for better performance. A current bug in the HBase interface
     // requires the use of Upsert.
-    if (hs_globals->isHbaseTable)
+    // For Hive tables the sample table used is a Trafodion table
+    if (hs_globals->isHbaseTable || hs_globals->isHiveTable)
```
DaveBirdsall: I gather that fixing the Hive stats performance problem is a one-line fix (just this line) and everything else in this change set pertains to removing obsolete DP2 sampling code?

DaveBirdsall:
Looks good to me.

PR #2, opened 2015-06-21
Migrate log4cpp to log4cxx

mattieubro:
Fix for TRAFODION-2 Migrate log4cpp to log4cxx. Please do not merge until we can verify that libexpat is linked correctly as shown in the following
$ ldd monitor
        linux-vdso.so.1 =>  (0x00007ffff7ffe000)
        librt.so.1 => /lib64/librt.so.1 (0x000000332fa00000)
        libsblogalt.so => /h/myuser/log4cxx/core/sqf/export/lib64d/libsblogalt.so (0x00007ffff7dfa000)
        libsqlite3.so.0 => /usr/lib64/libsqlite3.so.0 (0x000000333a600000)
        liblog4cxx.so.10 => /usr/lib64/liblog4cxx.so.10 (0x00007ffff7a0f000)
        libz.so.1 => /lib64/libz.so.1 (0x000000332f200000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x000000332ea00000)
        libstdc++.so.6 => /usr/lib64/libstdc++.so.6 (0x0000003333200000)
        libm.so.6 => /lib64/libm.so.6 (0x000000332f600000)
        libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x0000003332200000)
        libc.so.6 => /lib64/libc.so.6 (0x000000332e600000)
        libdl.so.2 => /lib64/libdl.so.2 (0x000000332ee00000)
        /lib64/ld-linux-x86-64.so.2 (0x000000332e200000)
        libaprutil-1.so.0 => /usr/lib64/libaprutil-1.so.0 (0x0000003330e00000)
        libldap-2.4.so.2 => /lib64/libldap-2.4.so.2 (0x0000003343000000)
        liblber-2.4.so.2 => /lib64/liblber-2.4.so.2 (0x0000003342400000)
        libexpat.so.1 => /lib64/libexpat.so.1 (0x0000003332600000)
        libdb-4.7.so => /lib64/libdb-4.7.so (0x00007ffff7697000)
        libapr-1.so.0 => /usr/lib64/libapr-1.so.0 (0x00007ffff746a000)
        libuuid.so.1 => /lib64/libuuid.so.1 (0x0000003334200000)
        libcrypt.so.1 => /lib64/libcrypt.so.1 (0x000000333d400000)
        libresolv.so.2 => /lib64/libresolv.so.2 (0x0000003330600000)
[...]

PR #1, opened 2015-06-20
Allow scan esps to take # of partitions.

nonstop-qfchen:
Fix [TRAFODION-1]: the number of Scan level ESP processes is too restrictive.

PR #186, opened 2015-11-24
[TRAFODION-1641] Fix MDAM costing bugs

DaveBirdsall:
This set of changes is a partial fix to [TRAFODION-1641]. It does resolve the issue described in the JIRA (that is, the query now gets a better choice of MDAM plans), however as in many Optimizer changes, there are side effects that still remain to be fully evaluated. Follow-on work will address these if judged to be significant.

JIRA 1641 describes two bugs. This set of changes fixes bug #1: MDAM costing now takes into account the cumulative cost of each key column materialized. This set of changes partially fixes bug #2: It removes a bit of code that set the stop column for all MDAM plans to the last column (which caused MDAM plans to always traverse all keys, which can be grossly inefficient).

For queries not containing RangeSpec predicates with OR subtrees, these changes will allow MDAM to traverse a leading proper subset of columns, which often will be far more efficient. (For RangeSpec predicates containing OR subtrees, there is a heuristic in the code that presently causes all key columns to be traversed. This could be improved, and is part of the follow-on work.)

A side effect of these changes is that the cost computed for MDAM plans of 2 or more key columns will be somewhat higher (and much higher when traversing key columns with high UEC). This may cause MDAM to be chosen less often than previously.

One example of this occurs in the regression test suites (seabase/TEST010): A single query there now chooses a full table scan instead of an index scan using MDAM. Not using MDAM is a good decision for this particular plan. I have not investigated, however, why a full table scan is chosen over a full index scan.

### DIFF
--- a/core/sql/optimizer/ScanOptimizer.cpp
+++ b/core/sql/optimizer/ScanOptimizer.cpp
@@ -553,11 +553,22 @@ private:
   // Estimated rows iff multiple probes
   CostScalar multiProbesDataRows_;
   // >>>>>>>>>>>>>>>>> Current prefix related members <<<<<<<<<<<<<<<<<
-  // # of subsets of each effective probe
+  // # of subsets of each effective probe at the current level
   CostScalar prefixSubsets_;
+  // cumulative # of subsets of each effective probe
+  // Why do we care? MDAM is a recursive algorithm. It first materializes
+  // values for the first key column. For each of those, it materializes 
+  // values for the second key column. And so on. Each of these levels adds
+  // progressively more cost which we must take into account. If we look
+  // only at prefixSubsets_ (that is, the current column level), we may be
+  // misled into thinking that adding more levels of column traversal is
+  // free. Which it is not. Moreover, as the number of rows approaches the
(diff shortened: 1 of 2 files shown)


zellerh:
The fix looks ok to me, but after spending an hour or two trying to understand what's going on in MDAM costing I still don't have much of a clue. The code seems nicely structured and modular, but the comments mostly explain what each method does, not what doing that means. What it does I can see from the code, too, but figuring out what it means is much harder. The ColumnOrderList::getSingleSubsetOrderForMDAM() method (file mdam.cpp) seems to allow any range specs in the prefix column. This seems good to me, but it also seems to contradict what the method does with IN lists (which are just another flavor or RangeSpecs). All this may be wrong, due my limited understanding, and it is not directly related to this specific fix.

nonstop-qfchen:
I checked the case with the OSIM and found setting CQD MDAM_APPLY_RESTRICTION_CHECK to 'ON' can be used as a workaround. 

Together with Dave's fix, we can also work on an improvement in this restriction heuristic to avoid MDAM scan against high UEC key column without a key predicate. 

--Qifan

PR #189, opened 2015-11-25
[TRAFODION-1647]Put unified copyrigit into Trafci

kevinxu021:
minus changes for copyright


core/conn/trafci/build.xml
```diff
@@ -148,6 +148,8 @@
 
        <replace file="${utils.dir}/trafci.cmd" token="@prodver@" value="${my_env.TRAFODION_VER_PROD}"/>
        <replace file="${utils.dir}/trafci.cmd" token="@copyrighthdr@" value="${my_env.PRODUCT_COPYRIGHT_HEADER}"/>
+       <replace file="${utils.dir}/trafci.cmd" token="##prodver##" value="${my_env.TRAFODION_VER_PROD}"/>
+       <replace file="${utils.dir}/trafci.cmd" token="##copyrighthdr##" value="${my_env.PRODUCT_COPYRIGHT_HEADER}"/>
```
hegdean: The 2 new lines are redundant as the previous 2 lines (149 and 150) do the same thing

core/conn/trafci/utils/trafci.sh
```diff
@@ -27,6 +27,8 @@
 #
 
 TRAFCI_CLASSPATH="##TRAFCI_CLASSPATH##"
+export TRAFODION_VER_PROD="##prodver##"
+export PRODUCT_COPYRIGHT_HEADER="##copyrighthdr##"
```
hegdean: How is ##prodver## being substituted?

kevinxu021: Thanks for comments. Already make changes in build.xml.

arvind-narain:
Thanks for the change Kevin. 

Since trafci.sh is a checked-in/versioned file, build step should not modify it. You could follow the same steps as trafci.cmd (144-147) by creating a versioned template file or some other way to avoid the error ( Error: Build modified 1 versioned file(s) ).

hegdean:
you need to make additional changes to build.xml so that we don't package and distribute the template file along with the final jar file. Please  fix the .gitignore file as well 

Thanks

kevinxu021:
Thanks for the comments. It'll has nothing with installation output. But ok, the changes will be made.
For build.xml, have you noticed trafci_install.sh? It specified particular files which will be installed into BIN folder, then *-tmpl file will not be included.
For .gitignore file, trafci.sh will not be tracking again and only if you run 'git add'. Assume you add a new temp file aaa.txt in utils/, you don't have to add it into .gitignore and only if you do 'git add aaa.txt', it will be tracking. Please correct me if it's incorrect. Thanks!

arvind-narain:
Thanks for the change Kevin - will merge it shortly.

Regarding .gitignore - I believe it is used to make sure that files untracked by Git remain untracked.

Following commands are executed after a build to check on tracked/untracked files:
   modified=$(git ls-files -m | wc -l)
   unignored=$(git ls-files -o --exclude-standard | wc -l)

PR #192, opened 2015-11-26
[TRAFODION-1558]Tableau returns to tables or data

xwq:
[TRAFODION-1558]Used driver connection with TCP:<ip>/<port>. Connection
succeeded but no schemas or tables are listed. It's was casued by the
driver use long (4 bytes) to store the points (8 bytes) which caused data
loss. The fix is use LONG_PTR insteaded.

### DIFF
--- a/win-odbc64/odbcclient/drvr35/cstmt.cpp
+++ b/win-odbc64/odbcclient/drvr35/cstmt.cpp
@@ -3389,7 +3389,7 @@ BOOL CStmt::setFetchOutputPerf(long rowsFetched, SQL_DataValue_def*& outputDataV
     m_FetchDataValue.numberOfRows = rowsFetched;
     if(rowsFetched > 0)
     {
-        m_FetchDataValue.rowAddress = new unsigned long[rowsFetched];
+        m_FetchDataValue.rowAddress = new LONG_PTR[rowsFetched];
         if (m_FetchDataValue.rowAddress != NULL)
             return TRUE;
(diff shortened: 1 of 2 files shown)


win-odbc64/odbcclient/drvr35/cstmt.cpp
```diff
@@ -3389,7 +3389,7 @@ BOOL CStmt::setFetchOutputPerf(long rowsFetched, SQL_DataValue_def*& outputDataV
     m_FetchDataValue.numberOfRows = rowsFetched;
     if(rowsFetched > 0)
     {
-        m_FetchDataValue.rowAddress = new unsigned long[rowsFetched];
+        m_FetchDataValue.rowAddress = new LONG_PTR[rowsFetched];
```
selvaganesang: Should it be VOID_PTR or (void *) instead.

xwq: The issue here is that the long is 4 bytes in Windows x64.  windows can ensure the LONG_PTR is 8 bytes in x64 system and 4 bytes in x32 system. I use LONG_PTR because of a lot Windows API use it . There is no VOID_PTR type.
I can use (void *) instead if you insist .

selvaganesang: Thanks for your prompt response. Use of LONG_PTR is justified very well. No need to change it to void *

PR #194, opened 2015-11-27
[TRAFODION-1645] add support to retrieve region stats through sql queries

anoopsharma00:
Details in JIRA.

Examples of sql commands in testcase regress/seabase/TEST002,EXPECTED002


anoopsharma00:
retest

anoopsharma00:
Jenkins retest

svarnau:
jenkins, retest

zellerh:
I didn't look at the details of how the TDB is implemented, but the overall idea of a "get region stats" looks good to me. Another way would have been to do this as a TMUDF, that would have avoided the "using" clause, but our current TMUDF interface may not have all the necessary interfaces. Once we support SQL in TMUDFs, we could put a simple TMUDF on top of this function.

sandhyasun:
In the regression test,  you have some calls to "hbase shell flush" . Looks like you're using swhbase . But that  may not work on clusters or official build machines. You may need to resplace those with "regrhbase" calls.

core/sql/comexe/ComTdbExeUtil.h
```diff
@@ -3452,6 +3453,122 @@ class ComTdbExeUtilHBaseBulkUnLoad : public ComTdbExeUtil
   NABasicPtr   snapshotSuffix_;                         // 30 - 37
   char         fillersExeUtilHbaseUnLoad_[16];          // 38 - 53
 };
+
+static const ComTdbVirtTableColumnInfo comTdbRegionStatsVirtTableColumnInfo[] =
+  {
+    { "CATALOG_NAME",                   0, COM_USER_COLUMN, REC_BYTE_F_ASCII,  256, FALSE, SQLCHARSETCODE_UTF8 , 0, 0, 0, 0, 0, 0, 0, COM_NO_DEFAULT, "",NULL,NULL, COM_UNKNOWN_DIRECTION_LIT, 0  },
```
DaveBirdsall: Just curious: Why fixed length CHAR instead of VARCHAR?

core/sql/executor/ExExeUtilGet.cpp
```diff
@@ -5315,4 +5358,745 @@ ExExeUtilHiveMDaccessPrivateState::~ExExeUtilHiveMDaccessPrivateState()
...
+					(char*)"", //exe_util_tdb.zkPort(),
+                                        jniDebugPort,
+                                        jniDebugTimeout);
+
+  regionInfoList_ = NULL;
+  
+  tableName_ = new(glob->getDefaultHeap()) char[2000];
+
+  // get hbase rootdir location
+  hbaseRootdir_ = new(glob->getDefaultHeap()) char[1000];
```
DaveBirdsall: Path lengths in Linux can be 1024; will 1000 be enough?

core/sql/executor/ExExeUtilGet.cpp
```diff
@@ -5315,4 +5358,745 @@ ExExeUtilHiveMDaccessPrivateState::~ExExeUtilHiveMDaccessPrivateState()
...
+  return 0;
+}
+
+short ExExeUtilRegionStatsTcb::populateStats
+(Int32 currIndex, NABoolean nullTerminate)
+{
+  str_pad(stats_->catalogName, sizeof(stats_->catalogName), ' ');
+  str_cpy_all(stats_->catalogName, catName_, strlen(catName_));
+  if (nullTerminate)
+    stats_->catalogName[strlen(catName_)] = 0;
```
DaveBirdsall: We declared this field as 256 bytes; should we have declared them 257 to allow for the null?

core/sql/executor/ExExeUtilGet.cpp
```diff
@@ -5315,4 +5358,745 @@ ExExeUtilHiveMDaccessPrivateState::~ExExeUtilHiveMDaccessPrivateState()
...
+    {
+      str_cpy_all(stats_->regionName, regionInfo, 
+                  (Lng32)(sep1 - regionInfo)); 
+
+      if (nullTerminate)
+        stats_->regionName[sep1 - regionInfo] = 0;
+    }
+  
+  char * sepStart = sep1;
+  stats_->numStores = getEmbeddedNumValue(sepStart, '|', FALSE);
```
DaveBirdsall: This code seems to be assuming a particular format of text output (from HBase I assume). What happens if that format changes? Seems like we don't detect that.

core/sql/parser/ParKeyWords.cpp
```diff
@@ -1030,6 +1031,7 @@ ParKeyWord ParKeyWords::keyWords_[] = {
   ParKeyWord("STATEMENT",          TOK_STATEMENT,   NONRESTOKEN_),
   ParKeyWord("STATIC",             TOK_STATIC,      NONRESTOKEN_),
   ParKeyWord("STATISTICS",         TOK_STATISTICS,  SECOND_|NONRESTOKEN_),
+  ParKeyWord("STATS",              TOK_STATS,       NONRESTOKEN_),
```
DaveBirdsall: Should we use the STATISTICS keyword, since it is already available?

DaveBirdsall:
Are we ready to commit this?

sandhyasun:
One minor issue - the  regression test still uses swhbase - I had commented on this earlier. Anoop needs to check if this will be an issue on machines that don't use install_local_hadoop. Normally we have to use regrhbase.ksh

DaveBirdsall:
Just checking: Sandhya, did Anoop's latest change satisfy your concern? If so, I'm guessing this is ready to commit?

sandhyasun:
Yes ready to commit.

PR #195, opened 2015-11-30
fix JIRA 1654 for Trafodion

nonstop-qfchen:
Hi,

Could you please review the fix? 

The idea is to compute the product of UECs for key columns without predicates and consider MDAM when that product is within a fraction of the total UEC. 

CQDs re-arranged/created for this new and the existing restriction. 

MDAM_APPLY_RESTRICTION_CHECK:
  // Specify whic additional restriction check to apply
  //  0: no check
  //  1: apply majority of keys with predicates check
  //  2: apply total UECs on keyless key columns check (default)
  //  3: apply both 1) and 2)

MDAM_TOTAL_UEC_CHECK_UEC_THRESHOLD:
 // A threshold of total UECs on keyless key columns above which MDAM will not be considered.
  // The threshold is expressed as a percentage of the total RC.
 Default to 1 percent of the total RC. 

MDAM_TOTAL_UEC_CHECK_MIN_RC_THRESHOLD:
// A threshold of minitmal RC above which the above total UEC check will be applied. Default to 10,000 rows

### DIFF
--- a/core/sql/optimizer/ScanOptimizer.cpp
+++ b/core/sql/optimizer/ScanOptimizer.cpp
@@ -317,7 +317,14 @@ void MdamTrace::setLevel(enum MdamTraceLevel l)
 #endif // if MDAM_TRACE
 // LCOV_EXCL_STOP
 
-static NABoolean checkMDAMadditionalRestriction(const ColumnOrderList& keyPredsByCol,const CollIndex&  lastColumnPosition, CollIndex&  noOfmissingKeyColumns, CollIndex&  presentKeyColumns)
+enum restrictCheckStrategy { MAJORITY_WITH_PREDICATES=1, TOTAL_UECS=2, BOTH=3 };
+
+static NABoolean checkMDAMadditionalRestriction(
+    const ColumnOrderList& keyPredsByCol,
+    const CollIndex& lastColumnPosition, 
+    const Histograms& hist,
+    restrictCheckStrategy strategy,
+    CollIndex&  noOfmissingKeyColumns, CollIndex&  presentKeyColumns)
 {
    KeyColumns::KeyColumn::KeyColumnType typeOfRange = KeyColumns::KeyColumn::EMPTY;
    CollIndex index = 0;
@@ -331,8 +338,21 @@ static NABoolean checkMDAMadditionalRestriction(const ColumnOrderList& keyPredsB
    if ( mtd_mdam_uec_threshold < 0 )
      checkLeadingDivColumns = FALSE;
 
+   CostScalar totalRC = hist.getRowCount().getCeiling();
+
+   float totalUEC_threshold = 1;
+
+   Lng32 minRC = (ActiveSchemaDB()->getDefaults()).getAsLong(MDAM_TOTAL_UEC_CHECK_MIN_RC_THRESHOLD);
+
+   if ( totalRC > minRC )
+     (ActiveSchemaDB()->getDefaults()).getFloat(MDAM_TOTAL_UEC_CHECK_UEC_THRESHOLD, totalUEC_threshold);
+
+   totalUEC_threshold *= totalRC.getValue();
+
    NABoolean isLeadingDivisionColumn = FALSE;
    NABoolean isLeadingSaltColumn = FALSE;
+
+   CostScalar totalUecsForKeylessKeyColumns = 1;
 
    for (index = 0; index < lastColumnPosition; index++)
    {
@@ -367,6 +387,11 @@ static NABoolean checkMDAMadditionalRestriction(const ColumnOrderList& keyPredsB
            )
            noOfmissingKeyColumns++;
 
+        // accumulate the product of uecs for columns without predicates
+                   
+        totalUecsForKeylessKeyColumns *= hist.getColStatsForColumn(
+                   keyPredsByCol.getKeyColumnId(index)).getTotalUec().getCeiling();
+
      } else {
 
         checkLeadingDivColumns = FALSE;
@@ -374,8 +399,21 @@ static NABoolean checkMDAMadditionalRestriction(const ColumnOrderList& keyPredsB
 
      }
    }     
-   if(presentKeyColumns > noOfmissingKeyColumns) return TRUE;
-   return FALSE;
+   switch ( strategy ) {
+     case MAJORITY_WITH_PREDICATES:
+        return (presentKeyColumns > noOfmissingKeyColumns);
+
+     case TOTAL_UECS:
+        return ( totalUecsForKeylessKeyColumns < totalUEC_threshold );
+
+     case BOTH:
+        return ( presentKeyColumns > noOfmissingKeyColumns &&
+                 totalUecsForKeylessKeyColumns < totalUEC_threshold );
+
+     default:
+       return FALSE;
+  }
+  return FALSE;
 }
 
 // stack allocated only
@@ -8710,12 +8748,25 @@ void MDAMCostWA::computeDisjunct()
 
   CollIndex noOfmissingKeyColumnsTot = 0;
   CollIndex presentKeyColumnsTot = 0;
+
+
+  const IndexDesc *idesc = optimizer_.getFileScan().getIndexDesc();
+  const ColStatDescList& csdl = idesc->getPrimaryTableDesc()->getTableColStats();
+  Histograms hist(csdl);
+	 
+  Lng32 checkOption = (ActiveSchemaDB()->getDefaults()).getAsLong(MDAM_APPLY_RESTRICTION_CHECK);
+
   if(CURRSTMT_OPTDEFAULTS->indexEliminationLevel() != OptDefaults::MINIMUM
      && (!mdamForced_)
 	 && (CmpCommon::getDefault(RANGESPEC_TRANSFORMATION) == DF_ON )
-	 && (CmpCommon::getDefault(MDAM_APPLY_RESTRICTION_CHECK) == DF_ON )
-	 &&
-   (!checkMDAMadditionalRestriction(keyPredsByCol,optimizer_.computeLastKeyColumnOfDisjunct(keyPredsByCol),noOfmissingKeyColumnsTot,presentKeyColumnsTot))
+	 && checkOption >= 1 
+	 && (!checkMDAMadditionalRestriction(
+                                    keyPredsByCol,
+                                    optimizer_.computeLastKeyColumnOfDisjunct(keyPredsByCol),
+                                    hist,
+                                    (restrictCheckStrategy)checkOption,
+                                    noOfmissingKeyColumnsTot,
+                                    presentKeyColumnsTot))
    )
   {
         
--- a/core/sql/sqlcomp/nadefaults.cpp
+++ b/core/sql/sqlcomp/nadefaults.cpp
@@ -2160,7 +2160,7 @@ SDDkwd__(ISO_MAPPING,           (char *)SQLCHARSETSTRING_ISO88591),
   XDDflt__(MC_SKEW_SENSITIVITY_THRESHOLD,        "0.1"),
 
 
-  DDkwd__(MDAM_APPLY_RESTRICTION_CHECK,	"OFF"),
+  DDui___(MDAM_APPLY_RESTRICTION_CHECK,	            "2"),
   DDflt0_(MDAM_CPUCOST_NET_OVH,			"2000."),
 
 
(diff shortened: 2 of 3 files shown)


DaveBirdsall:
The motivation for this change seems to be that when there is a key column lacking predicates, MDAM is not taking this into account. Please take a look at the fix to JIRA [TRAFODION-1641] (here: https://github.com/apache/incubator-trafodion/pull/186) as I believe that addresses this problem.

core/sql/optimizer/ScanOptimizer.cpp
```diff
@@ -331,9 +338,22 @@ static NABoolean checkMDAMadditionalRestriction(const ColumnOrderList& keyPredsB
...
+
+   if ( totalRC > minRC )
+     (ActiveSchemaDB()->getDefaults()).getFloat(MDAM_TOTAL_UEC_CHECK_UEC_THRESHOLD, totalUEC_threshold);
+
+   totalUEC_threshold *= totalRC.getValue();
+
    NABoolean isLeadingDivisionColumn = FALSE;
    NABoolean isLeadingSaltColumn = FALSE;
 
+   CostScalar totalUecsForKeylessKeyColumns = 1;
```
DaveBirdsall: Not sure what a keyless key column is. I'm guessing you meant, "key column lacking predicates"?

nonstop-qfchen:
Yes, we have two defenses to assure MDAM is suitable: the heuristics and
the cost.  Both are needed as both are not perfect due to various reasons.

This fix is to improve the heuristics part of the defense.

On Mon, Nov 30, 2015 at 11:06 AM, DaveBirdsall <notifications@github.com>
wrote:

> The motivation for this change seems to be that when there is a key column
> lacking predicates, MDAM is not taking this into account. Please take a
> look at the fix to JIRA TRAFODION-1641 as I believe that addresses this
> problem.
>
> —
> <https://github.com/apache/incubator-trafodion/pull/195#issuecomment-160691364>
> .
>



-- 
Regards, --Qifan

core/sql/optimizer/ScanOptimizer.cpp
nonstop-qfchen: Good catch. Will make a change.

On Mon, Nov 30, 2015 at 11:08 AM, DaveBirdsall <notifications@github.com>
wrote:

> In core/sql/optimizer/ScanOptimizer.cpp
> <https://github.com/apache/incubator-trafodion/pull/195#discussion_r46171726>
> :
>
> >     NABoolean isLeadingDivisionColumn = FALSE;
> >     NABoolean isLeadingSaltColumn = FALSE;
> >
> > +   CostScalar totalUecsForKeylessKeyColumns = 1;
>
> Not sure what a keyless key column is. I'm guessing you meant, "key column
> lacking predicates"?
>
> —
> <https://github.com/apache/incubator-trafodion/pull/195/files#r46171726>.
>



-- 
Regards, --Qifan

core/sql/optimizer/ScanOptimizer.cpp
```diff
@@ -374,8 +399,21 @@ static NABoolean checkMDAMadditionalRestriction(const ColumnOrderList& keyPredsB
 
      }
    }     
-   if(presentKeyColumns > noOfmissingKeyColumns) return TRUE;
-   return FALSE;
+   switch ( strategy ) {
+     case MAJORITY_WITH_PREDICATES:
+        return (presentKeyColumns > noOfmissingKeyColumns);
```
DaveBirdsall: I am not sure I see the point of this strategy. Suppose I have a 6 column key and a query with predicates on 4 columns. It makes a huge difference whether the predicates are on the leading columns (better) vs. the trailing columns (worse). This heuristic does not take the column order into account.

core/sql/optimizer/ScanOptimizer.cpp
```diff
@@ -374,8 +399,21 @@ static NABoolean checkMDAMadditionalRestriction(const ColumnOrderList& keyPredsB
...
+
+     case TOTAL_UECS:
+        return ( totalUecsForKeylessKeyColumns < totalUEC_threshold );
```
DaveBirdsall: This seems like a reasonable heuristic.

DaveBirdsall:
I'm good with these changes, though I am not sure I see the value in the MAJORITY_WITH_PREDICATES heuristic.

core/sql/optimizer/ScanOptimizer.cpp
nonstop-qfchen: I agree and we are going to use the new strategy (total UEC) as the default
one.

Thanks

On Mon, Nov 30, 2015 at 11:23 AM, DaveBirdsall <notifications@github.com>
wrote:

> In core/sql/optimizer/ScanOptimizer.cpp
> <https://github.com/apache/incubator-trafodion/pull/195#discussion_r46173600>
> :
>
> > @@ -374,8 +399,21 @@ static NABoolean checkMDAMadditionalRestriction(const ColumnOrderList& keyPredsB
> >
> >       }
> >     }
> > -   if(presentKeyColumns > noOfmissingKeyColumns) return TRUE;
> > -   return FALSE;
> > +   switch ( strategy ) {
> > +     case MAJORITY_WITH_PREDICATES:
> > +        return (presentKeyColumns > noOfmissingKeyColumns);
>
> I am not sure I see the point of this strategy. Suppose I have a 6 column
> key and a query with predicates on 4 columns. It makes a huge difference
> whether the predicates are on the leading columns (better) vs. the trailing
> columns (worse). This heuristic does not take the column order into account.
>
> —
> <https://github.com/apache/incubator-trafodion/pull/195/files#r46173600>.
>



-- 
Regards, --Qifan

nonstop-qfchen:
It was put into the place for HPIT queries.

On Mon, Nov 30, 2015 at 11:30 AM, DaveBirdsall <notifications@github.com>
wrote:

> I'm good with these changes, though I am not sure I see the value in the
> MAJORITY_WITH_PREDICATES heuristic.
>
> —
> <https://github.com/apache/incubator-trafodion/pull/195#issuecomment-160697942>
> .
>



-- 
Regards, --Qifan

core/sql/sqlcomp/nadefaults.cpp
```diff
@@ -2180,6 +2180,10 @@ SDDkwd__(ISO_MAPPING,           (char *)SQLCHARSETSTRING_ISO88591),
   XDDkwd__(MDAM_SCAN_METHOD,			"ON"),
 
   DDflt0_(MDAM_SELECTION_DEFAULT,		"0.5"),
+
+  DDflt0_(MDAM_TOTAL_UEC_CHECK_MIN_RC_THRESHOLD, "10000"),
+  DDflt0_(MDAM_TOTAL_UEC_CHECK_UEC_THRESHOLD,	 "0.01"),
```
zellerh: IMHO the 1% threshold used here is too low. We are saying that if MDAM can eliminate "only" 98% of the values then it isn't even worth thinking about the possibility of using MDAM? Or am I misunderstanding how the heuristic works?

The other problem with this heuristic is of course that it doesn't care about which predicates we have. Let's say we have 3 leading key columns, a,b,c, all with integer values 0...98 (UEC=99). Let's assume the total row count is 100,000.

Case 1:
  - Predicate is WHERE c = 55
  - UEC of columns w/o pred (a, b) is 99*99 = 9801 (no MC stats considered!)
  - Threshold: 1,000 (1 % of 100,000)
  - MDAM is not considered

Case 2:
  - Predicate is WHERE b between 1 and 90
  - UEC of columns w/o pred (a) is 99
  - Threshold is again 1,000
  - MDAM is considered, even though this query is not obviously better suited for MDAM than case 1.

Do we really need these heuristics, or could we just let costing decide? If we do, should we a) consider MC stats for the predicate-less columns and b) consider increasing MDAM_TOTAL_UEC_CHECK_UEC_THRESHOLD to a higher value?

DaveBirdsall: Hi Hans,

Actually, your Case 2 is better suited for MDAM than Case 1. In Case 2, we'll materialize each of the values of A (99 of them) and do a begin/end subset from B = 1 to B = 90. Just 99 subsets. In Case 1, we'll materialize each of the values of A (done 99 times), and for each value of A, materialize the values of B (9801 times, or less if there are functional dependencies which MC stats would show), and then for each of these, do a begin/end subset on C. Each materialization is potentially a random I/O, not to mention some significant path length (compared to the path length of just traversing to the next row sequentially). But more to your point, the costing code takes into account these differences.

nonstop-qfchen: Hi Hans and Dave,

Thanks a lot for the review comments.

There is a piece of detail that I need to describe.

The method checkMDAMadditionalRestriction() takes the argument
lastColumnPosition which is 1 plus the index of the stopping column to
consider the heuristic.  In the call, the actual argument is 6, or to the
column BRUKRS that has the predicate defined.

For the two cases for discussion, the index of the stopping columns will be
3 for case 1 and to 2 for case 2.  In other words, the level of protection
would be different.

On the use of MC, I believe MC stats will further reduce the total UEC.
However, currently, the MCs are generated by default for any prefixes of
the key columns. So if we want to integrate the MC stats, the RC count can
be reduced only by the MC on the longest prefix of key columns that are
free of predicate. I think this is a good suggestion and will try to see if
it can be arranged.

Third, I am open to suggestion to fine tune the threshold (if needed, say
to 10%?).

Lastly,  on whether this heuristic should be put into the place.  My view
here is that the costing is never perfect for MDAM and therefor extra
orthogonal protections are always good. In this special customer case, the
MDAM is rejected even if the threshold CQD is raised to 100%. This
indicates that the heuristic works.

Thanks --Qifan

 key_columns ............ _SALT_, RYEAR, DOCLN, DOCNR, RLDNR, RBUKRS, RCLNT
  executor_predicates .... (RYEAR = '2014') and (RBUKRS = 'A019')
  begin_key .............. (_SALT_ = (\:_sys_HostVarLoHashPart Hash2Distrib
                             32)), (RYEAR = '2014'), (DOCLN = '<min>'),
                             (DOCNR = '<min>'), (RLDNR = '<min>'),
                             (RBUKRS = 'A019'), (RCLNT = '<min>')
  end_key ................ (_SALT_ = (\:_sys_HostVarHiHashPart Hash2Distrib
                             32)), (RYEAR = '2014'), (DOCLN = '<max>'),
                             (DOCNR = '<max>'), (RLDNR = '<max>'),
                             (RBUKRS = 'A019'), (RCLNT = '<max>')



#0  checkMDAMadditionalRestriction (keyPredsByCol=...,
lastColumnPosition=@0x7ffffffeb8ec,
    hist=..., strategy=TOTAL_UECS, noOfmissingKeyColumns=@0x7ffffffeb8e4,
    presentKeyColumns=@0x7ffffffeb8e0) at ../optimizer/ScanOptimizer.cpp:329
329   KeyColumns::KeyColumn::KeyColumnType typeOfRange =
KeyColumns::KeyColumn::EMPTY;
(gdb) p lastColumnPosition
$1 = (const CollIndex &) @0x7ffffffeb8ec: 6

On Mon, Nov 30, 2015 at 5:56 PM, DaveBirdsall <notifications@github.com>
wrote:

> In core/sql/sqlcomp/nadefaults.cpp
> <https://github.com/apache/incubator-trafodion/pull/195#discussion_r46222783>
> :
>
> > @@ -2180,6 +2180,10 @@ SDDkwd__(ISO_MAPPING,           (char *)SQLCHARSETSTRING_ISO88591),
> >    XDDkwd__(MDAM_SCAN_METHOD,			"ON"),
> >
> >    DDflt0_(MDAM_SELECTION_DEFAULT,		"0.5"),
> > +
> > +  DDflt0_(MDAM_TOTAL_UEC_CHECK_MIN_RC_THRESHOLD, "10000"),
> > +  DDflt0_(MDAM_TOTAL_UEC_CHECK_UEC_THRESHOLD,	 "0.01"),
>
> Hi Hans,
>
> Actually, your Case 2 is better suited for MDAM than Case 1. In Case 2,
> we'll materialize each of the values of A (99 of them) and do a begin/end
> subset from B = 1 to B = 90. Just 99 subsets. In Case 1, we'll materialize
> each of the values of A (done 99 times), and for each value of A,
> materialize the values of B (9801 times, or less if there are functional
> dependencies which MC stats would show), and then for each of these, do a
> begin/end subset on C. Each materialization is potentially a random I/O,
> not to mention some significant path length (compared to the path length of
> just traversing to the next row sequentially). But more to your point, the
> costing code takes into account these differences.
>
> —
> <https://github.com/apache/incubator-trafodion/pull/195/files#r46222783>.
>



-- 
Regards, --Qifan

zellerh: Thanks, Qifan, you make a good point. Sorry, my example isn't very convincing. Just wanted to run these thoughts by you, if you feel that 0.01 is the right number then I'm ok to go with it. Certainly it's a lot better than just counting the columns! Changing the number later will be relatively easy.

sureshsubbiah:
Merging since all comments have been addressed to reviewers satisfaction and tests pass.

PR #196, opened 2015-11-30
TRAFODION-1626 fix for dtmci statis -j

trinakrug:
Errors were not being checked after the TMSTATS call for the "json" case.

### DIFF
--- a/core/sqf/src/tm/tools/dtmci.cpp
+++ b/core/sqf/src/tm/tools/dtmci.cpp
@@ -472,23 +472,22 @@ void process_tmstats_node(bool pv_reset, int32 pv_nid, bool json)
 
     lv_error = TMSTATS(pv_nid, &lv_stats, pv_reset);
 
-    if(json==false) {
-        if (lv_error)
-            printf("Node %d\t**Error %d.", pv_nid, lv_error);
-        else
-        {
-            printf("Node %d:", pv_nid);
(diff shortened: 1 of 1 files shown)


core/sqf/src/tm/tools/dtmci.cpp
```diff
@@ -472,23 +472,22 @@ void process_tmstats_node(bool pv_reset, int32 pv_nid, bool json)
...
-            // Pool statistics
-            printf("\n  Txn Pool:\t");
-            print_poolStats(lv_stats.iv_transactionPool_stats);
-            printf("\n  Thrd Pool:\t");
-            print_poolStats(lv_stats.iv_threadPool_stats);
-            printf("\n  RMMsg Pool:\t");
-            print_poolStats(lv_stats.iv_RMMessagePool_stats);
-        }
+    if (lv_error) {
+        printf("Node %d\t**Error %d.\n", pv_nid, lv_error);
```
DaveBirdsall: Should the error be reported as a JSON document in the JSON case?

venkat1m:
The JSON output is primarily used by a python script parses the stdout and will try to parse as json. If the result is not a valid JSON, the python script will mark that as an error and continue to the next iteration.

A normal json response with values as 0 will be misleading. The python script tries to convert the values as numbers. So a null will not work in the python code.

So best to just return a error text that is not json. The script that uses the JSON response can handle the output.

PR #197, opened 2015-12-01
TRAFODION-1643

mashengchen:
after hbase-0.98.8(include 0.98.8), the instantiation process of ScanInfo(class) is different from before, which need KeepDeletedCells(enum) as parameter instead of boolean.

### DIFF
--- a/core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/regionserver/transactional/TrxTransactionState.java
+++ b/core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/regionserver/transactional/TrxTransactionState.java
@@ -34,7 +34,6 @@
 package org.apache.hadoop.hbase.regionserver.transactional;
 
 import java.io.IOException;
-
 import java.lang.Class;
 
 import java.lang.reflect.Constructor;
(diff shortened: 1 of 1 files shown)


mashengchen:
jenkins, retest

DaveBirdsall:
Looks good to me. Can someone with experience with HBase APIs please review?

DaveBirdsall:
Just checking: Is this ready to commit?

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/regionserver/transactional/TrxTransactionState.java
```diff
@@ -126,15 +130,49 @@
...
-
 	if (sb_sqm_98_1) {
 	    LOG.info("Got info of Class ScanQueryMatcher for HBase 98.1");
 	}
 	if (sb_sqm_98_4) {
 	    LOG.info("Got info of Class ScanQueryMatcher for HBase 98.4");
 	}
+	
+	
+	Class scaninfoClazz = null;
```
selvaganesang: I believe scaninfoclazz should also be made static. Can you please check if the scanInfoConstructor becomes invalid if scanInfoclazz gets gc-ed.

selvaganesang:
I will be merging this PR by end of day. Please review and send in your comments before.

DaveBirdsall:
Looks good to me.

PR #198, opened 2015-12-01
[TRAFODION-1659] Need Batch test cases in JDBC Phoenix tests

ryzuo:
1. BatchInsert positive test case added into phoenix
2. BatchInsert negative test case added into phoenix


DaveBirdsall:
Looks good to me.

PR #199, opened 2015-12-01
fix JIRA 1654 (rework)

nonstop-qfchen:
Hi,

This rework set addresses the following review comments. 

1. A variable is not properly named
2. MC UEC is not used in computation of the total UEC. 
3. Adjustment of the threshold CQD MDAM_TOTAL_UEC_CHECK_UEC_THRESHOLD to 20%

Please review. 

Thanks


core/sql/optimizer/ScanOptimizer.cpp
```diff
@@ -387,28 +390,61 @@ static NABoolean checkMDAMadditionalRestriction(
...
         checkLeadingDivColumns = FALSE;
         presentKeyColumns++;
 
+        // If the set of key columns without predicate is not empty, fetch the MC UEC 
+        // for the entire set. If the MC UEC exists, replace the current accumualted
+        // total UEC with the MC UEC.
+        //
+        // We will set the set to empty so that the fetching MC UEC logic will not kick in 
+        // until a new key column without predicates is seen.
+        if ( currentPredicatelessKeyColumnGroup.entries() > 0 ) {
```
zellerh: I guess we could use > 1 here, to skip single columns?

zellerh:
Thanks for the rework, I hope this will prove to be a much better heuristic than the current version! I have one minor comment, but I would suggest fixing that later, if you want to, no need for another commit now.

nonstop-qfchen:
Hi Hans,

Thanks a lot of the review.  I'll submit a fix (>0 ==> >1) in next commit.

--Qifan

On Tue, Dec 1, 2015 at 4:15 PM, Hans Zeller <notifications@github.com>
wrote:

> Thanks for the rework, I hope this will prove to be a much better
> heuristic than the current version! I have one minor comment, but I would
> suggest fixing that later, if you want to, no need for another commit now.
>
> —
> <https://github.com/apache/incubator-trafodion/pull/199#issuecomment-161112997>
> .
>



-- 
Regards, --Qifan

DaveBirdsall:
Looks good to me. I can commit it, or perhaps Hans, you'd like the practice?

PR #202, opened 2015-12-03
[TRAFODION-1656]Remove DISCONNECTED& refactor some code

kevinxu021:
It's useless to be DISCONNECTED since nothing is done between DISCONNECTED and AVALIBLE.
And seems to be not reasonable to sleep 30s.
Tested on while(true){connect/close} without sleeping and it successfully did hundreds times.

### DIFF
--- a/dcs/src/main/java/org/trafodion/dcs/master/listener/ConnectReply.java
+++ b/dcs/src/main/java/org/trafodion/dcs/master/listener/ConnectReply.java
@@ -163,29 +163,20 @@ class ConnectReply {
             for(int i=0; i < randomPicks; i++){
                 while(true){
                     index = random.nextInt();
-                    index = index > 0? index : -index;
+                    index = Math.abs(index);
                     index %= length;
                     if (indexArr[index] != index) break;
                 }
                 indexArr[index] = index;
-                maxIndex = index > maxIndex ? index : maxIndex;
+                maxIndex = Math.max(maxIndex, index);
                 server = servers.get(index);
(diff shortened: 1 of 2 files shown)


DaveBirdsall:
Looks OK to me.

dcs/src/main/java/org/trafodion/dcs/master/listener/ConnectReply.java
```diff
@@ -163,29 +163,21 @@ boolean buildConnectReply  (Header hdr, ConnectionContext cc, SocketAddress clie
...
-                    else {
-                        found = true;
-                        break;
-                    }
-                }
-                else
-                    continue;
+                if(LOG.isDebugEnabled())
+                    LOG.debug(clientSocketAddress + ": " + "server selected in search 1 " + server );
+                data = isServerAvalible(nodeRegisteredPath);
```
arvind-narain: Thanks for the refactor Kevin. Code looks good now.
Question - Doesn't seem to be a difference between the new log message (175) for this case and the next case (191) - intentional ? I think earlier the index was being printed.

Code is good to be checked in. The next set of change (for disconnect) will allow tests to pass without holding onto the connections - thanks for that.

DaveBirdsall:
Is this ready to commit now?

kevinxu021:
Arvind, any more comments? If no, please merge it.

arvind-narain:
Thanks Kevin for the change - will merge it now.

PR #203, opened 2015-12-03
[TRAFODION-1668]initialize spj for upload a customized jar

kevinxu021:


core/conn/spj_init/src/main/java/com/traf/mgmt/JarFileMgmt.java
```diff
@@ -0,0 +1,435 @@
...
+				fos = new FileOutputStream(fname, (appendFlag == 0));
+
+				fos.write(Snappy.uncompress(Arrays.copyOf(data, data.length)));
+				fos.flush();
+			} finally {
+				if (fos != null)
+					fos.close();
+			}
+
+			syncJar(userPath, fileName);
```
zellerh: Should we also set execute permissions for C++ DLLs? Probably ok to do this unconditionally for jars also.

core/conn/spj_init/src/main/java/com/traf/mgmt/JarFileMgmt.java
```diff
@@ -0,0 +1,435 @@
...
+		} catch (Throwable t) {
+			LOG.error(t.getMessage(), t);
+			throw new SQLException(t.getMessage());
+		}
+	}
+
+	private static void checkFileName(String fileName) throws SQLException {
+		if (fileName.contains("/") || fileName.contains("\\"))
+			throw new SQLException("Illegal file name: " + fileName
+					+ ". File name must not contain \"/\".");
```
zellerh: It's probably not a problem right now with any of the calls, but I'm wondering whether we should also disallow "..", at least if that's the entire file name. Again, I don't have any example where this would be a problem with the current code.

core/conn/spj_init/src/main/resources/init_spj.sql
```diff
@@ -0,0 +1,156 @@
+#!/bin/bash
```
zellerh: Since this is a shell script, should the name really end in ".sql"? Not sure what conventions are used in the conn project, but I usually use ".sql" for scripts that are obeyed in sqlci or trafci.

core/conn/spj_init/src/main/resources/init_spj.sql
```diff
@@ -0,0 +1,156 @@
...
+      LIBRARY SPJMGMT
+      LANGUAGE JAVA
+      PARAMETER STYLE JAVA
+      READS SQL DATA
+      ;
+      GRANT EXECUTE ON PROCEDURE HELP TO PUBLIC;
+      
+      CREATE PROCEDURE PUT (
+      IN FILEDATA VARCHAR(102400) CHARACTER SET ISO88591,
+      IN FILENAME VARCHAR(256) CHARACTER SET ISO88591,
```
zellerh: I wonder whether we should use UTF-8 for the file names, so that we can also support file names in Chinese and other languages. Do you know whether this is supported by Linux? I would suggest VARCHAR(256 BYTES) CHARACTER SET UTF8 in that case - but then I'm not 100% sure whether SPJs support that, we might need to use UCS2?

core/conn/spj_init/src/main/java/com/traf/mgmt/JarFileMgmt.java
```diff
@@ -0,0 +1,435 @@
...
+		return conn;
+	}
+
+	private static String getCurrentUser(Connection conn) throws SQLException {
+		Statement st = null;
+		ResultSet rs = null;
+		String user = null;
+		try {
+			st = conn.createStatement();
+			rs = st.executeQuery("values(current_user)");
```
zellerh: Might be better to use "session_user" here, since we want the user who is connected. Right now it probably doesn't make a difference, but it will, once we have definer's rights (current_user would then be 'DB__ROOT').

core/conn/spj_init/src/main/java/com/traf/mgmt/JarFileMgmt.java
```diff
@@ -0,0 +1,435 @@
...
+	}
+
+	private static String getCodeFilePath(Connection conn) throws SQLException {
+		String user = getCurrentUser(conn);
+		String root = System.getenv("MY_SQROOT");
+		if (root == null || "".equals(root.trim())) {
+			LOG.error("Cant get your traf installation path!");
+			throw new SQLException("Cant get your traf installation path!");
+		}
+		File file = new File(root + "/spj/" + user);
```
zellerh: Can we make this root + "/udr/lib/" + user? This will be used for UDFs as well. Eventually, we want to replace this method with libraries that are stored in BLOBs and are cached, probably in directory root + "/udr/cache/".

core/conn/spj_init/src/main/java/com/traf/mgmt/JarFileMgmt.java
```diff
@@ -0,0 +1,435 @@
...
+			}
+			if (st != null) {
+				try {
+					st.close();
+				} catch (Exception e) {
+					LOG.warn(e.getMessage(), e);
+				}
+			}
+		}
+
```
zellerh: Probably we need to sanitize the user name as well, to make sure we do something about user "Hans/../../../../../../etc/rc.d". Also, we may need to replace some characters that are acceptable in user names but not in file names (not sure what those are). Maybe at least replace any '/' and '\0' characters in the user name with something like '_'.

core/conn/spj_init/src/main/java/com/traf/mgmt/JarFileMgmt.java
```diff
@@ -0,0 +1,435 @@
...
+			LOG.info("Put " + fileName + ", length: " + data.length + ", file string length:" + fileData.length());
+			String userPath = getCodeFilePath(conn);
+			close(conn);
+			String fname = userPath + fileName;
+			checkFile(fname, data.length);
+			FileOutputStream fos = null;
+			try {
+				fos = new FileOutputStream(fname, (appendFlag == 0));
+
+				fos.write(Snappy.uncompress(Arrays.copyOf(data, data.length)));
```
zellerh: We had some problems on systems that didn't have Snappy installed (for using Snappy to compress HFiles). Do you think this might be an issue with client systems, or do we include Snappy in our installation on the client side?

zellerh:
Thanks, looks great overall! See comments/questions on some details.

core/conn/spj_init/src/main/java/com/traf/mgmt/JarFileMgmt.java
kevinxu021: Yes, right. The JARs will be loaded by UDR. The policy should be applied on loading. JAVA SecurityManager is for the whole JVM.

core/conn/spj_init/src/main/java/com/traf/mgmt/JarFileMgmt.java
kevinxu021: '/' and '\' is not allowed, so I think ".." may be ok because "../xxx" is not allowed and "..filename" is ok to be a file name.

core/conn/spj_init/src/main/resources/init_spj.sql
kevinxu021: yes, right. will fix it.

core/conn/spj_init/src/main/java/com/traf/mgmt/JarFileMgmt.java
zellerh: What I meant was to add this:

> Set<PosixFilePermission> perms = new HashSet&lt;PosixFilePermission&gt;();
>
> perms.add(PosixFilePermission.OWNER_READ);
> perms.add(PosixFilePermission.OWNER_WRITE);
> perms.add(PosixFilePermission.OWNER_EXECUTE);
> perms.add(PosixFilePermission.GROUP_READ);
> perms.add(PosixFilePermission.GROUP_EXECUTE);
> perms.add(PosixFilePermission.OTHERS_READ);
> perms.add(PosixFilePermission.OTHERS_EXECUTE);
> setPosixFilePermissions(Path.get(fname),perms);

core/conn/spj_init/src/main/resources/init_spj.sql
```diff
@@ -0,0 +1,156 @@
...
+#
+# @@@ END COPYRIGHT @@@
+
+SERVER_JAR=${MY_SQROOT}/export/lib/spj_init.jar
+CI_SERVER_LOG=${MY_SQROOT}/ciserver/cislog
+SQLCI=${MY_SQROOT}/export/bin32/sqlci
+CP=/bin/cp
+MANAGEABILITY_CATALOG=TRAFODION
+CIS_SCHEMA=DEFAULT_SPJ
+
```
robertamarton: You are creating a new schema, I assume this is a system schema so you want to make sure the schema name begins and ends with an underscore.  This prevents a problem if someone already has a schema named DEFAULT_SPJ, and it works with other system schemas.

core/conn/spj_init/src/main/resources/init_spj.sql
```diff
@@ -0,0 +1,156 @@
...
+function dropAndCreateSchema {
+    echo "Creating Schema for CI Server Component"
+    sqlci << sqlciEOF
+
+      cqd CAT_IGNORE_ALREADY_EXISTS_ERROR 'on';
+      cqd CAT_IGNORE_DOES_NOT_EXIST_ERROR 'on';
+
+      DROP SCHEMA $MANAGEABILITY_CATALOG.$CIS_SCHEMA CASCADE;
+      CREATE SCHEMA $MANAGEABILITY_CATALOG.$CIS_SCHEMA;
+
```
robertamarton: Do you want to add an authorization clause to the CREATE SCHEMA, for example AUTHORIZATION DB__ROOT.  This will make sure DB__ROOT is the schema owner.

core/conn/spj_init/README.md
```diff
@@ -0,0 +1,24 @@
...
+E.g.   
+trafci>set param ?p1 help;  
+trafci>call DEFAULT_SPJ.HELP(?p1)  
+PUT - Upload a JAR. SHOWDDL PRODURE DEFAULT_SPJ.PUT for more info.  
+LS - List JARs. SHOWDDL PRODURE DEFAULT_SPJ.LS for more info.  
+LSALL - List all JARs. SHOWDDL PRODURE DEFAULT_SPJ.LSALL for more info.  
+RM - Remove a JAR. SHOWDDL PRODURE DEFAULT_SPJ.RM for more info.  
+RMREX - Remove JARs by a perticular pattern. SHOWDDL PRODURE DEFAULT_SPJ.RMREX for more info.  
+GETFILE - Upload a JAR. SHOWDDL PRODURE DEFAULT_SPJ.GETFILE for more info.  
+```
```
robertamarton: Do you need to run as the Trafodion ID to enable this feature?  If someone like Webroot wants to initialize this, how would this be done?  I assume someone would need to run the init_sqj script.  Will this be automated somehow?  Can someone other than the Trafodion ID perform this operation if they have the correct database privileges?

core/conn/spj_init/src/main/java/com/traf/mgmt/JarFileMgmt.java
kevinxu021: By default, it's OWNER_READ, OWNER_WRITE, GROUP_READ, GROUP_WRITE and OTHERS_READ. I think that's enough. Execute permission is not needed since it will be executed in JVM.

core/conn/spj_init/src/main/resources/init_spj.sql
```diff
@@ -0,0 +1,156 @@
...
+      cqd CAT_IGNORE_DOES_NOT_EXIST_ERROR 'off';
+
+      exit;
+procEOF
+    
+}
+
+dropAndCreateSchema
+createProcedures
+echo "Successfully completed installation of SPJs."
```
robertamarton: Do you really want to grant execute privilege to PUBLIC for all these  procedures?  Not sure what they all do but something called PUT sure sounds like something you would want to restrict.  It is important to restrict who can add foreign code to Trafodion.

core/conn/spj_init/src/main/java/com/traf/mgmt/JarFileMgmt.java
kevinxu021: Ok, is there any system variable for "/udr/lib/" or "/udr/cache/"? Otherwise to hardcode it? It seems there is variable  "$MY_UDR_ROOT".

core/conn/spj_init/src/main/java/com/traf/mgmt/JarFileMgmt.java
kevinxu021: For snappy, not yet. Somebody who want to call this procedure should involve Snappy related JAR into their project.

core/conn/spj_init/src/main/java/com/traf/mgmt/JarFileMgmt.java
zellerh: What I am trying to say is that your new SPJs will also be used to deploy DLLs for UDFs written in C or C++. Those DLLs need to have execute privilege or they will not work. You could try to find out whether a file is a jar file or a DLL and set the permissions accordingly, or just always set execute permissions.

core/conn/spj_init/src/main/java/com/traf/mgmt/JarFileMgmt.java
zellerh: Good point. I would probably just use something like $MY_UDR_ROOT/lib. However, there is a bug: MY_UDR_ROOT is set to $MY_SQROOT in the regular shell environment (used for sqlci). It is set to $MY_SQROOT/udr for tdm_udrserv (set in $MY_SQROOT/etc/ms.env file). So, if we want to use $MY_UDR_ROOT we would need to fix that first, or just use $MY_SQROOT to make the name.

core/conn/spj_init/src/main/java/com/traf/mgmt/JarFileMgmt.java
zellerh: IMHO it would be better to just forget about Snappy and send the data uncompressed. Requiring Snappy has caused us a lot of trouble on the server side. Requiring it on the client and server side (probably with compatible versions) is not worth it for this use case that's about convenience, not performance.

kevinxu021: Agree, will remove it.

core/conn/spj_init/src/main/resources/init_spj.sql
zellerh: Good point, I forgot that aspect. The PUT tries to sandbox the files it puts into a specific directory, so it should be relatively safe, but it may still be nice to be able to restrict this to a specific role. Roberta, is it possible that this initialization script creates a role, something like "LIBRARY_CREATOR", so that we can grant execution privilege to that role instead of PUBLIC? Then the DBA of the customer can grant this role to whoever they want.

core/conn/spj_init/src/main/java/com/traf/mgmt/JarFileMgmt.java
kevinxu021: Sure, i'll add all permission you listed to non-JAR files.

venkat1m:
1. shouldn't the class names be "org.trafodion...." instead of "com.trafodion..."

2. In the pre-apache product, we had 2 levels of security. a) component privileges that controlled who can upload jar or create library or create procedure etc. b) execute privilege on these system SPJs. 

3. It is not clear to me if we need to sandbox the SPJs in user specific directories. In the pre-apache product we had tried user and role specific folders and it was a nightmare to share and manage the SPJs by different admin users and we had reverted back to a single base folder. With the combination of the component privileges and SQL privileges we can control security.

core/conn/spj_init/README.md
kevinxu021: For build, Trafodion ID is not require. But the only required thing is to upload the jar into every node on cluster for "create library".  And Trafodion ID should have READ permission on the JAR.
For initialization, anywhere SQLCI is available, whatever it's Tafodion ID. By the way, sqlci can be replaced by trafci, then it can be initialized on client-side.

core/conn/spj_init/src/main/java/com/traf/mgmt/JarFileMgmt.java
kevinxu021: Ok, will use "$MY_SQROOT/udr/lib" +user

core/conn/spj_init/src/main/java/com/traf/mgmt/JarFileMgmt.java
kevinxu021: Sure, session_user should be better.

core/conn/spj_init/src/main/java/com/traf/mgmt/JarFileMgmt.java
```diff
@@ -0,0 +1,435 @@
...
+* Unless required by applicable law or agreed to in writing,
+* software distributed under the License is distributed on an
+* "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+* KIND, either express or implied.  See the License for the
+* specific language governing permissions and limitations
+* under the License.
+*
+* @@@ END COPYRIGHT @@@
+ */
+package com.traf.mgmt;
```
hegdean: Package name needs to be changed from  com.traf  to  org.trafodion.spj.

core/conn/spj_init/src/main/java/com/traf/mgmt/JarFileMgmt.java
kevinxu021: Sure, replace '/' and '\0' with '_'. Maybe using something like MD5 to have a unique name without any particular character.

core/conn/spj_init/src/main/resources/init_spj.sql
kevinxu021: Good point! Will change it. Rename to "_SPJ_".

core/conn/spj_init/src/main/resources/init_spj.sql
kevinxu021: Whatever the user you want to manage the procedures. The following scripts will grant public privileged on procedures, then it can be performed by any user.

core/conn/spj_init/Makefile
```diff
@@ -0,0 +1,37 @@
...
+all:
+	@$(MAVEN) package
+	${CP} target/spj_init-*.jar ${INSTALL_LIBDIR}/spj_init.jar
+	${CP} src/main/resources/init_spj.sql ${INSTALL_SCRIPTSDIR}/
+	@chmod +x ${INSTALL_SCRIPTSDIR}/init_spj.sql
+clean:
+	@rm ${INSTALL_LIBDIR}/spj_init.jar
+	@rm -f ${INSTALL_SCRIPTSDIR}/init_spj.sql
+	@$(MAVEN) clean
+cleanall: clean
```
hegdean: This needs to be incorporated to the top level Makefile. so spj mgmt jar files are built automatically  when 'make  all' is executed from the very top level

core/conn/spj_init/src/main/resources/init_spj.sql
kevinxu021: Yes, a JAR name could be Chinese, but usually we don't do like that. Yes, Linux supports Chinese. Will make changes on it - CHARACTER SET UTF8

core/conn/spj_init/src/main/java/com/traf/mgmt/JarFileMgmt.java
kevinxu021: Good! Will be spjmgmt.

core/conn/spj_init/src/main/resources/init_spj.sql
```diff
@@ -0,0 +1,156 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+# @@@ END COPYRIGHT @@@
+
+SERVER_JAR=${MY_SQROOT}/export/lib/spj_init.jar
+CI_SERVER_LOG=${MY_SQROOT}/ciserver/cislog
+SQLCI=${MY_SQROOT}/export/bin32/sqlci
```
hegdean: remove CI_SERVER_LOG if  not being used
We don't build 32 bit anymore. Need to change it to 64 bit and reference  debug or  release version

core/conn/spj_init/src/main/resources/init_spj.sql
```diff
@@ -0,0 +1,156 @@
...
+CP=/bin/cp
+MANAGEABILITY_CATALOG=TRAFODION
```
hegdean: Can we change the name from MANAGEABILITY_CATALOG to something else(CATALOG_NAME)

core/conn/spj_init/src/main/resources/init_spj.sql
```diff
@@ -0,0 +1,156 @@
...
+CIS_SCHEMA=DEFAULT_SPJ
+
+
+function dropAndCreateSchema {
+    echo "Creating Schema for CI Server Component"
```
hegdean: remove the echo statement or change the description

core/conn/spj_init/Makefile
kevinxu021: Yes, will link to top level Makefile

core/conn/spj_init/src/main/resources/init_spj.sql
robertamarton: We do have special roles today that we use to own external schemas for HIVE and HBASE.  They are called DB__HIVEROLE

robertamarton: and DB__HBASEROLE.  If you want to create a role, you could create one and call it DB__LIBROLE.  However, we already have a component privilege called MANAGE_LIBRARIES that could be granted to someone or some role.  This is a system level privilege grant.

core/conn/spj_init/src/main/resources/init_spj.sql
robertamarton: This relates to the role comment Hans made below.  We could create a role and change the AUTHORIZATION clause to be the role.  But this is only valid when authorization is enabled.  So you need logic to handle to set the correct authorization clause and code is needed when authorization is enabled and disable to change the schema owner.  I would, for simplicity, make DB__ROOT or DB__ROOTROLE the owner and require the user to set privileges to whatever role or user they want to manage privileges.

kevinxu021:
Listed all changes included in the latest code:
1. Remove snappy
2. Add permissions chmod 755 xxx
3.com.trafodion -> org.trafodion
4. Use "$MY_SQROOT/udr/lib" +user for base dir
5. Replace current_user with session_user 
6. For user name in path, replace '/' and '\' with '_'
7. Rename init_spj.sql to init_spj.sh
8. Rename schema name to _SPJ_.
9. Make changes on file name with CHARACTER SET UTF8
10. Rename the project name "spjmgmt"
11. Link to top level Makefile
12. Remove CI_SERVER_LOG and rename sqlci path to bin64
13. Rename MANAGEABILITY_CATALOG  to CATALOG_NAME
14. Change echo messages echo"Creating Schema for CI Server Component

zellerh:
Thanks for making all these changes, looks good!

zellerh:
Looking at Roberta's comments and at the naming and location of these SPJs, two more comments: We still have the question who should have execute rights for these SPJs. We do not want PUBLIC to have this right, since we are uploading code to the server. The other is naming: org.trafodion.mgmt is probably too generic, this is about UDR library management. IMHO, it would be good to put the code under core/sql/src/main/java/org/trafodion/udr/library_management. Roberta, could you propose the GRANT and possibly CREATE ROLE statements that we should use?

robertamarton:
For privileges:



In order to create a library, you need MANAGE_LIBRARY privilege.  Today, we
set this privilege as:



CREATE COMPONENT PRIVILEGE MANAGE_LIBRARY AS 'ML' ON SQL_OPERATIONS SYSTEM

  DETAIL 'Allow grantee to manage libraries';



--  GRANT COMPONENT PRIVILEGE MANAGE_LIBRARY

ON SQL_OPERATIONS TO DB__ROOT

WITH GRANT OPTION;



GRANT COMPONENT PRIVILEGE MANAGE_LIBRARY

ON SQL_OPERATIONS TO DB__ROOTROLE

WITH GRANT OPTION;



In order to create a procedure, you need CREATE_PRIVILEGE privilege.  This
can be done by the schema owner, or by granting an entity the CREATE or
CREATE_PRIVILEGE privilege. Today we set this privilege as:



CREATE COMPONENT PRIVILEGE CREATE_PROCEDURE AS 'CP' ON SQL_OPERATIONS SYSTEM

  DETAIL 'Allow grantee to create procedures';



--  GRANT COMPONENT PRIVILEGE CREATE_PROCEDURE

ON SQL_OPERATIONS TO DB__ROOT

WITH GRANT OPTION;



GRANT COMPONENT PRIVILEGE CREATE_PROCEDURE

ON SQL_OPERATIONS TO DB__ROOTROLE

WITH GRANT OPTION;



In order to execute the procedure you need the EXECUTE object level
privilege.



I would leave privilege setting, as much as possible, to the  customer.
They are ultimately responsible for setting security.

So, I would just create the procedures as DB__ROOT and grant DB__ROOTROLE
the execute privilege with grant option.



The customer would then decide who (user or role) should be assigned this
privilege and perform appropriate grant statements.



   Roberta



*Sent:* Tuesday, December 15, 2015 8:53 AM
*Subject:* Re: [incubator-trafodion] [TRAFODION-1668]initialize spj for
upload a customized jar (#203)



Looking at Roberta's comments and at the naming and location of these SPJs,
two more comments: We still have the question who should have execute
rights for these SPJs. We do not want PUBLIC to have this right, since we
are uploading code to the server. The other is naming: org.trafodion.mgmt
is probably too generic, this is about UDR library management. IMHO, it
would be good to put the code under
core/sql/src/main/java/org/trafodion/udr/library_management. Roberta, could
you propose the GRANT and possibly CREATE ROLE statements that we should
use?

—
<https://github.com/apache/incubator-trafodion/pull/203#issuecomment-164825207>
.

core/conn/spj_mgmt/Makefile
```diff
@@ -0,0 +1,37 @@
...
+# under the License.
+#
+# @@@ END COPYRIGHT @@@
+
+LIBDIR := lib
+JAVAC := $(JAVA_HOME)/bin/javac
+INSTALL_LIBDIR := ${MY_SQROOT}/export/lib
+INSTALL_SCRIPTSDIR := ${MY_SQROOT}/sql/scripts
+CP :=/bin/cp -f
+INSTALL_INCLUDEDIR := ${MY_SQROOT}/export/include
```
hegdean: Do we need these 2 variables LIBDIR and INSTALL_INCLUDEDIR?

core/conn/spj_mgmt/src/main/java/org/trafodion/mgmt/JarFileMgmt.java
```diff
@@ -0,0 +1,434 @@
...
+* Unless required by applicable law or agreed to in writing,
+* software distributed under the License is distributed on an
+* "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+* KIND, either express or implied.  See the License for the
+* specific language governing permissions and limitations
+* under the License.
+*
+* @@@ END COPYRIGHT @@@
+ */
+package org.trafodion.mgmt;
```
hegdean: can we call the package name same as the directory name org.trafodion.spjmgmt.

core/conn/spj_mgmt/src/main/resources/init_spj.sh
```diff
@@ -0,0 +1,155 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+# @@@ END COPYRIGHT @@@
+
+SERVER_JAR=${MY_SQROOT}/export/lib/spj_mgmt.jar
+SQLCI=${MY_SQROOT}/export/bin32/sqlci
```
hegdean: Is SQLCI being used? if not can you please remove it

hegdean:
Do we have developer regression for spj mgmt code. Can we make sure we add them to regression test folder.

kevinxu021:
Hans, A little question about the source code path, do you mean to merge it into trafodion-sql.jar?
Changes included:
1. rename package name to spjmgmt
2.create a role SPJ_EXECROLE for spj execution

zellerh:
Kevin, yes, this could be included in trafodion-sql.jar, if you move the source files.

zellerh:
The current name, org.trafodion.spjmgmt is also ok (although I have pointed out before that it is not SPJ management, but library management). The name of the role, SPJ_EXECROLE is confusing, though. This is about a few SPJs for library management, the role sounds as it it were used for all SPJs.

venkat1m:
Yes, the SPJ_EXECROLE name seems to suggest it is for ALL spjs on the system.  Calling the role DB__LIBRARYMGRROLE or DB__LIBMGRROLE might be appropriate since it clearly states the scope of these system SPJs which are solely to mange the underlying code files used by the library objects. And the DB__ prefix is because this is a system created role.

kevinxu021:
Thanks for suggestion. How about keeping it as a project since UDR might load every class at a time for each call, the smaller, the better.

venkat1m:
README.md file is missing the Apache copyright header and that is causing the build static check to fail.

kevinxu021:
Prefix "DB__" is reserved, the object with it cant be deleted. so renamed role name

zellerh:
Sure, Kevin, keeping it as a separate project is fine.

svarnau:
The 1/4/16 test result was spurious message from jenkins migration.

core/sql/lib_mgmt/README.rst
```diff
@@ -0,0 +1,57 @@
...
+   # @@@ END COPYRIGHT @@@
+
+=============================
+Trafodion Core README
+=============================
+
+How to build and initialize default SPJs 
+
+  Build
+cd $MY_SQROOT/../conn/spj_init
```
hegdean: Please change the contents of this README to match directory/file name change

arvind-narain:
One untracked file and missing version info in new jar:

2016-01-06 09:30:30 Error: Build created 1 untracked file(s)
2016-01-06 09:30:30        Update .gitignore files
2016-01-06 09:30:30 ===========
2016-01-06 09:30:30 sqf/sql/scripts/init_libmgmt.sh

2016-01-06 09:32:18 Error: version info is missing (sqvers)
2016-01-06 09:32:19 $MY_SQROOT/export/lib/lib_mgmt.jar missing version string

kevinxu021:
jekin, test

kevinxu021:
jekins, retest

kevinxu021:
please dont merge, will fix version issue soon and add more API for create/alt/remove libraries.

kevinxu021:
Please review the code again. if no more comments, please feel free to merge.

core/sql/lib_mgmt/README.rst
```diff
@@ -0,0 +1,63 @@
...
+LS - List JARs. SHOWDDL PRODURE [SCHEMA NAME.]LS for more info.
+ 
+LSALL - List all JARs. SHOWDDL PRODURE [SCHEMA NAME.]LSALL for more info. 
+ 
+RM - Remove a JAR. SHOWDDL PRODURE [SCHEMA NAME.]RM for more info.
+
+RMREX - Remove JARs by a perticular pattern. SHOWDDL PRODURE [SCHEMA NAME.]RMREX for more info.
+
+GETFILE - Download a JAR. SHOWDDL PRODURE [SCHEMA NAME.]GETFILE for more info.
+
```
robertamarton: Should this say PROCEDURE instead of PRODURE?

core/sql/lib_mgmt/src/main/resources/init_libmgmt.sh
```diff
@@ -0,0 +1,196 @@
...
+      cqd CAT_IGNORE_ALREADY_EXISTS_ERROR 'on';
+      cqd CAT_IGNORE_DOES_NOT_EXIST_ERROR 'on';
+
+      DROP SCHEMA $CATALOG_NAME.$LIB_SCHEMA CLEANUP CASCADE;
+      CREATE SCHEMA $CATALOG_NAME.$LIB_SCHEMA;
+      set schema $CATALOG_NAME.$LIB_SCHEMA;
+      DROP LIBRARY ${LIB_NAME} CASCADE;
+      CREATE LIBRARY ${LIB_NAME} FILE '${SERVER_JAR}';
+      CREATE ROLE ${DB__LIBMGRROLE};
+
```
robertamarton: Were we going to create the special schema starting with an underscore to make it a system owned schema?  For example: "_DB_LIBMGR_"?
Also, is DB_LIBMGRROLE a system role?  If so, it should begin with "DB__" (2 underscores) to be compatible with other system defined roles.

core/sql/lib_mgmt/src/main/resources/init_libmgmt.sh
```diff
@@ -0,0 +1,196 @@
...
+      cqd CAT_IGNORE_ALREADY_EXISTS_ERROR 'off';
+      cqd CAT_IGNORE_DOES_NOT_EXIST_ERROR 'off';
+
+      exit;
```
robertamarton: This does not concern your code change but if you want initialize/upgrade trafodion to create the the DB_LIBMGR schema, let me know.  It looks like we may be supporting an upgrade.  Also, if you want initialize/upgrade authorization to create the DB_LIBMGRROLE, I can add it also.  I have been asked to add another system role so it would be simple enough to do.

core/sql/lib_mgmt/src/main/resources/init_libmgmt.sh
robertamarton: For some reason the underscores disappeared - "_DB_LIBMGR_".

robertamarton: Apparently, git is removing the underscores,  the name " (underscore)DB_LIBMGR(underscore)"

robertamarton:
Changes look really good, very easy to read.  My only concern is the name for the schema and role.

robertamarton:
One other thing.  I did not notice any regression tests related to security.  I assume that you have run tests with security (at least authorization) enabled.  We want to make sure that privileges work and that file permissions are set correctly. If you plan to add any unit tests related to security in SQL regressions, please add them to the new privs2 test suite.

btw - I try to add unit tests for new features at the time I deliver for two reasons - one, if someone else wants to use the new feature, they can look at the tests to understand it better and two, a subsequent delivery will not cause a regression.

core/sql/lib_mgmt/README.rst
kevinxu021: Yes, typo issue.

core/sql/lib_mgmt/src/main/resources/init_libmgmt.sh
kevinxu021: For role name, DB_LIBMGRROLE should be fine, since with prefix "DB__" cant be deleted. This project is working as a plugin which is nice to have. Schema name and role name might be modified.
For the schema name, now it's back to double underscore because it's removable.

core/sql/lib_mgmt/src/main/resources/init_libmgmt.sh
kevinxu021: Thanks. If this script could be running with "initialize trafodion" or with some options, it would be helpful. But it will slow initialization a little bit. Let's ask @zellerh for some comments.

kevinxu021:
Good suggestion. Before running test cases, this project should be initialized with Trafodion initialization. Otherwise the test cases will failed on Jenkins.

traflm:
hi,
Is there any update on this one? What is the current pending issues for this one to be merged?
thanks,
Ming

arvind-narain:
Could the reviewers please comment if all the concerns have been addressed ? One was regarding creation and usage of system roles. Other was related to unit tests with security enabled.

core/sql/lib_mgmt/src/main/resources/init_libmgmt.sh
```diff
@@ -0,0 +1,206 @@
...
+# specific language governing permissions and limitations
+# under the License.
+#
+# @@@ END COPYRIGHT @@@
+
+SERVER_JAR=${MY_SQROOT}/export/lib/lib_mgmt.jar
+CI=sqlci
+CATALOG_NAME=TRAFODION
+LIB_SCHEMA="DB__LIBMGR"
+DB__LIBMGRROLE=DB_LIBMGRROLE
```
zellerh: Did you want it the other way round, one underscore on the left and two underscores on the right? Also change the code below where we reference ${DB__LIBMGRROLE}?

core/sql/lib_mgmt/src/main/resources/init_libmgmt.sh
zellerh: Yes, this could add a bit of time to initialize trafodion, but I found that creating routines is relatively fast, because they don't create any HBase objects. So, I hope it should be fine. If we can avoid a separate initialization step for this it will be good.

zellerh:
Looks good to me. Having a schema DB__LIBMGR and granting privileges only to DB__LIBMGRROLE is a good solution, I think. I am ready to commit. I am also sorry for any delay I have caused!

robertamarton:
Initialize spjs:

Changes will start shortly to add initialize spj code.  This code can be committed before the initialize changes are completed.  A new JIRA will be written for this work - here are the planned changes:

Add an “initialize trafodion, init_procedures” option – If someone want this feature, they will just need to perform this command
 
(Since the schema below is a system schema, to follow the naming convention of other system schemas, the name should be something like "_LIBMGR_" but I could also support "DB__LIBMGR".)

The init_procedures option:
  
 Creates a private schema DB__LIBMGR  owned by DB__ROOT
 Creates library called DB__LIBMGRNAME in the DB__LIBMGR schema. The corresponding jar file is located: ${MY_SQROOT}/export/lib/lib_mgmt.jar
      (assume this jar file exists when the library is created – it not operation fails)
  Creates all the routines described in the init_libmgmt.sh script
  If authorization is enabled: call “initialize authorization, upgrade” to complete the request
 
  If the schema, library, routines, and privileges are already setup, this command is a no-op and returns “okay”.
 
Change “initialize trafodion, upgrade” command - The next time we support an upgrade, this will upgrade a system that did not yet perform an “init_procedures” - this can also be postponed until we support an official upgrade
 
Perform the same steps as init_procedures does – create the schema, libraries, and procedures.  Then upgrade authorization.
 
Change “initialize trafodion” command – this is a fresh installation
 
Before initializing authorization,  create the schema, library and routines
Then continue with initializing authorization
 
Change “initialize authorization, upgrade” - 
 
Create the DB_LIBMGRROLE – if not exists
Grant execute privilege on routines, if the routines exist
 
Change “initialize authorization” to 
 
Create the DB_LIBMGRROLE – if not exists
Grant execute privilege on routines, if the routines exist
 
 
Since SPJ’s are initialized during “initialize trafodion”, all regressions run through daily builds will work.
However, for environments that do not perform an “initialize trafodion”, then a manual ““initialize trafodion, init_procedures” is required.  Any regression tests that use this feature should include the “initialize trafodion, init_procedures” command.  If they are already initialized, this is a no-op.
 
Since routines may change more frequently than upgrades, need to think a bit about how to make it easy to add, drop and modify these functions and procedures.

PR #204, opened 2015-12-03
TRAFODION-1570 and TRAFODION-1667 fixes to sqvers -u display

zellerh:
Fix sqvers -u to treat all the jars from a build as a single version (JDBC jars not yet integrated).
Add version display for SuSE Linux to sqvers.
Remove LD_BIND_NOW environment variable, should be for debugging only.


DaveBirdsall:
Looks good to me.

PR #205, opened 2015-12-03
[[TRAFODION-1655]] RMS reports incorrect hbase io time and

selvaganesang:
0 for OperCpuTime for operators hosted in ESP.

Isolate the timer for Operator Cpu Time from HBase IO Timer.
Statistics enabling flag was not set though statistics area was
created and some of the metrics were collected. Operator Cpu time
metrics collection requires statistics to be enabled. It is now
enabled when the operators are hosted in ESPs.


DaveBirdsall:
Looks good to me.

PR #210, opened 2015-12-04
Stats warnings

nonstop-qfchen:
Hi,

I updated the regression tests with an increased value for CQD HIST_ROWCOUNT_REQUIRING_STATS to take care of missing stats warnings. 

A minor improvement also included to consider MC uecs only when the # of key columns without predicates is larger than 1. This is for a previous fix JIRA 1665. 

Thanks


zellerh:
Thanks, looks good! Will commit this once the check tests pass.

PR #211, opened 2015-12-07
Add a new case to BatchTest of Phoenix test

ryzuo:
1. A new test case of inserting duplciate primary key value with
"insert" statement.
2. Changed the initializtion of expected value array of upsert
duplicate primary key values, previous code may cause incorrect test
result.

### DIFF
--- a/tests/phx/src/test/java/org/trafodion/phoenix/end2end/BatchTest.java
+++ b/tests/phx/src/test/java/org/trafodion/phoenix/end2end/BatchTest.java
@@ -44,7 +44,9 @@ public class BatchTest extends BaseTest {
     
     public  static final int ROW_COUNT = 10000;
     private int[] expectedStatusArray = null;
+    private int[] statusArray = null;
 
+    private static final String strInsert = "INSERT INTO " + BATCH_TEST_TABLE + " VALUES (?,?)";
     private static final String strUpsert = "UPSERT USING LOAD INTO " + BATCH_TEST_TABLE + " VALUES (?,?)";
     private static final String strUpdate = "UPDATE " + BATCH_TEST_TABLE + "SET (ID, NAME) = (?,?)";
     private static final String strDelete = "DELETE FROM " + BATCH_TEST_TABLE;
@@ -91,7 +93,7 @@ public class BatchTest extends BaseTest {
             upsertStmt.setString(2, "Traf The World " + i);
             upsertStmt.addBatch();
         }
-        int[] statusArray = upsertStmt.executeBatch();
+        statusArray = upsertStmt.executeBatch();
         long endTime = System.currentTimeMillis();
         System.out.println("Time consumption for batch inserting "
                 + ROW_COUNT + " rows is " + (endTime - startTime)  + " milli seconds");
@@ -110,7 +112,7 @@ public class BatchTest extends BaseTest {
     }
 
     @Test
-    public void testBatchInsertNegative() throws Exception {
+    public void testBatchInsertDuplicate() throws Exception {
         printTestDescription();
 
         createTestTable(BATCH_TEST_TABLE);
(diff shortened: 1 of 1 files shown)


tests/phx/src/test/java/org/trafodion/phoenix/end2end/BatchTest.java
```diff
@@ -118,69 +120,126 @@ public void testBatchInsertNegative() throws Exception {
...
+        } catch(SQLException sqle) {
+            assertEquals("Batch update failed. See next exception for details", sqle.getMessage());
+            SQLException e = null;
+            e = sqle.getNextException();
+            do {
+                assertTrue(e.getMessage().contains("ERROR[8102] The operation is prevented by a unique constraint"));
+            } while((e = e.getNextException()) != null);
+        }
+        
+        //assertArrayEquals(expectedStatusArray, statusArray);
```
zellerh: This is commented out, doesn't that mean the test is not working?

tests/phx/src/test/java/org/trafodion/phoenix/end2end/BatchTest.java
```diff
@@ -118,69 +120,126 @@ public void testBatchInsertNegative() throws Exception {
...
+
+        int rowCount = 0;
+        ResultSet rs = conn.createStatement().executeQuery(strSelect);
+        while(rs.next()) {
+            //System.out.println("ID = " + rs.getString(1) + ", Name = " + rs.getString(2));
+            //assertEquals(expectedIdArray[rs.getRow()-1], rs.getInt(1));
```
zellerh: Same question here, why is this commented out? Also, I think "strSelect" should use an ORDER BY clause, otherwise this test may fail.

ryzuo: Hi Hans, thanks for the reminding here. For now, the result I observed is because of the duplicate primary key, the whole batch failed, thus there's no status array generated, neither no row is inserted. My understanding before was a failure of  individual row should not cause the whole batch fails, and as I recall, SQ worked that way, but now Traf does not.

ryzuo: And not just JDBC T2/T4 drivers, ODBC does not work that as well,when the rowsets passed to SQL if there's duplicated row, it just fails the entire rowsets. So I'm not sure is Traf supposed to work this way? Or just it does not have this ability yet, or it's just the "unique constrain" works this way. So I just kept the assert as commented. I will run more tests about the invalid value of individual row in a batch with other conditions. And, you are right, I will change the select use an order by.

zellerh: Maybe you can add a comment to the test explaining that, so we know why it's currently commented out? Sorry for my late reply.

DaveBirdsall:
Just checking: Is this change ready to merge?

zellerh:
I can merge it and add a comment. According to Rao, this is a "feature". We cannot detect duplicate rows in a batch at this point.

DaveBirdsall:
Go for it! Thanks.

zellerh:
Ok, did the following: Added the JIRA, TRAFODION-1659 to the commit text. Also created a new JIRA, 1701, for the missing ability to handle batch inserts with duplicates. And also added a comment to the test, explaining why it does not check for the results right now.

Ryzuo, can you add the ORDER BY to the test in a future commit?

ryzuo:
Sure thing. I'm running more tests to see how the rowsets are processed by sql, and I was planning update this with more test cases.
For now, seems with constraints violation within a batch, the whole batch will fail, while if there is just individual value invalid nothing related to any constraint, such as a string value is longer than the defined value of a column, the batch will insert successfully with just the individual row fails.
I will file JIRA for the issues I found, and commit another PR to update the test cases.

PR #212, opened 2015-12-07
[TRAFODION-1663]

sbroeder:
Prepare, Commit, and Abort requests from TransactionManager now carry a participant number that helps
track the specific requenst in the RegionEndpointCoprocessor causing a problem when an
UnknownTransactionException is thrown.


shangshengtung:
The code change looks good.

PR #218, opened 2015-12-09
[TRAFODION-1581] TMUDF for JDBC queries

zellerh:
This pull request has 3 initial commits:

[TRAFODION-1581] TMUDF for JDBC queries
[TRAFODION-1672] Failures when running regressions twice
[TRAFODION-1695] ORDER BY on divisioned table w/o sort

The 1581 commit also contains:

[TRAFODION-1582] Optional Drill install for local hadoop

Other small fixes:

- install_local_hadoop now picks ports in the non-ephemeral
  range when doing install_local_hadoop -p fromDisplay.
  Port number range starts at 24000 + 200*display number.
  Stay below 42 for your display to avoid the ephemeral range
  and pick a different display # if you run into port conflicts.

- Setupdir step (this includes building libhdfs, if needed)
  now logs its output with a suffix ##(setupdir) like most
  other components

  core/Makefile

- Fixed a bug causing a core when a TMUDF produced no
  output columns and another bug with a VARCHAR parameter
  at the beginning of a parameter list.

  core/optimizer/UdfDllInteraction.cpp

- Fixed a bug in parsing two consecutive patterns in sqlci
  input, like $$a$$$$b$$

  core/sqlci/SqlCmd.cpp

- Small fixes to doxygen documentation: With the new web
  page structure, make doxygen version match the Trafodion
  version (at least the initial release that applies), also
  fix links to wiki to point to the Apache wiki.

  core/sql/sqludr/doxygen_tmudr.1.6.config
  core/sql/sqludr/sqludr.cpp
  core/sql/sqludr/sqludr.h

- Make hive-exec dependency for SQL explicit (would sometimes
  produce an error depending on the sequence in which things
  are built)

  core/sql/pom.xml


core/sqf/sql/scripts/install_local_hadoop
```diff
@@ -385,10 +385,13 @@ else
...
-    MY_START_PORT=`expr 50000 + $MY_START_PORT '*' 100`
+    # Pick a port range starting at 24000 + 200 * display number.
+    # Choose a display number of up to 42 to stay below the
+    # ephemeral port range which typically starts at 32768.
+    MY_START_PORT=`expr 24000 + $MY_START_PORT '*' 200`
   elif [ $MY_START_PORT == "rand" ]; then
-    # pick a random number between 9000 and 49000 that is divisible by 200
-    MY_START_PORT=`expr $RANDOM '%' 200 '*' 200 + 9000`
+    # pick a random number between 12000 and 24000 that is divisible by 200
+    MY_START_PORT=`expr $RANDOM '%' 50 '*' 200 + 12000`
```
DaveBirdsall: Looks like you're picking a port between 12000 and 22000 (which is fine). If you want to go up to 24000 you could use 60 instead of 50.

core/sql/optimizer/GroupAttr.cpp
```diff
@@ -327,6 +327,38 @@ void GroupAttributes::addConstraint(ItemExpr *c)
...
+             constraints_.next(occ);
+             constraints_.advance(occ))
+          {
+            if (occ.getItemExpr()->getOperatorType() ==
+		ITM_CHECK_OPT_CONSTRAINT)
+              {
+                const ValueIdSet &occPreds =
+                  ((CheckOptConstraint *) occ.getItemExpr())->getCheckPreds();
+
+                if (occPreds.contains(cc->getCheckPreds()))
```
DaveBirdsall: A question about how this works (more from my own ignorance): Does "contains" do a semantic comparison? Or just a ValueId comparison? For example, if I code two CHECK CONSTRAINTS with predicates A < B and A < B, is the ValueId of the '<' the same for both? And how is that accomplished? Similarly, if I have two CHECK CONSTRAINTS with predicates A < B and B > A, will the second be normalized back to A < B and get the same ValueId?

core/sql/optimizer/GroupAttr.cpp
```diff
@@ -580,6 +612,127 @@ void GroupAttributes::addSuitableCompRefOptConstraints(
...
+      if (col.getItemExpr()->getOperatorType() == ITM_VEG_REFERENCE)
+        {
+          const ValueIdSet &vegMembers =
+            static_cast<VEGReference *>(col.getItemExpr())->
+            getVEG()->getAllValues();
+          for (ValueId b=vegMembers.init();
+               vegMembers.next(b);
+               vegMembers.advance(b))
+            if (b.getItemExpr()->getOperatorType() == ITM_BASECOLUMN)
+              col = b;
```
DaveBirdsall: So, you'll get the last ITM_BASECOLUMN in the ValueIdSet vegMembers. I gather that any base column will do? Or must it be from a specific table? (I'm wondering about a join predicate.)

core/sql/optimizer/RelRoutine.cpp
```diff
@@ -1081,33 +1085,37 @@ NARoutine * PredefinedTableMappingFunction::getRoutineMetadata(
...
-      }
+      externalName = "TRAF_CPP_TIMESERIES";
+      libraryPath += getenv("SQ_MBTYPE");
+      break;
+
+    case REL_TABLE_MAPPING_BUILTIN_JDBC:
+      lang = COM_LANGUAGE_JAVA;
+      paramStyle = COM_STYLE_JAVA_OBJ;
+      externalName = "org.trafodion.sql.udr.predef.JDBCUDR";
+      libraryPath += "/trafodion-sql-";
```
DaveBirdsall: Why do we add a slash for libraryPath in this case but not in the previous two cases? (Is there something different about slashes for C++ and Java?)

core/sqf/sql/scripts/install_local_hadoop
zellerh: Thanks for catching that, I'll fix it.

core/sql/sqludr/sqludr.cpp
```diff
@@ -7761,7 +7761,7 @@ void UDR::processData(UDRInvocationInfo &info,
  *
  *  This method is called in debug Trafodion builds when certain
  *  flags are set in the UDR_DEBUG_FLAGS CQD (CONTROL QUERY DEFAULT).
- *  See https://wiki.trafodion.org/wiki/index.php/Tutorial:_The_object-oriented_UDF_interface#Debugging_UDF_code
+ *  See https://cwiki.apache.org/confluence/display/TRAFODION/Tutorial%3A+The+object-oriented+UDF+interface#Tutorial:Theobject-orientedUDFinterface-DebuggingUDFcode
```
DaveBirdsall: Is the tutorial going to move to the new Trafodion website or will it stay on the Confluence wiki? (Check with Gunnar.)

core/sql/optimizer/RelRoutine.cpp
zellerh: Yes, for C++ it's a Linux directory, $MY_SQROOT/export/lib${SQ_MBTYPE}. For Java it's a jar file, like $MY_SQROOT/export/lib/trafodion-sql-1.3.0.jar.

core/sql/src/main/java/org/trafodion/sql/udr/predef/JDBCUDR.java
```diff
@@ -0,0 +1,690 @@
...
+ *    <connection string>,
+ *    <user name>,
+ *    <password>,
+ *    <statement_type>,
+ *    <sql statement 1>
+ *    [ , <sql statements 2 ...n> ] )) ...
+ *
+ * The first 7 arguments are required and must be
+ * string literals that are available at compile
+ * time.
```
DaveBirdsall: Must be available at compile time? Does this mean I have to hard-code my password?

DaveBirdsall:
Looks good to me. A lot of work went into this.

core/sql/sqludr/sqludr.cpp
zellerh: Gunnar commented on those wiki pages that are moving, saying that no more changes should be made. He did not comment on this one, so I assume it will stay on the wiki.

core/sql/optimizer/GroupAttr.cpp
zellerh: Yes, good point, this check is nearly useless, probably comes from cloning the logic from another case. I could remove it entirely or add a comment that explains it. I'm leaning towards something like this comment:

"Note that the check for inclusion of value ids of the predicates is not very useful. A more useful check would be whether the existing predicates imply the new ones or vice versa, but that would be much more complicated code and is not required at this point."

zellerh: Sorry, just realize that below, the code may add the same check constraint multiple times, so we do need the simple guard against that. Therefore the check is not nearly useless, it's just not as smart as it could be.

core/sql/optimizer/GroupAttr.cpp
zellerh: The ScanKey::isAKeyPredicateForColumn method below relies on ScanKey::expressionContainsColumn() as a test. That method should be ok with any VEG member, index or base column, just not with the VEGReference itself. This brings up some interesting questions, whether this code is necessary and also whether there is a potential issue with equating rows of another table with a constant that is not in a VEG (at this time I think we are ok on that one). I'm investigating this a bit further and will add a fix and/or another comment here.

core/sql/src/main/java/org/trafodion/sql/udr/predef/JDBCUDR.java
zellerh: Unfortunately, yes.

core/sql/optimizer/GroupAttr.cpp
zellerh: In the meantime I looked more at the code and tried a few options. Removing this code makes the test case fail. In the commit for the rework, I'll replace the current check (VEG member is a base column) with a check whether the VEG member is referenced by the check constraints. That should ensure that we use the correct value id (I found that the value selected is actually an index column).

About the issue of applying transitive closure incorrectly, when we have a predicate T1.a=const and a VEG(T1.a, T2.b): I'll add a comment to the code explaining why I think this won't happen.

DaveBirdsall:
Looks good to me. I think this is ready to commit.

PR #219, opened 2015-12-09
[TRAFODION-1688] Changes to support recovery following RegionServer f…

sbroeder:
…ailure

### DIFF
--- a/core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/TrxRegionEndpoint.java
+++ b/core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/TrxRegionEndpoint.java
@@ -2931,7 +2931,7 @@ CoprocessorService, Coprocessor {
         this.skipWal = config.getBoolean(CONF_SKIP_WAL, DEFAULT_SKIP_WAL);
         this.fullEditInCommit = config.getBoolean(CONF_COMMIT_EDIT, DEFAULT_COMMIT_EDIT);
 	m_regionName = this.regionInfo.getRegionNameAsString();
-	m_isTrafodionMetadata = m_regionName.contains("_MD_");
+	m_isTrafodionMetadata = m_regionName.contains("TRAFODION._MD_.");
 
         LOG.info ("TRX Endpoint coprocessor, " 
 		  + " region: " + m_regionName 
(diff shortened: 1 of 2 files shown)


shangshengtung:
The change looks good.

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/TrxRegionEndpoint.java
```diff
@@ -2911,14 +2918,28 @@ public void start(CoprocessorEnvironment env) throws IOException {
...
         this.scannerThreadWakeFrequency = config.getInt(HConstants.THREAD_WAKE_FREQUENCY, 10 * 1000);
 
         this.cleanTimer = config.getInt(SLEEP_CONF, DEFAULT_SLEEP);
         this.memoryUsageThreshold = config.getInt(MEMORY_THRESHOLD, DEFAULT_MEMORY_THRESHOLD);
         this.memoryUsagePerformGC = config.getBoolean(MEMORY_PERFORM_GC, DEFAULT_MEMORY_PERFORM_GC);
+        this.asyncWal = config.getInt(CONF_ASYNC_WAL, DEFAULT_ASYNC_WAL);
+        this.skipWal = config.getBoolean(CONF_SKIP_WAL, DEFAULT_SKIP_WAL);
+        this.fullEditInCommit = config.getBoolean(CONF_COMMIT_EDIT, DEFAULT_COMMIT_EDIT);
+	m_regionName = this.regionInfo.getRegionNameAsString();
+	m_isTrafodionMetadata = m_regionName.contains("_MD_");
```
DaveBirdsall: This could be fooled, right? If my table name was _MD_, this would return true? Also if my schema name was ABC_MD_DEF?

DaveBirdsall:
Apart from one question, looks good to me.

sbroeder:
I suppose you are right Dave and this would result in some extra info level tracing events.  How about I change the test to "TRAFODION._MD_.", would that suffice?

sbroeder:
Interesting that the underscore character disappeared in my post and converted the MD to italics, but I think you understand my point.  It should be TRAFODION.<underscore>MD<underscore>.

sbroeder:
Blah.  That didn't work either.  Imagine there is an underscore character on either side of the MD

DaveBirdsall:
Got it, Sean. And I think your suggestion will work fine.

sbroeder:
Thanks Dave. That's a plan then

On Dec 9, 2015, at 2:49 PM, DaveBirdsall <notifications@github.com<mailto:notifications@github.com>> wrote:


Got it, Sean. And I think your suggestion will work fine.

DaveBirdsall:
Looks good to me. I think this is ready to commit.

sbroeder:
Thanks Dave.

PR #221, opened 2015-12-10
TRAFODION-1650

mashengchen:
after hbase-0.98.9(include 0.98.9), the instantiation process of ScanQueryMatcher(class) is different from before, which add one long class as parameter.


DaveBirdsall:
jenkins, retest

sbroeder:
Looks good to me.

PR #222, opened 2015-12-11
[TRAFODION-1699] Update website publishing instructions

svarnau:
There is more to this JIRA. Just adding one small documentation update.


DaveBirdsall:
I just committed this, but evidently the mirroring is backed up so it still shows as Open.

PR #223, opened 2015-12-11
[TRAFDOION-1666]script to update win-odbc version

xwq:
update the scirpt to make it can update the version number in build_os.bat
and the inno setup script.

### DIFF
--- a/win-odbc64/odbcclient/update_version.pl
+++ b/win-odbc64/odbcclient/update_version.pl
@@ -71,12 +71,12 @@ sub update_file {
             print OUTFILE;
         }
         elsif( /SOFTWARE\\\\ODBC\\\\ODBCINST.INI\\\\TRAF ODBC /) {
-        	s/(\d+)\.(\d+)/$major.$minor/;
-        	print OUTFILE;
+            s/(\d+)\.(\d+)/$major.$minor/;
+            print OUTFILE;
         }
         elsif( /DRIVER_NAME\[\] = "TRAF ODBC / ) {
-        	s/(\d+)\.(\d+)/$major.$minor/;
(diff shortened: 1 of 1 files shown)


win-odbc64/odbcclient/update_version.pl
```diff
@@ -97,4 +97,80 @@ sub update_file {
...
+        else {
+            print OUTFILE;
+        }
+    }
+
+
+    close( INFILE ) or warn "Warning: Can't close $infile - $!";
+    close( OUTFILE ) or warn "Warning: Can't close $outfile - $!";
+
+    unless ( rename $outfile, $infile ) {
```
DaveBirdsall: Does this preserve the permissions of the file?

win-odbc64/odbcclient/update_version.pl
```diff
@@ -97,4 +97,80 @@ sub update_file {
...
+        }
+        else {
+            print OUTFILE;
+        }
+    }
+
+    close( INFILE ) or warn "Warning: Can't close $infile - $!";
+    close( OUTFILE ) or warn "Warning: Can't close $outfile - $!";
+
+    unless ( rename $outfile, $infile ) {
```
DaveBirdsall: Same question as above.

win-odbc64/odbcclient/update_version.pl
xwq: In my environment, new file will has the same permissions as the old. However the developer who use the tool to update the version should check the new file again before committing them. 
Since we have too many files need to be updated in windows odbc driver when we update the Traofion version, it's very easy to miss some file if we don't use a tool.

PR #227, opened 2015-12-15
[TRAFODION-1700] Upsert with omitted default value columns leave the aligned format table in corrupted state.

selvaganesang:
The user is now required to supply the default values in upsert/upsert with load command. When the columns with default value is omitted, the following error is returned.

*** ERROR[7001] Default value is missing for upsert or is not valid for column <column_name>


zellerh:
Is this fix is just temporary, until we put in the real fix in the binder? Do you want to use TRAFODION-1700 for the real fix or create a new one?

selvaganesang:
Hans, It is not clear to me if there is any other option other than select the row first. It it exists replace the omitted value with the value in the old row or populate with the default rows. This would make upsert to perform at the speed of scan. Do you think we should do this?

selvaganesang:
Also, this restriction is for the upsert command. It is ok to omit the column in case of insert.

zellerh:
Hi Selva, couldn't we handle this better? Upsert must be working by matching the clustering key of a new and existing row. If the row is new, then we insert default values for those columns that are not specified. If the row already exists, I would argue that we should overwrite the entire row, with the new one, choosing default values for those columns that the user didn't specify in the upsert. If users want something different they would have to use a merge statement. What do you think?

selvaganesang:
With regular hbase format, the omitted columns are never put into hbase either at the time of insert or upsert whether the row exists or not. The select substitutes the default value of the column if it is missing in hbase table.  So, this works well in regular hbase format when the upsert omits the column, the old value is returned with subsequent select or the default value when the column is missing. We can either put the default value or give an error whichever makes sense.with the aligned format for now. The aligned format table behavior will deviate from the regular hbase format unless aligned format is enhanced to omit the missing column in the row hdr info somehow. By giving out error, we let user to decide to use upsert into table with DEFAULT values or use merge statement in case of aligned format. However, I am also ok with silently substituting with default values when the row exists.

zellerh:
Yes, I am thinking of a fix in the binder that just adds all the missing columns, if the table uses aligned format.

selvaganesang:
It is currently being done in generator in addDefaultColumns in case of insert for regular hbase format and aligned format.Do you think it should be done in Binder only.

selvaganesang:
More changes expected. Please don't merge this change

DaveBirdsall:
Is this ready for commit?

zellerh:
Sorry for not seeing this earlier. Very good that you were able to make this fix. Just a question: For non-aligned format, I assume this won't cause us to write all the columns, even those with a NULL as default value?

selvaganesang:
I should have added this comment earlier. Yes. That's correct Hans. But upon further testing, the upsert command  for non-aligned format table has problems with omitted columns that has current_timestamp  as the default value when the row or the timestamp column doesn't exist already. So, I am working on a fix. The final fix that is being proposed is

Upsert for non-aligned format  will be transformed into merge when the column with current_timestamp as the default value is omitted. For rest of the cases, the current implementation works great.

Upsert for aligned format will be transformed to merge when a new CQD TRAF_UPSERT_DEFAULT_COLUMN_VALUE_OVERWRITE is 'OFF'.  When this CQD is set to 'ON', the upsert command will use the default value for the omitted columns whether the row exists or not. Need some opinion what should be the default for this CQD. Should it be 'ON' or 'OFF' by default.

So, I will be pushing another commit when I am ready with the fix.

zellerh:
Thanks, sounds very good! For TRAF_UPSERT_DEFAULT_COLUMN_VALUE_OVERWRITE, I would probably go with "OFF", the space-saving method. If people want to gain time at the expense of space, they can turn this on. But that's not a strong opinion.

svarnau:
The 1/4/16 test result was spurious message from jenkins migration.

DaveBirdsall:
Hi @selvaganesang, I think the merges (700 files) bring some risk of including stuff you might not want. And it makes it impossible to review. I wonder if you could close this pull request, then put your changes on a fresh branch and create a new pull request with just those changes?

selvaganesang:
Closing this PR because it has become too old. Opened a new PR #340

PR #228, opened 2015-12-15
[TRAFODION-1670] upsert and upsert with load doesn't report the rows affected information in Trafci [TRAFODION-1709] trafci does not deallocate the statement

hegdean:


selvaganesang:
Looks good to me. Will merge to master by end of the day.

DaveBirdsall:
Looks good to me also. Thanks, Selva, for taking on the commit.

PR #229, opened 2015-12-15
[TRAFODION-1618] Fix row estimation to scale to more than 255 columns

DaveBirdsall:
See the JIRA for details.

### DIFF
--- a/core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
+++ b/core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
@@ -1032,11 +1032,58 @@ public class HBaseClient {
         float defCacheFraction = 0.4f;
         return config.getFloat("hfile.block.cache.size",defCacheFraction);
     }
+
+    // if we make the method below public later, should think about whether this is the
+    // right class to host this method
+
+    // compares two qualifiers as unsigned, lexicographically ordered byte strings
+    static private boolean isQualifierLessThanOrEqual(KeyValue nextKv,
+                                                      KeyValue currKv)
+    {
+       int currLength = currKv.getQualifierLength(); 
+       int currOffset = currKv.getQualifierOffset();
+       byte [] currQual = currKv.getQualifierArray();
+       int nextLength = nextKv.getQualifierLength(); 
+       int nextOffset = nextKv.getQualifierOffset();
+       byte [] nextQual = nextKv.getQualifierArray();   
+
+       // If we later decide we need a performance-critical version of this method,
+       // we should just use a native method that calls C memcmp.
+
+       int minLength = nextLength;
+       if (currLength < nextLength)
+         minLength = currLength;
+
+       for (int i = 0; i < minLength; i++) {
+         // ugh... have to do some gymnastics to make this an
+         // unsigned comparison
+         int nextQualI = nextQual[i+nextOffset];
+         if (nextQualI < 0)
+           nextQualI = nextQualI + 256;
+         int currQualI = currQual[i+currOffset];
+         if (currQualI < 0)
+           currQualI = currQualI + 256;
+
+         if (nextQualI < currQualI)
+           return true;
+         else if (nextQualI > currQualI)
+           return false;
+         // else equal, move on to next byte
+       }
+
+       // the first minLength bytes are the same; the shorter array
+       // is regarded as less
+
+       boolean rc = (nextLength <= currLength);      
+
+       return rc;
+    }
+
     // Estimates row count for tblName by iterating over the HFiles for
     // the table, extracting the KeyValue entry count from the file's
     // trailer block, summing the counts, and dividing by the number of
     // columns in the table. An adjustment is made for the estimated
-    // number of missing (null) values by sampling the first several
+    // number of missing values by sampling the first several
     // hundred KeyValues to see how many are missing.
     public boolean estimateRowCount(String tblName, int partialRowSize,
                                     int numCols, long[] rc)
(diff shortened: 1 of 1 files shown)


core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
```diff
@@ -1032,11 +1032,58 @@ public float getBlockCacheFraction()
...
+       }
+
+       // the first minLength bytes are the same; the shorter array
+       // is regarded as less
+
+       boolean rc = (nextLength <= currLength);      
+
+       return rc;
+    }
+
```
sureshsubbiah: This method looks good me. An alternative may be to use the compareTo method in ByteBuffer http://docs.oracle.com/javase/1.5.0/docs/api/java/nio/ByteBuffer.html#compareTo(java.nio.ByteBuffer)
We use ByteBuffer in a few places in $MY_SQROOT/../sql/src/main/java/org/trafodion/sql and $MY_SQROOT/../sql/src/main/java/org/trafodion/sql/udr.
In this case we may have to use the wrap method to get a ByteBuffer of just the relevant parts and then call compareTo on that?
http://docs.oracle.com/javase/1.5.0/docs/api/java/nio/ByteBuffer.html#wrap(byte[], int, int)

I have been fascinated by the ByteBuffer class and could not pass up an opportunity to mention them here. Current code is good too.

core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
```diff
@@ -1088,36 +1139,65 @@ public boolean estimateRowCount(String tblName, int partialRowSize,
...
-            // their frequency in the full file set.
+
+            // Trafodion column qualifiers are ordinal numbers, but are represented
+            // as varying length unsigned little-endian integers in lexicographical
+            // order. So, for example, in a table with 260 columns, the column
+            // qualifiers (if present) will be read in this order: 
+            // 1 (x'01'), 257 (x'0101'), 2 (x'02'), 258 (x'0201'), 3 (x'03'),
+            // 259 (x'0301'), 4 (x'04'), 260 (x'0401'), 5 (x'05'), 6 (x'06'), 
+            // 7 (x'07'), ...
+            // We have crossed the boundary to the next row if and only if the
```
zellerh: Would this break if/when we support multiple column families? Would it be better to compare the row keys instead?

zellerh: I guess the same problem (if it is a problem) exists in the old code.

core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
```diff
@@ -1088,36 +1139,65 @@ public boolean estimateRowCount(String tblName, int partialRowSize,
...
+            // next qualifier read is less than or equal to the previous, 
+            // compared unsigned, lexicographically.
+
```
sureshsubbiah: Current code is good, though I am confused as to why we do not try something simpler like
comparing the Key in consecutive KeyValue objects till it changes? There is this method on KeyValue that will return the key as a string https://hbase.apache.org/devapidocs/org/apache/hadoop/hbase/KeyValue.html#getKeyString()
Maybe we just compare strings then?
Is the idea that keys can be longer strings and are more expensive to compare?

core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
```diff
@@ -1088,36 +1139,65 @@ public boolean estimateRowCount(String tblName, int partialRowSize,
...
+            // Trafodion column qualifiers are ordinal numbers, but are represented
+            // as varying length unsigned little-endian integers in lexicographical
+            // order. So, for example, in a table with 260 columns, the column
+            // qualifiers (if present) will be read in this order: 
+            // 1 (x'01'), 257 (x'0101'), 2 (x'02'), 258 (x'0201'), 3 (x'03'),
+            // 259 (x'0301'), 4 (x'04'), 260 (x'0401'), 5 (x'05'), 6 (x'06'), 
+            // 7 (x'07'), ...
+            // We have crossed the boundary to the next row if and only if the
+            // next qualifier read is less than or equal to the previous, 
+            // compared unsigned, lexicographically.
```
sureshsubbiah: Some simplication is possible for aligned format though it is not clear if additional complexitu in code is warranted. With aligned format we know that each row will have exactly one column and I suppose null/default values will always be included. So the number of KeyValues we get from the header is the number of rows in the table.

sureshsubbiah:
Changes look good Dave. Does this address TRAFODION-823 too? Should we open a separate, lower priority JIRA on the discovery you made that estimated rowcount can differ by about 10% depending on when it taken, before/after flush/compaction? It is an interesting problem, though I am not certain if the difference in estimate is sufficient for more attention

core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
selvaganesang: I wonder if it is possible to estimate based on the getEntries(), the number of columns in the table and the number of default value/nul columns.  We also have avgKeyLen, avgValueLen and fileSize that might aid in row estimation.

DaveBirdsall:
Hans is correct; this code will need further revision to support multiple column families. I didn't think of that when coding this solution. I was following the spirit of the algorithm that was there. Comparing keys would achieve multi-column family support. To Suresh's comment, hopefully HBase provides methods already to compare keys, so we can use those when we get to that point. Selva's ideas of using other metrics are worth exploring as well. I am not sure what to do about the memstore issue. The good news is that it throws things off by only 10-20%, whereas before things could be off by orders of magnitude. I think I'll go ahead and commit this change since it is an improvement on what's there. I will file a follow-on JIRA for multi-column families. Or if there is an existing JIRA for that, I'll make a note that this needs to be addressed there.

DaveBirdsall:
To Suresh's question about JIRA TRAFODION-823, this issue may well be related. Further research is needed to confirm.

DaveBirdsall:
To Suresh's comment about aligned row format, I had considered this too. I did a unit test to make sure that row estimates were correct for the aligned row format case. I asked myself the question of whether making a fast path for that case was reasonable, but decided that the overhead seemed small and I didn't want to introduce more complexity at this stage. But it is worth considering in the future.

DaveBirdsall:
To Suresh's question about JIRA TRAFODION-823, it turns out this is not the cause. The tables in that JIRA all have less than 255 columns.

PR #231, opened 2015-12-17
[TRAFODION-1680] Dcs needs to provide support for HA on AWS

hegdean:


hegdean:
jenkins, retest

RAT_README.txt
```diff
@@ -60,6 +60,7 @@ $MY_SQROOT/../sql/bin/SqlciErrors.txt
 
 dcs/conf/servers 
 dcs/conf/backup-masters 
+dcs/conf/master
                  :  Apache Trafodion DCS default configuration file
                  -> configuration file do not handle comments
```
kevinxu021: Please describe those files separately. it's not very clear.

core/sqf/sql/scripts/cleanlogs
```diff
@@ -0,0 +1,92 @@
...
+fi
+
+function usage() {
+   prog=`basename $0` 
+   echo ""
+   echo "$prog { all | dcs | rest | core }"
+   echo "    all  --- Remove files from core, dcs and rest logs folder"
+   echo "    dcs  --- Remove files from $DCS_INSTALL_DIR/logs folder"
+   echo "    rest --- Remove files from $REST_INSTALL_DIR/logs folder"
+   echo "    core --- Remove log files residing in $MY_SQROOT/logs folder"
```
kevinxu021: dcs and rest are component name. Could you please list out what kinds of logs included in core? For DEV, it's very good, but for production it's dangerous that it may lost lots of important info. A little confused that the logs has rotation rules why to clear them?

hegdean: cleanlogs script is only a helpful utility. This cleanlogs script is not invoked by any other script and used only on demand on need basis.  The logs for core are all the log files that reside in MY_SQROOT/logs folder

RAT_README.txt
hegdean: dcs/conf/master is another configuration file which contains the hostname.. It is  just like backup-masters file and as described in the comments it  does not handle comments

core/sqf/sql/scripts/cleanlogs
kevinxu021: it's might be helpful that save them before remove, or give users some tips and tell them the logs will be removed. if they insist on, go ahead to remove them. yes, it's a utility, but some users might use it, right? we know what they are in MY_SQROOT/logs folder, but the users may not, so those info should be included. otherwise, users may make mistakes.

install/installer/setup_ssh_key
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
kevinxu021: dcs/conf/master seems to be an empty file. is that possible to remove it?

hegdean: This has to be an empty file that needs to exist and will be updated during installation

RAT_README.txt
kevinxu021: could you please list out what those three files for separately. "Apache Trafodion DCS default configuration file" is not clear.

install/installer/setup_ssh_key
kevinxu021: curious, nothing in it. it could be created during installation.

RAT_README.txt
hegdean: This is not documentation but a short description for each component  for list of files that excludes the copyright information as mentioned in the initial section of the this file.

core/sqf/sql/scripts/cleanlogs
hegdean: This will be used only during troubleshooting not otherwise

install/installer/setup_ssh_key
hegdean: This is is similar to our backup-masters file which is also empty and does no harm but it helps users to know that they can modify this file if they require DcsMaster to start on a specific node. Dcs documentation will be updated in the subsequent PR

PR #234, opened 2015-12-17
A new option "-p available" to install_local_hadoop. 

nonstop-qfchen:
With this option, the port range is chosen randomly from the list returned reported by findPort.py script.

### DIFF
--- a/core/sqf/sql/scripts/findPort.py
+++ b/core/sqf/sql/scripts/findPort.py
@@ -104,3 +104,4 @@ if not foundOne:
 print
 
 exit(exitCode)
+
--- a/core/sqf/sql/scripts/install_local_hadoop
+++ b/core/sqf/sql/scripts/install_local_hadoop
@@ -322,7 +322,7 @@ MY_CMD=$0
 MY_IMPLICIT_Y=
 NOACTION=
 VERBOSE=
-MISSING_P_VAL="ERROR: -p option should be followed by <start port num>, rand or fromDisplay"
+MISSING_P_VAL="ERROR: -p option should be followed by <start port num>, rand, fromDisplay or available"
 
 # process command line arguments
 ################################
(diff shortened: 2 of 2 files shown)


zellerh:
Looks good to me!

zellerh:
Actually, here is one comment I want to make: We currently use the following ranges:

-p rand uses ports between 12000 and 24000
-p fromDisplay uses ports above 24000. If the display is less than 42, it stays below the boundary of the ephemeral range, which usually starts at 32K.

Since -p fromDisplay is meant to be deterministic, -p available should not use ports above 24000, since that would break the idea that once you have a display running, that guarantees a corresponding port range. Could we make -p available stay in the range between 12000 and 24000?

nonstop-qfchen:
Hi Hans,

Thanks for the comment. I have submitted the rework. Thanks --Qifan

On Thu, Dec 17, 2015 at 5:34 PM, Trafodion Automated Test <
notifications@github.com> wrote:

> Test Passed. https://jenkins.esgyn.com/job/Check-PR-master/218/
>
> —
> <https://github.com/apache/incubator-trafodion/pull/234#issuecomment-165615753>
> .
>



-- 
Regards, --Qifan

nonstop-qfchen:
Good to know that the push event went through.  I pushed another one to
make sure it is the case.

#220 should contain the latest rework change.

thanks --Qifan



On Thu, Dec 17, 2015 at 6:51 PM, Trafodion Automated Test <
notifications@github.com> wrote:

> Previous Test Aborted. New Check Test Started:
> https://jenkins.esgyn.com/job/Check-PR-master/220/
>
> —
> <https://github.com/apache/incubator-trafodion/pull/234#issuecomment-165626834>
> .
>



-- 
Regards, --Qifan

zellerh:
Note that this code is not exercised by the Jenkins test, so no need to wait for it (probably could do a skip test).

PR #235, opened 2015-12-20
[TRAFODION-1699] Ported the Trafodion Client Install Guide to asciidoc.

gtapper:
Build the document using the pom.xml in docs/client_install:

   clean site post-site

The build will generate the web book as well as the PDF file for the
guide.

The generated PDF file will be copied to docs/target and
docs/target/1.3.0 as Trafodion_Client_Install_Guide.PDF.

The generated web book will be copied to docs/target/client_install and
docs/target/1.3.0/client_install. The index.html file in both these
directories invokes the web book.

Please test this level of automation. If it works, then we can make
similar changes to other asciidoc-generated documents thereby populating
docs/target automatically.

docs/src/markdown/documentation.md has NOT been updated to point to the
new web book. That'll be done as soon as this checkin has been verified by
a committer.

I don't understand why the win-odbc64 directory was deleted as part of this check-in -- hopefully, a result of git fetch --all?

jenkins, skip test


gtapper:
Withdrawn due to messed up branch.

PR #236, opened 2015-12-21
[TRAFODION-1685]Fix catalog api for getCatalog and getSchema

kevinxu021:


core/conn/jdbc_type2/native/SrvrSmd.cpp
```diff
@@ -357,6 +357,7 @@ void odbc_SQLSvc_GetSQLCatalogs_sme_(
 			}
 
 			SQLValue = (SQLValue_def *)outputValueList->_buffer;
+			/* //useless and doesn't be included in MXOS server
```
hegdean: Can we remove the code in comments?  Is there a reason why we need to keep it?

DaveBirdsall:
Is this change ready to be committed?

hegdean:
We need to make sure we have regression tests added to the test library?   Kevin, did we add test case for this  fix to our test library?

kevinxu021:
Dave, please feel free to commit.
Anu, For QA test case, i'm not sure. For me, have already done the testing.

hegdean:
Would it be possible just add junit test case for this fix.  Thanks

PR #237, opened 2015-12-21
[TRAFODION-1718] Add Batch test cases to dcs/src/test/jdbc_test folder

ryzuo:
New test cases added:
1. Batch insert with string overflow.
2. Batch update with string overflow.
3. Batch update with ID duplication.
4. Order the result set for test result assertion.


DaveBirdsall:
jenkins, retest

hegdean:
In jdbc_test folder as we already have TestBatch.java program can we rename the BatchTest.java to BatchDMLTest.java as it covers the various dml statements. Also we don't want to duplicate this tests that got merged to  phoenix folder,  may be you could remove the same test from phoenix test folder.

ryzuo:
@hegdean Hi Anu, I don't quite understand your point. First, if the new tests need be named as BatchDMLTest.java, does it mean that the old TestBatch.java is for DDL test? But I don't see there's DDL test, batch operations are all just DML(insert/upsert/update).
Second, the code in TestBatch.java is really hard to read and maintain, lots of hardcode, test case granularity is too big (too many columns in one single table of one single test case, hard to debug), unclear test objective, nearly no comment. The main test objective I see in that code is just different data types. But Rowsets operation essentially is transaction, data type is not really the major focus. Honestly, I rather to remove this TestBatch.java, and re-write them all.

So I will just keep the new file name just like it is BatchTest.java now, rename it won't tell any difference, and later future I will replace the old TestBatch.java with more clear test cases.

And sure thing, the ones in phoenix can be removed, I will do that.

arvind-narain:
I'll merge this by end of day ( PST ) if there are no further comments.

PR #238, opened 2015-12-21
[TRAFODION-1714] Sample table population plan now printed before exec…

DaveBirdsall:
…ution

### DIFF
--- a/core/sql/ustat/hs_cli.cpp
+++ b/core/sql/ustat/hs_cli.cpp
@@ -112,6 +112,49 @@ private:
 #pragma warn(770)  // warning elimination
 
 // -----------------------------------------------------------------------
+// DESCRIPTION: Call SQL CLI to execute a SQL statement. The caller is
+//          responsible for all error checking and recovery.
+// INPUTS:  stmt = statement descriptor for the SQL query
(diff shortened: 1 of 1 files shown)


DaveBirdsall:
jenkins, retest

DaveBirdsall:
Re: "just wonder..." Which line did you have in mind?

selvaganesang:
It looks like this function might be having a resource leak already. The stmt, srcDesc are not getting deallocated from SQL. If so, it would be good to fix it.

zellerh:
Looks good to me, too.

DaveBirdsall:
Hi Selva. I think the original author covered the resource leak problem. In line 207 of the new file, an HSRef object is created referencing the statement descriptor and the source text input descriptor. The HSRef destructor deallocates those objects. This is a nice technique actually because the HSRef object is created as a stack variable so its destructor will always be called (well, except for a longjmp).

DaveBirdsall:
jenkins, retest

DaveBirdsall:
I will commit this change as soon as tests pass.

selvaganesang:
Thanks Dave for your explanation. Yes. It is a clever technique.

PR #241, opened 2015-12-25
[TRAFODION-1724]Fixes for getTypeInfo()

kevinxu021:


xwq:
jenkins, retest

kevinxu021:
jekins,retest

mashengchen:
jenkins, retest

kevinxu021:
jenkins,retest

DaveBirdsall:
I'm guessing this code change is OK; I don't know enough about this particular code to be certain. I wanted to ask also: I noticed references to NONSTOP_SQLMX_NSK.MXCS_SCHEMA in the code (not in your changes, but earlier in SrvrCommon.cpp). I'm wondering if those are now obsolete? Do they cause problems?

core/conn/jdbc_type2/native/SrvrCommon.cpp
```diff
@@ -1991,36 +1972,14 @@ short do_ExecFetchAppend(
...
-            "NONSTOP_SQLMX_NSK.MXCS_SCHEMA.CATANSIMX",
-            SQLCLI_ODBC_MODULE_VERSION,
-            1234567890,
-            sqlStmtType,
-            false,true);
-    }
+    pSrvrStmt = createSrvrStmt(dialogueId,
+        stmtLabel,
+        &sqlcode,
+        "NONSTOP_SQLMX_NSK.MXCS_SCHEMA.CATANSIMX",
```
hegdean: Do we still need  the MXCS_SCHEMA reference?

core/conn/jdbc_type2/native/CSrvrStmt.cpp
```diff
@@ -99,7 +99,7 @@ SRVR_STMT_HDL::SRVR_STMT_HDL(long inDialogueId)
     moduleName[0] = '\0';
     inputDescName[0] = '\0';
     outputDescName[0] = '\0';
-    isClosed = TRUE;
+    isClosed = FALSE;
```
selvaganesang: Changing the default value for isClosed from TRUE to FALSE can have other repercussions. We might miss throwing error. Can you please confirm this change.

core/conn/jdbc_type2/native/SrvrCommon.cpp
selvaganesang: This is the module name parameter. Can this point to NULL instead.

selvaganesang:
In general, it would be good if you can mention how this change solved the issue briefly. Neither JIRA nor the commit message didn't mention it unless I missed it somehow.

kevinxu021:
Hi Selva, flag isClosed should be false on statement initialization. And it's an internal statement that is why the last parameter of createSrvrStatement should be true. The issue is involved by those two flags.

arvind-narain:
jenkins, retest

selvaganesang:
Kevin, Thanks. I see the code below in EXECUTE method in SqlInterface.cpp. This code below expects the isClosed flag in SRVR_STMT_HDL to be set to TRUE after initialization and the execute to set to FALSE upon successful execution. I think the path taken by the getTypeInfo() call to execute the query is not setting the flag to FALSE after execution. That needs to be fixed rather than setting this flag to FALSE in the constructor. Logically, the SRVR_STMT_HDL is in closed state till the query is prepared and executed. The execution of the query changes it to open state.

if(GlobalInformation::getSQLMX_Version() == CLI_VERSION_R2)
        {
                pSrvrStmt->isClosed = FALSE;            // Assume we are open after EXEC_Exec

                if (pSrvrStmt->sqlStmtType & TYPE_SELECT) retcode = CLI_Exec(pStmt, pInputDesc, 0);
                else retcode = CLI_ExecFetch(pStmt, pInputDesc, 0);
        }
        else                            // VER 2.1
        {
                // Should be Closed. If Opened assert
                DEBUG_ASSERT(pSrvrStmt->isClosed, ("Server Statement is Open before execute."));
                pSrvrStmt->isClosed = FALSE;            // Assume we are open after EXEC_Exec
                switch (pSrvrStmt->getSqlQueryStatementType())
                {

kevinxu021:
Selva, Thanks for comments. Yes, it indicates the status of server-side statement. But nothing impact whatever it's true or false since it'll be set to true after prepare( if it's false, it'll try to close once more.) It has nothing with this issue, so I have set it back. Maybe in the future, it should be separated for the concept of statement status from T2 statement and SQL statement.

selvaganesang:
Thanks Kevin. I will be merging this PR after the check-PR test passes

arvind-narain:
Selva faced some problems merging. Will try to merge now.

PR #243, opened 2015-12-31
[TRAFODION-1726]Support getIndexInfo

kevinxu021:


svarnau:
jenkins, test

core/conn/jdbc_type2/native/SrvrCommon.cpp
```diff
@@ -1634,6 +1635,53 @@ short do_ExecSMD(
...
+		    "cast(-2 as smallint) TYPE, cols.COLUMN_NUMBER ORDINAL_POSITION, "
+		    "substr(cols.COLUMN_NAME,0,CHAR_LENGTH(cols.COLUMN_NAME)) COLUMN_NAME, "
+		    "cast((case when keys.ORDERING=0 then 'A' else 'D' end) as varchar(4)) ASC_OR_DESC,"
+		    "cast(-2 as int) CARDINALITY, cast(-2 as int) PAGES, cast(NULL as varchar(10)) FILTER_CONDITION "
+		    "from TRAFODION.\"_MD_\".OBJECTS obj "
+		    "left join TRAFODION.\"_MD_\".INDEXES index on obj.object_uid=index.BASE_TABLE_UID "
+		    "left join TRAFODION.\"_MD_\".OBJECTS idxobj on index.index_uid=idxobj.object_UID "
+		    "left join TRAFODION.\"_MD_\".COLUMNS cols on cols.column_name<>'SYSKEY' and idxobj.object_uid=cols.object_uid "
+		    "left join TRAFODION.\"_MD_\".KEYS keys on cols.object_uid=keys.object_uid and cols.COLUMN_NAME=keys.COLUMN_NAME and keys.column_name<>'SYSKEY' "
+		    "where 1=1 %s"
```
arvind-narain: Could this %s ( uniqueness ) be added just before FOR READ UNCOMMITTED .. ? or is this done for performance reason ?
 " where  
 " (obj.SCHEMA_NAME ...
 ..
 " %s FOR READ ..."

core/conn/jdbc_type2/native/SrvrSmd.cpp
```diff
@@ -300,6 +301,15 @@ void odbc_SQLSvc_GetSQLCatalogs_sme_(
 		inputParam[2] = schemaNmNoEsc;
 		inputParam[3] = expSchemaNm;
 	    inputParam[4] = NULL;
+	    if(uniqueness == 0)
```
arvind-narain: Can you please add some comment that this is related to SQL_INDEX_UNIQUE(0) or use the same ?

core/conn/jdbc_type4/build.xml
```diff
@@ -27,10 +27,11 @@
 	</tstamp>
 
 	<!-- input -->
-        <exec executable="/bin/sh">
-            <arg line="getBuildInfo"/>
-        </exec>
-        <property file="buildId"/>
+	<exec executable="git" outputproperty="bldId">
```
arvind-narain: is this change related to this Jira ?

core/conn/jdbc_type2/native/SrvrCommon.cpp
```diff
@@ -1634,6 +1635,53 @@ short do_ExecSMD(
...
+		    inputParam[4] = NULL;
+		    snprintf((char *)sqlString->dataValue._buffer, totalSize,
+		    "select "
+		    "obj.CATALOG_NAME TABLE_CAT, obj.SCHEMA_NAME TABLE_SCHEM, obj.OBJECT_NAME TABLE_NAME,"
+		    "cast((case when index.IS_UNIQUE=0 then 1 else 0 end) as smallint) NON_UNIQUE, "
+		    "cast(NULL as varchar(32)) INDEX_QUALIFIER, idxobj.OBJECT_NAME INDEX_NAME, "
+		    "cast(-2 as smallint) TYPE, cols.COLUMN_NUMBER ORDINAL_POSITION, "
+		    "substr(cols.COLUMN_NAME,0,CHAR_LENGTH(cols.COLUMN_NAME)) COLUMN_NAME, "
+		    "cast((case when keys.ORDERING=0 then 'A' else 'D' end) as varchar(4)) ASC_OR_DESC,"
+		    "cast(-2 as int) CARDINALITY, cast(-2 as int) PAGES, cast(NULL as varchar(10)) FILTER_CONDITION "
```
arvind-narain: Are these defaults ok ( -2 for cardinality, pages, type ) - just checking - I'm sure you have checked what these should be as per the API.

The other request is to add a test case for this api and port the change to mxosrvr also.

core/conn/jdbc_type2/native/SrvrCommon.cpp
```diff
@@ -1634,6 +1635,53 @@ short do_ExecSMD(
...
+		    "from TRAFODION.\"_MD_\".OBJECTS obj "
+		    "left join TRAFODION.\"_MD_\".INDEXES index on obj.object_uid=index.BASE_TABLE_UID "
+		    "left join TRAFODION.\"_MD_\".OBJECTS idxobj on index.index_uid=idxobj.object_UID "
+		    "left join TRAFODION.\"_MD_\".COLUMNS cols on cols.column_name<>'SYSKEY' and idxobj.object_uid=cols.object_uid "
+		    "left join TRAFODION.\"_MD_\".KEYS keys on cols.object_uid=keys.object_uid and cols.COLUMN_NAME=keys.COLUMN_NAME and keys.column_name<>'SYSKEY' "
+		    "where 1=1 %s"
+		    " and (obj.SCHEMA_NAME = '%s' or trim(obj.SCHEMA_NAME) LIKE '%s' ESCAPE '\\')"
+		    " and (obj.OBJECT_NAME = '%s' or trim(obj.OBJECT_NAME) LIKE '%s' ESCAPE '\\')"
+		    " FOR READ UNCOMMITTED ACCESS order by INDEX_NAME, ORDINAL_POSITION;",
```
arvind-narain: should this be ordered by NON_UNIQUE, TYPE, INDEX_NAME, and ORDINAL_POSITION ?

dcs/src/main/java/org/trafodion/dcs/servermt/serverHandler/ServerApiGetCatalogs.java
```diff
@@ -353,7 +353,7 @@ else if (columnType == ServerConstants.SQL_ROWVER){
                     case ServerConstants.SQL_API_SQLSTATISTICS:
                         if(LOG.isDebugEnabled())
                             LOG.debug(serverWorkerName + ". getIndexInfo (catalogNm :" + catalogNm + ", schemaNm :" + schemaNm + ", tableNm :" + tableNm + ", uniqueness :" + ((uniqueness == 1)? true : false) + ", true)");
-                        rs = dbmd.getIndexInfo(catalogNm, schemaNm, tableNm, (uniqueness == 1)? true : false, true);
+                        rs = dbmd.getIndexInfo(catalogNm, schemaNm, tableNm, (uniqueness == 0)? true : false, true);
```
arvind-narain: Thanks for the catch. Could you use SQL_INDEX_UNIQUE(0) or SQL_INDEX_ALL(1) ?

core/conn/jdbc_type2/native/SrvrCommon.cpp
kevinxu021: Says in standard API doc: 
CARDINALITY int => When TYPE is tableIndexStatistic, then this is the number of rows in the table; otherwise, it is the number of unique values in the index. 
PAGES int => When TYPE is tableIndexStatisic then this is the number of pages used for the table, otherwise it is the number of pages used for the current index. 
TYPE short => index type: ◦ tableIndexStatistic - this identifies table statistics that are returned in conjuction with a table's index descriptions 
◦ tableIndexClustered - this is a clustered index 
◦ tableIndexHashed - this is a hashed index 
◦ tableIndexOther - this is some other style of index

core/conn/jdbc_type2/native/SrvrCommon.cpp
kevinxu021: i don't think it's reasonable because ORDINAL_POSITION is the order the indexes which should be the first level.

dcs/src/main/java/org/trafodion/dcs/servermt/serverHandler/ServerApiGetCatalogs.java
```diff
@@ -353,7 +353,7 @@ else if (columnType == ServerConstants.SQL_ROWVER){
                     case ServerConstants.SQL_API_SQLSTATISTICS:
                         if(LOG.isDebugEnabled())
                             LOG.debug(serverWorkerName + ". getIndexInfo (catalogNm :" + catalogNm + ", schemaNm :" + schemaNm + ", tableNm :" + tableNm + ", uniqueness :" + ((uniqueness == 1)? true : false) + ", true)");
-                        rs = dbmd.getIndexInfo(catalogNm, schemaNm, tableNm, (uniqueness == 1)? true : false, true);
+                        rs = dbmd.getIndexInfo(catalogNm, schemaNm, tableNm, (uniqueness == ServerConstants.SQL_INDEX_ALL)? true : false, true);
```
arvind-narain: Thanks for the updates. Here shouldn't this be SQL_INDEX_UNIQUE (0 - like your previous change). You could update line 355 also.

kevinxu021:
Thanks Arvind, it's a mistake.

core/conn/jdbc_type2/native/SrvrCommon.cpp
arvind-narain: I'm checking on this one Kevin - I'll let you know.

core/conn/jdbc_type2/native/SrvrCommon.cpp
arvind-narain: Sorry - missed on this reply earlier. I was just quoting from the following for the order:
https://docs.oracle.com/javase/7/docs/api/java/sql/DatabaseMetaData.html#getIndexInfo(java.lang.String,%20java.lang.String,%20java.lang.String,%20boolean,%20boolean)

Retrieves a description of the given table's indices and statistics. They are ordered by NON_UNIQUE, TYPE, INDEX_NAME, and ORDINAL_POSITION.

core/conn/jdbc_type2/native/SrvrCommon.cpp
```diff
@@ -1634,6 +1635,53 @@ short do_ExecSMD(
...
+		    "cast(-2 as smallint) TYPE, cols.COLUMN_NUMBER ORDINAL_POSITION, "
+		    "substr(cols.COLUMN_NAME,0,CHAR_LENGTH(cols.COLUMN_NAME)) COLUMN_NAME, "
+		    "cast((case when keys.ORDERING=0 then 'A' else 'D' end) as varchar(4)) ASC_OR_DESC,"
+		    "cast(-2 as int) CARDINALITY, cast(-2 as int) PAGES, cast(NULL as varchar(10)) FILTER_CONDITION "
+		    "from TRAFODION.\"_MD_\".OBJECTS obj "
+		    "left join TRAFODION.\"_MD_\".INDEXES index on obj.object_uid=index.BASE_TABLE_UID "
+		    "left join TRAFODION.\"_MD_\".OBJECTS idxobj on index.index_uid=idxobj.object_UID "
+		    "left join TRAFODION.\"_MD_\".COLUMNS cols on cols.column_name<>'SYSKEY' and idxobj.object_uid=cols.object_uid "
+		    "left join TRAFODION.\"_MD_\".KEYS keys on cols.object_uid=keys.object_uid and cols.COLUMN_NAME=keys.COLUMN_NAME and keys.column_name<>'SYSKEY' "
+		    " where 1=1 "
```
arvind-narain: If you do land up touching the code, then "1=1 and " could be removed.

core/conn/jdbc_type2/native/SrvrCommon.cpp
arvind-narain: I checked on the query. With the current one if table name is not specified ( all indexes in a schema ) then I get index_names with NULL and get extra entries - please do check. 

Following query worked for me:

select
trim(obj.CATALOG_NAME) TABLE_CAT, trim(obj.SCHEMA_NAME) TABLE_SCHEM, trim(obj.OBJECT_NAME) TABLE_NAME,
cast((case when index.IS_UNIQUE=0 then 1 else 0 end) as smallint) NON_UNIQUE,
cast(NULL as varchar(1)) INDEX_QUALIFIER, 
trim(idxobj.OBJECT_NAME) INDEX_NAME, 
cast(3 as smallint) TYPE, 
cast(cols.COLUMN_NUMBER as smallint) ORDINAL_POSITION,
trim(cols.COLUMN_NAME) COLUMN_NAME, 
cast((case when keys.ORDERING=0 then 'A' else 'D' end) as varchar(4)) ASC_OR_DESC,
cast(NULL as int) CARDINALITY, cast(NULL as int) PAGES, cast(NULL as varchar(1)) FILTER_CONDITION
from TRAFODION."_MD_".OBJECTS obj
join TRAFODION."_MD_".INDEXES index on obj.OBJECT_UID=index.BASE_TABLE_UID 
join TRAFODION."_MD_".KEYS keys on index.INDEX_UID = keys.OBJECT_UID
join TRAFODION."_MD_".COLUMNS cols on obj.OBJECT_UID = cols.OBJECT_UID and keys.COLUMN_NUMBER = cols.COLUMN_NUMBER
join TRAFODION."_MD_".OBJECTS idxobj on index.index_uid=idxobj.OBJECT_UID  
where
(obj.SCHEMA_NAME = '%s' or trim(obj.SCHEMA_NAME) LIKE '%s' ESCAPE '\\')
and (obj.OBJECT_NAME = '%s' or trim(obj.OBJECT_NAME) LIKE '%s' ESCAPE '\\')
and (obj.object_type = 'BT')
%s FOR READ UNCOMMITTED ACCESS order by NON_UNIQUE, TYPE, INDEX_QUALIFIER, INDEX_NAME, ORDINAL_POSITION;

Just looked at the documentation for SQLSTATISTICS - ASC_OR_DESC should be char(1). For defaults I chose TYPE as 3 and others NULL 

@zellerh - Hans - could you look at the query for getting index information from metadata and advice - I referred to smdixquery in regression tests.

kevinxu021: For table name is null, API says table name is required.
Parameters:
catalog a catalog name; must match the catalog name as it is stored in this database; "" retrieves those without a catalog; null means that the catalog name should not be used to narrow the search
schema a schema name; must match the schema name as it is stored in this database; "" retrieves those without a schema; null means that the schema name should not be used to narrow the search
table a table name; must match the table name as it is stored in this database
unique when true, return only indices for unique values; when false, return indices regardless of whether unique or not
approximate when true, result is allowed to reflect approximate or out of data values; when false, results are requested to be accurate

kevinxu021: In my testing(jdbcT4), it should have an exception when table name is null.
For TYPE, if the values of TYPE depands on column ASC_OR_DESC, it's duplicated. By the way, 3 also made me confused. -2 also is not a reasonable value, just want to say it is not use right now waiting for a meaningful value in the future.

arvind-narain: 1. Sorry for the confusion and the delay Kevin. I was referring to the output when all tables (%) under a schema are specified. In my testing the query in the code returned some rows with empty index name and in some cases temporary table names etc. This can be fixed in a separate JIRA if needed.

 <Row id="39">
   <TABLE_CAT>TRAFODION</TABLE_CAT>
   <TABLE_SCHEM>MTD</TABLE_SCHEM>
   <TABLE_NAME>MTD0</TABLE_NAME>
   <NON_UNIQUE>0</NON_UNIQUE>
   <INDEX_QUALIFIER>NULL</INDEX_QUALIFIER>
   <INDEX_NAME>NULL</INDEX_NAME>
   <TYPE>-2</TYPE>
   <ORDINAL_POSITION>NULL</ORDINAL_POSITION>
   <COLUMN_NAME>NULL</COLUMN_NAME>
   <ASC_OR_DESC>D</ASC_OR_DESC>
   <CARDINALITY>-2</CARDINALITY>
   <PAGES>-2</PAGES>
   <FILTER_CONDITION>NULL</FILTER_CONDITION>
 </Row>
 <Row id="40">
   <TABLE_CAT>TRAFODION</TABLE_CAT>
   <TABLE_SCHEM>MTD</TABLE_SCHEM>
   <TABLE_NAME>__SCHEMA__</TABLE_NAME>
   <NON_UNIQUE>0</NON_UNIQUE>
   <INDEX_QUALIFIER>NULL</INDEX_QUALIFIER>
   <INDEX_NAME>NULL</INDEX_NAME>
   <TYPE>-2</TYPE>
   <ORDINAL_POSITION>NULL</ORDINAL_POSITION>
   <COLUMN_NAME>NULL</COLUMN_NAME>
   <ASC_OR_DESC>D</ASC_OR_DESC>
   <CARDINALITY>-2</CARDINALITY>
   <PAGES>-2</PAGES>
   <FILTER_CONDITION>NULL</FILTER_CONDITION>
 </Row>

2. I did look at the corresponding ODBC API. Ordinal position should start from 1, so please fix this one and I'll check this in. The following should be changed -

     cols.COLUMN_NUMBER ORDINAL_POSITION

to

   keys.KEYSEQ_NUMBER ORDINAL_POSITION

or with a cast if needed.

3. Also could you please create a separate enhancement JIRA - I think for  column_class 'S' -  "DIVISION BY" column_name (_DIVISION_n) and SALT we  should have an optional column at the end - maybe "REMARKS" or some appropriate heading which should give the text from text.text - null for non-system columns (column_class <> 'S')

kevinxu021: hi Arvind, sorry for missing your comments. It's in jdbcT2, so right now it's following JDBC API. And i tried getIndexInfo("%", "%", "%", false, true), it works fine. the first 3 parameters are required.

arvind-narain:
Not a problem Kevin. This has got delayed, hence want to get this in as soon as possible.

I do still think ORDINAL_POSITION should start from 1 - used to be the same for both ODBC and JDBC in the other platform supported. 

Do please look into the following - I think if we start with 0 then it will break applications trying to utilize this API. Hence would recommend to make that correction in the current checkin and then handle any other defect/enhancement observed via a different JIRA. Do let me know if you think differently after looking at this information. If so we can create a separate JIRA for this issue for investigation.

From oracle - this does indicate that it starts from 1 ( to distinguish from tableIndexStatistic )

ORDINAL_POSITION short => column sequence number within index; zero when TYPE is tableIndexStatistic

Other sites - have example where the ordinal position is utilized:

https://technet.microsoft.com/en-us/library/ms378646(v=sql.105).aspx
http://support.nuodb.com/hc/en-us/articles/204825560-JDBC-DatabaseMetaData-getIndexInfo-ordinal-position-

selvaganesang:
@kevinxu021 and @arvind-narain Is this PR good to go?

PR #244, opened 2016-01-04
[TRAFODION-1727]Fixes for DatabaseMetaData.getProcedures and getProcedureColumns

kevinxu021:


core/conn/jdbc_type2/native/SrvrCommon.cpp
```diff
@@ -1634,6 +1634,94 @@ short do_ExecSMD(
...
+
+                    snprintf((char *)sqlString->dataValue._buffer, totalSize,
+                        "select obj.CATALOG_NAME PROCEDURE_CAT, obj.SCHEMA_NAME PROCEDURE_SCHEM, "
+			"obj.OBJECT_NAME PROCEDURE_NAME, cast(NULL as varchar(10)) R1,cast(NULL as varchar(10)) R2,"
+                        "cast(NULL as varchar(10)) R3, cast(NULL as varchar(10)) REMARKS, cast(0 as smallint) PROCEDURE_TYPE, "
+                        "obj.OBJECT_NAME SPECIFIC_NAME "
+                        " from TRAFODION.\"_MD_\".OBJECTS obj "
+                        " where "
+                        " (obj.SCHEMA_NAME = '%s' or trim(obj.SCHEMA_NAME) LIKE '%s' ESCAPE '\\')"
+                        " and obj.OBJECT_TYPE='UR' "
```
zellerh: This query will select UDFs in addition to stored procedures. I hope that's ok.

core/conn/jdbc_type2/native/SrvrCommon.cpp
```diff
@@ -1634,6 +1634,94 @@ short do_ExecSMD(
...
+lse 0 end) as smallint) COLUMN_TYPE, "
+                            "cols.FS_DATA_TYPE DATA_TYPE, cols.SQL_DATA_TYPE TYPE_NAME, "
+                            "cols.COLUMN_PRECISION \"PRECISION\", cols.COLUMN_SIZE LENGTH, cols.COLUMN_SCALE SCALE, "
+                            "cast(1 as smallint) RADIX, cols.NULLABLE NULLABLE, cast(NULL as varchar(10)) REMARKS, "
+                            "cols.DEFAULT_VALUE COLUMN_DEF, cols.FS_DATA_TYPE SQL_DATA_TYPE, cast(0 as smallint) SQL_DATETIME_SUB, "
+                            "cols.COLUMN_SIZE CHAR_OCTET_LENGTH, cols.COLUMN_NUMBER ORDINAL_POSITION, "
+                            "cols.NULLABLE IS_NULLABLE, cols.COLUMN_NAME SPECIFIC_NAME"
+                            " from TRAFODION.\"_MD_\".OBJECTS obj "
+                            " left join TRAFODION.\"_MD_\".COLUMNS cols on obj.OBJECT_UID=cols.OBJECT_UID "
+			    " left join TRAFODION.\"_MD_\".KEYS keys on cols.COLUMN_NAME=keys.COLUMN_NAME and cols.OBJECT_UID=keys.OBJ
```
zellerh: Is this outer join with the keys table really needed?

svarnau:
jenkins, test

svarnau:
Build flagged syntax error:

native/SrvrCommon.cpp:1705: error: missing terminating " character	##(JDBC_TYPE2)
native/SrvrCommon.cpp:1706: error: missing terminating " character	##(JDBC_TYPE2)
native/SrvrCommon.cpp:1715: error: missing terminating " character	##(JDBC_TYPE2)
native/SrvrCommon.cpp:1716: error: missing terminating " character	##(JDBC_TYPE2)
native/SrvrCommon.cpp: In function â€˜short int do_ExecSMD(void*, const CEE_handle_def*, ExceptionStruct*, ERROR_DESC_LIST_def*, long int, const char*, long unsigned int, short int, const char*, short int, const char*, const char*, const char*, const char*, const char**, const char**, SQLItemDescList_def*, long int*)â€™:	##(JDBC_TYPE2)
native/SrvrCommon.cpp:1706: error: expected â€˜)â€™ before â€˜lseâ€™	##(JDBC_TYPE2)
native/SrvrCommon.cpp:1706: error: expected â€˜;â€™ before â€˜asâ€™	##(JDBC_TYPE2)

kevinxu021:
Thanks for comments.
Changes:
1. Remove "left join keys".
2. Fix compile error.

arvind-narain:
Thanks for the much needed change Kevin - great work.

Some question/comments:
1.in SQLPROCEDURES can  PROCEDURE_TYPE be used to indicate if it is an SPJ or UDF ( or maybe REMARKS ) ?
2. Can inputParam[2] and inputParam[3] be utilized to search on a wild card procedure name ?
3. This change will be needed on the server end (mxosrvr) to support T4 and ODBC for non-multiDCS env. Would that get handled via this Jira too ?
4. Could you also add some test cases for getProcedures and getProcedureColumns ?

Thanks.

core/conn/jdbc_type2/native/SrvrCommon.cpp
kevinxu021: This query usually used for UI tools like DBVisualizer, SQuirreL SQL Client and so on. It might be  good if you UDF could be listed as a SPJ, i think.

kevinxu021:
1. Not sure since it's ok to list them out.
2. Sure.
3. I'd like to have another JIRA since MXOS is in use that we should treat it more careful.
4. Where to add? do you mean DEV unit test or QA test suit?

arvind-narain:
1. If the select is listing both spjs and udfs then it will help to distinguish between the two. Seen some other products that use PROCEDURE_TYPE for the same.
2. Thanks for the change.
3. Sure - that should be ok.
4. Did mean DEV unit tests.

core/conn/jdbc_type2/native/SrvrCommon.cpp
```diff
@@ -1634,6 +1634,92 @@ short do_ExecSMD(
...
+                    if ((!checkIfWildCard(catalogNm, catalogNmNoEsc) || !checkIfWildCard(schemaNm, schemaNmNoEsc) || !checkIfWildCard(
+tableNm, tableNmNoEsc)) && !metadataId)
+                    {
+                        executeException->exception_nr = odbc_SQLSvc_GetSQLCatalogs_ParamError_exn_;
+                        executeException->u.ParamError.ParamDesc = SQLSVC_EXCEPTION_WILDCARD_NOT_SUPPORTED;
+                        FUNCTION_RETURN_NUMERIC(EXECUTE_EXCEPTION,("EXECUTE_EXCEPTION"));
+                    }
+
+                    if (strcmp(catalogNm,"") == 0)
+                        strcpy(tableName1,SEABASE_MD_CATALOG);
```
selvaganesang: Should this be changed to the default catalog in the connection when the catalogNm is NULL?  Default to SEABASE instead of SEABASE_MD_CATALOG when it is not set.  Ditto for all other APIs.

arvind-narain:
Regarding PROCEDURE_TYPE, following is what is described for odbc SQLPROCEDURES. I know this is for Java T2 but could we use the same to distinguish SPJs and UDF ? @zellerh Han's any opinion ?

SQL_PT_UNKNOWN (0): It cannot be determined whether the procedure returns a value.
SQL_PT_PROCEDURE (1): The returned object is a procedure; that is, it does not have a return value.
SQL_PT_FUNCTION (2): The returned object is a function; that is, it has a return value.

zellerh:
Yes, I agree, it would be good to set PROCEDURE_TYPE to SQL_PT_PROCEDURE for SPJs and to SQL_PT_FUNCTION for scalar UDFs and TMUDFs. You may need to join the OBJECTS table to the PROCEDURES table to do that. Here is a query I tried:

    select object_name, object_type, udr_type,
           case when udr_type = 'P' then 1
                when udr_type = 'F' or udr_type = 'T' then 2
                else 0 end procedure_type
    from objects join routines
      on objects.object_uid = routines.udr_uid
    where object_type = 'UR';

core/conn/jdbc_type2/native/SrvrCommon.cpp
kevinxu021: SEABASE is a schema name, but SEABASE_MD_CATALOG is catalog name. Anything i missed?

selvaganesang: Only TRAFODION can be the catalog. Please change it to TRAFODION.

kevinxu021: It's TRAFODION actually.

kevinxu021:
Thanks for comments. Already done.

arvind-narain:
Thanks for the updates Kevin - I'll merge this shortly.

Selva, if there is a concern that user supplied catalog name should only be TRAFODION and the method should error out if not, or it should override the supplied catalog name with TRAFODION then we could fix this for all APIs via a separate JIRA. Hope that is ok.

Kevin, do please plan to add unit tests as part of JIRA TRAFODION-1741 or a separate JIRA to cover these new APIs for JDBC.

core/conn/jdbc_type2/native/SrvrCommon.cpp
selvaganesang: I am sorry. Please ignore my comment. I mistook it for schema name somehow

PR #246, opened 2016-01-05
[TRAFODION-1737]Fixed jar confiction

kevinxu021:


svarnau:
jenkins, add user

arvind-narain:
Hi Kevin, could you please mention the scenario and the error seen ?

kevinxu021:
Hi Arvind, as you know, DCSServer is a JAVA program and CLASSPATH is required. if you check CLASSPATH, you would notice that there are 2 kinds of jsp-api jar. One has ExpressionFactory.newInstance(), but the other one doesn't.  if JVM loaded the one doest have newInstance(), it will cause the error. Unfortunately, it occurs on my VM. Anyway, it's not reasonable to include more than one version for the same project JAR.

arvind-narain:
Thanks for the explanation. Suggest to include such information in future Jiras/pull request. Will merge in shortly.

PR #247, opened 2016-01-05
TRAFODION-1736 REST server reports duplicate DCS connections

venkat1m:
Fix for TRAFODION-1736 where REST server was reporting duplicate entries for servers after a dcs restart.

Also fixed these other errors reported by eclipse:
- Fixed the java package name for the GetStatusResponse.java and ConfTool.java
- Removed the @override annotation for some of the methods which were either implementing interface methods or abstract methods with no real implementation in the super class. Touched classes are ZkClient , JAXBContextResolver and PlainTextMessageBodyProducer

### DIFF
--- a/core/rest/src/main/java/org/trafodion/rest/GetStatusResponse.java
+++ b/core/rest/src/main/java/org/trafodion/rest/GetStatusResponse.java
@@ -20,7 +20,7 @@
 *
 * @@@ END COPYRIGHT @@@                                                          */
 
-package org.trafodion.dcs.rest;
+package org.trafodion.rest;
 
 import java.io.IOException;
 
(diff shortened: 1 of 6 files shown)


venkat1m:
The fix was verified using Postman and looking at the JSON output of the /servers/dcs/connections REST url. After a DCS restart, the JSON output only showed one set of entries.

DaveBirdsall:
Looks good to me.

arvind-narain:
Thanks Venkat. Looks good.

hegdean:
Thanks Venkat. Changes look good to me. 

For future release we should think about adding test framework for REST component

arvind-narain:
merging now.

PR #251, opened 2016-01-07
[TRAFODION-1680] Dcs needs to provide support for HA on AWS

hegdean:
Documentation and installer changes


dcs/bin/scripts/dcsbind.sh
```diff
@@ -354,7 +354,7 @@ dcsEcho "gv_float_external_ip :" $gv_float_external_ip
 dcsEcho "gv_float_internal_ip :" $gv_float_internal_ip
 
 #Check if AWS_CLOUD environment variable defined
-if [[ -z $AWS_CLOUD ]]; then
+if [[ $AWS_CLOUD == "false" ]]; then
```
arvind-narain: wasn't the earlier check ok ? Does AWS_CLOUD get set to false also or just true/empty ?

arvind-narain:
Merging this shortly if there are no further comments.

arvind-narain:
Some differences noted between the version tested on AWS and this. Will merge after the change is checked in.

arvind-narain:
Merging now. Some issues seen with single node installs for AWS will be handled separately.

PR #252, opened 2016-01-07
[TRAFODION-1742] Failure of certain Hive regression tests on HDP alone

sureshsubbiah:


core/sql/executor/hiveHook.cpp
```diff
@@ -319,7 +319,7 @@ struct hive_sd_desc* populateSD(HiveMetaData *md, Int32 mainSdID,
   
   // TODO : loop over SDs
   if (findAToken(md, tblStr, pos, "sd:StorageDescriptor(", 
-                 "getTableDesc::sd:StorageDescriptor(###"))
+                 "getTableDesc::sd:StorageDescriptor(###"),FALSE)
```
sandhyasun: Is the idea to suppress the error and see if the hive operation succeeds ? All the other calls above this seem to default to TRUE?

sureshsubbiah: Unlike other findAToken calls, this one expects to not find the token. If a token is indeed found we return NULL and will not create a NATable object. In the typical case, findAToken will return False and control will pass to line 325. If the last argument to findAToken, raiseError uses the default value of TRUE we will add an error detail text to the HiveMetaDB object. This error detail information will be used in NATable.cpp when error 1192 is raised. I think in the problem case we have the error detail text as always (which is usually harmless) but then later in the code path we get triggered onto raising an error in NAtable.cpp and use this older error detail text, thus maybe hiding the real error.

The metadata string returned by Hive in HDP when the test fails is the exact same as what we usually get.

sandhyasun: Ok looks like the right fix then. We don't want this error text to show up for some other NATable error.

zellerh:
Thanks, looks good to me!

sandhyasun:
I am having permission issues while doing a "git push" right now and unable to merge. The change is ready to be merged.

DaveBirdsall:
I will try to merge it

DaveBirdsall:
Merged. Sandhya, I'm wondering if you were using the wrong password for the "git push"? I use my apache password. Perhaps you were using your github password?

PR #254, opened 2016-01-08
[TRAFODION-1747] Avoid generating SQL errors in successful statements.

zellerh:
Using the "reportError" parameter in the underlying method to avoid
generating an error when we just want to test whether a certain
object name is registered in the OBJECTS table.

### DIFF
--- a/core/sql/optimizer/NATable.cpp
+++ b/core/sql/optimizer/NATable.cpp
@@ -4768,13 +4768,14 @@ ULng32 hashColPosList(const CollIndexSet &colSet)
 
 
 // ----------------------------------------------------------------------------
-// method: lookupObjectUid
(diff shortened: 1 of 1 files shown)


DaveBirdsall:
Looks good to me. Code is simpler than before. And I imagine will be less tedious to debug.

selvaganesang:
Looks good to me. But, do you think we should do the same with other callers of getObjectUID?

sureshsubbiah:
Looks good to me. Thanks.

zellerh:
Thanks to the reviewers. Selva, what I changed is a static function and I also renamed it, so all the callers of this function have been changed. The other calls to getObjectUID are calls to a method.

PR #256, opened 2016-01-08
Fix for JIRA TRAFODION-1748

sbroeder:
UnknownTransactionException in prepare causes error 97


DaveBirdsall:
Looks OK to me. I take it the fix to the bug is the change in TransactionRegionLocation.java, while the other changes are simply to give more logging detail?

sbroeder:
Thanks Dave.  You're correct.  The real fix is to realize that if a new region's start and end keys are contained withing the existing region's start and end keys it is the same region.

svarnau:
jenkins, add user

narendragoyal:
Looks good to me.

sbroeder:
Thanks Narendra.

PR #257, opened 2016-01-09
for jira 1720, add support to convert gbk into utf8

traflm:
This pull request is for a fix to JIRA 1720, add support of GBKTOUTF8 translate function. Please help to review it.


svarnau:
jenkins, ok to test

core/sql/common/csconvert.cpp
```diff
@@ -1275,3 +1279,41 @@ char * findStartOfChar( char *someByteInChar, char *startOfBuffer )
...
+  return outlen;
+}
+/* from gbk to utf8 */
+int gbk2utf8(char *inbuf,size_t inlen,char *outbuf,size_t outlen)
+{
+  return code_convert("gbk","utf-8",inbuf,inlen,outbuf,outlen);
+}
+
+int gbkToUtf8(char* gbkString, size_t gbklen, 
+              char* result ,size_t outlen, int addNullAtEnd)
```
DaveBirdsall: Would bool be a better data type for addNullAtEnd?

core/sql/common/csconvert.cpp
```diff
@@ -1275,3 +1279,41 @@ char * findStartOfChar( char *someByteInChar, char *startOfBuffer )
...
+{
+
+   int finalLength = gbk2utf8 ( gbkString, gbklen,  result, outlen);
+   
+   if (finalLength == -1 ) return 0;
+   
+   if ( addNullAtEnd > 0 )
+      result[finalLength] = 0;
```
DaveBirdsall: Could we have a buffer overrun here?

core/sql/exp/exp_conv.cpp
```diff
@@ -9321,6 +9321,32 @@ convDoIt(char * source,
...
 
+// gb2312 -> utf8
+  case CONV_GBK_F_UTF8_V:
+  {
+    char * targetbuf = new char[sourceLen*4+1];
+    size_t sl = sourceLen;
+    int convLen = gbkToUtf8( source, sl, targetbuf, sl*4);
+    int copyLen = 0;
+    if (convLen > 0) {
+      copyLen = (convLen< targetLen) ? convLen: targetLen;
```
DaveBirdsall: Might this truncate part of a UTF-8 character? I think there are already functions in the product that will copy UTF-8 strings safely without risk of truncation.

core/sql/common/csconvert.cpp
traflm: This 'addNullAtEnd is defined as int for all other similar functions, so following the convertion. But you are right, I will try to change it to bool.

core/sql/common/csconvert.cpp
traflm: yes, let me double check. A len checker is better here. will change it.

core/sql/exp/exp_conv.cpp
traflm: yes, truncation is need to be considered. I overlook this part. Will add it asap. 
thanks Dave for all your valuable comments.

DaveBirdsall:
I'm OK with committing this, so long as the truncated character problem in exp_conv.cpp is taken care of in a later check-in.

core/sql/exp/exp_clause_derived.h
```diff
@@ -1542,7 +1542,9 @@ enum conv_case_index {
   CONV_UTF8_F_UCS2_V                   =248,
 
   CONV_BLOB_BLOB                       =249,
-  CONV_BLOB_ASCII_F                    =250
+  CONV_BLOB_ASCII_F                    =250,
+
+  CONV_GBK_F_UTF8_V                    =251
```
sureshsubbiah: Do we also need a case for CONV_GBK_F_UTF8_F ? I am not quite how these case indexes work though. Maybe my question is if we have the user giving us GBK characters for a CHAR(N) UTF8 column, will it work with this change? Or is change applicable only when the target column in of type VARCHAR(N) UTF8

core/sql/exp/exp_conv.cpp
sureshsubbiah: What if there is a real overflow, i.e. user has given a GBK value which when converted to UTF8 exceeds the max varchar length? Will that be covered by this error or is that checked in a different place? Some how I don't see max length of target column in here...

core/sql/optimizer/SynthType.cpp
```diff
@@ -5207,6 +5207,18 @@ const NAType *Translate::synthesizeType()
          err4106arg = SQLCHARSETSTRING_UTF8;
        break;
 
+     case GBK_TO_UTF8:
+       if (translateSource->getCharSet() == CharInfo::GBK || translateSource->getCharSet() == CharInfo::UnknownCharSet )
+         charsetTarget = CharInfo::UTF8;
```
sureshsubbiah: Why do we have || CharInfo::UnknownCharSet here? I do not see it it in neighbouring statements. This is just for my understanding.

core/sql/optimizer/SynthType.cpp
```diff
@@ -5207,6 +5207,18 @@ const NAType *Translate::synthesizeType()
...
+       else
+       {
+            if( CmpCommon::getDefaultString(HIVE_FILE_CHARSET) ==  CmpCommon::getDefaultString(HIVE_DEFAULT_CHARSET) )
+              err4106arg = SQLCHARSETCODE_GB2312;
+            else
+             charsetTarget = CharInfo::UTF8;
```
sureshsubbiah: I suppose this if/else logic is also something basic to this change, but I do not understand it. A comment will be helpful.

core/sql/exp/exp_conv.cpp
DaveBirdsall: My guess is that the targetbuf is always long enough, since it allows 4 bytes for every byte in the source. But you are right, there could be truncation when copyLen is calculated. I'm wondering if type synthesis for TRUNCATE guarantees that the output length is always large enough, though, and it is only upon assignment to a target column that we'd truncate?

core/sql/sqlcomp/nadefaults.cpp
```diff
@@ -1958,6 +1958,7 @@ SDDkwd__(EXE_DIAGNOSTIC_EVENTS,		"OFF"),
 
   DDkwd__(HIVE_DEFAULT_CHARSET,            (char *)SQLCHARSETSTRING_UTF8),
   DD_____(HIVE_DEFAULT_SCHEMA,                  "HIVE"),
+  DD_____(HIVE_FILE_CHARSET,            (char *)SQLCHARSETSTRING_UTF8),
```
sureshsubbiah: If we have to join two Hive tables, one in UTF8 and the other in GBK, I suppose this method will have to be extended? Does Hive metadata store charset info? If not would a comment next to the tablename help? This is just a minor suggestion for something you may have already considered, not a request to add this now.

sureshsubbiah:
Change looks really good. Thanks. Please consider adding a query to the hive regression test suite so that developers have a clear idea on how to use this feature. Additionally adding something to the wiki will also help.

One more thing, do we need changes explain and showplan? Or maybe they already work OK.

core/sql/common/csconvert.cpp
```diff
@@ -1275,3 +1279,47 @@ char * findStartOfChar( char *someByteInChar, char *startOfBuffer )
...
+  return outlen;
+}
+/* from gbk to utf8 */
+int gbk2utf8(char *inbuf,size_t inlen,char *outbuf,size_t outlen)
+{
+  return code_convert("gbk","utf-8",inbuf,inlen,outbuf,outlen);
+}
+
+/* convert gbk string into UTF8 */
+int gbkToUtf8(char* gbkString, size_t gbklen,
```
zellerh: Having two methods with almost the same name and almost the same parameter list is very hard to read. I find that very confusing. Could we give them more different names? Or, maybe just have a single method with an optional parameter?

core/sql/exp/exp_clause_derived.h
zellerh: That may come when we support the CHAR datatype in Hive? Since right now we only support STRING in Hive, maybe we never have any CHAR columns in a Hive table?

core/sql/exp/exp_conv.cpp
zellerh: If we have to check for truncation, see functions lightValidateUTF8Str() and findStartOfChar() in file core/sql/common/csconvert.h.

core/sql/exp/exp_conv.cpp
```diff
@@ -9321,6 +9321,33 @@ convDoIt(char * source,
   };
   break;
 
+// gb2312 -> utf8
+  case CONV_GBK_F_UTF8_V:
+  {
+    char * targetbuf = new char[sourceLen*4+1];
```
zellerh: This should call CharInfo::getMaxConvertedLenInBytes() in file core/sql/common/charinfo.h. This class already knows about GBK. Also, maybe we could add an optimization? An ASCII character in GBK maps into a single-byte character in UTF-8. As far as I can tell, GBK does not go beyond the Unicode Basic Multi-lingual Plane (BMP), meaning that a 2 byte GBK sequence makes at most a 3 byte Unicode character. So, the target length is 3/2 * source length, rounded up to the nearest integer. If targetLen is sufficient for that conversion, maybe we can avoid allocating a buffer and write directly into target? Doing a new and a delete is expensive in this performance-critical context.

core/sql/optimizer/SynthType.cpp
```diff
@@ -5207,6 +5207,18 @@ const NAType *Translate::synthesizeType()
...
+       }
+       break;
+
```
zellerh: Note that we also need to calculate the length of the new type, that is done below by calling CharInfo::getMaxConvertedLenInBytes(). Again, the optimization I mentioned on line 9327 would be useful here, to avoid wasting space. However, I see a problem in Translate::synthType here, in that it always generates an output type with a size specified in characters, not bytes. That's inefficient and we should consider fixing it. For GBK->UTF8, this means that a 1000 byte input buffer will result in a 4000 byte output buffer, even though we can show that 1500 bytes would be enough.

core/sql/sqlcomp/nadefaults.cpp
zellerh: Yes, but that's also a problem with the current implementation, with fully supported charsets. We can't join two tables with ISO and UTF8, either. IMHO the solution to that is called iconv.

zellerh:
Changes look good, but I am missing a part: What I would have expected is some code that will automatically convert data from the HIVE_FILE_CHARSET to HIVE_DEFAULT_CHARSET, if the two are different.

core/sql/sqlcomp/nadefaults.cpp
zellerh: The default value for HIVE_FILE_CHARSET should not be UTF8, since that would be wrong if we have a file with ISO8859-1 characters and set HIVE_DEFAULT_CHARSET to ISO88591. Maybe better to set it to an empty string, which would mean that HIVE_FILE_CHARSET is the same as HIVE_DEFAULT_CHARSET. Given the way we implement it, maybe we should only allow an empty string and GBK as values for HIVE_FILE_CHARSET.

core/sql/exp/exp_conv.cpp
traflm: The new commit check the copyLen and targetLen. If targetLen is not enough, instead of doing truncate, it will raise a SQL Error. 
This task is primarily designed to support bulk data loading of GBK data source. If the Trafodion do implicit truncation, it may not be a good idea. For example, a 8 char string will become 7 chars after loading, and may cause very hard to detect issue later.
And it is also complex to implement UTF8 truncate method, thanks Hans point out the two APIs , I will study them carefully.
However, given the 'raise error' handling, I think we don't need to care about Truncate in the Middle issue. Just report error and stop data loading.

core/sql/exp/exp_clause_derived.h
traflm: Let me add this support. I didn't realize what the '_F' means in the beginning. Hive does not have CHAR or VARCHAR, it only support 'string'. So I assume Trafodion treats it as VARCHAR always, but let me try to add support of CONV_GBK_F_UTF8_F as well.

core/sql/optimizer/SynthType.cpp
traflm: I need to remove the 'unknowncharset' checking. It is wrong.
In the beginning, I assumed Trafodion will treat the column charset of hive table as unknown, since it seems to me Hive doesn't have charset info in the column definition. So I add this checking, but it turns out Trafodion will treat Hive column charset as 'HIVE_DEFAULT_CHARSET', so it should never be 'unknown'. I will remove this checking, it is confusing and wrong assumption.

core/sql/optimizer/SynthType.cpp
traflm: yes, I will add comments. 
Related to what Hans suggested below, this logic need to change. 
It is the basic for this whole task. I will change the logic and add comments

core/sql/sqlcomp/nadefaults.cpp
traflm: yes, Hans is correct. The current logic cannot work well if in some cases, HIVE_DEFAULT_CHARSET is not UTF8. It is better to change HIVE_FILE_CHARSET as empty and change the checking logic in the synthType.cpp logic accordingly.

core/sql/common/csconvert.cpp
traflm: yes, I will remove dup functions.

core/sql/exp/exp_conv.cpp
traflm: very good point, new/delete is better to be avoided. let me check this. As a change for 'truncate concern' before, I let TRANSLATE raise SQL error if the target buffer is not long enough. So a better approach is to use the targetbuf directly and let iconv() to do all the dirty works. If the targetbuf is long enough, iconv() can success, if targetbuf is not long enough, iconv() will fail, we don't need to treat all details. iconv is GLibc API, tested and tuned for years. If iconv() return an error, Trafodion just raise SQL error based on the error code. This way, we no longer need to new a temp buffer , and do memcpy, should be a very good optimization.

traflm:
hi, Hans,
I cannot find a good way to do converting automatically. The current implicit converting is triggered only  when source and target data column charset is different. So if we can change HIVE_DEFAULT_CHARSET, which control the source charset, it will do an automatic converting with current code base. But since change that will have side-effect, so that is not supposed to change. Then the current implicit converting code will not be triggered.
In this case, I think ask the admin to explicit do a TRANSLATE make sense.

traflm:
hi, Suresh,
I would like to add new hive test case later in another pull request.
I don't think we need to add EXPLAIN change for this yet. We will start a bigger project to support GBK totally, EXPLAIN change belong to that task I hope.

selvaganesang:
I seem to recall LocaleToUTF8 kind of function getting called at the time of Hive to Trafodion loading. Do we need such conversion in this case.

core/sql/exp/exp_clause_derived.h
zellerh: Please see my comment below on automatically making this conversion. Given that, support for fixed characters may not be needed yet.

zellerh:
Ming, you said you don't see a good place to add the automatic conversion. Here is what I would suggest: Look at method FileScan::codeGenForHive() in file core/sql/generator/GenRelScan.cpp. This method builds two lists of CAST expressions, one for columns needed in predicates and one for columns needed only in the select lists. It makes CAST nodes by calling HbaseAccess::createAsciiColAndCastExpr(). By the way "ASCII" in this name is not quite correct. This produces an expression cast(NATypeToItem()). The NATypeToItem() represents the Hive column and this is the only place where we would support GBK for now (plus an equivalent place for inserts). If we wanted to translate a value, we would need to assign character set HIVE_FILE_CHARSET to "asciiType" in this method and produce cast(translate(NATypeToItem())) in this method. This would be needed only if a) this is for a Hive table, b) HIVE_FILE_CHARSET is non-empty (i.e. GBK) and not equal to HIVE_DEFAULT_CHARSET. For inserts, you would need to do something similar or disable inserts into GBK tables for now. You could add an optimization to skip the translate for numeric, datetime and interval columns that consist of ASCII characters only, if the translation leaves ASCII unchanged.

zellerh:
About the missing regression test that Suresh mentioned: Not having a test makes it harder to review the code, since the change is not complete. I would prefer a complete change with a test case, or at least a description of what works and what is not yet working.

core/sql/optimizer/SynthType.cpp
```diff
@@ -5207,6 +5207,24 @@ const NAType *Translate::synthesizeType()
...
+         charsetTarget = CharInfo::UTF8;
+       else
+       {
+           /* this is a solution to support GBK before Trafodion can support GBK in total
+            * see jira 1720 for more details
+            * the logic here is:
+            * when HIVE_FILE_CHARSET is not empty, it means the real charset in Hive table is not same as HIVE_DEFAULT_CHARSET
+            * in this case, allow the converting , ignoring the source charset checking above
+            */
+            if( CmpCommon::getDefaultString(HIVE_FILE_CHARSET) ==  "" ) //CmpCommon::getDefaultString(HIVE_DEFAULT_CHARSET) )
```
zellerh: See my other comment about where to put the automatic conversion. If you do that, this check will not be needed. Otherwise, this logic will now allow a TRANSLATE(GBKTOUTF8) anywhere in SQL statements. Allowing this just gives people an opportunity to shoot themselves in the foot, so it may be better to avoid it.

traflm: totally agree. 
I will add a test case, it will take some time to understand the test suite.
And I will look at the codeGen, I didn't understand the whole path clearly. Yes, it is the right place to do the converting. My current change allow translate function to be used in unexpected cases, so, I will have another version asap. 
Thanks for your patient review and guidance, it is very appreciated! There will be more commits in near future.

DaveBirdsall:
Just checking... are we expecting further changes for this?

traflm:
Yes, Dave, there will be more changes, and I will cancel this pull request and create a new one later. since there will be some big changes and will take some time, just keep this open for comments reference. Thanks all for the comments, it helps a lot.

traflm:
hi, all, again,

I made some new changes, and it seems to be better now.
1. add more comments
2. make the translate implicit from GBK to UTF8 in HIVE_SCAN
3. add a test case, modify the hive/TEST005 to test this new change

Now, TRANSLATE( col using GBKTOUTF8) cannot be used directly, only when select from hive table, and HIVE_FILE_CHARSET is set to 'GBK'. So no other side effort is possible.

Please help to review again and comment.

thanks,
Ming

core/sql/generator/GenRelScan.cpp
```diff
@@ -214,22 +215,32 @@ int HbaseAccess::createAsciiColAndCastExpr(Generator * generator,
       ((CharType*)newGivenType)->setUpshifted(FALSE);
     }
 
+  if (newGivenType->getTypeQualifier() == NA_CHARACTER_TYPE &&
+      CmpCommon::getDefaultString(HIVE_FILE_CHARSET) == "GBK")
```
zellerh: It might be good to add a check here that HIVE_FILE_CHARSET must be an empty string or equal to HIVE_DEFAULT_CHARSET or GBK. In all other cases we would probably not do what the user expected.

core/sql/generator/GenRelScan.cpp
```diff
@@ -787,6 +798,7 @@ short FileScan::codeGenForHive(Generator * generator)
   const Int32 executorPredTuppIndex = 3;
   const Int32 asciiTuppIndex = 4;
   ULng32 asciiRowLen; 
+  ULng32 translateRowLen;
```
zellerh: Looks like this is not needed? Same for transVids on line 811 and transVal on lines 1518 and 1614?

core/sql/generator/GenRelScan.cpp
zellerh: This may need to upshift the default value first, to catch the case where the user sets the CQD to lower case 'gbk'?

zellerh:
Good to see that you were able to do the automatic conversion. The implementation looks good, but it is restricted to a narrow case of converting Hive tables from GBK to UTF-8. It might be good to add checks for that specific case (see comments above).

zellerh:
jenkins, retest

core/sql/common/csconvert.cpp
```diff
@@ -1275,3 +1279,45 @@ char * findStartOfChar( char *someByteInChar, char *startOfBuffer )
...
+{
+   int originalOutlen = outlen;
+   int finalLength = charsetConvert( "gbk","utf-8", gbkString, gbklen,  result, outlen);
+   
+   if (finalLength == -1 ) 
+     return -1;
+   
+   if ( addNullAtEnd )
+   {
+     if(originalOutlen >= finalLength )
```
DaveBirdsall: Wouldn't it need to be >, rather than >=? (Or does the caller leave an extra byte for null terminator?)

core/sql/common/csconvert.cpp
```diff
@@ -1275,3 +1279,45 @@ char * findStartOfChar( char *someByteInChar, char *startOfBuffer )
...
+       result[finalLength] = 0;
```
DaveBirdsall: If there isn't room for a null terminator, should we return -1? (That is, does the caller always expect a null terminator to be added when "addNullAtEnd" is true?)

core/sql/exp/exp_conv.cpp
```diff
@@ -9321,6 +9321,32 @@ convDoIt(char * source,
   };
   break;
 
+// gb2312 -> utf8
+// JIRA 1720
+  case CONV_GBK_F_UTF8_V:
+  {
+    int copyLen = 0;
+    int convLen = gbkToUtf8( source, sourceLen, target, targetLen);
+    if (convLen > 0) {
```
DaveBirdsall: Should this be >=? (What happens when we convert an empy string?)

DaveBirdsall:
I found a few possible off-by-one kinds of bugs, but other than that (and the comments others raised) it looks good to me.

core/sql/generator/GenRelScan.cpp
traflm: will change this.

core/sql/common/csconvert.cpp
traflm: Yes, I agree with you, This should return -1. The target buffer allocate length of buffer according to the target attribute, no special consideration an extra byte for NULL char.

core/sql/exp/exp_conv.cpp
traflm: yes, this is correct, thanks a lot, if the source is empty string, the target will be empty string as well and the convLen will be 0. So it is a valid case.

DaveBirdsall:
Thanks for making the updates. You've worked very hard on getting this feature in, which will be very valuable. The changes look good to me now.

traflm:
thanks Dave, 
Hi, Hans, I have changed the checking logic to make more restriction:

  if (newGivenType->getTypeQualifier() == NA_CHARACTER_TYPE &&
 +      (CmpCommon::getDefaultString(HIVE_FILE_CHARSET) == "GBK" || CmpCommon::getDefaultString(HIVE_FILE_CHARSET) == "gbk") && CmpCommon::getDefaultString(HIVE_DEFAULT_CHARSET) == "UTF8" )
 +        needTranslate = TRUE;

So it will only do a translate when the HIVE_FILE_CHARSET is set to 'GBK'/'gbk' and HIVE_DEFAULT_CHARSET is 'UTF8'. 

thanks,
Ming

DaveBirdsall:
Just wanted to check: Is this ready for commit?

zellerh:
Yes, I think it is and I'll commit it shortly.

PR #258, opened 2016-01-11
TRAFODION-1729

mashengchen:
change the coprocessor deployment method, to avoid intrusive into underlying HBase system.


mashengchen:
lose one file

sureshsubbiah:
jenkins, add user

mashengchen:
jenkins, retest

DaveBirdsall:
jenkins, retest

mashengchen:
jenkins, retest

mashengchen:
jenkins, retest

mashengchen:
jenkins, retest

mashengchen:
jenkins, retest

mashengchen:
jenkins, retest

mashengchen:
i will optimize it, do not merge

mashengchen:
jenkins, retest

mashengchen:
jenkins, add user

mashengchen:
jenkins, retest

mashengchen:
jenkins, stop

mashengchen:
jenkins, retest

mashengchen:
seems have some version conflect..

mashengchen:
jenkins, add user

mashengchen:
jenkins, add user

sureshsubbiah:
jenkins, add user

zellerh:
The code looks good to me, but like others commented in the JIRA, I'm not quite convinced that this really helps a lot. Dave pointed out that it does not allow transactional operations on native HBase tables. Also, we still need to load Trafodion code into the region server, so in that sense we are still intrusive, and probably always will be.

This code allows us to use specific coprocessors for individual tables. That may be a good thing, but it could also be problematic, e.g. if we want to change all the coprocessors.

About the need for restarting HBase: If we add a new jar, that would have to be in the classpath of the region server. Wouldn't that require a restart of HBase, regardless of whether we change the hbase_coprocessor_region_classes configuration entry or not?

mashengchen:
this is the first step of jira1729, just peel hbase_coprocessor_region_classes configuration from hbase-site.xml. the point of "native hbase table can't use tranactional" still need to do some researches. i will do it in the next pull request.

DaveBirdsall:
I would prefer that we not merge this code so long as it removes transactional capabilities for native HBase tables. That feature is something that is advertised as a Trafodion feature, so we should not regress it. If we can fix this issue, then I'd be glad to merge this.

DaveBirdsall:
Looks good to me. Would like to hear what other reviewers say also.

zellerh:
I'm sitting somewhat on the fence. Part of me thinks that this may be a step in the right direction and that it shouldn't harm, part of me is still questioning how this will work in the long term. Could the author (and others, of course) comment on these questions?

- Do we have a plan to get rid of all of the Trafodion-related HBase configuration entries? If not, does it make sense to do work to eliminate this one?
- I assume that there is no upgrade required. Existing tables will have the coprocessors configured globally, new ones (after this change) will have per-table coprocessors?
- How do we switch coprocessors, e.g. going from MVCC to another method to handle concurrency? Today, this is done by editing the config file?
- Even this new method has dependencies. How are we going to manage the jar files required - is that going to be different from what we do today? Today this is done by the installer.
- Can this change avoid a restart of HBase? If we add a table-specific coprocessor that requires a jar file, won't that jar file have to be in the classpath of the region server? And, adding it to the classpath would require a restart?

Sorry, some of my questions come from a lack of understanding how this all works.

traflm:
hi, Hans,
I have some feedback:
•Do we have a plan to get rid of all of the Trafodion-related HBase configuration entries? If not, does it make sense to do work to eliminate this one?
  Yes, this is the first effort to get rid of all Trafodion-related HBase configuration entries. Another big one is to avoid overloading HRegion, HBase 1.2 already changes it for Trafodion. So all left are some configuration items, some I am rather sure we can do in Trafodion code instead of hbase-site.xml. Still 2 or 3 need investigation. But this single change will benefit still in some special cases, like CDH regionsServer group.

•I assume that there is no upgrade required. Existing tables will have the coprocessors configured globally, new ones (after this change) will have per-table coprocessors?
  No upgrade required. After this change, in fact, nothing changes . Trafodion still have the hbase-site.xml modified by installer, and if hbase-site.xml exist, it will take higher priority and all these new logic will not take effect. Just when hbase-site.xml doesn't have the coprocessor configured. These new code will take effect. So when we are confident enough , we can remove hbase-site.xml coprocessor configs.

•How do we switch coprocessors, e.g. going from MVCC to another method to handle concurrency? Today, this is done by editing the config file?
  Today, one must edit the ms.env and restart Trafodion to switch.

•Even this new method has dependencies. How are we going to manage the jar files required - is that going to be different from what we do today? Today this is done by the installer.
  No different about jar file handling.

•Can this change avoid a restart of HBase? If we add a table-specific coprocessor that requires a jar file, won't that jar file have to be in the classpath of the region server? And, adding it to the classpath would require a restart?
  If Trafodion doesn't have other hbase-site.xml dependencies, like overload the HRegion, this will avoid a restart of HBase. If one coprocessor needs a jar , admin MUST put it into the CLASSPATH that HBase RegionServer started with, even without this change, it is same. It will not need a restart. I will write a small demo to test it and share with you later. That can help better to understand this topic.

thanks,
Ming

zellerh:
Hi Ming, thanks for answering the questions. I think your answers show that this change doesn't really get us closer to eliminating the dependency between HBase and Trafodion. BTW, I totally agree that overloading the region class would be very good to avoid.

Since the change also doesn't make the dependency worse and since you have a plan to go forward I am ok with that.

PR #262, opened 2016-01-12
Trafodion 1699

gtapper:
Ported the following documentation to use asciidoc thereby being part of the Trafodion source tree: Client Installation Guide, Command Interface Guide, odb User Guide, and SQL Reference Guide. Also began the creation of a new Installation Guide, which is not activated yet.

Each document is part of the docs subdirectory under the docs directory. Builds are done using maven under each document's subdirectory: 

* *clean suite* builds the web book and the PDF version.
* *post-site* copies the generated files under the correct directories under docs/target, which is assumed to exist; that is, the *clean site* step under the master pom.xml should be run before building the documents.

Documents are now located under docs/target/docs under:

1. The document's directory, which is based on the source directory. This directory contains the latest version of the document.
2. Under <version number>/<document directory>, which contains the release version copy of the document.

All this is explained in the document.html file on the website.

This change means that the DCS and REST documents need to be copied to their respective places. Please refer to documentation.md, which shows where the documents should be placed on the target website and in which directory.

Previous .pdf versions on the website can be removed. Previous document subdirectories can be remove, too. This last statement is specific for this release only. In future releases, the latest versions will be overwritten and new version directories will be created and populated with the SPECIFIC assumption that the previous versioned directories are kept in place.


docs/.gitignore
```diff
@@ -1 +0,0 @@
-/target/
```
svarnau: Why is gitignore file being removed?

svarnau:
jenkins, add user

gtapper:
I haven't touched any .gitignore file?

svarnau:
Missing license statement per build check in trafodion/docs/css/trafodion-theme.yml

gtapper:
Fixed missing license statement in yaml file. Please rerun build.

svarnau:
You say you did not touch the .gitignore file, but your pull request definitely deletes that file. Please put it back.

gtapper:
Added back the gitignore file. Please rebuild and update website.

svarnau:
I was trying to read and follow instructions in document.md. It was a bit hard in raw form, but finally went to github location to see render: https://github.com/gtapper/incubator-trafodion/blob/TRAFODION-1699/docs/src/site/markdown/document.md 
Looks here like one of the tables in the build section is not formatted properly.  At least GH does not render it right.
Also, I was confused by references to MS Word docs, because I could not find any. I think it was oversight that needs to be removed now that all docs are asciidoc.

svarnau:
You refer to putting all these steps together in a makefile, but they are all maven build steps. Why not just put them together in the top-level pom file to build the various docs as sub-projects?  Not an expert in that, but I'd think better to have all maven.

svarnau:
The DCS and REST documents get generated as 2.0.0 per the main trafodion release version, but the other docs all get generated as 1.3.0.  So we have a mis-match.

gtapper:
I removed the reference to Word document. I have no idea why DCS and REST
are building 2.0.0 documents in the 1.3.0 release?

On Wed, Jan 13, 2016 at 11:42 AM, Steve Varnau <notifications@github.com>
wrote:

> The DCS and REST documents get generated as 2.0.0 per the main trafodion
> release version, but the other docs all get generated as 1.3.0. So we have
> a mis-match.
>
> —
> <https://github.com/apache/incubator-trafodion/pull/262#issuecomment-171393417>
> .
>



-- 
Thanks,

Gunnar
*If you think you can you can, if you think you can't you're right.*

PR #263, opened 2016-01-12
[TRAFODION-1752] JVM goes out-of-memory for hundreds of upsert statem…

selvaganesang:
…ent in one session

HTableClient object was not getting released under certain conditions in batch get operation.
Fixed the code to release the object under these conditions.

### DIFF
--- a/core/sql/executor/ExHbaseAccess.h
+++ b/core/sql/executor/ExHbaseAccess.h
@@ -845,7 +845,7 @@ public:
     , DONE
     , ALL_DONE
     , COMPLETE_ASYNC_INSERT
-    , DONE_WITH_CLOSE
+    , CLOSE_AND_DONE
 
   } step_;
 
@@ -1158,7 +1158,7 @@ public:
     , APPLY_PRED
     , RETURN_ROW
     , COMPLETE_ASYNC_OPERATION
-    , DONE_WITH_CLOSE
+    , CLOSE_AND_DONE
   } step_;
 
   ExHbaseAccessSQRowsetTcb( const ExHbaseAccessTdb &tdb,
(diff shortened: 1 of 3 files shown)


DaveBirdsall:
Looks good. Does this bug occur in any of the other HBase tcbs?

selvaganesang:
It is possible that it might happen with other tcbs too. It is important that the HTableClient is released from the JNI/ executor side for get, batch get and scan operations. The HTableClient object is released on the java side in case of Put and Delete operations for non-async operations. I will go through the code where getRowOpen, getRowsOpen and scanOpen is being called and fix it in those places too

DaveBirdsall:
Just checking: Selva, are you planning any more changes? If not, I think this change is ready to commit.

selvaganesang:
Yes. I have observed there is a possibility of leaking HTableClient objects in other places. I will fix them too.

DaveBirdsall:
Looks good to me. I might have chosen a different name for "DONE_WITH_CLOSE". To my ears, this suggests that we are done closing, but from the code I see that it means, "done, except we still need to do a close". "DONE_EXCEPT_CLOSE" might have been a little better. But I don't think this should hold back the commit.

selvaganesang:
I can change it into CLOSE_AND_DONE if needed because this is what this step does. Isn't?

DaveBirdsall:
That sounds good.

DaveBirdsall:
Looks good to me. Let's ship this puppy.

selvaganesang:
Can you please commit this tomorrow if you don't hear any more review comments

DaveBirdsall:
Yes



*From:* selvaganesang [mailto:notifications@github.com]
*Sent:* Wednesday, January 20, 2016 2:23 PM
*Cc:* DaveBirdsall <dave.birdsall@esgyn.com>
*Subject:* Re: [incubator-trafodion] [TRAFODION-1752] JVM goes
out-of-memory for hundreds of upsert statem… (#263)



Can you please commit this tomorrow if you don't hear any more review
comments

—
<https://github.com/apache/incubator-trafodion/pull/263#issuecomment-173381058>
.

PR #264, opened 2016-01-12
[TRAFODION-1720] Rework, misplaced paranthesis

sureshsubbiah:


DaveBirdsall:
The title for this change seems to be incorrect; should it be [TRAFODION-1742] Rework, ...

zellerh:
+1, this is a tough one to find!

sandhyasun:
Looks good !  Tricky one for the eyes indeed especially  when there was no compilation error :-)

PR #265, opened 2016-01-14
[TRAFODION-1682]MT-DCS protocol issue fix

xwq:
This commit will fix following issues.
1. [TRAFODION-1682][MTDCS] ODBC: SQLFetch returned invalid cursor
state. It's caused by the length of descriptor2List is not correct and
the sqlQueryType is not correct either. For sqlQueryType,
    SqlUtils.getSqlStmtType(String str) will be used instead of
    SqlUtils.getSqlStmtType(int stmtType) since it can't return the
    correct value.
2. [TRAFODION-1653][MTDCS] coast tests core dumped at 'operator new'
when Multithread DCS is on. It's caused by the EndTransaction message
missing a int which indicate 0 excaption when there is no exception.
3. [TRAFODION-1658][MTDCS] with MultiThread DCS on, coast core dumped
at memcpy when there is parameter in prepare statement. It's caused by
the length of the descriptor2List is not correct.

### DIFF
--- a/dcs/src/main/java/org/trafodion/dcs/servermt/serverDriverInputOutput/Descriptor2List.java
+++ b/dcs/src/main/java/org/trafodion/dcs/servermt/serverDriverInputOutput/Descriptor2List.java
@@ -80,7 +80,7 @@ public class Descriptor2List {
         if (descCount == descNumber){
             Descriptor2 desc = null;
             if (oldFormat == false)
-                descLength = 3 * ServerConstants.INT_FIELD_SIZE;
+                descLength = 2 * ServerConstants.INT_FIELD_SIZE;
             else
                 descLength = ServerConstants.INT_FIELD_SIZE;
             varLength = 0;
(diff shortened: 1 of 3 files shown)


arvind-narain:
Nice work Weiqing. Looks good to me. 

I'm assuming the change for TRAFODION-1653 is only needed for end transaction API.  

Will checkin by evening if there are no further comments.

xwq:
For now, I can only make sure the change for TRAFODION-1653 is needed by end transaction. But we still have many bugs when we run odbc driver test with MT-DCS. I will check all the APIs when I try to fix these bugs. 

However I think It's OK to checkin this change since the core dump will hold back other cases. I also add some comments to avoid confusing other people.

arvind-narain:
Thanks. Will merge in. If this change is needed by other APIs then the currrent change can be moved into insertInfoByteBuffer and lengthOfData.

PR #267, opened 2016-01-14
TRAFODION-1754 Showddl component does not display grants correctly

robertamarton:
Showddl should be displaying grants on an operation in a manner that allows
the privileges to be recreated.  Currently, they are returned in the order
in which they are read from the metadata.

While working on this JIRA, several issues were fixed including:

- Consolidated privilege values into one location - NAUserId.h. They used to be
  stored in NAUserId.h, ComSmallDefs.h, and PrivMgrDefs.h
- The check for getting the next available role ID was not correct.
- PrivMgrPrivileges::hasColWGO has incorrect indexing
- PrivMgrComponentPrivileges::selectAllWhere has incorrect error checking


robertamarton:
Haven't had any comments in 6 days so will go ahead and commit this pull request.

PR #268, opened 2016-01-14
[TRAFODION-1753] Wrong HASH2 partitioning for salted tables

zellerh:
Fix to create the HASH2 partitioning function for salted tables
with the number of salt buckets as the number of partitions and
with only the salt columns as the partitioning key.


DaveBirdsall:
Looks good to me.

PR #270, opened 2016-01-15
Fix reopened JIRA 1736

venkat1m:
The earlier fix caused only the mxosrvrs associated with last DCS servers to be displayed in the REST dcs connections output. This was because the fix was put inside a loop which cleared the dcsservers discovered for the previous nodes.

The fix is now moved outside the for loop and so preserves all the discovered mxosrvrs. Tested and verified this on a 2 node cluster.


DaveBirdsall:
Looks good to me. I will commit at the end of the day if there are no objections.

arvind-narain:
Looks good Venkat. Thanks for catching it.

PR #279, opened 2016-01-21
TRAFODION-1648 Region Split/Move work

trinakrug:
Rework of Region Split/Move code.  Enhanced error handling and ability to split/move with active transactions.


DaveBirdsall:
Looks good to me. But it is out of my area of expertise. Will commit this if someone from within this area of expertise gives it the OK.

sbroeder:
Looks good, although the formatting seems weird in places.  Looks like there are a lot of white space errors (tabs).  Logic is good.

trinakrug:
Most the formatting changes you see are actual differences in spacing.  It has to do with adding an if statement that surrounds an existing block of code, so all the code in those blocks were indented by 4 spaces.

sbroeder:
Thanks for clarifying Trina!

PR #280, opened 2016-01-22
[TRAFODION-1774]use hashCode instead of random

xwq:
[TRAFODION-1774][MT-DCS]use hashCode() instead of random to make the
stmtHandle is unique

### DIFF
--- a/dcs/src/main/java/org/trafodion/dcs/servermt/serverSql/TrafStatement.java
+++ b/dcs/src/main/java/org/trafodion/dcs/servermt/serverSql/TrafStatement.java
@@ -68,8 +68,7 @@ public class TrafStatement {
     public TrafStatement(String serverWorkerName, String stmtLabel, Connection conn, String sqlString, int sqlStmtType) throws SQLException {
         init();
         this.stmtLabel = stmtLabel;
-        stmtHandle = random.nextInt(1000);
-        stmtHandle = (stmtHandle < 0)? -stmtHandle : stmtHandle;
+        stmtHandle = this.hashCode();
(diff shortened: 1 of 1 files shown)


xwq:
Now, MT-DCS use random.nextInt(1000) to generator a number for statement which used as stmtHandle. However it can't make sure the value it unique for echo statement. hashCode() could be used instead of random.

xwq:
I have test it.In the test When using random.nextInt() it will generator a reduplicate stmtHandle after I creating 20 statements.

DaveBirdsall:
The change looks fine. I'm wondering, though, if hashcode() could return duplicate values? (Perhaps it does so much less frequently.)

xwq:
Jvm can make sure it is unique.

xwq:
JVM can guarantee it's unique. However I think it may return duplicate if there are more than 2^32 objects since it's a INT value. But I don't think we need consider this situation。

DaveBirdsall:
While I agree that a duplicate is less likely with this change, I'm wondering what the impact is when there is a duplicate. Is this still a serious problem from a user perspective? If the objective is simply to get a unique statement identifier, would it make sense to detect that a duplicate occurred and then add a retry loop that would just increment the value until a unique statement identifier is achieved?

xwq:
In many cases, TrafConnection need use the stmtHandle to identify a statement, so we need it is unique. 
TrafStatement itself can't know the stmtHandle of other statements, so we need detect the duplicate in TrafConnection and then set it to TrafStatement. 
I use hashCode() since it make the changes simple, and in Java, hashCode() is a native method , we don't need to worry about the performance.

PR #283, opened 2016-01-25
[TRAFODION-1778] Update command when HBASE_VSBB_ROWSET_OPT 'ON' fails…

selvaganesang:
… with internal error at times

Direct buffer used in update command was not getting reset correctly when exact number of
rows as the maximum number of rows set using CQD HBASE_VSBB_ROWSET_SIZE is buffered for rowset operation. This caused the subsequent rowset operation to fail.


DaveBirdsall:
Looks good to me. New code is simpler which is nice.

PR #284, opened 2016-01-25
[TRAFODION-1420] Use ClientSmallScanner for small scans to improve performance

eowhadi:
Hbase implements an optimization for small scan (defined as scanning less than a data block ie 64Kb) resulting in 3X performance improvement. The underlying trick is about cutting down RPC calls from 3 (OpenScan/Next/Close) to 1, and use pread stateless instead of seek/read state-full and locking method to read data. This JIRA is about improving the compiler who can be aware if a scan will be acting on single data block (small) or not, and pass this data to executor so that it can use the right parameter for scan. (scan.setSmall(boolean)).
reference:
https://issues.apache.org/jira/browse/HBASE-9488
https://issues.apache.org/jira/browse/HBASE-7266


DaveBirdsall:
Looks good to me.

DaveBirdsall:
I will commit this at the end of the day today if there are no further comments.

PR #287, opened 2016-01-27
[TRAFODION-1779] Non-optimal plan chosen for a query against the crea…

selvaganesang:
…ted table

in the same session

The CQD TRAF_RELOAD_NATABLE_CACHE can be turned on to invalidate the cache entry.

The usage will be as follows:
create table t1 ..
load into table t1 .. -->This adds the table t1 to NATable cache with estimated row count as 100
cqd traf_reload_natable_cache 'on' ;
select * from t1 -->The non-metadata table accessed by this query will be invalidated from
the NATable cache and reloaded
cqd traf_reload_natable_cache 'off'

Also removed the use of this CQD in the DDL portion of the codes because it was never used.
This CQD was used prior to introduction of query invalidation feature.

### DIFF
--- a/core/sql/optimizer/NATable.cpp
+++ b/core/sql/optimizer/NATable.cpp
@@ -7864,12 +7864,9 @@ NATable * NATableDB::get(CorrName& corrName, BindWA * bindWA,
       table = NULL;
     }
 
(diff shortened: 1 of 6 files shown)


DaveBirdsall:
Looks good to me.

PR #288, opened 2016-01-27
[TRAFODION-1784]remove closed statement from HashMap

xwq:
[TRAFDOION-1784][MTDCS]statement shoud be removed from HashMap when it's closed.
TrafConnection uses a HashMap to store the statements in MTDCS code.
When a statement is closed, it should be removed from the HashMap,
otherwise the JVM garbage collection couldn't delete the unused objects.


DaveBirdsall:
Looks good to me.

PR #290, opened 2016-01-28
TRAFODION-1595 Privilege manager tables missing from HBase

robertamarton:


DaveBirdsall:
Looks good to me. The logging should be especially helpful.

PR #291, opened 2016-01-28
[TRAFODION-1791] UPSERT with bare one row subquery in VALUES clause fails [TRAFODION-1792] UPSERT with parameters in VALUES clause fails

sureshsubbiah:


core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -10272,8 +10273,14 @@ RelExpr* Insert::xformUpsertToMerge(BindWA *bindWA)
...
   ((Join*)re)->setTSJForMergeWithInsert(TRUE);
+  ((Join*)re)->setTSJForMergeUpsert(TRUE);
   ((Join*)re)->setTSJForWrite(TRUE);
-  if (bindWA->hasDynamicRowsetsInQuery())
+
+  // if Inputs of current insert are empty (i.e. we have no params/rowsets)
+  // then there will be no pull up of inputs during transform and the join will
+  // not see the inputs of the mergeUpdate due to intermediate nodes. So
+  // add inputs directly to join and use elimination to remove extra inputs
+  if (NOT getGroupAttr()->getCharacteristicInputs().isEmpty())
```
zellerh: I'm having trouble understanding this if statement. As far as I can see, _myOuterRefs_ are the source values, coming from the child of the original insert. After the transformation they will come from the left child of the tuple flow, _re_, and be passed to the right child, _mu_. So, I would have added them unconditionally as char. inputs of _mu_. If the insert (_this_) char. inputs are non-empty, then I would have add those to _re_ and maybe even to _mu_.

core/sql/optimizer/NormRelExpr.cpp
```diff
@@ -1697,11 +1709,26 @@ void Join::recomputeOuterReferences()
       // If it is a TSJ don't add the outputs of the left child to
```
zellerh: Could also fix the comment.

core/sql/optimizer/NormRelExpr.cpp
```diff
@@ -7050,8 +7077,8 @@ void RelRoot::recomputeOuterReferences()
       child(0)->getGroupAttr()->getCharacteristicInputs().
                   getLeafValuesForCoverTest(leafValues, emptyGA, emptySet);
       CMPASSERT((getGroupAttr()->getCharacteristicInputs().contains
-                (child(0)->getGroupAttr()->getCharacteristicInputs())) || 
-                (getGroupAttr()->getCharacteristicInputs().contains (leafValues)));
+		 (child(0)->getGroupAttr()->getCharacteristicInputs())) ||
```
zellerh: Message from the TAB police, if you can, configure your editor to put spaces instead of tabs. The good news is that they are configured as 8 spaces :-)

core/sql/regress/executor/EXPECTED015.SB
```diff
@@ -1721,4 +1799,70 @@ A            B                     C
...
+--- SQL command prepared.
+>>
+>>prepare s8 from UPSERT INTO DE (ID, dataElementConceptID, valueDomainID)  
++>VALUES  (?[10], ?[10], ?[10] ) ;
+
+--- SQL command prepared.
+>>
+>>prepare s9 from UPSERT INTO DE (ID, dataElementConceptID, valueDomainID)  
++>VALUES  (1,  (select d.id from DEC1 d where d.codeValue = 'aa'),  3 ) ;
+
```
zellerh: Just surviving the PREPARE of these is good enough for the test? Would it make sense to confirm the transformation with something like this: select count(*) from table(explain(null, 'S9')) where operator like '%MERGE%';

zellerh:
Code looks good to me. I have a couple of questions, no need to answer them before you commit if you don't see an issue with them.

core/sql/optimizer/BindRelExpr.cpp
sureshsubbiah: In RelExpr::recomputeOuterReferences and most other implementations of recomputeOuterReferences char inputs are pulled up from the child to the parent, but this pullup code is executed only if the parent has inputs to start with. If parent has no inputs to start with, then most recomputeOuterRefences implementations do little. I was seeing that if inputs were added to "mu" and the UPSERT had no params (i.e. when all inputs mentioned in this method are internal ones, those that are required because the "mu" is now on the right side of a tuple_flow, not those that will come in from the cli through the input_desc) they were not getting pulled up, and inlining nodes were not getting the values they needed.
I tried for a long to time to avoid adding any inputs explicitly to both "mu" and "re" and letting inputs be managed by existing code in Transform/Normalize. I could not get that to work as they started failing an assert in optimizer. Now I have adopted the idea of doing the minimum necessary. Add only to "mu" as far as possible, however in the no input_desc char. inputs I did have to add it to "re".

core/sql/optimizer/NormRelExpr.cpp
sureshsubbiah: Will fix my TABs, likely after calling you. I use emacs and am not sure how to set this correctly.

core/sql/regress/executor/EXPECTED015.SB
sureshsubbiah: Will make this change in my next delivery.

sureshsubbiah:
Thank you Hans for the review.

core/sql/optimizer/BindRelExpr.cpp
zellerh: Thanks for the explanation!

core/sql/optimizer/NormRelExpr.cpp
zellerh: You can add this to your ~/.emacs file:

(setq-default indent-tabs-mode nil)

PR #292, opened 2016-01-29
[TRAFODION-1775][TRAFODION-1777] UPDATE STATS usability improvements

DaveBirdsall:
For [TRAFODION-1775], added a check to UPDATE STATISTICS to return a new error (9243) if UPDATE STATISTICS is attempted on a large table (100,000,000 or more rows) without a SAMPLE clause. Added an explicit NO SAMPLE clause to the syntax to override this check. The value 100,000,000 is not hard-coded; it can be changed via CQD USTAT_YOULL_LIKELY_BE_SORRY.

For [TRAFODION-1777], fixed a minor annoyance. UPDATE STATISTICS on a non-existing object would return three error messages that the object doesn't exist. Changed it to return just one.

### DIFF
--- a/core/sql/sqlcomp/DefaultConstants.h
+++ b/core/sql/sqlcomp/DefaultConstants.h
@@ -709,6 +709,7 @@ enum DefaultConstants
                                             // CQDs are delimited by ","
   USTAT_MIN_CHAR_UEC_FOR_IS,     // minimum UEC for char type to use internal sort
   USTAT_MIN_DEC_BIN_UEC_FOR_IS,  // minimum UEC for binary types to use internal sort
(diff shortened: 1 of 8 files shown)


core/sql/sqlcomp/DefaultConstants.h
```diff
@@ -709,6 +709,7 @@ enum DefaultConstants
                                             // CQDs are delimited by ","
   USTAT_MIN_CHAR_UEC_FOR_IS,     // minimum UEC for char type to use internal sort
   USTAT_MIN_DEC_BIN_UEC_FOR_IS,  // minimum UEC for binary types to use internal sort
+  USTAT_YOULL_LIKELY_BE_SORRY, // minimum row count where explicit NO SAMPLE clause is required
```
zellerh: :-)

zellerh:
Looks great to me.

PR #296, opened 2016-01-29
Additional changes for JIRA [TRAFODION-1703].

sbroeder:
Verifying the state of the record before deleting


core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/generated/TrxRegionProtos.java
```diff
@@ -26558,6 +26558,16 @@ public Builder clearHasException() {
...
      */
     long getAuditSeqNum();
+
+    // required bool ageCommitted = 5;
+    /**
+     * <code>required bool ageCommitted = 5;</code>
+     */
+    boolean hasAgeCommitted();
+    /**
+     * <code>required bool ageCommitted = 5;</code>
```
DaveBirdsall: What is the point of the repeated <code>required bool ageCommitted = 5;</code> comments? There seems to be several of them.

DaveBirdsall:
Looks OK to me.

sbroeder:
I'm not exactly sure Dave, but there were conflicts needing resolution when I did the merge.  This is in the generated code area, so I probably shouldn't try to manually resolve conflicts there.  I should instead resolve the conflicts in the source files and rebuild the generated code.  Let me see if I can do that and recreate the pull request.

sbroeder:
I just rebuilt the generated/TrxRegionProtos.java file and after reionserting the copyright block at the head of the file it is exactly the same, so I believe it is correct.  There are various auto generated methos got building the protobufs for the requests and replies and these all have the required 5th parameter which is set in the TrxRegion.proto file.  I think this is normal.

Thanks for raising this Dave.  It's always good to verify.

Sean

PR #297, opened 2016-02-01
fix 1795: findPort.py may find some ports that are in CLOSE_WAIT state

nonstop-qfchen:
Hi,

Could you please do a quick review of the fix to findPort.py?  In the fix, a new method to find the port numbers in use is added. This method utilizes the command "sudo lsof -i" and then do a simple parsing to grab all these port numbers. If the sudo command is unable to run, then the existing method is called. 

Thanks --Qifan

### DIFF
--- a/core/sqf/sql/scripts/findPort.py
+++ b/core/sqf/sql/scripts/findPort.py
@@ -63,23 +63,39 @@ args = parser.parse_args()  # exits and prints help if args are incorrect
 
 exitCode = 0
 
-# get the set of lines to process ( netstat | grep localhost )
 
-p1 = subprocess.Popen(["netstat"], stdout=subprocess.PIPE)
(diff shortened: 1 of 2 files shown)


DaveBirdsall:
Looks good to me.

zellerh:
Ok with me, too. Now that we use only non-ephemeral port ranges, do we really need this?

nonstop-qfchen:
Yes.  I was not able to install Trafodion successfully without going
through non-ephemeral ports that are currently occupied (as revealed by
sudo lsof).

This automates the process.

Thanks

On Mon, Feb 1, 2016 at 12:02 PM, Hans Zeller <notifications@github.com>
wrote:

> Ok with me, too. Now that we use only non-ephemeral port ranges, do we
> really need this?
>
> —
> <https://github.com/apache/incubator-trafodion/pull/297#issuecomment-178099309>
> .
>



-- 
Regards, --Qifan

PR #298, opened 2016-02-01
[TRAFODION-1749] Bundle log4cxx library in Trafodion build

svarnau:
Since log4cxx is not available in all Linux distros, we make it a
build-time dependency only, and bundle the shared library for runtime.
This eliminates the installer dependency to check log4cxx package.

The build-time dependency can be satisfied either by installing log4cxx
and log4cxx-devel RPM packages, if available, or by building library
from source.  The traf_tools_setup.sh script is enhanced to download
source code and build this dependency.

@robertamarton @amandamoran

### DIFF
--- a/core/conn/odbc/src/odbc/nsksrvr/Makefile
+++ b/core/conn/odbc/src/odbc/nsksrvr/Makefile
@@ -121,7 +121,7 @@ DBT_LIBS =
 
 EARLY_LIBS = -ljsig
 
-COMMON_LIBS = -lsbfs -lsbms -levlsq -lwin -ltdm_sqlcli -larkcmp_dll -L../nsksrvrcore -lmxocore -L$(LOG4CXX_LIB_DIR) -llog4cxx
+COMMON_LIBS = -lsbfs -lsbms -levlsq -lwin -ltdm_sqlcli -larkcmp_dll -L../nsksrvrcore -lmxocore -llog4cxx
 
 ifeq ($(SQ_MTYPE),64)
 LOC_JSIG=$(JAVA_HOME)/jre/lib/amd64/server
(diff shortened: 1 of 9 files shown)


DaveBirdsall:
Looks good to me.

robertamarton:
Looks great.

Does the build instructions on the web page need to change to install apr and apr-utils? 

Also, as part of testing release 1.3.0, I tested the build by not using the TOOLSDIR but setting up each tool location as a separate environment variable export. So I created a set of exports for LLVM, ICU, MAVEN, etc.  Should we make sure this works with adding CXX to the picture.

svarnau:
Yes, I will update the wiki page on setting up the build environment, once this change is in. 

Yes, you could definitely have the LOG4CXX location be somewhere other than TOOLSDIR. Did you want me to add the variables to core/sqf/LocalSettingsTemplate.sh to highlight that?

robertamarton:
Changing LocalSettingsTemplate.sh would be good.



    Roberta



*From:* Steve Varnau [mailto:notifications@github.com]
*Sent:* Monday, February 1, 2016 1:47 PM
*Subject:* Re: [incubator-trafodion] [TRAFODION-1749] Bundle log4cxx
library in Trafodion build (#298)



Yes, I will update the wiki page on setting up the build environment, once
this change is in.

Yes, you could definitely have the LOG4CXX location be somewhere other than
TOOLSDIR. Did you want me to add the variables to
core/sqf/LocalSettingsTemplate.sh to highlight that?

—
<https://github.com/apache/incubator-trafodion/pull/298#issuecomment-178209637>
.

DaveBirdsall:
Is this ready to commit?

svarnau:
Yes, I think I've addressed Roberta's suggestion. Should be good to go.

PR #301, opened 2016-02-02
[TRAFODION-1799] create index failed with ERROR[3047]

sureshsubbiah:
Create index was failing when bulk load is used to populate the newly created index
Bulk load uses RangePartitionFunction on source table. We get an error when 
source table has timestamp column in key and rangepartitionfunction is used.
Fix is to avoid casting <min>/<max> used internally to TIMESTAMP type.
Change in ItemExpr.cpp

Change in NormRelExpr.cpp (tabs) and executor/test015 is rework for previous PR.

### DIFF
--- a/core/sql/optimizer/ItemExpr.cpp
+++ b/core/sql/optimizer/ItemExpr.cpp
@@ -10519,8 +10519,11 @@ ConstValue * ConstValue::castToConstValue(NABoolean & negate_it)
 
 Int32 ConstValue::getArity() const { return 0; }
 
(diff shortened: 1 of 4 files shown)


DaveBirdsall:
I'm missing something. In optimizer/ItemExpr.cpp, I see that a variable was renamed. And in optimizer/NormRelExpr.cpp, I see that some code was re-indented. And I see changes to regression tests. But that's it. What code change fixed the bug? What did I miss?

DaveBirdsall:
Never mind. I see it in the ItemExpr.cpp code. I missed the added 'if'. Code change looks good to me.

PR #304, opened 2016-02-02
Fix for JIRA [TRAFODION-1713]

sbroeder:
IDs returned from IDTMSRV process should be periodically re-calibrated
so that they are a better approximation of a logical time

### DIFF
--- a/core/sqf/src/tm/idtmsrv.cpp
+++ b/core/sqf/src/tm/idtmsrv.cpp
@@ -37,9 +37,12 @@
 #include "seabed/ms.h"
 #include "seabed/pctl.h"
 #include "seabed/pevents.h"
-
+#include "seabed/timer.h"
+#include "seabed/thread.h"
 #include "idtmsrv.h"
 
+short          gv_tleid;
+short          gv_time_refresh_delay; // time in tics (10ms)
 char           ga_name[BUFSIZ];
 char          *gp_shm;
 unsigned long *gp_shml;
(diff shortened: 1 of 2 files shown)


core/sqf/src/tm/idtmsrv.cpp
```diff
@@ -53,6 +56,37 @@ void do_reply(BMS_SRE *pp_sre, char *pp_reply, int pv_len, short pv_ec);
...
+void reset_time_counter() {
+
+    struct timespec lv_new_ts;
+
+    clock_gettime(CLOCK_REALTIME, &lv_new_ts);
+    unsigned long lv_new_tsl = ((unsigned long) lv_new_ts.tv_sec << 20) |
+                               ((unsigned long) lv_new_ts.tv_nsec / 1000);
+
+    unsigned long lv_existing_tsl = __sync_add_and_fetch_8(gp_shm, 0);
+    __sync_add_and_fetch_8(gp_shm, lv_new_tsl - lv_existing_tsl);
```
DaveBirdsall: Can time go backwards here? And would that cause a problem?

sbroeder:
Great question.  Let me explain what we currently see, and why the proposed change as part of my explanation.

The server hands out 64 bit IDs based on an AtomicLong it stores internally.  This value is initialized at startup to the current system time and is monotonically increased with each request.  What we see today is that the ID is very strongly correlated to system time at startup, but over time the ID values begin to trail very substantially from the system clock.  This is because the requests don't come in with the frequency of time clicks.

The proposed change is to try to periodically refresh the internal ID based on the current system time so that it stays closer to the system time.  This would have the observable effect of the returned IDs making periodic leaps FORWARD, not backward.

The same would be true of IDs following a system shutdown.  The new system time is ahead of the previous system time or ID and could never go backwards... unless...

Unless the system clocks are reset.  If the system clocks are reset, then yes the IDs could go backwards and this would have the same effect as setting the clocks backwards today.  It would depend on many variables - how far back did the time get set, what was the latest time used in the database, etc.  In practice I don't think this is very likely to affect anyone.

DaveBirdsall:
Code changes look good to me. Thanks for the explanation.

PR #306, opened 2016-02-04
[TRAFODION-1614] Improve time to build external dependencies

svarnau:
This script is for one-time set-up of build environment (aka TOOLSDIR),
but an hour or more is excessive, per the JIRA. So to shrink the time, we kick off all
the builds in background.  All the builds are independent, except
that LLVM depends on udis86.

LLVM is the long pole in build time. Biggest gain is building the
two flavors of LLVM in parallel. All the other tools can be
downloaded and built while we are waiting for LLVM.

Factoring download speed out of the equation (pre-downloaded the source tar files), it took me 26 minutes to build everything. Roughly 50% improvement.

### DIFF
--- a/install/traf_tools_setup.sh
+++ b/install/traf_tools_setup.sh
@@ -93,13 +93,38 @@ function downloadSource
   fi
 }
 
+LOGDIR=$(pwd)
+LOGFILE=$LOGDIR/traf_tools_setup.log
+BLDLOG=$LOGDIR/traf_tools_bld
+
+# -----------------------------------------------------------------------------
+# execute build functions in background
+# $1 - short name of component (single word, also used in logfile name)
+# $2 - first make target (usually empty string for the default target)
+# $3 - second make target
+# $4 - dir/file to check for success
+# -----------------------------------------------------------------------------
+function bkgBuild
+{
+  LOG=${BLDLOG}-${1}.log
+  echo "INFO:   Starting background make and install for $1" | tee -a $LOGFILE
+
+ (make $2 > $LOG 2>&1
+  echo "INFO:   make completed" > $LOG
+  make $3 > $LOG 2>&1
+  if [[ ! -e $4 ]]; then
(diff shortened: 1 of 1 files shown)


zellerh:
Looks good! Thanks for making this faster.

install/traf_tools_setup.sh
```diff
@@ -93,13 +93,38 @@ function downloadSource
...
+ ) &
+}
 
 # main code
 
 TOOLSDIR=
 BASEDIR=
-LOGDIR=`pwd`
-LOGFILE=$LOGDIR/traf_tools_setup.log
 rm $LOGFILE 2>/dev/null
```
robertamarton: where does the LOGFILE variable get set?

robertamarton:
Looks good -

traflm:
llvm, bison is also avialble via yum installation. But it will install them into 'system default' folder, like /usr/lib and /usr/include, so the Makefile need change as well. But if that is doable, I think it will further greatly improve the speed.
It needs some non-trivial effort, so just my 2 cents.

traflm:
llvm, bison is also avialble via yum installation. But it will install them into 'system default' folder, like /usr/lib and /usr/include, so the Makefile need change as well. But if that is doable, I think it will further greatly improve the speed.
It needs some non-trivial effort, so just my 2 cents.

traflm:
llvm, bison is also avialble via yum installation. But it will install them into 'system default' folder, like /usr/lib and /usr/include, so the Makefile need change as well. But if that is doable, I think it will further greatly improve the speed.
It needs some non-trivial effort, so just my 2 cents.

svarnau:
@traflm, I agree that would be nice.
I actually tried to use LLVM from yum install, but the version was newer and I could not get compilation to work, even after several code changes, and I gave up.
As for bison, our requirement is for 3.x, and the one available in the distro is 2.x, so I did not even try to make that work.

PR #307, opened 2016-02-04
[TRAFODION-1733][TRAFODION-1734] License text improvements

svarnau:
These changes are due to review feedback received from release 1.3.

    [TRAFODION-1734] License info improvements from reviews of 1.3 release

    Added info on particular type of BSD & MIT licenses used.
    Removed .txt file suffixes.
    Made copy of license rather than give URL.
    Removed license for protobuf, since we merely use the tool rather than
    include their software.

    [TRAFODION-1733] Remove unneeded info from license file

    The FBstring.h file is Apache license, not a different license, so
    no need to call it out.

    ASCIIdoc is a tool that is used to generate files in build process
    but the software is not distributed with Traodion software, so it
    does not apply.


DaveBirdsall:
Looks good to me.

LICENSE
```diff
@@ -307,11 +286,6 @@ For details see http://jquery.org/license
```
robertamarton: Do you need to change this to point to the license file you copied to licenses directory?

LICENSE
```diff
@@ -307,11 +286,6 @@ For details see http://jquery.org/license
 
 ++++++++++++++++++++++++++++++
 
-Apache Trafodion uses asciidocs to create its documentation which is licensed
-under MIT 
-
-For details see http://asciidoctor.org/
```
robertamarton: Just to confirm - an original comment we received for our first release attempt (that did include docs) was that we needed to add asciidocs to our license file.  When we released 1.3.0, we did not include docs so we then received a comment that this text was not needed.  Are we going to include docs in our source tree?  If so, will we need this text?

LICENSE
svarnau: Agreed. I'll fix that oversight.

LICENSE
svarnau: No, I don't think so. The docs do not include any ASCIIdoc software. It is merely used as a build tool to generate other formats (html, pdf) from asciidoc formatted files.

svarnau:
jenkins, retest -- I don't think sqstart is dependent on the license file!

DaveBirdsall:
Is this ready for commit?

svarnau:
If Roberta is satisfied with my response to her comments, then yes ready to go.

robertamarton:
Go for it.



   roberta



*Sent:* Friday, February 5, 2016 4:21 PM
*Subject:* Re: [incubator-trafodion] [TRAFODION-1733][TRAFODION-1734]
License text improvements (#307)



Is this ready for commit?

—
<https://github.com/apache/incubator-trafodion/pull/307#issuecomment-180632754>
.

PR #308, opened 2016-02-05
TRAFODION-1813  add commons-httpclient depenedency into dcs/pom.xml

traflm:


traflm:
This is strange, I made a brand new building machine , install a CentOS6.5, Then follow the build instruction and it failed at DCS build. I made this change and it can build. But without this change, DCS can be built before, so I am not sure if this is required, just send it out for expert to review.
At least it is good thing to add dependency in the pom.xml.

traflm:
Not an issue , thanks Arvind for help.

PR #309, opened 2016-02-05
TRAFODION-1798- Enable DDL transaction infrastructure by default.

prashanth-vasudev:
This is Step 0 ( of 4 steps ) as outlined in the jira TRAFODION-1798


DaveBirdsall:
jenkins, add user

DaveBirdsall:
Looks good to me. To be clear, this change enables DDL transactions but doesn't turn them on by default, correct? (For the latter, we'd need to change the default for CQD DDL_TRANSACTIONS to 'ON'?)

prashanth-vasudev:
That is correct. 
The thought is for this code to be exercised in daily build and regression tests. 

Parser flag and cqd need to be set to use this feature in a controlled environment. ( see regression test for the parser flag and cqd setting )

Regards,
Prashanth

shangshengtung:
The change looks good.

PR #310, opened 2016-02-05
Fixes to JIRAs 1793,1807,1808. Details below.

anoopsharma00:
1807: timezone specification support in datetime values
1808: query cache bug fix with time/timestamp values with fraction
1793: get stmt with views now include schema names


core/sql/common/DateTimeType.cpp
```diff
@@ -171,6 +172,8 @@ NABoolean DatetimeType::isEncodingNeeded() const
 
 NABoolean DatetimeType::operator==(const NAType& other) const
 {
+  DatetimeType &otherDT = (DatetimeType&)other;
```
DaveBirdsall: Not sure why this variable is defined. We don't use it anywhere. Doesn't hurt anything though.

core/sql/common/DateTimeType.cpp
```diff
@@ -1399,18 +1403,69 @@ NABoolean SQLMPDatetime::isCompatible(const NAType& other, UInt32 * flags) const
...
+  flags |= CONV_NO_HADOOP_DATE_FIX;
+  short rc = 
+    ExpDatetime::convAsciiToDatetime((char*)strValue, strlen(strValue), 
+                                     dstValue, 100,
+                                     startField, endField, 
+                                     ExpDatetime::DATETIME_FORMAT_NONE,
+                                     trueFP,
+                                     NULL, &diags, flags);
+  if (rc)
+    return;
```
DaveBirdsall: fractionPrecision will be uninitialized if we take this return

DaveBirdsall:
Looks good to me, apart from one error path that might be a concern.

DaveBirdsall:
jenkins, add user

core/sql/common/DateTimeType.cpp
anoopsharma00: That return is for error paths. The Datetime value being constructed will be invalid and
not be used. isValid() method will return true.
But we can initialize fractionPrecision to zero on entry to this method.

anoopsharma00: correction: isValid() method will return FALSE on error path.

PR #312, opened 2016-02-09
TRAFODION-1757 Improvements to DCS infrastructure:

arvind-narain:
1. handle partially up and operational Trafodion environment.

2. If zookeeper interface gives an error during dcsstop, then explicitly kill the DCS master and servers.


hegdean:
Changes look good to me.

DaveBirdsall:
Looks good to me too.

PR #313, opened 2016-02-09
[TRAFODION-1819] Disallow UPDATE STATS in a user transaction

DaveBirdsall:
UPDATE STATS sometimes does DDL and DML under the covers, and likes to start its own transactions for this purpose. Trafodion doesn't support mixed DDL and DML in the same transaction. Therefore, it makes sense for UPDATE STATS to check for a user transaction up front, and raise an error if there is one running.


robertamarton:
Looks good to me

PR #314, opened 2016-02-09
[TRAFODION-1818] Parallel build with maven caused error

svarnau:
Add dependencies in sqf/Makefile to prevent collision of maven
builds downloading dependencies.

Replace spaces in hbasetmlib2/Makefile with tabs. Makefile syntax
requires initial tab for shell commands.


zellerh:
Thanks for fixing that! Looks good.

DaveBirdsall:
Looks good to me.

PR #317, opened 2016-02-10
TRAFODION-1798 This change is item 2( of 4 items ) listed in jira

prashanth-vasudev:
Fixes the following scenario which resulted in a hang.
Begin work;
insert into table t  values( 1);
drop table t;
commit work;

In this scenario, as part of commit, the drop table t gets hadmin:disabled in prepare phase followed by hadmin delete in commit phase. However, because there was a insert to table, commit phase encounters the disabled table. Fix is to detect this exception in the commit thread and validate against the drop table list maintained in TmDDL log.


sbroeder:
Looks good to me.

shangshengtung:
The change looks good.

sandhyasun:
Looks good. Merging this.

selvaganesang:
It is very strange that state is maintained as a string. Consider changing it to a constant to reduce path length and unnecessary object creation sometime in future

prashanth-vasudev:
Thanks Selva. Will include this change in next set of changes.

PR #321, opened 2016-02-12
Update fix for TRAFODION-1814

venkat1m:
The fix earlier did not fully work when pstack was requested for a specific program.
The sqpstack output had a line with "--" string between the lines with "pstack-ing" and the "ssh -q <host> pstack <pid>" lines and this was not handled in the logic.
An example sqpstack output:
pstack-ing 24162
–
ssh -q node2 pstack 24162

The new fix ignores adding a new line if the string is just "--"


DaveBirdsall:
Looks good to me. Just curious: Why were there two copyrights in the old versions of the files?

arvind-narain:
Looks good Venkat.

venkat1m:
Seems to be an oversight in the task that fixed the copyright headers to proper Apache headers.

PR #322, opened 2016-02-13
[TRAFODION-1771] Enabled TESTRTS again

selvaganesang:
TESTRTS was failing at times with error 8926. This is due to
absence of ROOT_OPER_STATS in statistics area, but the stats descriptor
was incorrectly indicating that there is a ROOT_OPER_STATS.

In future, I will get rid of setting the accumulated statistics as part of
cleanup effort of the statistics collection.


DaveBirdsall:
Looks good to me.

PR #325, opened 2016-02-17
TRAFODION-1832: Add privilege regression test 132

robertamarton:
Activated privs1 regression test 132. This tests privileges on various 
utilities.


DaveBirdsall:
Looks good to me.

PR #327, opened 2016-02-17
[TRAFODION-1467][TRAFODION-1472] Histograms for key prefixes in salte…

DaveBirdsall:
…d tables

This patch changes the default for the CQD USTAT_ADD_SALTED_KEY_PREFIXES_FOR_MC to 'ON'. So, multi-column histograms for prefixes on salted tables will now be generated by default.


zellerh:
Looks very good, very nice that the fix is that simple.

PR #330, opened 2016-02-19
[TRAFODION-1834] fix issue in ESP colocation

traflm:


DaveBirdsall:
Looks good. One question though: ExHDPHash::hash is deterministic. If beginPos is always the same value, then we've replaced a bias toward one node with a bias toward some other. So, it might be wise to check on the behavior of beginPos.

traflm:
thanks Dave, the beginPos will change for each call of this method. It is the region index of the first region of a group of regions.
 If a table has 28 regions, and Trafodion decides to start 4 ESPs for this query, each ESP then needs to handle 7 regions. So the first ESP will be responsible for region 0 - 6, second 7- 13, etc... So the beginPos will be 0, 7, 14, 21 . I am not sure if hash will generate same value for this sequence?
Even so, in fact, which regionserver is most popular for a given range, for example, 0 -7, it is already  random. This randomization is an attempt, if it doesn't work as we wish, it doesn't harm, and we can continue to design some new algorithm,

zellerh:
Looks good to me. Hopefully the pseudo-random result will make diagnosing and debugging problems easier.

PR #331, opened 2016-02-19
JIRA TRAFODION-1825. Fix for core caused by input datetime values.

anoopsharma00:
When a datetime value is input as a param or used in a literal, 
it needs to exactly match the corresponding datatype.
For ex: '2010-01-01' is the correct format for DATE datatype.
But it is not the right format for TIMESTAMP datatype.
This is done for values that are input from sqlci, jdbc/odbc params, trafci
and literals.

On the other hand, if a value is converted using the cast function, then
the source value is extended with trailing zeroes.
For ex:  cast ('2010-01-01' as timestamp) will result in '2010-01-01 00:00:00.000'.

These 2 cases are now handled correctly. It was causing a memory overflow
which resulted in a core.

This checkin also fixes an error display bug that was not displaying the correct
datatype.


DaveBirdsall:
Looks good to me.

PR #332, opened 2016-02-19
[TRAFODION-1840] Disable bundling of Qt libs into packaging

svarnau:
Also clean up symlink to obsolete directory.

### DIFF
--- a/core/sqf/LocalSettingsTemplate.sh
+++ b/core/sqf/LocalSettingsTemplate.sh
@@ -67,6 +67,8 @@ LOG4CXX_INC_DIR="$TOOLSDIR/apache-log4cxx-0.10.0/include"
 
 # Explicitly set QT_TOOLKIT here if Qt is installed and you want to build the SqlCompilerDebugger
 # QT_TOOLKIT="$TOOLSDIR/Qt-4.8.5-64"
+# SqlCompilerDebugger ibrary will be dynamically linked with Qt libs.
+# Note that Qt uses LGPL license. It is not to be re-distributed under Apache license.
 
 # HBASE*, HIVE*, HADOOP* locations may be overridden here
 # uncomment the following lines to describe where your Hadoop locations exist.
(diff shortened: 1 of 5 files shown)


core/sql/SqlCompilerDebugger/mk.sh
```diff
@@ -21,6 +21,9 @@
 export PATH=$QT_TOOLKIT/bin:$PATH
 export LD_LIBRARY_PATH=$QT_TOOLKIT/lib:$LD_LIBRARY_PATH
 
+# Important: QT_TOOLKIT must not be bundled in Trafodion build
+#     QT License is not compatible with ASF license policy
+
 
 qmake -o Makefile SqlCmpDbg.pro
 make
```
robertamarton: I guess, if someone wants to use the debugger, then they will have to figure out how to download the toolkit and get it working.  This is probably okay.  If it become important, then someone can create a JIRA and we can work on a change that will make this process easier.

core/sqf/LocalSettingsTemplate.sh
```diff
@@ -67,6 +67,8 @@ LOG4CXX_INC_DIR="$TOOLSDIR/apache-log4cxx-0.10.0/include"
 
 # Explicitly set QT_TOOLKIT here if Qt is installed and you want to build the SqlCompilerDebugger
 # QT_TOOLKIT="$TOOLSDIR/Qt-4.8.5-64"
+# SqlCompilerDebugger ibrary will be dynamically linked with Qt libs.
+# Note that Qt uses LGPL license. It is not to be re-distributed under Apache license.
```
robertamarton: spelling error "ibrary"

robertamarton:
Changes look fine to me.

DaveBirdsall:
Looks good to me. Echoing Roberta though, I wonder if it would be good to include directions on how to get the library and use it, say, on our wiki page?

zellerh:
Looks good to me, too. Steve explained to me that someone who wants to use this GUI would need to have the Qt libraries in their LD_LIBRARY_PATH, which seems like the best solution to me. Agree, a wiki page might help to explain which RPMs (or other things) to install and possibly what additions to make to LD_LIBRARY_PATH. Hopefully someone with a new machine that doesn't have Qt installed yet can try it and provide that info.

PR #333, opened 2016-02-20
[TRAFODION-1838] Send CQDs down to child process on CREATE TABLE LIKE

DaveBirdsall:
Fix two bugs:

1. UPDATE STATISTICS on a table with a nullable key column failed, because the CQD ALLOW_NULLABLE_UNIQUE_KEY_CONSTRAINT was not propagated to a child compiler process by the CREATE TABLE LIKE logic. This has been fixed.

2. A buffer overrun can occur when using USTAT_CQDS_ALLOWED_FOR_SPAWNED_COMPILERS with a single CQD. This bug was encountered while debugging the first bug above. This has been fixed.

### DIFF
--- a/core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
+++ b/core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
@@ -356,6 +356,9 @@ void CmpSeabaseDDL::createSeabaseTableLike(
         query += keyClause;
       }
 
+  // send any user CQDs down 
+  Lng32 retCode = sendAllControls(FALSE, FALSE, TRUE);
+
   ExeCliInterface cliInterface(STMTHEAP, NULL, NULL, 
   CmpCommon::context()->sqlSession()->getParentQid());
 
--- a/core/sql/ustat/hs_update.cpp
+++ b/core/sql/ustat/hs_update.cpp
@@ -355,7 +355,6 @@ Lng32 UpdateStats(char *input, NABoolean requestedByCompiler)
            LM->Log(LM->msg);
         }
 
-        char *buf =  new (CmpCommon::statementHeap()) char[allowedCqdsSize];
         char* filterString = new (STMTHEAP) char[allowedCqdsSize+1];
         // We need to make a copy of the CQD value here since strtok
         // overwrites delims with nulls in stored cqd value.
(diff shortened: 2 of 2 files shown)


core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
```diff
@@ -356,6 +356,9 @@ void CmpSeabaseDDL::createSeabaseTableLike(
         query += keyClause;
       }
 
+  // send any user CQDs down 
+  Lng32 retCode = sendAllControls(FALSE, FALSE, TRUE);
+
```
zellerh: Just wondering whether it's safe to send all CQDs to the child compiler. As this missing CQD is causing trouble, maybe some other user CQDs that now get sent could also cause the statement to fail? Also, if I have a table with a nullable unique column, should I have to set this CQD to create another table like it, or should the CREATE TABLE LIKE work without the CQD?

DaveBirdsall: I think not. And here is why: If I do the CREATE TABLE LIKE from sqlci, the CREATE TABLE LIKE logic executes in the master and the under-the-covers vanilla CREATE TABLE executes in a tdm_arkcmp child. In sqlci, any CQD done in the master is also done in the child. So, in the sqlci case, the CQDs will be identical. As far as I know this has no ill effect. CQDs have an anomaly though: They are not propagated more than one level down. UPDATE STATS runs in a tdm_arkcmp child of sqlci (or MXOSRVR). The CREATE TABLE LIKE that UPDATE STATS executes runs in a tdm_arkcmp child of that process (that is, a grandchild of the sqlci). This seems to be intentional: UPDATE STATS wants to tightly control what CQDs execute in that process. The under-the-covers vanilla CREATE TABLE runs in a tdm_arkcmp child of the one running CREATE TABLE LIKE (that is, a great-grandchild of sqlci). UPDATE STATS itself executes a controlled set of CQDS; these reach the grandchild tdm_arkcmp (that is, the one running CREATE TABLE LIKE) but not the great-grandchild. But it is an accident of implementation that the vanilla CREATE TABLE runs one more process down. So, sending the CQDs down seems harmless (as it works that way from sqlci now) and appropriate (as UPDATE STATS has already controlled the CQDs run by the grandchild). Apologies for all the words; this would be much clearer if drawn in a picture.

DaveBirdsall: And to answer your second question: I am not sure. The current behavior of the code is that the CQD is required if doing CREATE TABLE LIKE from sqlci. I think one could make a case for changing that behavior. On the other hand, one might want to prevent casual propagation of nullable primary keys. Which would be an argument for always requiring the CQD. I don't feel strongly either way.

zellerh:
Thanks for the analysis, I lost track of how CQD propagation works. Hope I understand this correctly, that user CQDs are not going to contaminate the context that is used for metadata operations (like reading from the "_MD_".OBJECTS table). About my section question, I feel the same way, am not sure whether we should require the CQD or not, I also don't have a strong opinion.

PR #335, opened 2016-02-22
[TRAFODION-1765] trafci would throw an error when set statistics on

arvind-narain:
TRAFODION-1765 - trafci would throw an error when set statistics on 
TRAFODION-1757 - 1) DCS stop script still not handling zookeeper error properly
                              2) No indication in DCS logs for "No Available Servers" unless
                                  debug level is enabled.

### DIFF
--- a/dcs/src/main/java/org/trafodion/dcs/master/listener/ConnectReply.java
+++ b/dcs/src/main/java/org/trafodion/dcs/master/listener/ConnectReply.java
@@ -279,7 +279,6 @@ class ConnectReply {
                 LOG.error(clientSocketAddress + ": " + "No Available Servers");
             else
                 LOG.error(clientSocketAddress + ": " + "No Available Servers - exception thrown");
-            }
         } else {
             
             if (cc.datasource.length() == 0)
--- a/dcs/src/main/java/org/trafodion/dcs/master/listener/ConnectReply.java
+++ b/dcs/src/main/java/org/trafodion/dcs/master/listener/ConnectReply.java
@@ -1,25 +1,25 @@
-/**
-* @@@ START COPYRIGHT @@@
+//
+// @@@ START COPYRIGHT @@@
+//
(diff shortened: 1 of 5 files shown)


hegdean:
Changes looks good to me.

dcs/src/main/java/org/trafodion/dcs/master/listener/ConnectReply.java
```diff
@@ -275,12 +275,10 @@ boolean buildConnectReply  (Header hdr, ConnectionContext cc, SocketAddress clie
         if (found == false || exceptionThrown == true){
             exception.exception_nr = ListenerConstants.DcsMasterNoSrvrHdl_exn; //no available servers
             replyException = true;
-            if(LOG.isDebugEnabled()){
-                if (found == false)
-                    LOG.info(clientSocketAddress + ": " + "No Available Servers");
-                else
-                    LOG.info(clientSocketAddress + ": " + "No Available Servers - exception thrown");
-            }
+            if (found == false)
```
DaveBirdsall: It looks like both (found == false) and (exceptionThrown == true) could be true. If they are both true, you will log "No Available Servers", but won't record that an exception is thrown. Is this what you want?

DaveBirdsall:
Looks good to me.

dcs/src/main/java/org/trafodion/dcs/master/listener/ConnectReply.java
arvind-narain: Thanks Dave. Existing logic. Can I address this in the next checkin ?

DaveBirdsall: Sure; I was just asking the question not suggesting that it was right or wrong.

arvind-narain:
Yes Selva - I did confirm. - thanks for looking into this.

PR #339, opened 2016-02-24
[TRAFODION-1514] Stop using custom suffix (Nv44) of ICU libs

svarnau:
Merged work from RuoYu's @ryzuo pull request, and changed name of ICU
directory, to allow different paths for old/new versions of code.

**Trafodion Developers**: With this change, you may need to update your build environment. If you do not have $TOOLSDIR/icu4c_4.4, then run install/traf_tools_setup.sh to add it.  The old path ($TOOLSDIR/icu4.4) will be needed for older versions of code prior to this change.


DaveBirdsall:
Looks good to me. Hopefully @ryzuo will also comment?

arvind-narain:
jenkins, retest

Looks good to me Steve - thanks.

We'll have to update the following once this change is in to not specify the suffix - 
https://cwiki.apache.org/confluence/display/TRAFODION/How-To#How-To-ICU

svarnau:
@arvind-narain, good point. Now that those instructions have been scripted in traf_tools_setup,sh, I'm thinking we should update the wiki to refer to that script as the definitive build instruction for the TOOLSDIR depedencies.

svarnau:
jenkins, retest -- strange that the hbase start-up on test slaves seems to be flakey

ryzuo:
Oh... I nearly forgot this, the change looks good to me, thanks a lot for doing this, Steve. Agree with Arvind, we definitely need to update the wiki instruction.

PR #340, opened 2016-02-25
[TRAFODION-1700]  [TRAFODION-1847] fixes Upsert with omitted default value columns leave the …

selvaganesang:
[TRAFODION-1700] Upsert with omitted default value columns leave thealigned format table in corrupted state.
[TRAFODION-1847] Upsert with omitted timestamp columns having current_timestamp as default in a non-aligned format
table returns wrong value for this column

For TRAFODION-1847, the upsert is transformed into merge.
For TRAFODION-1700, the upsert is transformed into merge when the CQD
TRAF_UPSERT_WITH_INSERT_DEFAULT_SEMANTICS is set to OFF. By default this CQD is
set to OFF. When this CQD is ON, the upsert will add a new row with the omitted
columns filled with default values always.

### DIFF
--- a/core/sql/comexe/ComTdbHbaseAccess.cpp
+++ b/core/sql/comexe/ComTdbHbaseAccess.cpp
@@ -282,6 +282,7 @@ ComTdbHbaseAccess::ComTdbHbaseAccess(
   listOfUpDeldColNames_(NULL),
   listOfMergedColNames_(NULL),
   listOfIndexesAndTable_(NULL),
+  listOfOmittedColNames_(NULL),
 
   keyInfo_(NULL),
   keyColName_(NULL),
@@ -438,6 +439,7 @@ Long ComTdbHbaseAccess::pack(void * space)
   listOfFetchedColNames_.pack(space);
   listOfUpDeldColNames_.pack(space);
   listOfMergedColNames_.pack(space);
+  listOfOmittedColNames_.pack(space);
   listOfIndexesAndTable_.pack(space);
   keyInfo_.pack(space);
   keyColName_.pack(space);
@@ -508,6 +510,7 @@ Lng32 ComTdbHbaseAccess::unpack(void * base, void * reallocator)
   if(listOfFetchedColNames_.unpack(base, reallocator)) return -1;
   if(listOfUpDeldColNames_.unpack(base, reallocator)) return -1;
   if(listOfMergedColNames_.unpack(base, reallocator)) return -1;
+  if(listOfOmittedColNames_.unpack(base, reallocator)) return -1;
   if(listOfIndexesAndTable_.unpack(base, reallocator)) return -1;
   if(keyInfo_.unpack(base, reallocator)) return -1;
   if(keyColName_.unpack(base)) return -1;
--- a/core/sql/comexe/ComTdbHbaseAccess.h
+++ b/core/sql/comexe/ComTdbHbaseAccess.h
@@ -610,8 +610,9 @@ public:
   Queue* listOfDeletedColNames() { return listOfUpDeldColNames_; }
   Queue* listOfMergedColNames() { return listOfMergedColNames_; }
   Queue* listOfIndexesAndTable() { return listOfIndexesAndTable_; }
-
+  Queue* listOfOmittedColNames() { return listOfOmittedColNames_; }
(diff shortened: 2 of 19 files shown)


DaveBirdsall:
Code changes look OK to me. A question: When would we set TRAF_UPSERT_WITH_INSERT_DEFAULT_SEMANTICS to 'ON'? Is this something that would be done under the covers by some utility, say, LOAD?

selvaganesang:
Non-aligned format tables

For a table having a column with current_timestamp, we had issues when this column was omitted in a upsert statement. It wouldn't put a cell with current timestamp into hbase table.Then, a wrong value for the timestamp is displayed when the row is selected. In addition, the raw hbase table would have cells for the omitted default value columns at times. It wasn’t consistent when the cell with default value is inserted into raw hbase table.

When there is a omitted COM_CURRENT_DEFAULT column, the upsert is changed to merge always.  All other times it would be just upsert/insert.  All other default columns when it is not user specified is not written to hbase table. It is resolved at the time of select using the default values.

Aligned format:
There was corruption in the aligned format table row because the omitted columns with default values were not populated correctly.

Fix:
By default, upsert is always is changed to merge.  If the row doesn’t exist,  a new row is inserted with omitted  default values filled in.  If the row exists, the omitted columns are filled with old values.

With the cqd TRAF_UPSERT_WITH_INSERT_DEFAULT_SEMANTICS set to ON the omitted columns will always be filled with default values when upserted.

Yes, It is possible that I can use the above CQD to avoid merge in case of non-aligned format. But to ensure that we get the default value when this CQD is set to ON, we need store the default values when omitted always. We wouldn’t be able to take advantage of the feature of skipping cells when omitted and still get the default value upon select without actually storing the data in hbase.  Hence the above CQD applies to aligned format alone.

selvaganesang:
Merge is an expensive operation than just upsert/insert. If the user is ok to just to replace the default value for omitted columns or is sure that there wouldn't any matching row in the table, he/she can set this CQD to get better throughput numbers.

selvaganesang:
Sorry. I closed this PR inadvertently

sureshsubbiah:
Change looks great. Good work Selva.

selvaganesang:
I was able to run full regressions applying the patch of PR #345 .  All tests passed except charsets/TEST012.  I am looking into this issue.

selvaganesang:
This PR is ready to be merged. The recent commit have no effect on the check-PR job.

DaveBirdsall:
It can't be merged yet; github suggests there are conflicts.

selvaganesang:
I am attempting to merge resolving the conflicts. I hope it is ok to do that.

zellerh:
Didn't look at the code in all detail, but at a high level it looks good to me. For me, the question is what the SQL semantics of an upsert are supposed to be. I think it's ok to define it in the way that TRAF_UPSERT_WITH_INSERT_DEFAULT_SEMANTICS 'ON' does. So, I would argue that we could leave this CQD set to ON as the built-in default. The current solution, OFF, is ok with me as well.

My reason to argue for the ON semantics is this: If the user wanted to merge new and existing rows, they should use the MERGE statement, not UPSERT. UPSERT, IMHO, should insert or replace entire rows, not merge old and new rows. That's cleaner semantics.

selvaganesang:
I am also ok with either of them. Let's get some more votes from the community and then I will change it to 'ON' by default.  It also have the added advantage of  performing better than merge.

svarnau:
The risk with doing that is that the merge changes don’t go through testing
before hitting the branch. So there is possibility of breaking the branch.
So, how “ok” it is depends on how big/risky the conflicts are.



--Steve



*From:* selvaganesang [mailto:notifications@github.com]
*Sent:* Tuesday, March 1, 2016 9:57 AM
*Subject:* Re: [incubator-trafodion] [TRAFODION-1700] [TRAFODION-1847]
fixes Upsert with omitted default value columns leave the … (#340)



I am attempting to merge resolving the conflicts. I hope it is ok to do
that.

—
<https://github.com/apache/incubator-trafodion/pull/340#issuecomment-190832983>
.

selvaganesang:
Oh. I didn't think from that angle at all. Fortunately, this conflict was in DefaultConstants.h because this change included a new CQD. It is a minor conflict in the sense both changes should be included.

PR #343, opened 2016-02-26
[TRAFODION-1522] Consolidate HBase build dependencies

svarnau:
Centralized a few variables in sqenvcom.sh for version number of dependencies
(Hive, Thrift, HBase). These are items pulled in as shared libraries and also
as jars (via maven). We want to make sure these are consistent.

Each flavor of hbase-trx depends on distro-specific HBase for compatibility.
Other components (SQL, TM, hbase_utilities) have general dependency, so
we choose one hbase and hbase-trx to build against. We specify this in single
location (sqenvcom.sh) for consistency.

Also removed two pom.xml files no longer used.

Minor changes to setupdir target to clarify what is happening.

Tested both methods of building: with and without install_local_hadoop prior
to build.


DaveBirdsall:
Looks good to me, but I'm not an expert in this area.

PR #344, opened 2016-02-26
JIRA TRAFODION-1844:  support for 'alter table alter column datatype'

anoopsharma00:
Regression test seabase/TEST027 has been enhanced with alter col tests.


core/sql/parser/StmtDDLAlterTableAlterColumn.cpp
```diff
@@ -55,16 +55,22 @@ StmtDDLAlterTableAlterColumn::getText() const
...
-StmtDDLAlterTableAlterColumnDatatype::StmtDDLAlterTableAlterColumnDatatype( 
-     const NAString &columnName 
-     , NAType * natype
-     , CollHeap *heap)
+StmtDDLAlterTableAlterColumnDatatype::StmtDDLAlterTableAlterColumnDatatype(
+     ElemDDLNode * pColumnToAlter
+     ,CollHeap    * heap)
      : StmtDDLAlterTableAlterColumn(DDL_ALTER_TABLE_ALTER_COLUMN_DATATYPE,
-                                    columnName,
+                                    NAString(""),
```
DaveBirdsall: Do we need to keep this parameter? (Since the information about the changed column is in pColumnToAlter)

core/sql/regress/seabase/EXPECTED027
```diff
@@ -756,8 +756,381 @@ create index t027t01i2 on t027t01("cf2".b);
...
+*** ERROR[8839] Transaction was aborted.
+
+*** ERROR[1404] Column C cannot be altered. Reason: Old data could not be updated using the altered column definition.
+
+--- SQL operation failed with errors.
+>>alter table t027t7 drop column e;
+
+*** ERROR[1009] Column E does not exist in the specified table.
+
+--- SQL operation failed with errors.
```
DaveBirdsall: Is it OK to change the character set of a column?

core/sql/sqlcomp/CmpDescribe.cpp
```diff
@@ -200,11 +200,13 @@ short CmpDescribeSeabaseTable (
...
 
-                             // used to add or remove column definition from col list.
-                             // valid for 'createLike' mode. Used for 'alter add/drop col'.
+                             // used to add,rem,alter column definition from col list.
+                             // valid for 'createLike' mode. 
+                             // Used for 'alter add/drop/alter col'.
                              char * colName = NULL,
-                             NABoolean isAdd = FALSE,
-                             const NAColumn * nacol = NULL);
+                             short ada = 0, // 0,add. 1,drop. 2,alter
```
DaveBirdsall: An enum would be nice.

DaveBirdsall:
Looks good to me.

core/sql/regress/seabase/EXPECTED027
anoopsharma00: Yes, it is ok. During alter, values are update from old to the new column definition. If a character set has changed, then an internally generated TRANSLATE function is used to convert src to tgt. This is the same behavior as in any insert...select or update stmt. If an error occurs during this conversion, then the alter stmt fails.

core/sql/parser/StmtDDLAlterTableAlterColumn.cpp
anoopsharma00: yes, you are right, it is not needed for this particular case. But we would need it for other column based operations like renaming a column (there is another jira for rename). So we can keep it around and use it for rename implementation.

core/sql/sqlcomp/CmpDescribe.cpp
anoopsharma00: will add it for a subsequent checkin.

PR #345, opened 2016-02-28
Sean traf

sbroeder:
This addresses the issues of TRAFODION-1857 by eliminating the creation/close of connections associated with individual methods.


DaveBirdsall:
Looks good to me. Is this the old Java problem of the destructor not being called on objects until they are garbage-collected? Which might be a long time?

arvind-narain:
Thanks Sean for fixing this issue. 

One question - the new connection (line 713) - recoveryTableConnection - is needed for a remote TLOG or could the existing connection be used here also ?

sbroeder:
Hi Arvind.  I think it might be able to use the existing connection, but I didn't have a chance to test it.  That code should be the exception, though so we should not see too many connections from there.  I can test this and provide another patch if it is successful.

sbroeder:
Dave, I don't think it's the destructor because I did explicitly call connection.close, but this seems to just put them in CLOSE_WAIT, instead of actually closing them.


arvind-narain:
Thanks Sean. Maybe the unknownTableConnection was leaking earlier, or deleteStaleConnection call didn't close these connections, since all connections in my environment were established.

PR #347, opened 2016-02-29
[TRAFODION-1858] RI predicate generating string should not contain _S…

traflm:
…ALT_ column

### DIFF
--- a/core/sql/optimizer/BindRI.cpp
+++ b/core/sql/optimizer/BindRI.cpp
@@ -408,6 +408,39 @@ void RefConstraint::getMatchOptionPredicateText(NAString &text,
   text += ")";
 }
 
+//helper function to check if the given column name is reserved hidden coloum
+static NABoolean isHiddenColumn(const char *colname)
+{
+  int len = strlen(colname);
+  if(strcmp(colname , "_SALT_") ==0) 
+    return TRUE;
+  //check for DIVISION column
+  //pattern _DIVISION_%d_
+  //must longer than 12
+  if(len >= 12) {
+    //must end with _
+    if(colname[len] == '_')
+    {
+      //if begin with _DIVISION_?
+      if(strncmp(colname,"_DIVISION_",10) == 0) 
+      {
+        //middle part are number
+        int allDigit = 1;
+        for(int i = 0; i< len-11; i++)
+        {
+          if(isdigit(colname[i+10]) == 0)
+          {
+            allDigit = 0;
+            break; //not digit
+          }
+        } 
+        if(allDigit == 1)
+          return TRUE; 
+      }
+    }
+  }
+  return FALSE;
+}
 
 // Writes a row-value-constructor consisting of fully qualified column names
 // in Ansi (external) format
@@ -417,12 +450,18 @@ void RefConstraint::getPredicateText(NAString &text,
 				     NAString *corrName) const
 {
   NAString tblText = ( (corrName == NULL) ? tblName.getQualifiedNameAsAnsiString() : *corrName);
+  int pos= 0;
   text += "(";
   for (CollIndex i = 0; i < keyColumns.entries(); i++)
     {
-      if (i)
+      if(isHiddenColumn(keyColumns[i]->getColName()) )
+        continue;
+      if (pos > 0)
+      {
         text += ",";
+      }
       text += tblText + "." + ToAnsiIdentifier(keyColumns[i]->getColName());
+      pos++; //move the pos
     }
   text += ")";
 }
--- a/core/sql/optimizer/BindRI.cpp
+++ b/core/sql/optimizer/BindRI.cpp
@@ -409,6 +409,10 @@ void RefConstraint::getMatchOptionPredicateText(NAString &text,
 }
 
 //helper function to check if the given column name is reserved hidden coloum
+//NOTE:
+//  this function hardcode the special name string for SALT, DIVSION columns
+//  if the naming convension of SALT/DIVISION column is changed,
+//  this function MUST be changed as well
 static NABoolean isHiddenColumn(const char *colname)
 {
   int len = strlen(colname);
@@ -419,7 +423,7 @@ static NABoolean isHiddenColumn(const char *colname)
   //must longer than 12
   if(len >= 12) {
     //must end with _
-    if(colname[len] == '_')
+    if(colname[len-1] == '_')
     {
       //if begin with _DIVISION_?
       if(strncmp(colname,"_DIVISION_",10) == 0) 
--- a/core/sql/regress/compGeneral/EXPECTED013.SB
+++ b/core/sql/regress/compGeneral/EXPECTED013.SB
@@ -367,7 +367,55 @@ SCAN
 
 --- 2 row(s) inserted.
 >>
+>>create table fksalta ( id int not null, PRIMARY KEY (id))SALT USING 4 PARTITIONS;
+
+--- SQL operation complete.
 >>
+>>CREATE TABLE fksaltb ( id int not null, val int,
++>  PRIMARY KEY (id),
++>  CONSTRAINT FKsalt FOREIGN KEY (val) REFERENCES fksalta (id))
++>  SALT USING 4 PARTITIONS;
+
+--- SQL operation complete.
+>>
+>>INSERT INTO fksalta VALUES ( 1 );
+
+--- 1 row(s) inserted.
+>>
+>>INSERT INTO fksaltb VALUES ( 1, 1);
+
+--- 1 row(s) inserted.
+>>
+>>INSERT INTO fksaltb VALUES ( 2, 2);
+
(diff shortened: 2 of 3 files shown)


core/sql/optimizer/BindRI.cpp
```diff
@@ -408,6 +408,39 @@ void RefConstraint::getMatchOptionPredicateText(NAString &text,
...
+{
+  int len = strlen(colname);
+  if(strcmp(colname , "_SALT_") ==0) 
+    return TRUE;
+  //check for DIVISION column
+  //pattern _DIVISION_%d_
+  //must longer than 12
+  if(len >= 12) {
+    //must end with _
+    if(colname[len] == '_')
```
DaveBirdsall: colname[len] will always be '\0' since len is the strlen of colname. I think you meant, colname[len-1] here.

DaveBirdsall:
Looks like the code will work for "_SALT_", but there is a bug there that I think will prevent it from working for "_DIVISION_nnn_".

core/sql/optimizer/BindRI.cpp
```diff
@@ -417,12 +450,18 @@ void RefConstraint::getPredicateText(NAString &text,
 				     NAString *corrName) const
 {
   NAString tblText = ( (corrName == NULL) ? tblName.getQualifiedNameAsAnsiString() : *corrName);
+  int pos= 0;
   text += "(";
   for (CollIndex i = 0; i < keyColumns.entries(); i++)
     {
-      if (i)
+      if(isHiddenColumn(keyColumns[i]->getColName()) )
```
selvaganesang: Wonder if we can use isSystemColumn() instead. SystemColumn includes SYSKEY column also. Also there is a method isSysKeyColumn(), just in case you want to allow SYSKEY COLUMN.

core/sql/optimizer/BindRI.cpp
sureshsubbiah: Second Dave's comment. I suppose a column with name DIVISION_1_ will still show the problem? Maybe adding a regression test also a good idea though certainly not a requirement.

sureshsubbiah:
Change looks good. Thank you.

core/sql/optimizer/BindRI.cpp
zellerh: Yes, I agree. There are more methods in NAColumn, These names can change and users can choose their own name for the division columns if they like (although I would not recommend that). Why are we excluding salt and division here? It would be best to choose the right one of the many methods in NAColumn:

  isUserColumn()
  isSystemColumn()
  isSyskeyColumn()
  isIdentityColumn()
  isIdentityColumnByDefault()
  isIdentityColumnAlways()
  isComputedColumn()
  isComputedColumnAlways()

traflm:
thanks all, I will try to use NAColumn() functions instead of creating a new buggy function...
And add a regression test.
To answer Hans question, when creating two tables, A and B, while B create a Foreign Key constraint on A.a, we found out if table A is created with SALT, the functionality is broken. So we have to create A without partition as workaround.
Debug for a while, here  is what I understand: To find out if RI is meet, the 'insert into b' statement will generate a query to search in table A, but this new query has a problem to add all key columns, if there are some Trafodion auto-generated colunns like '_SALT_', since at the left side of the predicate, it only add the FK column in table B, but at right side, it add referenced column with those system columns, but the predicate cannot calculate the SALT at this point. It can caculate SALT at this point, but it is much difficult, so simply remove those column is the simplest way to solve this.  
It is hard to explain, since I am not fully understand all the logic here, so yes, I should test more.

zellerh:
Thanks very much for the explanation. My guess is that we use the clustering key on the right side. What we should be using is the primary key - just those columns mentioned in the PRIMARY KEY clause. That would be the ideal fix, but using the NAColumn methods and exclusing salt and division columns should work as well.

Thanks for fixing this!

traflm:
It is not easy to use NAColumn's isXXX() method. The caller does not set attributes like is-salt-column properly. For the SALT column, isSaltColumn() return false ... ...
Where to create all the NAColumns is rather complex, I need much more time to investigate it.

DaveBirdsall:
Perhaps you could fix the bug with the "\_DIVISION\_" columns in BindRI.cpp, and then we can commit this request?

selvaganesang:
What do  isSystemColumn() and  isSysKeyColumn() methods return? It looks like isSaltColumn_ is not set other than when it comes via constructor.

zellerh:
isSystemColumn() should return whether the column is visible when doing a select *. isSysKeyColumn() should return whether this is the SYSKEY system column. In some cases these methods (especially isSaltColumn() and isDivisioningColumn()) may not return TRUE if we are looking at an alternate index, because such indexes just contain the primary key columns. These columns in the index do not always indicate actual salting or divisioning of the index.

Example:

    create unique index ix1 on t(a);          -- may contain a _SALT_ column, index is not salted
    create index ix2 on t(a) salt like table; -- also contains a _SALT_ column, this one is salted

Sorry, I'm not sure whether this is an issue in this particular case.

traflm:
to Dave: yes, let me first do more investigation to find a better solution, this is not a very urgent requirement. 
to Selva: inside this method, isSystemColumn() and others all return false, it is where NAColumn object created that not properly set these flags. I am trying to find the root.
to Hans: RI is in a different code path I think, I am still trying to understand it. The predicate to ensure RI is generated by compiler, not from the source sql, so there must be some problem in the code which generate the original column list to compare.

traflm:
hi, all,
I look more into this issue, and I feel I need to accept Dave's proposal as the best option to fix this issue. That is, fix current isHiddenColumn() 'len - 1' issue and proceed with it.  And I added regression test for this change.
The proposal made by Selva and Hans to use NAColumn method is good. However, it is very difficult to  implement. As far as I understand up to now, it needs many changes to current code.
First, inside this RefConstraint::getPredicateText() method, each NAColumn object in the 'keyColumns' will NEVER return differently for any NAColumn::isXXColumn(), except the isUserColumn(), all other methods return false. The reason is: these objects are not properly initialized. 
The root where these objects are created is at:
 AbstractRIConstraint::setKeyColumns()
This line of code is how these NAColumns are created:
  column = new (heap) NAColumn(colDesc->colname, colDesc->position, NULL, heap);
So any attributes are NOT set at this point. And look further, where all column info is get from inside this setKeyColumns() method? It is from a class named constrnts_desc_struct.

struct constrnts_desc_struct {
  char *constrntname;
  char *tablename;
  ConstraintType type;
  Int32  colcount;
  Int32  isEnforced;
  char *indexname;
  desc_struct *check_constrnts_desc;
  desc_struct *constr_key_cols_desc;
  desc_struct *referenced_constrnts_desc;
  desc_struct *referencing_constrnts_desc;
#ifdef NA_LITTLE_ENDIAN
  void encode() {}
#endif
};

As one can see, there is NO information about column details like SALT, DIVISION, SYSKEY , etc...
So if we do need to create a NAColumn object which can correctly behave on isSaltColumn(), we have to change this constrnts_desc_struct.

However, look deeper into the code, all information that filled in the constrnts_desc_struct are come from a query from METADATA table ‘TRAFODION._MD_.TABLE_CONSTRAINTS'. So I think we have to save the corresponding attributes like SALT,DIVISION,SYSKEY in this metadata table. This is complex and break backward compatibility. 

Another way, is when generate the column list in the ‘TRAFODION._MD_.TABLE_CONSTRAINTS', filter out SALT,DIVISION,SYSKEY columns. However, this will have side effect which I cannot study out in a short time. If going into this way, the effort to solve this simple defect will be so insane. 

So , I admit that creating a new isHiddenColumn() function have shortcomings:
1. Itself may be incorrect, not well tested in all scenarios.
2. it must change when SALT, DIVISION, SYSKEY name conversion changed.

For 1, we have to do more tests.
For 2, I add some comments in this function to indicate this issue. And I saw there is a JIRA TRAFODION-1852 that will standardize these names as KeyWords, so for compatibility, I assume, these names will not change unless there is a strong reason.

Given, all above, I decided to go with my original code and abandon the effort to try to use NAColumn.isSaltColumn(), etc.

Hope all can agree and get this merged. This is a real blocking issue for real customers. A not perfect fix , but better than no fix.

i will commit a new version very soon,

- fixing Dave's comments about 'len -1'
- add test cases
- add some comments for isHiddenColumn()

zellerh:
@traflm your solution sounds good to me. Could you file a new JIRA to improve how we handle this in the longer term, with your analysis above? Note that we currently allow users to choose the name of the division column, although this syntax is not documented and not recommended.

DaveBirdsall:
Looks good to me. Is this ready to commit?

traflm:
thanks all, TRAFODION-1880 is created to follow up. This commit is a temp/easy solution, and we can work on TRAFODION-1880 to find out a better solution in a longer term.
Please help to merge it if we think this one is ready.

zellerh:
Ready to commit IMHO.

selvaganesang:
Agree with Hans. In the long term we need to ensure that the NAColumn is created with correct attributes. For now, it should be ok to commit.  I will commit now

svarnau:
I think @traflm means TRAFODION-1881

PR #348, opened 2016-02-29
TRAFODION-1706

mashengchen:
support vanilla hbase1.0 & cdh 1.0.0-cdh5.4.4


mashengchen:
jenkins, retest

narendragoyal:
This will need some rework. Please do not merge.

mashengchen:
jenkins, retest

core/sqf/hbase_utilities/src/main/java/org/trafodion/utility/backuprestore/TrafExportSnapshot.java
```diff
@@ -114,7 +119,10 @@
   private static final String INPUT_FOLDER_PREFIX = "export-files.";
 
   // Export Map-Reduce Counters, to keep track of the progress
-  public enum Counter { MISSING_FILES, COPY_FAILED, BYTES_EXPECTED, BYTES_COPIED, FILES_COPIED };
+  public enum Counter {
```
zellerh: Should this use some enum imported from another class? org.apache.hadoop.hbase.snapshot.ExportSnapshot.Counter?

core/sqf/sql/scripts/install_local_hadoop
```diff
@@ -391,7 +391,19 @@ else
     MY_START_PORT=`expr 24000 + $MY_START_PORT '*' 200`
   elif [ $MY_START_PORT == "rand" ]; then
     # pick a random number between 12000 and 24000 that is divisible by 200
-    MY_START_PORT=`expr $RANDOM '%' 60 '*' 200 + 12000`
+    lv_number=0   #initialize
```
zellerh: Sorry, I don't quite understand the reason for this change. Isn't the new code doing what the old code did, just more complicated? Also, it extends the range up to 31200. We wanted to keep the range fro 24000 to 32000 reserved for installations that use the -p fromDisplay option. Please don't change this algorithm unless there is a very good reason for doing so.

core/sqf/sql/scripts/install_local_hadoop
```diff
@@ -579,17 +591,15 @@ fi
...
-else
-  HADOOP_MIRROR_URL=http://download.nextag.com/apache/hadoop/common/hadoop-2.5.2
-  HADOOP_TAR=hadoop-2.5.2.tar.gz
-fi
+HADOOP_MIRROR_URL=http://archive.cloudera.com/cdh5/cdh/5
+HADOOP_TAR=hadoop-2.6.0-cdh5.4.4.tar.gz
 
-if [[ "$SQ_HBASE_DISTRO" = "HDP" ]]; then
-    HADOOP_TAR=hadoop-2.6.0.2.2.0.0-2041.tar.gz
+if [[ "$HBASE_DISTRO" = "HDP" ]]; then
```
zellerh: Is this a new requirement, that we set HBASE_DISTRO to some value before running install_local_hadoop? Maybe this is optional, only if we want to get something other than CDH? If so, that sounds good. We can make Apache the default once we have tested it for a while.

core/sqf/sql/scripts/install_local_hadoop
```diff
@@ -579,17 +591,15 @@ fi
...
+    HADOOP_TAR=hadoop-2.7.1.2.3.2.0-2950.tar.gz
```
zellerh: Don't we also have to set HADOOP_MIRROR_URL to some Hortonworks site?

core/sqf/sql/scripts/install_local_hadoop
```diff
@@ -602,15 +612,24 @@ MYSQL_TAR=mysql-5.6.10-linux-glibc2.5-x86_64.tar.gz
 MYSQL_JDBC_URL=http://cdn.mysql.com/archives/mysql-connector-java-5.1
 MYSQL_JDBC_TAR=mysql-connector-java-5.1.23.tar.gz
 
-HIVE_MIRROR_URL=https://archive.apache.org/dist/hive/hive-0.13.1
-HIVE_PREFIX=apache-hive-0.13.1-bin
+HIVE_MIRROR_URL=http://archive.cloudera.com/cdh5/cdh/5
+HIVE_PREFIX=hive-1.1.0-cdh5.4.4
 HIVE_TAR=${HIVE_PREFIX}.tar.gz
 
+if [[ "$HBASE_DISTRO" = "APACHE" ]]; then
```
zellerh: For HBASE_DISTRO=HDP, should we use Apache Hive instead of Cloudera?

core/sqf/sql/scripts/install_local_hadoop
```diff
@@ -602,15 +612,24 @@ MYSQL_TAR=mysql-5.6.10-linux-glibc2.5-x86_64.tar.gz
...
+
+
 HBASE_MIRROR_URL=http://archive.cloudera.com/cdh5/cdh/5
+HBASE_TAR=hbase-1.0.0-cdh5.4.4.tar.gz
 
-HBASE_TAR=hbase-0.98.6-cdh5.3.0.tar.gz
-if [[ "$SQ_HBASE_DISTRO" = "HDP" ]]; then
-    HBASE_TAR=hbase-0.98.4.2.2.0.0-2041-hadoop2.tar.gz
+if [[ "$HBASE_DISTRO" = "HDP" ]]; then
+    HBASE_TAR=hbase-1.1.2.2.3.2.0-2950.tar.gz
```
zellerh: Same here, also set HBASE_MIRROR_URL? You may not see this as a problem if you are using cached objects.

core/sqf/sql/scripts/install_local_hadoop
mashengchen: it's just for someone why want to install hbase version beside CDH. they can config it by do 'export HBASE_DISTRO=APACHE/HDP'.

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/SsccTransactionalScanner.java.tmpl
```diff
@@ -21,6 +21,26 @@
 * @@@ END COPYRIGHT @@@
 **/
 
+// @@@ START COPYRIGHT @@@
```
zellerh: Probably need to remove one of the two copyright notices?

core/sqf/sql/scripts/install_local_hadoop
mashengchen: MY_LOCAL_SW_DIST is the defaule set. just put the tars under MY_LOCAL_SW_DIST

core/sqf/sql/scripts/install_local_hadoop
mashengchen: i will change it back

zellerh: Thanks, from what you told me off-line it looks that this may be a merge conflict from another branch and changing it back is the right conflict resolution.

core/sqf/sql/scripts/install_local_hadoop
zellerh: Yes, but this is optional. It also should work without MY_LOCAL_SW_DIST, where it downloads the files from the web.

core/sqf/sql/scripts/install_local_hadoop
zellerh: That's great, thanks.

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/regionserver/transactional/CleanOldTransactionsChore.java.tmpl
```diff
@@ -21,14 +21,27 @@
 * @@@ END COPYRIGHT @@@
 **/
 
+/**
```
zellerh: Another duplicated copyright notice

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/regionserver/transactional/MemoryUsageChore.java.tmpl
```diff
@@ -21,6 +21,16 @@
 * @@@ END COPYRIGHT @@@
 **/
 
+/**
```
zellerh: Same here.

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/regionserver/transactional/TrxTransactionState.java.tmpl
```diff
@@ -21,15 +21,27 @@
 * @@@ END COPYRIGHT @@@
 **/
 
+/**
```
zellerh: ...and here

core/sql/pom.xml.apache
```diff
@@ -0,0 +1,131 @@
+<?xml version="1.0"?>
```
zellerh: Rather than having multiple pom.xml files, could we use profiles in a single pom.xml? Just asking, I can imagine very well that it won't work.

zellerh:
A very big piece of work! Thanks very much for doing this, it's a big step forward in achieving our goals for the next release. I noted some smaller comments, but didn't review all the Java changes in detail, hopefully someone more familiar with T-Rex and HBase can do that.

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/regionserver/transactional/TrxTransactionState.java.tmpl
```diff
@@ -77,96 +94,29 @@
...
-    static Class keepDeletedCellsClazz = null;
-    static Class scaninfoClazz = null;
-    static Constructor scaninfoConstructor = null;
-    static Object[] scaninfoArgs = null;
+    static java.lang.reflect.Constructor c1_0 = null;
 
     static {
         String version = VersionInfo.getVersion();// the hbase version string, eg. "0.6.3-dev"
         LOG.info("Got info of Class ScanQueryMatcher for HBase version :" + version);
-
```
selvaganesang: Are we getting rid of the backward compatibility here by removing this code.

mashengchen: i don't think there will happen this rid.. cause we have a 1.3 branch to support backward version(hbase 0.98)

core/sql/pom.xml.apache
mashengchen: no idea to use one pom.xml, we have to make 3 version (vanilla,cdh,hdp), so have to use three pom.xml

core/sqf/sql/scripts/install_local_hadoop
mashengchen: yes. for cdh&vanilla hbase it's fine...but seemingly hdp don't support separate tar for hadoop & hbase & hive. the only tar.gz is integrated for all. and it's 2.3GB, very big. not very recommend to download at the installation time.

traflm:
Too much changes, it is very hard for me to so may changes. It may need Ma, Shengchen to explain it in some way about the basic idea of this change.

core/sqf/sql/scripts/install_local_hadoop
mashengchen: hdp has it's own hive, if use apache hive, i can't promise it can work well ,maybe have some compatible problem. btw the hdp hive tar is integrated in HDP-2.3.0.0-centos6-tars-tarball.tar.gz. i can't find a separate tar..maybe someone can give me one hand to give me the download url

core/sql/pom.xml.apache
zellerh: Generally, profiles are used in Maven to build different versions, but profiles may not be sufficient for this case. See http://maven.apache.org/guides/introduction/introduction-to-profiles.html

core/sqf/sqenvcom.sh
```diff
@@ -284,6 +297,8 @@ elif [[ -d /opt/cloudera/parcels/CDH ]]; then
   export CURL_INC_DIR=/usr/include
   export CURL_LIB_DIR=/usr/lib64
 
+  lv_hbase_cp=`hbase classpath`
```
amandamoran: This will need to be set for Cloudera packages install as well. 

Add lv_hbase_cp=`hbase classpath` after elif [[ -n "$(ls /usr/lib/hadoop/hadoop-*cdh*.jar 2>/dev/null)" ]]; then
(around line 340).

svarnau:
This needs to be merged up to latest master to resolve conflicts before good test can be run.

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/regionserver/transactional/TrxTransactionState.java.tmpl
selvaganesang: I am not sure if we would maintain two different Trafodion versions and provide updates on continuous basis. It is not clear to me who should make this call. Trafodion community?

mashengchen:
will have a new pull request

PR #349, opened 2016-02-29
TRAFODION-1856: Revoke - object and column privilege checks not integrated for constraints

robertamarton:
Today, when revoking the object REFERENCES privilege, the revoke fails if there
are any RI constraints that require the privilege.  However, there may be column
level privileges that exist that would still allow the constraint to be present.
Conversely, when revoking column REFERENCES privilege, the revoke does not 
check to see if REFERENCES privilege has been granted at the object level.


core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
```diff
@@ -6309,6 +6304,23 @@ void CmpSeabaseDDL::alterSeabaseTableAddRIConstraint(
...
+
+       // If the user/role does not have REFERENCES privilege at the object 
+       // level, check to see if the user/role has been granted the privilege 
+       // on all affected columns
+       if (noObjPriv)
+         {
+           PrivMgrUserPrivs* privs = refdNaTable->getPrivInfo();
+           if (!privs->hasColReferencePriv(refdNAC->getPosition()))
+             {
+                *CmpCommon::diags() << DgSqlCode(-4481)
```
DaveBirdsall: Is there an enum for -4481?

core/sql/sqlcomp/PrivMgrMD.h
```diff
@@ -56,35 +58,29 @@ typedef int             Lng32;
...
-    details += ", UID is ";
-    details += to_string((long long int) objectUID);
-    details += ", name is ";
-    details += objectName;
-    details += ", grantee is ";
-    details += to_string((long long int) granteeID);
-    details += ", is owner ";
-    details += (grantorIsSystem) ? "true " : "false ";
+    details = "column usage ";
+    details += " column number is ";
```
DaveBirdsall: You'll get two spaces between "column usage" and "column number is". Is that what you want?

core/sql/sqlcomp/PrivMgrMD.h
```diff
@@ -110,12 +106,35 @@ typedef struct {
...
+  : objectUID(0),
+    objectOwner(NA_UserIdDefault),
+    objectType(COM_UNKNOWN_OBJECT),
+    columnReferences(NULL),
+    updatedPrivs()
+  {}
+    
+  virtual ~ObjectReference ( void )
+  {
+    while(!columnReferences->empty())
```
DaveBirdsall: This will abend if columnReferences is null.

core/sql/sqlcomp/PrivMgrMD.h
```diff
@@ -110,12 +106,35 @@ typedef struct {
...
+      delete columnReferences->back(), columnReferences->pop_back();
```
DaveBirdsall: Unusual coding technique (comma expression). Allows you to save some braces.

core/sql/sqlcomp/PrivMgrMD.h
```diff
@@ -130,7 +149,70 @@ typedef struct {
...
+  PrivMgrDesc updatedPrivs;
+
+  ColumnReference * findColumn (int32_t columnOrdinal)
+  {
+    if (columnReferences == NULL)
+      return NULL;
+    for (int i = 0; i < columnReferences->size(); i++)
+    {
+      ColumnReference *pRef = (*columnReferences)[i];
+      if (pRef->columnOrdinal == columnOrdinal)
```
DaveBirdsall: Instead of using a <vector>, you could use a <map> and get direct lookup rather than serial search.

DaveBirdsall:
Aside from one place where there might need to be a null pointer check, these code changes look good to me.

sureshsubbiah:
Looks good to me Roberta. Thanks.

core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
robertamarton: No, there are no enums for this range that I could find.

core/sql/sqlcomp/PrivMgrMD.h
robertamarton: Will fix this.

core/sql/sqlcomp/PrivMgrMD.h
robertamarton: I was having issues getting this code to work and I searched the web for examples.  This was one suggestion and it worked.

core/sql/sqlcomp/PrivMgrMD.h
robertamarton: I will note this and look at it for a subsequent commit.

core/sql/sqlcomp/PrivMgrMD.h
DaveBirdsall: Maybe I'm missing something. Isn't "delete columnReferences->back(), columnReferences->pop_back()" equivalent to { delete columnReferences->back(); columnReferences->pop_back(); }?

core/sql/sqlcomp/PrivMgrMD.cpp
```diff
@@ -962,29 +979,41 @@ PrivStatus PrivMgrMDAdmin::getReferencingTablesForConstraints (
...
   // Select all the constraints that are referenced by the table
-  // create_time is included to order by the oldest to newest
   std::string selectStmt = "select distinct o.object_uid, o.object_owner, o.object_type, o.create_time, ";
   selectStmt += "trim(o.catalog_name) || '.\"' || ";
-  selectStmt += "trim (o.schema_name) || '\".\"' ||";
-  selectStmt += "trim (o.object_name)|| '\"' from ";
+  selectStmt +=   "trim (o.schema_name) || '\".\"' ||";
+  selectStmt +=   "trim (o.object_name)|| '\"' ";
+  selectStmt += ", c.column_number "; 
+  selectStmt += "from " + uniqueRefConstraintsMDTable + std::string(", ");
```
selvaganesang: Why is that ","  should be std::string?  I wonder if std::string will trigger memory allocation in the malloc heap space

robertamarton: The "+" operator for the c++ string class does not allow char * to be appended. (At least I have been able to do so), but if it is converted to a string first, it works.  As far as how the malloc heap is used, probably depends on how the compiler generates the code.

sandhyasun:
Looks ready to merge.

PR #351, opened 2016-02-29
[TRAFODION-1850] First N operator no longer returns extra rows.

zellerh:
This also fixes [TRAFODION-908].

This is the entire or a partial fix to the problem where
a select [first n] * from udf(...) can return more than n rows.
The problem was that neither FirstN nor the UDF operator fully
implemented the GET_N protocol. While that may be ok for the UDF,
the FirstN needed a fix.

### DIFF
--- a/core/sql/executor/ExFirstN.cpp
+++ b/core/sql/executor/ExFirstN.cpp
@@ -100,6 +100,8 @@ ExFirstNTcb::ExFirstNTcb(const ExFirstNTdb & firstn_tdb,
 
   workAtp_ = NULL;
   firstNParamVal_ = 0;
+  effectiveFirstN_ = -1;
+  returnedSoFar_ = 0;
   if (firstn_tdb.workCriDesc_)
     {
       workAtp_ = allocateAtp(firstn_tdb.workCriDesc_, space);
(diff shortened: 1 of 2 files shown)


sureshsubbiah:
Change looks good Hans. Thank you for offering to take on this JIRA.

zellerh:
jenkins, retest

core/sql/executor/ExFirstN.cpp
```diff
@@ -228,11 +232,11 @@ short ExFirstNTcb::work()
 		// GET_N request value and firstNRows_ to my child.
 		if ((pentry_down->downState.request != ex_queue::GET_N) ||
 		    (pentry_down->downState.requestValue == firstnTdb().firstNRows()))
-		  centry->downState.requestValue = firstNVal;
+		  centry->downState.requestValue = effectiveFirstN_;
```
DaveBirdsall: In the comment above this (line 230 to be precise) should we change "firstNRows_" to "effectiveFirstN_"?

DaveBirdsall:
Looks good to me.

core/sql/executor/ExFirstN.cpp
zellerh: Thanks, I'll fix that.

zellerh: Actually, that made me realize that on new line 234 the code also needs to be changed to effectiveFirstN_.

zellerh:
jenkins, retest

DaveBirdsall:
Looks good to me.

PR #353, opened 2016-03-02
[TRAFODION-1865] uninitialized buffer cause Monitor abort in CentOS 7

traflm:
We tested on CentOS 7.2, by applying this change, sqstart success.

### DIFF
--- a/core/sqf/monitor/linux/process.cxx
+++ b/core/sqf/monitor/linux/process.cxx
@@ -3353,6 +3353,7 @@ CProcessContainer::CProcessContainer( bool nodeContainer )
    
     //create & initialize existing semaphore
     char sem_name[MAX_PROCESS_PATH];
+    memset(sem_name, 0 , MAX_PROCESS_PATH);  //clear the buffer
     snprintf(sem_name,sizeof(sem_name), "/monitor.sem.%s", getenv("USER"));
     Mutex = sem_open(sem_name,O_CREAT,0644,0);
     if(Mutex == SEM_FAILED)


DaveBirdsall:
Looks good to me.

DaveBirdsall:
jenkins, retest

DaveBirdsall:
jenkins, retest

zellerh:
Just curious, is this a bug in sem_open? I'm having a hard time imagining why we should have to initialize this buffer. The man page for sem_open says it expects a null-terminated string. Anyway, initializing it doesn't do any harm, so I'm just asking, not complaining.

traflm:
yes, Hans, I think it again. You are quite right. This change doesn't make too much sense. The const string should contain a null at the end, and after that null, there will be some garbage. 
But when we test on CentOS, sqstart can never work well, monitor abort() exactly at this sem_open place, and if return value is SEM_FAILED, it will abort and Trafodion cannot start.
After make this change, rebuild and reinstall, sqstart works fine on CentOS 7. 
Maybe some change in libc in CentOS 7. Let me investigate it further. But this change does no harm, just may not properly fix the JIRA. I need to do more tests.

zellerh:
Thanks for the explanation. Could it be some security problem or some left-over semaphore that causes the problem - but of course that would not explain why your change fixed the problem.

If this fixes the problem, then maybe the best thing is not to ask too many questions :-)

If this does not fix the problem, we will find out sooner or later and can investigate further. So, I see no harm in committing the change and seeing what happens.

traflm:
hi, all,
Please don't merge this. I realize what Hans‘ gut is quite right. 
I may made some fundamental mistakes here about how Trafodion support CentOS 7.
Our code is built in CentOS 6.5 , with lower version of libpthread . And when run on CentOS 7, it may behave in unexpected way. I guess what ever statement I add before sem_open() will fix the issue.
So correct way to support Trafodion running on CentOS 7 is to build Trafodion from a CentOS 7 system. 
My 'fix' to this issue is wrong. I will cancel this pull request.

traflm:
abandon this change

PR #354, opened 2016-03-02
[TRAFODION-1668] Emergency build fix

svarnau:
The earlier pull request passed test, but then it was a long time
before it merged. By time work for 1668 merged, it caused a build error.
The branch is now broken for all builds without this small change.


DaveBirdsall:
Looks good to me. I suppose we should commit this as soon as tests pass?

PR #355, opened 2016-03-02
[TRAFODION-1866] Build should not require qmake

svarnau:


zellerh:
Looks good to me, thanks for fixing that!

zellerh:
Actually, at second thought: Could we separate the mk.sh and cp commands into two lines and just ignore the exit code of cp? It would be good to catch checkins that break the GUI build. We had that a while ago and had a hard time getting it back to work when we didn't notice the problem immediately.

svarnau:
@zellerh yes, that thought had occurred to me as well to make it into two lines. But you make a good point why that is necessary.  Will do.

DaveBirdsall:
The rework looks good to me.

zellerh:
Looks good to me, too, thanks for making that change.

PR #356, opened 2016-03-02
[TRAFODION-1862] Transaction ID is returned as -1 when using the in m…

selvaganesang:
…emory table for RMS stats


zellerh:
Looks good to me.

DaveBirdsall:
Looks good to me also.

PR #357, opened 2016-03-02
[TRAFODION-1867] Allow parallel plans for non-partitioned tables in U…

DaveBirdsall:
…PD STATS.

The problem is for large non-partitioned tables, UPDATE STATS will do a series of SELECTs to compute statistics for sets of columns. The plans for these SELECTs are presently forced to be serial for non-partitioned tables as a workaround for a long-ago compiler bug which seems to have been fixed. It's time to remove this workaround and thereby speed up UPDATE STATS.


zellerh:
Looks good to me.

selvaganesang:
Good to go

PR #358, opened 2016-03-03
[TRAFODION-1860,1875]

mkby:
remove dependency of ctime.pl

### DIFF
--- a/core/sqf/sql/scripts/gensq.pl
+++ b/core/sqf/sql/scripts/gensq.pl
@@ -21,7 +21,6 @@
 #
 # @@@ END COPYRIGHT @@@
 #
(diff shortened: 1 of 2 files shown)


traflm:
Looks very good for me. 
This not only help for CentOS 7, but also other Linux dist in the future. We should reduce unnecessary dependencies as much as possible.

svarnau:
jenkins, add user

svarnau:
jenkins, retest -- hbase re-start failed during trafodion install on only one machine.

DaveBirdsall:
Looks good to me. If there are no further comments, I will commit this tomorrow morning (USA time).

PR #359, opened 2016-03-03
Genstats script and fix to JIRA 1872 (looping compiler)

nonstop-qfchen:
Hi,

This is to push the genstats.sh and the looping compiler fix to the Trafodion branch. 

Please review and commit.

Thanks --Qifan


DaveBirdsall:
Looks good to me. Thanks for contributing the genstats script. I've already used it and find it very convenient.

PR #361, opened 2016-03-03
[TRAFODION-1836] Remove win-odbc license info from clients package

svarnau:
Removing TFODBC64 from the client packaging, since there is still
a license issue with pulling in redistributed code from visual
studio.  The existing license info is moved to a new file that is
not added to the LICENSE file during packaging. But it is retained
for future use when the issue can be resolved and win-odbc added back
into the binary distribution.


DaveBirdsall:
Looks good to me. I wonder if our mentor's license text sniffers will stumble upon the retained text and raise a flag. Though I suppose that the action then would be to read the file, and then the explanation there should be sufficient.

PR #363, opened 2016-03-04
Support for doing GC on LOB data files and removing functionality from mxlobsrvr process TRAFODION-1848

sandhyasun:
The following set of changes are being committed : 
1. Moved all calls made in mxlobsrvr process into the master. Removed interprocess sned/recieve calls and made them function calls all done within master. Retianed the minimum infrastructure to mxlobsrvr but it's not started anymore during sqstart. 
2. Add a new LOB call to perform GC. Made this a new CLI call so we could extend it later to be a utility that can be called offline. For now this CLI cll will be triggered if a certain threshold is met during an insert/update/append operation. This GC is done per lob column (i.e per lob data file). Each lob column's GC can be done independently and do not need to be done all at once. Dpeneds on how much space each lob column ins occupying.
3 . Removed all uses of hardcoded references ot lob location and hdfsServer and hdfs port. ALl these are set via CQDs and are passed in to needed Tdbs. 
4. A new testcase has been added to executor/TEST130  to exercise the new code.


svarnau:
jenkins, retest -- build got stuck behind an earlier job that hung, causing tests to time out

core/sql/cli/Cli.cpp
```diff
@@ -9671,7 +9622,12 @@ Lng32 SQLCLI_LOBcliInterface
 	if (cliRC < 0)
 	  {
 	    cliInterface->retrieveSQLDiagnostics(myDiags);
-	    
+	     cliInterface->retrieveSQLDiagnostics(myDiags);
```
arvind-narain: This got duplicated - could be removed.

DaveBirdsall:
So is the effect of this change that LOB garbage collection is no longer driven by a separate process but instead is done on a piecemeal basis as clients write to LOB columns? Will this have the effect of introducing performance hiccups into transactions involving LOBs?

core/sql/cli/Cli.cpp
```diff
@@ -9684,15 +9640,17 @@ Lng32 SQLCLI_LOBcliInterface
 	cliRC = cliInterface->executeImmediate(query);
 
 	currContext.resetSqlParserFlags(0x1);
-
+	
 	if (cliRC < 0)
 	  {
 	    cliInterface->retrieveSQLDiagnostics(myDiags);
-	    
+	    if (myDiags->containsError(-CAT_OBJECT_DOES_NOT_EXIST_IN_TRAFODION))
```
DaveBirdsall: What if there are other errors in myDiags along with CAT_OBJECT_DOES_NOT_EXIST_IN_TRAFODION; is it OK to ignore them as well?

core/sql/cli/Cli.cpp
```diff
@@ -9581,15 +9531,14 @@ Lng32 SQLCLI_LOBcliInterface
...
+	    
 	    cliInterface->retrieveSQLDiagnostics(myDiags);
 	    
 	    goto error_return;
 	  }
 
            	// create lob descriptor chunks table salted
-       	str_sprintf(query, "create ghost table %s (descPartnKey largeint not null, descSysKey largeint not null, chunkNum int not null, chunkLen largeint not null, intParam largeint, stringParam varchar(400), primary key(descPartnKey, descSysKey, chunkNum)) salt using 8 partitions",
+       	str_sprintf(query, "create ghost table %s (descPartnKey largeint not null, descSysKey largeint not null, chunkNum int not null, chunkLen largeint not null, dataOffset largeint, stringParam varchar(400), primary key(descPartnKey, descSysKey, chunkNum)) salt using 8 partitions",
 	lobDescChunksName);
```
DaveBirdsall: Does renaming a column imply an upgrade step when this software is installed?

sandhyasun:
Dave, there has not been any support for GC prior to this commit.  This is all new. You are right that it will introduce a  hiccup for IUA operations.  what moved out of the mxlobsrvr process is the regular processing of inser/updates/deletes itself. All LOB metadata operations were done in the mxlobsrvr process and all hdfs operations in the master (in a serial manner). There was no need for the process hop in that case. Hence moved that functionality into the master.
Now for GC itself, when have a new Util operator/syntax that can be issued offline, we will consider doing the GC in the mxlobsrvr process. For now it's part of the master. Is that clearer ? This was the way to inttroduce GC without introducing new syntax to do it.

core/sql/exp/ExpLOB.cpp
```diff
@@ -186,13 +173,114 @@ Lng32 ExpLOBoper::createLOB(void * lobGlob, void * lobHeap,
...
+  if (lobGlob == NULL)
+    {
+      rc = initLOBglobal(lobGlobL, lobHeap);
+      if (rc)
+	return -1;
+    }
+  else
+    lobGlobL = lobGlob;
+ 
+  if (rc)
```
DaveBirdsall: Harmless but unnecessary as this 'if' can never be true.

core/sql/exp/ExpLOB.cpp
```diff
@@ -186,13 +173,114 @@ Lng32 ExpLOBoper::createLOB(void * lobGlob, void * lobHeap,
...
+  void * lobGlobL = NULL;
+   if (lobGlob == NULL)
+    {
+      rc = initLOBglobal(lobGlobL, lobHeap);
+      if (rc)
+	return -1;
+    }
+  else
+    lobGlobL = lobGlob;
+  rc = initLOBglobal(lobGlobL, lobHeap);
```
DaveBirdsall: Not sure I understand the point of the 'if'. In both cases, lobGlobL will be equal to lobGlob when we call initLOBglobal. Why not just code "rc = initLOBglobal(lobGlob, lobHeap)" once and be done with it?

core/sql/exp/ExpLOB.cpp
```diff
@@ -186,13 +173,114 @@ Lng32 ExpLOBoper::createLOB(void * lobGlob, void * lobHeap,
...
+  void * lobGlobL = NULL;
+  if (lobGlob == NULL)
+    {
+      rc = initLOBglobal(lobGlobL, lobHeap);
+      if (rc)
+	return -1;
+    }
+  else
+    lobGlobL = lobGlob;
+  rc = initLOBglobal(lobGlobL, lobHeap);
```
DaveBirdsall: Same comment as in the prior method.

DaveBirdsall:
Hi @sandhyasun, yes your explanation makes it clear. The changes look good to me. My code comments are all minor.

core/sql/cli/Cli.cpp
sandhyasun: We will ignore only this particular  error. If it's any other error, we go to "error_return". That will return an error to the calling layer.

core/sql/cli/Cli.cpp
sandhyasun: Yes normally it will. It is safe to make this change now since the feature is not officially "externalized".

core/sql/exp/ExpLOB.cpp
sandhyasun: You found a defect. The if is needed and we cannot call initLOBglobal if lobGlob was passed in. Correcting that.

core/sql/cli/Cli.cpp
DaveBirdsall: Right, but myDiags could contain more than one error, right? What happens then?

sandhyasun:
Working on a "drop schema" issue that is exposed with these changes. Please do not merge this yet.

core/sql/cli/Cli.cpp
sandhyasun: Found a related issue with drop schema and ddl transactions. I am going to backout this change and not filter out any errors. It does not help with ddl transactions - the diagnostics not being cleared causes problems during transaction commit. So opened JIRA TRAFODION-1883 and will fix it shortly (as part of a separate pull request) .

selvaganesang:
In case of hbase we have moved away from specifying the server and port and let Hbase Client API get it from hbase-site.xml. It looks like we are introducing the server and port for hive back here,  I couldn't figure out the need for it from the changes. Can you please explain?

sandhyasun:
The port and server changes done here are simply to use the CQD values which were always there. We are no tintroducing anything new here.The problem was the CQD values were used only in a couple of places. All other places , the port and server were hardcoded to "default" and 0 (zero). SO made changes to remove the hardcoding and use the CQD value in all places. The changes are not for hive access. They are limited tot the places where we are doing LOB data reads and writes.  We use the libhdfsapi in all places to access hdfs here. If there is away to get the port/server from the hbase-site.xml  to pass into the libhdfs API, I will look into that for a future commit.

DaveBirdsall:
The reworks look good to me. Is this ready to commit?

sandhyasun:
Yes - addressed all needed issues for this commit. Thanks !

DaveBirdsall:
Are you committing it @sandhyasun or would you like me to?

sandhyasun:
I will do it

selvaganesang:
Did you mean the server or port changes are for hbase access? I was planning to remove these CQDs to avoid confusion and let Hbase API handle it transparently when time permits.  If these are meant for hbase, then these would be cleaned up as part of this effort.

sandhyasun:
The server and port settings are for LOB hdfs libhdfsAPI access - not Hbase access. So these  two  CQDs LOB_HDFS_PORT and LOB_HDFS_SERVER need to stay for now.

PR #364, opened 2016-03-04
[TRAFODION-1863] PR340 caused PushdownV2 issues with column having non null default

eowhadi:
since non null default columns are sometime resolved at select time, we need to disable pushdown for these.


eowhadi:
jenkins, retest

core/sql/optimizer/ValueDesc.cpp
```diff
@@ -529,14 +529,13 @@ NABoolean ValueId::isAddedColumnWithNonNullDefault() const{
   default:
       break;
   }
-  if (nac && nac->isAddedColumn() && nac->getDefaultValue())
+  if (nac &&  nac->getDefaultValue() && strcmp(nac->getDefaultValue(),"NULL") != 0)
```
selvaganesang: Will it be better to check for DefaultColumnClass == COM_NULL_DEFAULT instead.

eowhadi: why is it better? I know what I do work. I don't know about the alternative. But I know how much a re run of regression cost :-)

selvaganesang:
Agree. But, isn't good to the use the existing methods. Moreover, the COM_NULL_DEFAULT is used in many places already.

generator/GenExpGenerator.cpp:    case COM_NULL_DEFAULT:
optimizer/NATable.cpp:                             COM_NULL_DEFAULT  ,//defaultClass,
smdio/CmUtil.cpp:  {COM_NULL_DEFAULT, COM_NULL_DEFAULT_LIT},
sqlcomp/CmpDescribe.cpp:  else if (nac->getDefaultClass() == COM_NULL_DEFAULT)
sqlcomp/CmpSeabaseDDLcommon.cpp:          defaultClass = COM_NULL_DEFAULT;
sqlcomp/CmpSeabaseDDLcommon.cpp:              defaultClass = COM_NULL_DEFAULT;
sqlcomp/CmpSeabaseDDLtable.cpp:      colInfoArray[index].defaultClass = COM_NULL_DEFAULT;
sqlcomp/CmpSeabaseDDLtable.cpp:      else if (colInfo.defaultClass == COM_NULL_DEFAULT)
sqlcomp/CmpSeabaseDDLutil.cpp:  {COM_NULL_DEFAULT, COM_NULL_DEFAULT_LIT},

eowhadi:
OK so tell me if this change is correct and is better:
replace: if (nac &&  nac->getDefaultValue() && strcmp(nac->getDefaultValue(),"NULL") != 0)

by: if (nac && nac-> getDefaultClass() != COM_NO_DEFAULT &&
nac->getDefaultClass() != COM_ALWAYS_COMPUTE_COMPUTED_COLUMN_DEFAULT &&
nac->getDefaultClass() != COM_ALWAYS_DEFAULT_COMPUTED_COLUMN_DEFAULT &&
nac->getDefaultClass() != COM_NULL_DEFAULT)

or probably using a switch construct with same semantic...

selvaganesang:
I think the exact replacement of the existing code will be (nac && nac->getDefaultValue() && (nac->getDefaultClass() != COM_NULL_DEFAULT).

Returns TRUE for any column with the default value but with non-null default value. This is what I understood reading the code before. For anything else it would be false. 

You can add nac->getDefaultClass() != COM_NULL_DEFAULT && nac->getDefaultClass() != COM_CURRENT_DEFAULT() because the columns with current default will be in hbase table always even when the column is omitted.

core/sql/generator/GenPreCode.cpp
```diff
@@ -11386,7 +11386,7 @@ NABoolean HbaseAccess::isHbaseFilterPredV2(Generator * generator, ItemExpr * ie,
   }
   //check if not an added column with default non null
   if ((foundBinary || foundUnary)&& (NOT hbaseLookupPred)){
-        if (colVID.isAddedColumnWithNonNullDefault()){
+        if (colVID.isColumnWithNonNullDefault()){
```
zellerh: Would a check for added columns with non-null default be sufficient for aligned format?

core/sql/optimizer/ValueDesc.cpp
zellerh: Agree with Selva that checking the enum is safer and more maintainable. Maybe you can make this change in a future check-in, so you don't have to rerun the regressions just for that?

zellerh:
Looks good to me, with some small comments.

core/sql/generator/GenPreCode.cpp
eowhadi: this code is totally bypassed for align format

eowhadi:
Test passed , I guess good for merging?

PR #366, opened 2016-03-06
[TRAFODION-1854] Trafodion cannot start on nodes with uppercase hostname

traflm:
Monitor's CClusterConfig class treat all character as lower case when reading the cluster configuration file.
Monitor will not use hostname in normal operation, so convert the hostname into lowercase in shell.cxx will solve the issue

### DIFF
--- a/core/sqf/monitor/linux/shell.cxx
+++ b/core/sqf/monitor/linux/shell.cxx
@@ -4569,6 +4569,18 @@ bool start_monitor( char *cmd_tail, bool warmstart, bool reintegrate )
     // Ensure that we are on a node that is part of the configuration
     char mynode[MPI_MAX_PROCESSOR_NAME];
     gethostname(mynode, MPI_MAX_PROCESSOR_NAME);
+    //JIRA: TRAFODION-1854, hostname contains upper case letters
+    //change all char in mynode to lowercase
+    //since cluster config strings will all in lowercase , see CTokenizer::NormalizeCase()
+    if(!VirtualNodes)  // for VirtualNodes, it use same gethostname, so do not tolower
+    {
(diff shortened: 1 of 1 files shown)


amandamoran:
What if on my network I have a node called "Node1.amanda.com" and "node1.amanda.com"? If the monitor changes everything to lowercase then wouldn't there be confusion?

Maybe this isn't really an issue, but it came to my mind.

traflm:
Hi, Amanda,
When a cluster have two nodes " Node1.amanda.com" and "node1.amanda.com", the monitor will fail for sure. It cannot distinguish the two nodes anymore. But that is a very rare case, we'd better ask user not to do that instead of spending time to support this feature, it will be really some coding work :-)

But it will be very helpful to test more, I don't have a real cluster to test this, just use 'hostname something' to unit test it on my dev workstation.

DaveBirdsall:
Looks good to me.

zellerh:
Looks good to me. [DNS names are not case-sensitive](https://tools.ietf.org/html/rfc4343), so "Node1.amanda.com" a "node1.amanda.com" must be one and the same node.

amandamoran:
Thanks Hans. That was what I needed to know :)

Thanks!

On Tue, Mar 8, 2016 at 8:58 AM, Hans Zeller <notifications@github.com>
wrote:

> Looks good to me. DNS names are not case-sensitive
> <https://tools.ietf.org/html/rfc4343>, so "Node1.amanda.com" a "
> node1.amanda.com" must be one and the same node.
>
> —
> <https://github.com/apache/incubator-trafodion/pull/366#issuecomment-193866762>
> .
>



-- 
Thanks,

Amanda Moran

PR #367, opened 2016-03-07
TRAFODION-1706

mashengchen:
support vanilla hbase1.0 & cdh 1.0.0-cdh5.4.4


core/sqf/sqenvcom.sh
```diff
@@ -140,18 +139,41 @@ fi
...
-    export HBASE_TRX_JAR=hbase-trx-hdp2_2-${TRAFODION_VER}.jar
+export HBASE_TRX_JAR=${HBASE_TRX_ID_CDH}-${TRAFODION_VER}.jar
+if [[ "$HBASE_DISTRO" = "HDP" ]]; then
+    export HBASE_TRX_JAR=${HBASE_TRX_ID_HDP}-${TRAFODION_VER}.jar
 fi
-# set common version to be consistent between shared lib and maven dependencies
-export THRIFT_DEP_VER=0.9.0
-export HIVE_DEP_VER=0.13.1
+if [[ "$HBASE_DISTRO" = "APACHE" ]]; then
+    export HBASE_TRX_JAR=${HBASE_TRX_ID_APACHE}-${TRAFODION_VER}.jar
```
mkby: If using Apache Hadoop, the script will detect the system as an Apache Hadoop environment, so it will jump to line 487, and check line 618, the script will ``export HBASE_TRX_JAR=hbase-trx-hbase_98_4-${TRAFODION_VER}.jar``, will this overwrite the ``HBASE_TRX_JAR`` you set here?

traflm:
Is it possible to break down this task? It is too complex to finish all in one check in.
But if it can pass Jenkins, I will spend more time to try to review it.

svarnau:
Latest test got past the HDP install issue, but sqstart does not work on HDP, though CDH is working fine.
Also, merge up is needed, since there are branch conflicts.

DaveBirdsall:
I'm wondering how to move this change forward. Seems like the conflict needs to be resolved. (I'm guessing that https://github.com/apache/incubator-trafodion/pull/376 might be one source of conflict, so the fix from that pull request needs to be integrated into the changes here.) Also seems like we need to get more reviewers active. Looks like most of the changes are in the install and transaction areas. Is this correct?

mashengchen:
i can add the change#376 in my PR, but i don't think it's good idea.

DaveBirdsall:
You'll need to resolve the conflict somehow or the change cannot be merged.

svarnau:
The change also needs to pass tests. We've not been able to get this working on HDP test nodes yet.

mashengchen:
jenkins, retest

DaveBirdsall:
jenkins, retest

prashanth-vasudev:
These changes look good to me. Thanks.

DaveBirdsall:
Sounds like this is ready to commit. My understanding is that separately there are several issues with the regression suites this week and those issues are being addressed. As soon as that is clean, we'll commit this change. Thanks for all the hard work on this.

sandhyasun:
Agree with previous comment. The change to fix most of the regression failures shoudl be in tonight. After that is in, can you  pull in the latest code and then confirm that ALL the SQL  regression suites (not just core and seabase) run cleanly ? If there are a couple of failures please share them here. This is  big change so we need to verify that before it gets merged.

mashengchen:
got that ,i will do a pull tomorrow.

PR #369, opened 2016-03-07
Fix for a typo in runregr_udr, causing udr/TEST002 failure

zellerh:
The problem was reported by Eric over the weekend and Arvind found
the bug and suggested this fix.

The failure occurs when a user installed Hadoop using non-standard ports and if there is no running instance on the default ports.

### DIFF
--- a/core/sql/regress/tools/runregr_udr.ksh
+++ b/core/sql/regress/tools/runregr_udr.ksh
@@ -851,9 +851,9 @@ if [ $LINUX -ne 0 ]; then
 fi
 
 export JDBC_T4_URL="jdbc:t4jdbc://localhost:23400/:"
-if [ -r $MY_SQROOT/sql/scripts/swenv.sh ]; then
+if [ -r $MY_SQROOT/sql/scripts/sw_env.sh ]; then
   # use a custom port for the JDBC Type 4 driver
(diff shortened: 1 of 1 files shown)


DaveBirdsall:
Looks good to me.

core/sql/regress/tools/runregr_udr.ksh
```diff
@@ -851,9 +851,9 @@ if [ $LINUX -ne 0 ]; then
 fi
```
hegdean: Do we want to clean up the unused  env variables/conditional checks ( _RLD_LIB_PATH and CLASSPATH) ?

zellerh: There is a lot of cleanup that could be done here, but that would require more testing and take more time. Is it ok if we keep this PR to just the typo?

hegdean: Should be fine Hans.  Just did not want to see hpt4jdbc.jar being referenced anywhere incorrectly as our tyep4 jar file is now called jdbcT4.jar

zellerh: Thanks, now I understand why you are suggesting this. I'll push an update shortly.

arvind-narain:
Looks good Hans - thanks. Merging in now.

PR #372, opened 2016-03-08
TRAFODION [1879] - Integrate Library Management into Trafodion Metadata

robertamarton:
The initial release of library management has been delivered to Trafodion. This
is follow-on work that integrates library management operations into the
existing Trafodion infrastructure. Currently, with the initial release of
library management, the consumer needs to run a special script to set up
everything. This delivery incorporates the steps into existing SQL commands and
removes the need for the script.


robertamarton:
jenkins, retest

core/sql/common/ComUser.cpp
```diff
@@ -347,3 +347,57 @@ Int16 ComUser::getAuthNameFromAuthID(Int32   authID,
...
+//   (out) actualLen - the length of the returned role list
+//   ( in) maxLen - the size of the roleList allocated by the caller
+//   ( in) delimited - delimiter to use (defaults to single quote)
+//   ( in) separator - specified what separator to use (defaults to comma)
+//   ( in) includeSpecialAuths - includes the special auths (PUBLIC and _SYSTEM) 
+//
+//  Returns:  FEOK -- found
+//            FEBUFTOOSMALL -- space allocated for role list is too small 
+// ----------------------------------------------------------------------------
+Int32 ComUser::getRoleList (char * roleList,
```
DaveBirdsall: I wonder if we could return an NAString instead, then we'd never have to trouble ourselves with FEBUFTOOSMALL? (At the price of dynamic heap management though.)

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -1425,3 +1426,169 @@ short CmpSeabaseDDL::validateRoutine(ExeCliInterface *cliInterface,
...
+                         gluedQuery, gluedQuerySize);
+
+      param_[0] = getSystemCatalog();
+      param_[1] = SEABASE_LIBMGR_SCHEMA;
+      param_[2] = getSystemCatalog();
+      param_[3] = SEABASE_LIBMGR_SCHEMA;
+      param_[4] = SEABASE_LIBMGR_LIBRARY;
+
+      str_sprintf(queryBuf, gluedQuery, param_[0], param_[1], param_[2], param_[3], param_[4]);
+      NADELETEBASIC(gluedQuery, STMTHEAP);
```
DaveBirdsall: How do we know that queryBuf is large enough to hold the data str_sprintf writes to it?

DaveBirdsall:
Looks good to me.

core/sql/common/ComUser.cpp
robertamarton: I thought about it, I tried to keep the code similar to other functions in the same file.  But you have a point. Will think about it for a subsequent release.

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
robertamarton: 2000 is big enough but I changed the approximate size calculation based on the size of the catalog, schema, library, and glued query - will deliver in my next release - coming soon.

PR #373, opened 2016-03-08
[TRAFODION-1880] Do not build libhdfs dependency from source

svarnau:
Make sure various pom.xml files with hadoop dependency are all pulling in
the same version.

Make HDFS dependencies a build-time pre-req. As with other build
dependencies on shared libraries, make them part of the build environment
(TOOLSDIR).

Prior to this change, we were picking up these dependencies (2 shlib &
1 header) from environment (such as local_hadoop) or downloading the
source and building it on the fly. Building it made trafodion build more
than twice as long.

Now requiring a consistent build dependency, separate from the runtime
environment.

The 64-bit native libraries are now available from binary distro of
hadoop-common. So we can download binaries and extract files we need rather
than build them.  Dependencies can be updated via install/traf_tools_setup.sh,
but in case the environment is not updated, the build-time script get_hdfs_files
is also updated to downoad the distro. If you really want to build
hadoop-common, that portion of the script was retained under a new --source
option.

### DIFF
--- a/core/sqf/sql/scripts/get_libhdfs_files
+++ b/core/sqf/sql/scripts/get_libhdfs_files
@@ -43,7 +43,7 @@ LOGFILE=${LIBHDFS_TEMP_DIR}/build.log
 
 # Hadoop source tar file to build libhdfs from
 HADOOP_MIRROR_URL=https://archive.apache.org/dist/hadoop/common/hadoop-2.6.0
-HADOOP_ID=hadoop-2.6.0
+HADOOP_ID=hadoop-${HADOOP_DEP_VER}
 HADOOP_SRC_ID=${HADOOP_ID}-src
 HADOOP_SRC_TAR=${HADOOP_SRC_ID}.tar.gz
 HADOOP_BIN_TAR=${HADOOP_ID}.tar.gz
--- a/core/sqf/sql/scripts/get_libhdfs_files
+++ b/core/sqf/sql/scripts/get_libhdfs_files
@@ -42,8 +42,8 @@ fi
 LOGFILE=${LIBHDFS_TEMP_DIR}/build.log
 
 # Hadoop source tar file to build libhdfs from
-HADOOP_MIRROR_URL=https://archive.apache.org/dist/hadoop/common/hadoop-2.6.0
 HADOOP_ID=hadoop-${HADOOP_DEP_VER}
+HADOOP_MIRROR_URL=https://archive.apache.org/dist/hadoop/common/${HADOOP_ID}
 HADOOP_SRC_ID=${HADOOP_ID}-src
 HADOOP_SRC_TAR=${HADOOP_SRC_ID}.tar.gz
 HADOOP_BIN_TAR=${HADOOP_ID}.tar.gz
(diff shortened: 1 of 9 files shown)


core/sqf/sql/scripts/get_libhdfs_files
```diff
@@ -37,10 +42,11 @@ fi
 LOGFILE=${LIBHDFS_TEMP_DIR}/build.log
 
 # Hadoop source tar file to build libhdfs from
-HADOOP_SRC_MIRROR_URL=https://archive.apache.org/dist/hadoop/common/hadoop-2.6.0
+HADOOP_MIRROR_URL=https://archive.apache.org/dist/hadoop/common/hadoop-2.6.0
```
zellerh: Would it make sense to replace the 2.6.0 with ${HADOOP_DEP_VER}? Just in case this script survives for a longer time.

DaveBirdsall:
Looks good to me.

core/sqf/sql/scripts/get_libhdfs_files
```diff
@@ -110,113 +123,141 @@ if [[ $FORCE_BUILD == true || \
       exit 1
     fi
   fi
-
   cd $LIBHDFS_TEMP_DIR
 
-  PROTOBUF_VER=`protoc --version 2>/dev/null | cut -f 2 -d ' '`
+  if [[ $SOURCE_BUILD != true ]]; then
+    echo "Downloading Hadoop-common binary distro..." | tee -a ${LOGFILE}
```
zellerh: There might be a small chance that we would pick up 32 bit libraries, since older Hadoop binary packages contain those instead of 64 bit libraries. Probably ok, as we move away from those older versions.

core/sqf/sql/scripts/get_libhdfs_files
```diff
@@ -110,113 +123,141 @@ if [[ $FORCE_BUILD == true || \
...
+       $HADOOP_ID/lib/native/libhdfs\*so\* \
+       $HADOOP_ID/include/hdfs.h
+
+    cp -f ${HADOOP_ID}/include/hdfs.h ${TGT_INC_DIR}
+    cp -Pf ${HADOOP_ID}/lib/native/libhdfs*.so* ${TGT_LIB_DIR}
+    cp -Pf ${HADOOP_ID}/lib/native/libhadoop*.so* ${TGT_LIB_DIR}
+
+  else
+
+    PROTOBUF_VER=`protoc --version 2>/dev/null | cut -f 2 -d ' '`
```
zellerh: Those are all marked as new, but my guess is that they are just indented, so I didn't look at these lines in detail.

core/sqf/sql/scripts/get_libhdfs_files
zellerh: Never mind, just saw the check for 64 bit libraries at the end.

core/sqf/sql/scripts/get_libhdfs_files
svarnau: Right. Above this line are new, but the large section below was indented for the else clause.

core/sqf/sql/scripts/get_libhdfs_files
svarnau: Yes, makes sense.

zellerh:
Very nice, another improvement in our build!

PR #374, opened 2016-03-10
[TRAFODION-1868] Compatibility with gcc 4.8

traflm:
this is partial changes to Trafodion code base in order to make it compatible with gcc 4.8.
Gcc 4.8 generally is more strict than gcc 4.4, so Trafodion code need to fix all potential issues.
1. variable declared and initialized, but not used anymore
2. suspicious overload of same variable names
3. ld need the required library name to be explicitly added using '-l' option
4. some STL/Boost functions replaced with c++11 features in gcc 4.8, so remove unnecessary call into boost
So there are a lot of changes, to make this process easier, we decided to do this incrementally. Fix small amount of issues each time, easier to code review and test.
It must make sure same code compile and work well under gcc 4.4.

This is the first commit for the whole gcc 4.8 support changes.
We already have a successful gcc 4.8 build prototype, and it works well on CentOS 7. But there are still more changes, so this is a rather proven way to support gcc 4.8.

### DIFF
--- a/core/dbsecurity/auth/Makefile
+++ b/core/dbsecurity/auth/Makefile
@@ -73,7 +73,7 @@ INCLUDES	= -I. -I./inc -I ../shared/inc \
 	        -I ../../sql/common
 
 
-LINK_OPTIONS	= -L$(LIBEXPDIR) -lldap -lssl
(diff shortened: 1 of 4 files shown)


DaveBirdsall:
Looks good to me.

PR #375, opened 2016-03-10
TRAFODION [1696] - drop authorization doesn't drop all roles, and create role will run into internal error

robertamarton:
A while back a change was made in the PrivMgr constructors to assume that
authorization is enabled by default instead of the unknown state.  If the state 
is unknown, privilege manager performs I/O to determine its state, otherwise
no additional checking is performed. This was changed because one, the 
authorization check in PrivMgr is expensive and two, the majority of the callers 
already perform the authorization check by looking in the compile context set up 
during process startup.  Role code was not updated to handle this change 
correctly as described in TRAFODION-1696.  Changes were made to check compiler 
context to verify that authorization is enabled  for role commands.


DaveBirdsall:
Looks good to me.

PR #376, opened 2016-03-11
[TRAFODION-1886] Region Server fails to come up at times with a NULL …

selvaganesang:
…pointer exception in

org.apache.hadoop.hbase.coprocessor.transactional.TrxRegionObserver class

This issue is likely to happen when trafodion install fails for some reason

### DIFF
--- a/core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/SplitBalanceHelper.java
+++ b/core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/SplitBalanceHelper.java
@@ -304,11 +304,13 @@ public class SplitBalanceHelper {
           List<String> trafTables = ZKUtil.listChildrenNoWatch(zkw, zSplitBalPathNoSlash);
           List<String> hbaseTables = ZKUtil.listChildrenNoWatch(zkw, SplitBalanceHelper.zkTable);
           
-          for(String tableName : trafTables) {
+          if(trafTables != null && hbaseTables != null) {
+            for(String tableName : trafTables) {
               if(!hbaseTables.contains(tableName)) {
-                  if(LOG.isTraceEnabled()) LOG.trace("zkCleanup, removing " + zSplitBalPath + tableName);
-                  ZKUtil.deleteNodeRecursively(zkw, zSplitBalPath + tableName);
+                if(LOG.isTraceEnabled()) LOG.trace("zkCleanup, removing " + zSplitBalPath + tableName);
+                   ZKUtil.deleteNodeRecursively(zkw, zSplitBalPath + tableName);
               }
+            }
           }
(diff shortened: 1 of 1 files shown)


core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/SplitBalanceHelper.java
```diff
@@ -304,11 +304,13 @@ protected void zkCleanup() {
           List<String> trafTables = ZKUtil.listChildrenNoWatch(zkw, zSplitBalPathNoSlash);
           List<String> hbaseTables = ZKUtil.listChildrenNoWatch(zkw, SplitBalanceHelper.zkTable);
           
-          for(String tableName : trafTables) {
+          if(trafTables != null && hbaseTables != null) {
```
traflm: There is a possibility that trafTables is not null but hbaseTables is null, or vice versa. In this case, the cleanup will leave something.
I am not quite clear, but I think this method is to clean up every leftover znodes. So the '&&' may not be proper here?

selvaganesang: trafTables should be subset of hbaseTables.  If it is just a plain hbase table, do you think it should be the responsibility of this class to cleanup the zk nodes. If so, we might need to re-factor this code. If not, this change should be fine though it is bit more defensive.

traflm: you are right Selva, thanks to help me understand this code.
It now looks good for me.

DaveBirdsall:
Looks good to me.

shangshengtung:
The trafTables in this case seems just a set of trafodion regions with ongoing split/balance (since online split/rebalance is allowed to proceed with concurrent active transactions), not all the trafodion regions. Transactional contexts need to be re-established after the split/balance. The fix will work fine. One minor exception is will we have a set of leftover trafTables zk nodes (/hbase/Trafodion/splitbalance) not been cleaned up after a system crash (the original code will do except it does not handle the hbaseTables == null case).

DaveBirdsall:
Is this ready for commit?

selvaganesang:
Yes. It is ready to commit

DaveBirdsall:
Committing it now

DaveBirdsall:
I committed this a couple of hours ago but oddly this pull request has not yet been closed. Steve thinks the github replication is behind for unknown reasons.

DaveBirdsall:
This particular merge got hung up in github replication for unknown reasons. Perhaps an event was dropped or missed somewhere along the way. I added a blank to one line and pushed the change again. That re-triggered the infrastructure and this time the merge happened. A useful technique to know when this happens again.

PR #378, opened 2016-03-11
[[TRAFODION 1884 ]] CDH 5.5 and HDP 2.4 install support

amandamoran:
Added check to make sure hbase-trx*.jar is copied to all nodes.

Installer support only for CDH 5.5 and HDP 2.4.


mkby:
Hi Amanda, should we also rename the script 'traf_hortonworks/cloudera_mods98' to 'traf_hortonworks/cloudera_mods' ? The postfix '98' is always confused.

amandamoran:
Yes we should. 

I will do that on a later check in. 

Thanks!

DaveBirdsall:
Looks good to me.

PR #381, opened 2016-03-13
JIRA 1849 Add support for alter column rename (details below)

anoopsharma00:
-- support for 'add column c rename to newc'
-- fixed issue with ddl stmts and aqr, it now works correctly
  if ddl_transactions are being used
-- drop table with views now drops them in reverse creation order.
  That way dependent views are dropped correctly.
-- views are dropped before and recreated after an alter column
  statement.
-- dev regressions tests now show system columns in invoke.
-- index uid is now stored in desc and NATable structs


core/sql/common/NAString.cpp
```diff
@@ -280,6 +280,29 @@ NAString Int64ToNAString(Int64 l)
   return NAString(resultstr);
 }
 
+NAString &replaceAll(NAString &source, const NAString &searchFor,
+                     const NAString &replaceWith)
+{
+  size_t indexOfReplace = NA_NPOS;
+  indexOfReplace = source.index(searchFor);
+  if (indexOfReplace != NA_NPOS)
```
DaveBirdsall: The "if" is unnecessary. If you delete it, the code behaves exactly as it would have otherwise.

core/sql/regress/charsets/EXPECTED012
```diff
@@ -120,11 +121,11 @@ tomorrow_suffix
...
...------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
DaveBirdsall: I take it the character set synthesis rules changed somewhere? The new results seem to be twice as long as the old.

core/sql/sqlcomp/CmpDescribe.cpp
```diff
@@ -2691,7 +2668,7 @@ short CmpDescribeSeabaseTable (
  
   // display syscols for invoke if not running regrs
```
DaveBirdsall: Comment needs updating?

DaveBirdsall:
Looks good to me. A question: Suppose I have table t1 with columns a and b, and t2 with columns a and b also. Suppose I have a view, "create view v1 as select * from t1 where b in (select b from t2)". Suppose I do an "alter table t1 alter column b rename c". Does the logic correctly figure out which "b" in the view text to rename? Another question: Suppose I have the tables t1 and t2 as above, and I have a view, "create view v2 as select t1 natural join t2". And suppose I do an "alter table t1 alter column b rename c". What happens with view v2?

core/sql/common/NAString.cpp
anoopsharma00: yes, that is true. 
This method was moved from sqlcomp/CmpDescribe.cpp to common dir
and is not new code.
Will fix it in a later checkin.

anoopsharma00:
Before rename, we save all view definitions on the table and drop those views. After rename, those views are recreated using the saved definition. If that recreation doesn't succeed, then rename operation fails. Users will have to drop and recreate the views with the right column name.
Right now, we dont automatically change column names in view or check constraint definitions.

If column 'b' in table t1 is renamed to 'c', then the create view will fail during recreation
since table t1 no longer has a column 'b'.

Here is a run with that scenario:

>>create view v1 as select * from t1 where b in (select b from t2);

--- SQL operation complete.
>>alter table t1 alter column b rename to c;

*** ERROR[4003] Column TRAFODION.SEABASE.T1.B is not a column in table TRAFODION.SEABASE.T1, or, after a NATURAL JOIN or JOIN USING, is no longer allowed to be specified with a table correlation name.

*** ERROR[1404] Column B cannot be altered. Reason: Error occurred while recreating views due to dependency on older column definition. Drop dependent views before doing the alter.

DaveBirdsall:
That makes perfect sense. Thanks!

DaveBirdsall:
Is this ready to commit?

PR #382, opened 2016-03-13
[TRAFODION-1854] rework

traflm:
rework of the previous fix, tested on a real cluster using upper case hostname.


DaveBirdsall:
Looks good to me. I take it you simply found other places that needed downshifting?

PR #383, opened 2016-03-13
[TRAFODION-1868] Compatibility with gcc 4.8 , part 2, DTM changes

traflm:
This is the second part of gcc 4.8 compatibility changes.
Related files are from DTM component.
Major fixed issues:
 - unused var
 - old C++ array init method
 - link options compatible for gcc 4.8
 - use NULL as integer, but NULL is a pointer


DaveBirdsall:
Looks good to me.

PR #385, opened 2016-03-14
TRAFODION-1885 : online expansion

trinakrug:
Enhancement to allow online expansion for new nodes.  Details in JIRA TRAFODION-1885.


narendragoyal:
Code looks good to me

core/sqf/monitor/linux/cmsh.cxx
```diff
@@ -127,11 +127,14 @@ int CCmsh::GetClusterState( PhysicalNodeNameMap_t &physicalNodeMap )
...
             {
-                // TEST_POINT: to force state down on node name 
+               // TEST_POINT and Exclude List : to force state down on node name 
                 const char *downNodeName = getenv( TP001_NODE_DOWN );
-                if ( downNodeName != NULL && 
-                    !strcmp( downNodeName, nodeName.c_str() ) )
-                {
+                const char *downNodeList = getenv( TRAF_EXCLUDE_LIST );
+                if (( downNodeList != NULL && 
+                    (strstr(downNodeList,nodeName.c_str()))) ||
```
DaveBirdsall: I suppose TRAF_EXCLUDE_LIST is a debugging tool. It occurs to me though that strstr() could return false positives, for example if there were two nodes  n001.x.y and another nn001.x.y. If nn001.x.y were in the list, strstr of n001.x.y would return true. But maybe we don't care about this for debugging purposes?

core/sqf/monitor/linux/pnode.cxx
```diff
@@ -1528,6 +1528,14 @@ void CNodeContainer::AddNodes( )
     int pnid;
     int rank;
     int *sparePNids = NULL;
+    
+    //  only relevant on a workstation acting as a single node
+    const char* envVar = getenv("SQ_MAX_RANK"); 
+    int maxNode;
+    if (envVar != NULL)
+      maxNode = atoi (envVar);
```
DaveBirdsall: Again, I suppose SQ_MAX_RANK is a debugging tool, so we don't bother checking for non-numeric inputs such as might happen with "export SQ_MAX_RANK=ON" (where atoi would return 0 I think).

core/sqf/monitor/linux/replicate.cxx
```diff
@@ -68,7 +68,7 @@ void CReplObj::validateObj()
 // Determine the maximum size of a replication object (excluding CReplEvent)
 int CReplObj::calcAllocSize()
 {
-    return  max(max(max(max(max(max(max(max(max(max(max(max(max(max(max(max(max(max(sizeof(CReplSoftNodeUp),
+    return  max(max(max(max(max(max(max(max(max(max(max(max(max(max(max(max(max(max(max(sizeof(CReplSoftNodeUp),
```
DaveBirdsall: The max'es are fun. Another way to do this is to create a union of all these objects, and take the sizeof the union. The C compiler then takes the max.

core/sqf/monitor/linux/pnode.cxx
trinakrug: This is a testing tool only.

core/sqf/monitor/linux/replicate.cxx
trinakrug: Yes, very fun :-)  I kept the code model as is with my change since I wasn't sure if there was a reason to do it this way.

core/sqf/monitor/linux/cmsh.cxx
trinakrug: It is not a debugging tool and we do not want false positives.  I will fix this.  Good catch!

core/sqf/monitor/linux/reqnodename.cxx
```diff
@@ -0,0 +1,114 @@
...
+    : CExternalReq(msgType, pid, msg)
+{
+    // Add eyecatcher sequence as a debugging aid
+    memcpy(&eyecatcher_, "RQEZ", 4);
+}
+
+CExtNodeNameReq::~CExtNodeNameReq()
+{
+    // Alter eyecatcher sequence as a debugging aid to identify deleted object
+    memcpy(&eyecatcher_, "rqez", 4);
```
DaveBirdsall: Nice technique

core/sqf/monitor/linux/shell.cxx
```diff
@@ -3069,6 +3072,103 @@ const char *zone_type_string( ZoneType type )
...
+              return;
+           }
+           //remove and read
+           PhysicalNodeMap.erase(current_name);
+           pnmit = PhysicalNodeMap.insert( PhysicalNodeNameMap_t::value_type 
+                                            ( newPhysicalNode->GetName(), newPhysicalNode ));
+           if (pnmit.second == false)
+           {   // Already had an entry with the given key value.  
+                printf( "[%s] Error: Internal error while changing node name. Node name exists, node name=%s\n", MyName, new_name );
+                return;
```
DaveBirdsall: Should we do a delete on newPhysicalNode in this code path? (Will it leak?)

DaveBirdsall:
Looks good to me. Might be a memory leak in one of the error paths.

DaveBirdsall:
Looks good to me. Will there be any other review from the DTM side?

trinakrug:
This is unrelated to the DTM, so there need not be any reviews from the DTM side.

PR #387, opened 2016-03-15
[TRAFODION-1890] disable SSCC entirely in current Trafodion

traflm:
SSCC is long time not maintained and test, and at the time of its writing, it is for HBase 0.98.6.
Remove it for now from install_local_hadoop will help in some times that HBase for no reason to fail to start and just remove it will make it work fine.
Will add it back after more tests of SSCC.


DaveBirdsall:
Looks good to me.

zellerh:
Looks good, thanks!

PR #388, opened 2016-03-15
TRAFODION-1883.Changes to drop LOB dependent tables during drop schema. 

sandhyasun:
The LOB dependent tables were getting dropped before the actual LOB table was getting dropped in the dropSchema method. Changed that to drop these objects in 2 sequences to avoid this problem.


DaveBirdsall:
Looks good to me.

PR #390, opened 2016-03-17
JIRA 1852: disallow reserved names as user columns

anoopsharma00:


core/sql/common/ComMisc.cpp
```diff
@@ -228,3 +228,17 @@ NAString ComConvertTrafNameToNativeName(
   return convertedName;
 }
 
+NABoolean ComTrafReservedColName(
+     const NAString &colName)
```
DaveBirdsall: I take it that by the time we get here, colName has already been case-shifted to upper case?

DaveBirdsall:
Looks good to me.

core/sql/common/ComMisc.cpp
anoopsharma00: If column is not delimited, then it will be upcased when this method is called.
If column is delimited ("a"), then it will retain the original case.

PR #391, opened 2016-03-17
JIRA 1899, Various fixes. Details below.

anoopsharma00:
-- time datatype with no fraction precision was not getting converted
   correctly and was putting out null characters.

-- dateformat usa was not displaying AM/PM with time datatype

-- update of primary key that got transformed into delete/insert
   would incorrectly delete the row if insert ran into an error

-- rename cascade option is not supported. An error is now returned.

-- rename table with check constraint is not supported.
   An error is now returned.

-- add/drop of primary key constraint now gives consistent behavior.
   Primary key columns are used as clustering key if primary key constraint
   is specified during create table.
   Primary key constraint creates a unique index if it is specified
   after create table and no user specified primary or store by clause
   was specified during create.
   Drop of primary key constraint succeeds only if that primary key
   is not the clustering key.

-- invalid ddl with salt clause no longer crashes and returns an error instead.

-- parser now gives a syntax error on MV and Triggers queries since
   these features are not externalized.

-- regress/seabase/TEST031 has been added for misc fixes


DaveBirdsall:
Looks good to me.

PR #393, opened 2016-03-18
[TRAFODION-1896] CQD TRAF_UPSERT_WITH_INSERT_DEFAULT_SEMANTICS for no…

selvaganesang:
…n-aligned format

This CQD is now honored for non-aligned format by replacing the column
value with default values for omitted columns and put into hbase table.


selvaganesang:
Hans, can you please review this PR

selvaganesang:
I wonder if this PR is ready to commit,  Are there any reviewers comment that needs attention?

selvaganesang:
I have resolved the conflicts. I hope to get this in before it gets diverged again from the master resulting in merge conflicts.

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -10165,22 +10165,35 @@ base table then the old version of the row will have to be deleted from
...
+          ((CmpCommon::getDefault(TRAF_UPSERT_MODE) == DF_MERGE) &&
+            // omitted current default columns with non-aligned row format tables
+            // or omitted default columns with aligned row format tables 
+            (((NOT isAlignedRowFormat) && omittedCurrentDefaultClassCols) ||
+            (isAlignedRowFormat && omittedDefaultCols))) ||
+          // CQD is set to Optimal, for non-aligned row format with omitted 
+          // current columns, it is converted into merge though it is not
+          // optimal for performance - This is done to ensure that when the 
+          // CQD is set to optimal, non-aligned format would behave like 
+          // merge when any column is  omitted
```
sureshsubbiah: I suppose the comment in line 10192 should say "when any current default column is omitted" ?

A table some place in code that shows what the behaviour is with 3 cqd settings for aligned omitted default, aligned omitted default current non-aligned omitted default, non-aligned omitted current default will be helpful.

sureshsubbiah:
Change looks good. New tests are detailed. Thank you. Thank you also for collecting feedback from the community and finding common ground on the semantics.

core/sql/optimizer/BindRelExpr.cpp
selvaganesang: What I intended to say is this:
Non-aligned format table usually behaves like merge though it is not changed into MERGE statement because Trafodion engine doesn't store the omitted default column in the raw hbase table.
When the upsert statement has omitted current default columns, it is changed into MERGE statement though it is not optimal because we want it to behave like above for this case also. Otherwise, it might be confusing to the application.

selvaganesang: I will add the table in the comments section from the message in user dlist when I get a chance to update this file again

PR #394, opened 2016-03-18
[TRAFODION-1868] Compatibility with gcc 4.8 , part 3, mornitor, odbc …

traflm:
this is part 3 of gcc 4.8 changes.
some changes are very strange, for example:
core/sqf/monitor/linux/monitor.cxx
it will call glibc function getrlimit(), which defined in sys/resource.h 
in gcc 4.4, this code can be compiled, which I don't know why. This is not what I have been taught in school. :-)  If a function is not declared, compiler should report error, but gcc 4.4 not. Maybe there are complex nested header file include, but I cannot trace it.
So simply for code readability, we should add it. I still don't understand why this code can be built without issue for so long time :-) 

After I understand why gcc 4.8 disable '--no-copy-dt-needed-entries'
https://wiki.ubuntu.com/NattyNarwhal/ToolchainTransition
I realize this is just a temp solution for gcc 4.8 build. In the following changes, I will disable it again and fix Makefile link options.
But first let's fix these .cpp and .h issues. Then on Makefile.

### DIFF
--- a/core/conn/odbc/src/odbc/nsksrvr/Makefile
+++ b/core/conn/odbc/src/odbc/nsksrvr/Makefile
@@ -130,6 +130,8 @@ LOC_JSIG=$(JAVA_HOME)/jre/lib/i386/server
 endif
 
 
+#gcc 4.8 need explicit enable this flag, otherwise ld must explicitly link with all so using '-l'
+GCCMODE        += -Xlinker --copy-dt-needed-entries
 # Produce mxosrvr
 $(BINEXPDIR)/mxosrvr: $(OBJS) $(LIBEXPDIR)/libzookeeper_mt.so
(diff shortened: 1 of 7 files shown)


core/conn/odbc/src/odbc/nsksrvr/Makefile
```diff
@@ -130,6 +130,8 @@ LOC_JSIG=$(JAVA_HOME)/jre/lib/i386/server
 endif
 
 
+#gcc 4.8 need explicit enable this flag, otherwise ld must explicitly link with all so using '-l'
```
traflm: This is NOT a change of gcc4.8, but the CentOS 7. CentOS 7 build gcc with '--no-copy-dt-needed-entries' OFF by default. Turn it on is a simple but not good change. I will come back later when all .cpp/.h changes are merged.

DaveBirdsall:
Looks good to me.

PR #395, opened 2016-03-18
[TRAFODION-1900] Optimize MDAM scans with small scanner

eowhadi:
When doing MDAM scans, we are performing interlaced scan for PROBE and for real scan. The probes always return only 1 row, then we close the scanner immediately, therefore should use always small scanner. I will make it conditional on the existing CQD HBASE_SMALL_SCANNER (ether SYSTEM or ON). In addition, caching of blocks retrieved by probe should always we at least receiving one succesfull cache hit on the next MDAM scan, therefore forcing caching ON for MDAM prob is a good idea. Again, will make this forcing conditional on HBASE_SMALL_SCANNER SYSTEM or ON.
Then for the real scan part of MDAM, I will use the following heuristic: If previous scan fitted in one hbase block, then it is likelly than next will also fit in one hbase block, therefore enable small scanner for next scan. Again all this only if CQD above is ON or SYSTEM.
Also includes a fix where SMALL_SCANNER would be turned on for MDAM scan because the compiler for MDAM is not polulating the expected number of rows returned.
Results of using small scanner on MDAM when it make sense showed a 1.39X speed improvement...

### DIFF
--- a/core/sql/executor/ExHbaseSelect.cpp
+++ b/core/sql/executor/ExHbaseSelect.cpp
@@ -445,6 +445,15 @@ ExWorkProcRetcode ExHbaseScanSQTaskTcb::work(short &rc)
   rc = 0;
   Lng32 remainingInBatch = batchSize_;
   NABoolean isFirstBatch = false;
+  // isFirstInBatch is a stack variable for optimization reason. It is used for the mdam small scanner optimization heuristic that
+  // is performed at runtime. Since this function is invoke intensively for all scan (mdam or regular scan), minimizing CPU/memory access
+  // impact on runtime code to a strict minimum is attempted. Given that we are trying to detect if the actual scan is bellow the size
+  // of an HBase block, having the runtime logic performing the detection only affect the first work invoke looks like the right idea.
+  // and leveraging an existing counter (remainingInBatch) instead of creating a new one. The reasonable asumption to allow this is that
+  // 1- batchSize_ being 8K, most likely times the row size, we are good in assuming that first hbase block will fit in batchSize
+  // 2- parent buffer size will be large enough to deal with one HBAse_Block_size without having to rely on re-invoking work in the middle.
+  // and anyway, if none of the reasonable assumption is true, then that's fine, the heuristic won't work, and we will use regular scanner,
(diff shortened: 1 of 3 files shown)


core/sql/executor/ExHbaseSelect.cpp
```diff
@@ -444,6 +444,7 @@ ExWorkProcRetcode ExHbaseScanSQTaskTcb::work(short &rc)
   Lng32 retcode = 0;
   rc = 0;
   Lng32 remainingInBatch = batchSize_;
+  NABoolean isFirstBatch = false;
```
DaveBirdsall: Do you want this to be a stack variable? The work method might return at any time (example: when the parent queue fills up) and then get called again, which would cause this variable to become false again. I'm wondering if this variable should be a member instead?

eowhadi: I think you are right , I thought the only exit would be due to batch size exceeding, and was seeing batch size =*K, so was good. But now I see  	    if (tcb_->moveRowToUpQueue(tcb_->convertRow_, tcb_->convertRowLen_, 
 				       &rc, FALSE))		 				       &rc, FALSE))
 	      return 1;		 	      return 1;

I am guessing this is what you are talking about (parent queue fill up). Can this happen realistically before the first 8K?

DaveBirdsall: It depends on the queue size. Which is set by the compiler and can be modified by the executor. But just as a general design principle, work methods are designed to interrupt their work at any time (often due to some resource constraint) so it is best to use members for any state that is needed for the life of the operator.

eowhadi: OK so here this boolean is not needed for the full length of the operator. Only for the first 8K rows.
I can change the logic and maintain a row counter on the tcb. That will have impact memory write for the full length of the operator. By doing my initial trick, I was hopping that I could leverage the existing remainingInBatch counter and not create another one. But I guess if I cannot trust that the parent queue can fill up with less than an HBase_Buffer_size, I cannot use the optimization.

DaveBirdsall:
Looks good to me.

PR #396, opened 2016-03-18
TRAFODION [1895] - Unable to grant and revoke privileges on native hive tables

robertamarton:
There are actually two issues:
1 - when an external table is created, the wrong role is assigned ownership 
    which makes it impossible to grant privileges.
2 - Query invalidation is not removing external tables from caches after revoke
    completes; therefore, the user/role still has permissions until a restart

Also, added code to initialize authorization to set hive schema ownerships and
dependent objects to DB__HIVEROLE.  This fixes the issue of ownership. For
existing installations that has authorization enabled, run initialize 
authorization again. This acts as an upgrade and will fix the owner issue.

Removed Row.java, BaseRow.java and InsertRow.java from T2 driver code - these
are obsolete and their removal fixes a licensing issue.  T4 driver code is
not touched since it still uses these files.


DaveBirdsall:
Looks good to me.

PR #400, opened 2016-03-23
[TRAFODION-1720] rework, fix a bug in it

traflm:
It was found previous fix will cause abort if the src contains 'full-width' chars. The guess is the pointer length is not enough. So here, I simulate the same behavior as UTF8 before.
 asciiType =  new (h) SQLVarChar(sizeof(Int64) ... 
that means the max length is 8, not 4. This is a pointer to real data not the data itself.

With this change I test with:
ＲＥＡＬ
and it works well.


DaveBirdsall:
Looks good to me.

PR #402, opened 2016-03-25
[TRAFODION-1617]

narendragoyal:
Add a few convenience scripts (and associated helper scripts)

Note: All the scripts with an "ilh_" prefix work if one has used install_local_hadoop

ilh_trafinit: ckillall, restarts HBase(clobbers HBase data), sqstart and initializes trafodion

ilh_traf_restart: ckillall, restarts HBase, sqstart

trafnodestatus: Prints the trafodion-status (whether Trafodion is UP or DOWN on a node)
		of the nodes in the Trafodion cluster. Has a '-j' switch to print the
 		output in JSON format.
		Note: There is another script 'sqnodestatus' with a similar output but that
		prints the status of the nodes from a Linux perspective.


DaveBirdsall:
Looks good to me. Thanks for adding these.

core/sqf/sql/scripts/ilh_traf_restart
```diff
@@ -0,0 +1,117 @@
...
+if [ -e $MY_SQROOT/sql/scripts/hbcheck ]; then
+    run_util hbcheck 4 10
+else
+    run_util ilh_hbcheck
+fi
+echo "--------------------------------------"
+
+echo "Repairing HBase (hbase hbck -repair)"
+ilh_hbase_repair y
+
```
amandamoran: Can you explain, why a hbase -repair is needed? Is this only needed for install_local_hadoop but not in a clusters environment?

narendragoyal: On a single node env (with HMaster, RegionServer, Zookeeper in a single process), I saw instances where the HMaster would get stuck at startup. Restarting HBase multiple times wouldn't help but executing 'hbase hbck -repair' somehow took care of the issue. I am not sure what it cleaned up in the HBase space. 

So, I wrapped this in a little script. Currently, I would not suggest that 'ilh_hbase_repair' be used in a clustered/production env by itself.

PR #403, opened 2016-03-25
TRAFODION [1902] - Showddl continues to show function after it's dropped

robertamarton:
Fixed a problem during a drop library request, where functions and procedures
were being dropped but not removed from DBRoutineCache.
Added a new regression test (udr/TEST103) that tests DDL operations related to 
functions, procedures, and libraries.  The regression tests were missing.


DaveBirdsall:
Looks good to me.

anoopsharma00:
Change is good. 

Method NARoutineDB::removeNARoutine need to be enhanced the same
way as method NATableDB::removeNATable to handle ddl xns.
We can commit this change and then make the other change for ddl xns.

PR #405, opened 2016-03-29
TRAFODION-1910 mxosrvr crashes on Hive query after reconnect

zellerh:
NATableDB is caching a pointer to a HiveClient_JNI object
(HiveMetaData::client_), but that object gets deallocated when a JDBC
client disconnects.  Fixing this by keeping the HiveClient_JNI around
across sessions.

### DIFF
--- a/core/sql/cli/CliExtern.cpp
+++ b/core/sql/cli/CliExtern.cpp
@@ -6316,7 +6316,8 @@ Lng32 SQL_EXEC_DeleteHbaseJNI()
       threadContext->incrNumOfCliCalls();
 
       HBaseClient_JNI::deleteInstance();
-      HiveClient_JNI::deleteInstance();
+      // The Hive client persists across connections
+      // HiveClient_JNI::deleteInstance();
    }
    catch(...)
    {


core/sql/cli/CliExtern.cpp
```diff
@@ -6316,7 +6316,8 @@ Lng32 SQL_EXEC_DeleteHbaseJNI()
       threadContext->incrNumOfCliCalls();
 
       HBaseClient_JNI::deleteInstance();
-      HiveClient_JNI::deleteInstance();
+      // The Hive client persists across connections
+      // HiveClient_JNI::deleteInstance();
```
DaveBirdsall: Is there any security issue here? If we integrate with Hive security (and I don't know if we have or not) is there some notion of re-authentication at connection time?

zellerh: Good question, and I think any security issues are (or need to be) handled in the HiveMetaData object, which already persists between connections. The object that now also persists is HiveClient_JNI (and the corresponding Java object). Those don't keep a cache or state about Hive metadata, they just have methods to read and validate Hive metadata. So, in short, no, I don't believe there is a security concern with this change.

DaveBirdsall:
Looks good to  me.

zellerh:
Actually, there is one aspect that we probably need to consider. If we want to pass the Trafodion database user id on to the HiveClient_JNI object and the Java org.trafodion.sql.HiveClient object, then we should do this for every call, not during the init() method. That would allow us to keep the object across connections. Doing so would only make sense if Trafodion and Hive share the same user id space, like some space defined by Kerberos. Thanks again for getting me to think more about this.

core/sql/cli/CliExtern.cpp
selvaganesang: Whenever a context is dropped, both hbaseClient_JNI and hiveClient_JNI is getting deleted. This CLI call is being called from the two places shown below. I am not sure about the need for this calls in these functions. If it is needed, I would think there would be similar problem with HBaseClient_JNI::deleteInstance too. Or is it better to get rid of this CLI call.

arkcmp/CmpStatement.cpp:  SQL_EXEC_DeleteHbaseJNI();
sqlcomp/CmpSeabaseDDLcommon.cpp:  SQL_EXEC_DeleteHbaseJNI();

In case of mxosrvr, there is default context. In case of T2 driver, the CliContext would be deallocated and these objects would be deleted. I am assuming that it should be ok to do that.

sureshsubbiah: I would echo Selva's comment. Can you please explain how the HiveClient_JNI object is different from HBaseClient_JNI? I had tried to model HiveClient_JNI after HBaseClient_JNI and would like check if that assumption is incorrect in other places too.

zellerh: Sorry, I added a couple of comments to the JIRA, TRAFODION-1910 (https://issues.apache.org/jira/browse/TRAFODION-1910), should have added them here.

Let me try to summarize: The comments I made try to explain that we _could_ treat HiveClient_JNI and HBaseClient_JNI differently. Suresh is asking why we _should_ do that.

So, I think what you are telling me is that I should write a new fix that treats the two objects in a similar way. Hope to have that ready soon. Thanks for your input.

selvaganesang:
My thinking is that we should remove calling SQL_EXEC_DeleteHbaseClient_JNI and this CLI method. It is not clear to me why should this be done outside of SQL engine via a CLI call. It might have been some residue left behind from the earlier versions of the code. 

If we remove this call, NATableDB should be able to use this across connections. NATableDB would be shared across connections if the same user comes into the mxosrvr. I believe NATableDB and other structures would be deallocated, if a different user comes into mxosrvr to take care of privileges.

zellerh:
Yes, that sounds good to me. As far as I understand, this would mean that we treat both HBaseClient_JNI and HiveClient_JNI the same and that we would keep both across JDBC connections (assuming the same user id). That could give us a small performance boost when applications use a lot of short connections.

selvaganesang:
Thanks Hans. Can you please remove the CLI too so that it is not used in
future.



Selva

*Sent:* Wednesday, March 30, 2016 12:00 PM
*Cc:* selvaganesang <selva.govindarajan@esgyn.com>
*Subject:* Re: [apache/incubator-trafodion] TRAFODION-1910 mxosrvr crashes
on Hive query after reconnect (#405)



Yes, that sounds good to me. As far as I understand, this would mean that
we treat both HBaseClient_JNI and HiveClient_JNI the same and that we would
keep both across JDBC connections (assuming the same user id). That could
give us a small performance boost when applications use a lot of short
connections.

—
You are receiving this because you commented.
<https://github.com/apache/incubator-trafodion/pull/405#issuecomment-203583076>

zellerh:
Thanks, will do, and likely submit a new pull request and close this one.

zellerh:
Closing this one, submitted a new request, https://github.com/apache/incubator-trafodion/pull/408.

PR #406, opened 2016-03-29
[TRAFODION-1823] ESP idle timeout does not kick in, leading to too ma…

selvaganesang:
…ny ESPs on the system

Close message from the master was not sent to ESP by the platform
agnostic messaging layer in Trafodion. Fixed this bug and then ESP idle time
works as expected.

ESP_IDLE_TIMEOUT can now be given as a CQD. Internally, it would be
changed to SET SESSION DEFAULT and used by the IPC layer of Trafodion.
ESP_IDLE_TIMEOUT is set to 30 minutes by default in the CQD too.

Added test in core/TESTRTS to test ESP_IDLE_TIMEOUT


prashanth-vasudev:
This change looks good to me. Thanks.

sandhyasun:
Changes to propagate the value as  CQD look good too . Will merge this shortly.

PR #407, opened 2016-03-30
Fixed regression failure for udr/TEST103.  Added a FILTER and added the  missing Utils.java file.

robertamarton:


svarnau:
looks good to me

PR #408, opened 2016-03-31
TRAFODION-1910 mxosrvr crashes on Hive query after reconnect (take 2)

zellerh:
NATableDB is caching a pointer to a HiveClient_JNI object
(HiveMetaData::client_), but that object gets deallocated when a JDBC
client disconnects. Fixing this by keeping the HiveClient_JNI around
across sessions.

Selva and Suresh commented on the first fix and suggested to treat both
HBaseClient_JNI and HiveClient_JNI the same and to remove the CLI
interface that's used to delete these objects.

Therefore, the new fix is to remove this CLI call. It gets called from
two places, one is when an ODBC/JDBC connection closes and the other
is from "initialize trafodion, drop". We believe that neither of them
is needed. Note that we have only one object of each type per CLI
context, and that we delete both objects when we delete the context
(ContextCli::deleteMe()), so there are no leaks.

### DIFF
--- a/core/sql/arkcmp/CmpStatement.cpp
+++ b/core/sql/arkcmp/CmpStatement.cpp
@@ -1311,8 +1311,6 @@ CmpStatement::process (const CmpMessageEndSession& es)
       CURRENTQCACHE->makeEmpty();
     }
 
-  SQL_EXEC_DeleteHbaseJNI();
-
   return CmpStatement_SUCCESS;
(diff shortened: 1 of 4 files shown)


DaveBirdsall:
Looks good to me. "Delete code until it works!"

selvaganesang:
Good to go. I will merge this in

PR #414, opened 2016-04-05
[TRAFODION-1912]support automatic convert DOS format during bulkload

traflm:
Introduce a new CQD : HIVE_SCAN_SPECIAL_MODE
which is a bitmap, to allow different 'tolerance' suggested by JIRA 1911.
This patch is for DOS format handling. When HIVE_SCAN_SPECIAL_MODE set to 1, HIVE_SCAN will automatically replace \r\n to space\n

Please help to review.


core/sql/executor/ExHdfsScan.cpp
```diff
@@ -367,6 +367,9 @@ ExWorkProcRetcode ExHdfsScanTcb::work()
   char cursorId[8];
   HdfsFileInfo *hdfo = NULL;
   Lng32 openType = 0;
+  int hiveScanMode = 0;
+
+  hiveScanMode = CmpCommon::getDefaultLong(HIVE_SCAN_SPECIAL_MODE);
```
selvaganesang: Usually, CQDs are not accessed directly in executor operators directly. It should be passed as a flag in the TDB. It is possible that the query is compiled in a different process, then this CQD setting won't be available in the executor layer.

selvaganesang: Please get this CQD passed as an attribute in TDB

traflm: thanks Selva, this explains my puzzle. In the hive regress test, I have to 
cqd CALL_EMBEDDED_ARKCMP 'ON';
It must be due to the reason you mentioned, here, I will add this as a TDB flag.

core/sql/executor/ExHdfsScan.h
```diff
@@ -429,41 +432,92 @@ class ExOrcFastAggrTcb  : public ExOrcScanTcb
...
+   }
+  }
+  else
+  {
+   while (curr < end) {
+     if (*curr == c)
+     {
+         if(count>0 && c == '\n')
+         {
+           if(s[count-1] == '\r') s[count-1] = ' ';
```
selvaganesang: What happens if the column happens to be numeric column.  Is the space character tolerated in the conversion

core/sql/executor/ExHdfsScan.h
```diff
@@ -429,41 +432,92 @@ class ExOrcFastAggrTcb  : public ExOrcScanTcb
...
-       *rdSeen = FALSE;
-       return curr;
+  int count = 0;
+  if( (mode & HIVE_MODE_DOSFORMAT)>0 )  //check outside the while loop to make it faster
+  {
+    while (curr < end) {
+      if (*curr == rd) {
+         if(count>0 && rd == '\n')
+         {
+             if(s[count-1] == '\r') s[count-1] = ' ';
```
selvaganesang: Same comment as above

core/sql/executor/ExHdfsScan.h
traflm: I tested with DATE and INT, both works well with extra space. But maybe \0 is better. Let me try it out.

traflm: \0 will not work, I will keep it as space.

traflm:
jenkins, retest

traflm:
If the last column is a NULL, this change will add a space. Need to fix it again...

traflm:
I changed the implementation: not replace the \r with \n, but return back the previous pointer before \r. And in the caller, add that extra position when it needs to move to next field.
Test with INT/DATE/String, all works well, also test with last column as null, works fine too.

selvaganesang:
Change looks good to me. Can you please modify the new test to insert into trafodion table with date as a last column. Also add another test for numeric column as the last column and insert into trafodion table too.

traflm:
Thanks Selva, I will add the test for numeric/date data

PR #415, opened 2016-04-05
[TRAFODION-1917] Trafodion fails to build HDP related jar files

mashengchen:


core/sqf/sql/scripts/install_local_hadoop
```diff
@@ -614,7 +614,7 @@ HIVE_TAR=${HIVE_PREFIX}.tar.gz
 HBASE_MIRROR_URL=http://archive.cloudera.com/cdh5/cdh/5
 HBASE_TAR=hbase-${HBASE_DEP_VER_CDH}.tar.gz
 if [[ "$HBASE_DISTRO" = "HDP" ]]; then
-    HBASE_TAR=hbase-${HBASE_DEP_VER_HDP}.tar.gz
+    HBASE_TAR=hbase-1.1.2.2.3.2.0-2950.tar.gz
```
selvaganesang: Prashanth brought this up already. I suggested that It would be better to create a soft link is created at the local repository to point to the HBASE_DEP_VER_HDP rather than changing the install_local_hadoop.

mashengchen: agree with that, but how to decide the version for HBASE_DEP_VER_HDP, if it's a simplified version like 1.1.2, how would a new trafodion user know the actually version. or always use a fix version, like 1.1.2.2.3.2.0-2950?

selvaganesang:
The customers are expected to install hadoop using the chosen distro processes. They can choose any specific version of 1.1.2 because Trafodion is expected to work with any version starting with 1.1.2. Install_local_hadoop is meant for contributing organization like Esgyn or contributors. This is supposed to point to the local repository. In the local repository we can always soft link a more specific version to the major version like 1.1.2.

prashanth-vasudev:
The right thing to do is not hard code the tar file name instead create a softlink to tarfile wherever the tar file is stored on the system. So no additional change is required.

This additional change can be abandoned.

PR #416, opened 2016-04-05
[TRAFODION-1918] Remove disabling auto flush concept in Trafodion

selvaganesang:
The auto flush is on by default in insertRows method. But insertRows was
incorrectly called with the wrong parameter. In fixing this issue, it was
observed that auto flush is deprecated in HBase 1.x. It is also incorrect
to disable auto flush in Trafodion. Hence removed the code related to
auto flush, flush table and flush all tables in Trafodion.

All these are dormant code in Trafodion.

Also disabled executor/TEST106 for now because it seems to cause a hang
at the time of drop table.


selvaganesang:
jenkins, retest

DaveBirdsall:
Looks good to me.

sureshsubbiah:
Looks good. I suppose TEST106 is being disabled for some other reason? Merging...

selvaganesang:
TEST106 was disabled because of hang issue in drop statement. Prashanth will create a JIRA and work on it.

PR #417, opened 2016-04-05
[TRAFODION-1473] T4 support for new sqltypecode BLOB(-602) and CLOB(-…

arvind-narain:
…603)

Support in T2 and odbc drivers will be addressed in later checkin.
Automated tests for drivers will be checked in at that stage.

Please refer to LOB Interface document in the jira for ways to extract/update data from such columns. set/get lob interfaces will be supported in the next release.


hegdean:
Changes looks good to me

sandhyasun:
Looks like this is ready to commit.

PR #420, opened 2016-04-07
various fixes

anoopsharma00:
-- Implicit xns for ddl stmts are now started in master exe if
   autocommit is OFF
-- default values for varchar columns are now handled correctly
-- DATEFORMAT function on timestamp values are now handled correctly
-- CREATE/DROP PROCEDURE stmt are now handled correctly with DDL xns.
-- alter primary key stmt now returns an error if table already has a
   user defined primary/clustering key
-- regress/core/TEST116, seabase/TEST030,031 have been updated with new tests


anoopsharma00:
jenkins retest

svarnau:
Anoop, your commits almost always come from centos user.  That stays in the code history forever. It would be helpful that you could set your git user info when you set up new VM, before making commits. Thanks.

selvaganesang:
Can you please link [TRAFODION-1864] LONGWVARCHAR column support have problems with storage of the column in hbase and uncomment the tests in charsets/TEST012

anoopsharma00:
With the code fix, this charsets/TEST012 now passes without any known differences.
Is there something else that need to be uncommented in TEST012? Or did you mean
to remove the known diff file?

anoopsharma00:
varchar default values pbm is being tracked by JIRA [TRAFODION-1864]

core/sql/generator/GenExpGenerator.cpp
```diff
@@ -753,12 +752,42 @@ void ExpGenerator::copyDefaultValues(
...
-	  str_cpy_all(tgtDefVal, 
-		      srcAttr->getDefaultValue(), 
-		      srcAttr->getDefaultValueStorageLength());
+          Lng32 tgtDefLen = tgtAttr->getDefaultValueStorageLength();
+          Lng32 srcDefLen = srcAttr->getDefaultValueStorageLength();
+          char* srcDefVal = srcAttr->getDefaultValue();
+	  char * tgtDefVal = new(generator->getSpace()) char[tgtDefLen];
+          // if source and target def storage lengths dont match, then
+          // need to move each part (null, vclen, data) separately.
+          if (tgtDefLen != srcDefLen)
```
DaveBirdsall: Might we be fooled into thinking the tgtDefLen is equal to the srcDefLen, but src is nullable varchar(n-2) while tgt is non-nullable varchar(n)?

selvaganesang:
Thanks Anoop. Looks like I created the known diff file. If charsets/TEST012 passes without the diff file, then the issue is fixed

DaveBirdsall:
With one possible exception (see inline comment), the changes look good to me.

anoopsharma00:
jenkins retest

PR #421, opened 2016-04-07
test pom.xml.apache repo change

venkat1m:


zellerh:
Sounds good, if that helps. Is this a temporary change or should we permanently move to the new repo?

venkat1m:
The https://repo1.maven.org/maven2 is the commonly used central maven repository. So I would recommend this be the permanent repo.

PR #422, opened 2016-04-11
[TRAFODION-1421] Implement parallel scanner primitive

eowhadi:
new CQD: hbase_dop_parallel_scanner
- default 0.0 = disabled
- from 0.x to 1.0 -> the DOP will be dynamically assessed as a percentage of region to scan. If 10 region, and cqd is .3 (30%), then 3 threads will be used...
- from 2.0 and above, will take ceiling, and use it as fixed number of threads.
Advantage of parallel scanner parallelism over ESP is on resource consumption. Parallel scanner is just creating threads, while ESP are full proceses wth high memory consumption.
For now, this feature is just a primitive that can be manually exercised. The next obvious step is to make the compiler aware of it and pick it over ESP DOP when appropriate...


eowhadi:
BTW, I ran the full regression test with forcing the new CQD to 2.0 in sbdefs file and verified all the FAILS where understood and normal due to new line showing up on explain. The regular regression still does not have this cqd set as this is not maint to be one by default but only activated on a case by case basis, since there is no compiler support for this feature yet.

selvaganesang:
I am seeing that we instantiate HTable which is marked as @InterfaceAudience.Private.

Also, the note in the class says that "HTable is no longer a client API. Use Table instead. It is marked InterfaceAudience.Private indicating that this is an HBase-internal class as defined in Hadoop Interface Classification There are no guarantees for backwards source / binary compatibility and methods or class can change or go away without deprecation".

Are we ok with it?

eowhadi:
well, we have been OK so far to use non public API. Yes it causes trouble, it is one more. Unfortunately.
Given that this feature is off by default, worse case, we don't enable it...

eowhadi:
jenkins retest

core/sqf/src/seatrans/.gitignore
```diff
@@ -12,3 +12,6 @@ hbase-trx/src/main/java/org/apache/hadoop/hbase/regionserver/transactional/Clean
 hbase-trx/src/main/java/org/apache/hadoop/hbase/regionserver/transactional/KeyValueListScanner.java
 hbase-trx/src/main/java/org/apache/hadoop/hbase/regionserver/transactional/MemoryUsageChore.java
 hbase-trx/src/main/java/org/apache/hadoop/hbase/regionserver/transactional/TrxTransactionState.java
+hbase-trx/src/main/java/org/apache/hadoop/hbase/client/ClientScanner98.java
```
DaveBirdsall: Is the name ClientScanner98 significant? For example, does it work only on HBase 0.98? Is there a different implementation needed for HBase 1.0?

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/ClientScanner98.java.tmpl
```diff
@@ -0,0 +1,490 @@
...
+    protected boolean closed = false;
+    // Current region scanner is against.  Gets cleared if current region goes
+    // wonky: e.g. if it splits on us.
+    protected HRegionInfo currentRegion = null;
+    protected ScannerCallable callable = null;
+    protected final LinkedList<Result> cache = new LinkedList<Result>();
+    protected final int caching;
+    protected long lastNext;
+    // Keep lastResult returned successfully in case we have to reset scanner.
+    protected Result lastResult = null;
```
DaveBirdsall: Should the close method reset lastResult to null to aid in garbage collection?

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/ClientScanner98.java.tmpl
```diff
@@ -0,0 +1,490 @@
...
+   */
+  @Deprecated
+  public ClientScanner98(final Configuration conf, final Scan scan, final TableName tableName,
+      HConnection connection, RpcRetryingCallerFactory rpcFactory) throws IOException {
+    this(conf, scan, tableName, connection, rpcFactory, RpcControllerFactory.instantiate(conf));
+  }
+
+  /**
+   * Create a new ClientScanner for the specified table Note that the passed {@link Scan}'s start
+   * row maybe changed changed.
```
DaveBirdsall: Typo in the comments ("changed changed")

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/ClientScanner98.java.tmpl
```diff
@@ -0,0 +1,490 @@
...
+        byte [] stopRow = scan.getStopRow();
+        int cmp = Bytes.compareTo(stopRow, 0, stopRow.length,
+          endKey, 0, endKey.length);
+        if (cmp <= 0) {
+          // stopRow <= endKey (endKey is equals to or larger than stopRow)
+          // This is a stop.
+          return true;
+        }
+      }
+      return false; //unlikely.
```
DaveBirdsall: Why is false unlikely? Are we past the end key almost all the time? (Seems counter-intuitive.)

core/sqf/src/seatrans/.gitignore
eowhadi: In 1.0 (actually after the fact I discovered that early 1.0 also didn't have the feature), and even in a latter version of 98, hbase introduced replica RS. This feature makes the client scanner multithread and not working with parallel scanner. So I froze an old version of non multi thread scanner, that does not support replica to be used with parallel scanner. That is ClientScanner98. It works with our 1.x HBase version.

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/ClientScanner98.java.tmpl
eowhadi: the whole clientScanner98 will be de-referenced, to GC will get rid of lastResult as it is a member of Clientscanner98. This code is quazi copy past from existing client scanner code. So would like to minimize changes to strick minimum (make sure it does compile with newer hbase version...)

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/ClientScanner98.java.tmpl
eowhadi: I did not try to understand or alter the ClientScanner code. So I don't know about this comment. This code comes from earlier version of ClientScanner Hbase, just renamed 98 and adding missing overrides to comply with newer abstract class in hbase 1.0

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/ClientScanner98.java.tmpl
```diff
@@ -0,0 +1,490 @@
...
+              // invocation.
+              skipFirst = true;
+            }
+            if (e instanceof OutOfOrderScannerNextException) {
+              if (retryAfterOutOfOrderException) {
+                retryAfterOutOfOrderException = false;
+              } else {
+                // TODO: Why wrap this in a DNRIOE when it already is a DNRIOE?
+                throw new DoNotRetryIOException("Failed after retry of " +
+                  "OutOfOrderScannerNextException: was there a rpc timeout?", e);
```
DaveBirdsall: Well, it does make for entertaining debugging :-)

DaveBirdsall:
Looks good to me. I'm ready to commit if the author is also happy with it.

eowhadi:
I am happy :-)

PR #423, opened 2016-04-11
Trafodion 1928 Trafodion daily build regression in CDH distro fails… … in hive tests

selvaganesang:


svarnau:
Looks good to me.

PR #425, opened 2016-04-11
[TRAFODION-1924] Add license/notice files to jdbc driver jars

svarnau:
This will allow distribution of jars directly via maven repos.

Also fix build-time sqenvcom.sh issue. The default empty variables
of APACHE_*_HOME caused existence check of /lib directory, which of
course always exists. Changed default location to "None", so we do
not falsely think we have hbase and query it for classpath.

Removed error checks, since it is fine in the build case to not have those
components.


DaveBirdsall:
Looks good to me.

PR #427, opened 2016-04-12
Phoenix T2 tests failing with CDH

arvind-narain:
Partial-Bug: [TRAFODION-1929]
Temporary disable running tests with maven till root cause is known.


selvaganesang:
Merging this. I believe these tests will be run but not via maven.  We also should consider running a subset of T2 tests as part of PR check so that these issues are caught by the contributor

PR #430, opened 2016-04-13
[TRAFODION-1686][MTDCS] ODBC: SQLParamdata ran into error for date ty…

xwq:
…pe columns

Windows ODBC need the maxLen is exactly equal to the real length. So
update the maxLen in jdbc t2.

### DIFF
--- a/core/conn/jdbc_type2/native/SqlInterface.cpp
+++ b/core/conn/jdbc_type2/native/SqlInterface.cpp
@@ -246,14 +246,10 @@ SQLRETURN GetJDBCValues( SQLItemDesc_def *SQLItemDesc,
 		SQLItemDesc->signType      = TRUE;
 		break;
 	case SQLTYPECODE_DATETIME:
-		// Calculate the length based on YYYY/MM/DD HH:MM:SS.ffffff
-        SQLItemDesc->maxLen = 20 + 6;
 		switch (SQLItemDesc->datetimeCode)
 		{
 		case SQLDTCODE_DATE:
 			SQLItemDesc->ODBCDataType = SQL_TYPE_DATE;
-            // For DATE, update the maxLen since it only have YYYY/MM/DD
-            SQLItemDesc->maxLen = 10;
 			break;
 		case SQLDTCODE_TIME:
 			SQLItemDesc->ODBCDataType = SQL_TYPE_TIME;
--- a/core/conn/jdbc_type2/native/SqlInterface.cpp
+++ b/core/conn/jdbc_type2/native/SqlInterface.cpp
@@ -246,10 +246,14 @@ SQLRETURN GetJDBCValues( SQLItemDesc_def *SQLItemDesc,
 		SQLItemDesc->signType      = TRUE;
 		break;
 	case SQLTYPECODE_DATETIME:
+		// Calculate the length based on YYYY/MM/DD HH:MM:SS.ffffff
+        SQLItemDesc->maxLen = 20 + 6;
 		switch (SQLItemDesc->datetimeCode)
 		{
 		case SQLDTCODE_DATE:
 			SQLItemDesc->ODBCDataType = SQL_TYPE_DATE;
+            // For DATE, update the maxLen since it only have YYYY/MM/DD
+            SQLItemDesc->maxLen = 10;
 			break;
 		case SQLDTCODE_TIME:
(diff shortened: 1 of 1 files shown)


core/conn/jdbc_type2/native/SqlInterface.cpp
```diff
@@ -246,10 +246,14 @@ SQLRETURN GetJDBCValues( SQLItemDesc_def *SQLItemDesc,
...
+        SQLItemDesc->maxLen = 20 + 6;
 		switch (SQLItemDesc->datetimeCode)
 		{
 		case SQLDTCODE_DATE:
 			SQLItemDesc->ODBCDataType = SQL_TYPE_DATE;
+            // For DATE, update the maxLen since it only have YYYY/MM/DD
+            SQLItemDesc->maxLen = 10;
 			break;
 		case SQLDTCODE_TIME:
 			SQLItemDesc->ODBCDataType = SQL_TYPE_TIME;
```
DaveBirdsall: Should we add code to calculate maxLen for TIME as well? (It would be just HH:MM:SS)

xwq: Maybe we need update all the maxLen for different datetime type. However I need write some unit test case to make it work well for each datetime type.

  So, for this PR, I only update SQLDTCODE_DATE for this issue .

  If you think we need update all the maxLen, we can keep this PR here , I will take some days to test it and update this PR later.

DaveBirdsall: Doing it later in a separate PR is fine. Might be good to write a separate JIRA to keep track of it.

core/conn/jdbc_type2/native/SqlInterface.cpp
```diff
@@ -246,10 +246,14 @@ SQLRETURN GetJDBCValues( SQLItemDesc_def *SQLItemDesc,
 		SQLItemDesc->signType      = TRUE;
 		break;
 	case SQLTYPECODE_DATETIME:
+		// Calculate the length based on YYYY/MM/DD HH:MM:SS.ffffff
+        SQLItemDesc->maxLen = 20 + 6;
```
selvaganesang: SQLItemDesc->maxLen is used as input for some of the data types. Does this mean it is invalid for SQLItemDesc for date columns. If so, you might consider fixing where SQLItemDesc is obtained from SQL.

xwq: JDBC T2 use VARCHAR to handle DATE type.So , we need set the maxLen in T2. 

Maybe a better fix is using the binary time format. But I need more time to investigate how to use binary date format from SQLCLI .  So I want to submit this little fix to make the MT-DCS can run with windows ODBC.

selvaganesang: No need to use the internal format because SQLCLI accepts TIMESTAMP, DATE and TIME in character format. I think the point I am trying to make here is SQLItemDesc->maxLen isused by getMemoryAllocInfo to allocate memory buffer accordingly on the client side to be shipped to the JNI side. If so, I am curious how the parameters for date/time/timestamp worked till now?

xwq: The issue is not about getMemoryAllocInfo here. Since the maxLen is hard code to 26, so the length of the memory is always enough. 

When use MT-DCS, the maxLen will be passed to driver side. ODBC driver will check if the maxLen is equal to the real length of the value, for example , the maxLen of DATE should be 10.

core/conn/jdbc_type2/native/SqlInterface.cpp
xwq: I will create a new JIRA after this PR is merged.

DaveBirdsall:
Is this ready for commit?

selvaganesang:
I am not sure. I am not clear from the response if the date/time/timestamp never worked in any flavor. In other words what is the value maxLen as returned from SQL for DATE, TIME and TIMESTAMP columns

xwq:
@selvaganesang You are right. I check the value returned from SQL , it's 10 for DATE. So I change the  fix to  just remove the hardcode.  I don't know why jdbc t2 hardcode the maxLen to 26 before.

DaveBirdsall:
Looks good to me. Does this look right to you, @selvaganesang?

selvaganesang:
Great. Good to go in now

PR #431, opened 2016-04-13
[TRAFODION-1931]ODB tool support windows platform

xwq:
add Visual Studio project files to support compile odb in windows platform

### DIFF
--- a/.rat-excludes
+++ b/.rat-excludes
@@ -110,7 +110,6 @@ TCPIPV6_os.*
 resource.h*
 resource.hm*
 odb.sln
-odb.vcxporj*
 odb.vcxproj.*
 # BSD licensed files
 swscanf.cpp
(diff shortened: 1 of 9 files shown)


.rat-excludes
```diff
@@ -109,6 +109,9 @@ TCPIPV6_os.*
 .ism*
 resource.h*
 resource.hm*
+odb.sln
+odb.vcxporj*
```
DaveBirdsall: Is this line a typo? (I notice that the next line is odb.vcxproj*). I don't see any files that match this pattern. Perhaps we can remove this line?

xwq: Thank you.

I used to want to add "odb.vcxproj*". However I found the next line "odb.vcxproj.*" was enough . So I remove it.

DaveBirdsall:
Looks good to me.

PR #432, opened 2016-04-13
[[TRAFODION 1932]] Fixing sqnodestatus

amandamoran:
Removing tools directory for installer makefile (tools should only be used for dev)
Adding new acceptKeys script to easily create known_host file


DaveBirdsall:
Looks good to me.

PR #433, opened 2016-04-14
[TRAFODION-1929] Set correct working dir for phx dependency builds

svarnau:
Per the Jira, the testing automation now invokes the test script in
the trafodion user environment, but that also has the side effect of
making the current directory as trafodion home directory.

The test script already calculates the path where it was invoked,
so changing directory there makes the dependent builds work.


arvind-narain:
Thank Steve. Merging now.

PR #436, opened 2016-04-15
Fix for JIRA [TRAFODION-1935]

sbroeder:
TransactionManager needlessly retries in doPrepareX when an
UnknownTransactionException is caught

### DIFF
--- a/core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionManager.java
+++ b/core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionManager.java
@@ -549,13 +549,15 @@ public class TransactionManager {
                retry = false;
              }
           }
-          catch (UnknownTransactionException ute) {
-              LOG.error("Got unknown exception in doCommitX by participant " + participantNum
-            		  + " for transaction: " + transactionId + " " + ute);
-              transactionState.requestPendingCountDec(true);
-              throw new UnknownTransactionException();
-          }
           catch (Exception e) {
+             if(e.toString().contains("UnknownTransactionException")) {
+                String errMsg = new String("Got unknown exception in doCommitX by participant " + participantNum
+              		  + " for transaction: " + transactionId + " " + e);
+                LOG.error(errMsg);
+                transactionState.requestPendingCountDec(true);
+                throw new UnknownTransactionException(errMsg);
+             }
+
              LOG.error("doCommitX participant " + participantNum + " retrying transaction "
                       + transactionId + " due to Exception: " + e);
              refresh = true;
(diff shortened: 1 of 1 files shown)


core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionManager.java
```diff
@@ -549,13 +549,15 @@ public SsccCommitResponse call(SsccRegionService instance) throws IOException {
...
              }
           }
-          catch (UnknownTransactionException ute) {
-              LOG.error("Got unknown exception in doCommitX by participant " + participantNum
-            		  + " for transaction: " + transactionId + " " + ute);
-              transactionState.requestPendingCountDec(true);
-              throw new UnknownTransactionException();
-          }
           catch (Exception e) {
+             if(e.toString().contains("UnknownTransactionException")) {
```
DaveBirdsall: Is it conceivable that we might get a false positive with a string lookup? (Could the characters "UnknownTransactionException" occur for some other reason in the text, for example as part of a delimited identifier for a Trafodion table name?)

DaveBirdsall:
Just checking... is there further action on this pull request?

sbroeder:
I thought I responded to your previous question last week, but I don't see the comment.

I don't think the chance of a false positive is very high - having "UnknownTransactionException" in the string as part of a table name seems improbable.  But the important thing is IF it did occur, the worst thing would be we wouldn't retry for the current Exception that was caught.  In either case, an Exception did occur.

I don't think there is additional work for the pull request

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionManager.java
selvaganesang: Please consider using  e instanceof UnknownTransactionException instead to avoid false postives

sbroeder:
Thanks for the suggestion Selva!  I'll resumit

DaveBirdsall:
Looks good to me. @selvaganesang, are you happy with this too?

selvaganesang:
Looks good. However, there are surrounding codes where instanceof could have been used instead of checking the string contains the classname.

sbroeder:
Can you give me an example of the code you're referring to?  I think I used instance of whenever an Exception was caught.  In other cases where a String was returned I used contains.  I think that is the correct behavior since using instance of would require me to create a new Exception from that String and then attempt to determine if it's an instance of or not.  This doesn't seem appropriate to me, but perhaps this is what you're looking for?

selvaganesang:
My bad. I misunderstood getException returns Exception object. Change is good to go

sbroeder:
Thanks for reviewing guys

PR #437, opened 2016-04-15
Fix for [TRAFODION-1617]

narendragoyal:


DaveBirdsall:
Thanks for fixing this problem! I recently tried using ilh_cleanhb and it failed because the Zookeeper directories had changed.

PR #438, opened 2016-04-15
Support for [TRAFODION-1891] and [TRAFODION-1937]

sandhyasun:
Support for get statements to retrieve lob related info.
Support for  tracing/debug iformtion for LOB operations. 
Added a new test case to executor/TEST130 to cover the new "get" command.


core/sql/cli/Cli.cpp
```diff
@@ -10495,13 +10528,18 @@ Lng32 SQLCLI_LOB_GC_Interface
 
       // Restore original data file.
       Int32 rc2=ExpLOBoper::restoreLobDataFile(lobGlobals,tgtLobName, (void *)currContext.exHeap(),hdfsServer,hdfsPort,lobLocation);
+      if (rc2)
+        {
+          lobDebugInfo("restoreLobDataFile Failed",0,__LINE__,lobTrace);
```
DaveBirdsall: Is the value of rc2 interesting here? (Or is it always -1 on an error and therefore uninteresting?)

core/sql/executor/ExExeUtilGet.cpp
```diff
@@ -6116,4 +6116,588 @@ short ExExeUtilRegionStatsFormatTcb::work()
...
+  char *catName = NULL;
+  char *schName = NULL;
+  char *objName = NULL;
+  Int32 offset = 0;
+  char columnName[LOBINFO_MAX_FILE_LEN]= {'\0'};
+  char lobLocation[LOBINFO_MAX_FILE_LEN]={'\0'};
+  char lobDataFilePath[LOBINFO_MAX_FILE_LEN]={'\0'};
+  Int64 lobEOD=0;
+
+  char buf[1000];
```
DaveBirdsall: Looks like this buffer is plenty long enough. Though I did have to hunt for LOBINFO_MAX_FILE_LEN to be sure (it is 256). Something to consider in the future is to use an expression here involving LOBINFO_MAX_FILE_LEN + some constant to make it easier for a reader to judge correctness. (It also has the benefit that if you change the value of LOBINFO_MAX_FILE_LEN later, say to 1024, this code doesn't break.)

core/sql/executor/ExExeUtilGet.cpp
```diff
@@ -6116,4 +6116,588 @@ short ExExeUtilRegionStatsFormatTcb::work()
...
+    }
+  
+
+  // populate catName, schName, objName.
+  if (extractParts(tableName,
+                   &catName, &schName, &objName))
+    {
+      return -1;
+    }
+  str_pad(buf,1000,' ');
```
DaveBirdsall: Use sizeof(buf) instead of 1000 to be more robust

core/sql/executor/ExExeUtilGet.cpp
```diff
@@ -6116,4 +6116,588 @@ short ExExeUtilRegionStatsFormatTcb::work()
...
+  offset = (currLobNum-1)*LOBINFO_MAX_FILE_LEN; 
+  str_cpy_all(lobInfo_->columnName, &((getLItdb().getLobColList())[offset]),
+              strlen(&((getLItdb().getLobColList())[offset])));
+  
+  //lob location  
+  char *lobLocation = &((getLItdb().getLobLocList())[offset]);
+  str_cpy_all(lobInfo_->lobLocation, lobLocation, strlen(lobLocation));
+                          
+  // lobDataFile
+  char tgtLobNameBuf[256];
```
DaveBirdsall: Should this be LOBINFO_MAX_FILE_LEN?

DaveBirdsall:
Looks good to me.

core/sql/cli/Cli.cpp
sandhyasun: Yes it returns -1 on error.

core/sql/executor/ExExeUtilGet.cpp
sandhyasun: Will consider making this a constant too.

core/sql/executor/ExExeUtilGet.cpp
sandhyasun: Yes - that would be better.

PR #441, opened 2016-04-19
[TRAFODION-1920] suppress SQL error during HIVE_SCAN when encounter i…

traflm:
…nvalid value, assign null to the invalid value

This pull request try to solve the issue when Trafodion try to retrieve data from a HDFS text file which contains invalid data. Like:
space for a numeric value
non-numeric string for a numeric value

this change will use the HIVE_SCAN_SPECAIL_MODE '2' , when that CQD is set to 2, HIVE_SCAN will automaticlly convert the invalid data into null and continue, instead of error out immediately.

How it does?
HIVE_SCAN need to do the data type conversion via expression evaluation. There are two ways to evaluate an expression: pCode and clause's eval() function. Since the current HIVE_SCAN do conversion row by row, and details are encoded in corresponding pCode. So the simplest way I can find out is to fix the invalid data by set it to null and retry the whole evaluation.
In pCode and eval() function, right after a failure, the new code saves the offset of the offending source data pointer. In HIVE_SCAN work() function, once it got an evaluation error, find and fix the offending data to null , in form of "R2" , exploded format,(R2 , refer to comments for FileScan::codeGenForHive) and retry the conversion again. The maximum retry time is the number of columns. If run out of retry time, then raise an error.
To avoid unexpected side-effect: this change will only modify pCode and normal evaluation a little bit. After an converting error, it tries to remember the offset of offending source data, and that is all. So it will not affect any normal evaluation/execution.
In HIVE_SCAN, in a special mode controlled by CQD, the code will go into a retry-loop, after the expression evaluation failure.
 
It is easier and maybe faster to change pCode execution to set the target as null and continue, but since pCode is used in many cases, not only HIVE_SCAN move experession. So that change will introduce unknown side effect.

This retry logic does introduce overhead, but in normal case the code path is same as before, in special mode, if no invalid data, it will also go through same path as before, the overhead will be there only in case there is converting error.

This change only fix from HIVE_SCAN perspective, to fully support transparent Bulkloader during loading, we still need to change the bulkloader work() method to tolerate other source of failure , I suppose. 

However, this PR will make sure there will be no failure during HIVE_SCAN due to not-matching data type. Later, if there is CAST, or TRANSLATE or TRUNCATE issues, we still need to change bulkload code.
So this is the first part to fully support TRAFODION-1920, but it should handle a lot of real cases already.
Add some sample test cases in hive/TEST005 also, interesting reviewers can check the test for more details.
Please help to review, and this will be a help in some real use cases.


core/sql/executor/ExHdfsScan.cpp
```diff
@@ -1378,9 +1423,27 @@ ExWorkProcRetcode ExHdfsScanTcb::work()
...
+    {
+      attr = asciiSourceTD->getAttr(neededColIndex);
+      neededColIndex++;
+	  if(attr->getOffset() == colidx)
+	{
+		*(short *)&hdfsAsciiSourceData_[attr->getNullIndOffset()] = -1;
+	}
+    }
+  }
+}
```
DaveBirdsall: I'm still reviewing, and I'm not familiar with this code. I'm trying to understand how this works. Does hdfsAsciiSourceData_ point to an ASCII string? Why would putting two bytes of x'ff' into the string make a value look like a null? How does the move expression parse to the next field?

traflm: thanks Dave,
Please refer to the very excellent comments above src code in FileScan::codeGenForHive(). 

The hdfsAsciiSourceData_ is the pointer to row 'R2' referred to in those comments. It is in a special internal tuple format called 'exploded format row', each field in this tuple contains at least two parts as far as I know: first byte is null-indicator, when it is 0, means not null, when it is 0xFF (-1), it means null. After the one byte null indicator, it follows 8 bytes of value, but it is not the real value, just a pointer to R1. When need, it needs to deference the pointer.  R1 is a buffer containing raw data read from HDFS file. R3 is for predicate, R4 is for projection and return rows.
So set 0xff here will set required field in R2 as null, in this case, it is because it contains invalid data, like 'abc' for an Integer field. And the moveColsConvertExpr expression failed to convert it into an integer. So the logic go back here , calling this method to set the invalid data field to null. And try again. Because the evaluation is now done row by row, not column by column, so this is the simplest way to change the logic at present . There are better ways, but without more guidance, I cannot do those, for example: do this converting column by column, this requires a huge change in compiler to generate different pCode and expression tree, which beyond my knowledge.
I didn't change any current normal logic.

The move expression typically will generate a piece of pCode instructions, if the row has 3 columns, the generated pCode will have 3 sets of instructions like this(instruction name is not valid, just try to explain what I want to mean)
IFNULL
  INST1
ELSE
  INST2
I modified INST2 to save the offset of invalid data offset in 'R2', and everything else keep same.
If pCode is turned off, the expression will be evaluated in normal clause evaluation path. It is a list of clauses, each clause correspond to an operation, and each column has a convert clause to do the data converting if needed. This is generated inside compiler, I don't know the details, parser should generate a tree of ItemExpr, and gencode will generate correct expression.

selvaganesang:
I am volunteering to review this code. Please give me some time to read this change.

core/sql/executor/ExHdfsScan.cpp
selvaganesang: I wonder if we need to change it to NULL at line no 1585 when convertExpr() returns an error  if the column is nullable and the cqd is set to convert to null. It is not clear to me the need for other changes.

selvaganesang: I think both convertExpr() and moveColsConvertExpr() deals at a row level.

traflm: Thanks Selva, I agree, will change it as well. It will support more converting cases. If simply select * from hive, convertExpr() will be null. So I ignore it. But if one add some cast in the select list, it will be invoked, so same logic should apply to it as well.
I will have an update asap.

traflm:
jenkins, retest

selvaganesang:
I wonder if the following solution might be cleaner. Introduce a new built-in function NULLIFERROR('value_expression') similar to NULLIFZERO('value_expression'). Implicitly add this function for HdfsScan if the CQD HIVE_SCAN_SPECTIAL_MODE is set 2.

selvaganesang:
To ensure that this change works as expected, can you please add more tests.with CQD set to 2
a) insert into traf_table ...select * from hive table 
b) consider having some columns have not nullable with the hive columns returning an error
c) consider having all possible data types so that conversion errors are handled as expected

traflm:
Yes, Selva, I will add more tests. b) should already covered by current test. for c), I do need a list of supported HIVE data types to test fully. I now test with (float, string, int, largeint, smallint). TinyInt failed, but I don't think Trafodion can support tinyint maybe. I will raise this question in the more proper place: mail list maybe.
I agree that NULLIFERROR will be a very clean and good solution, but that requires more knowledge of compiler, that will be a rather big project for me. So I suggest we can go with this not-optimal solution first, and implement the NULLIFERROR as phase II and clean up these code. In the field, we do need this feature. This is a prototype you can call it :-) 
I will add tests asap, and then start to study how to write a new built-in function in Trafodion.

traflm:
I just take a brief look at the NULLIFZERO, I feel I can copy it. So let me rework this from scratch to use NULLIFERROR approach, it will solve more problem current change cannot solve.
So let me rework on this , that needs some time though.

traflm:
Hi, Selva,
I didn't go with the NULLIFERROR approach at first place because I feel some difficulty and confusing about it, it doesn't looks possible for me instinctively. 
Now, I try to really spend time on this approach, that confusion become more concrete: As far as I understand, a function must have scalar value as operand. In this case, we cannot know the converting result until try to do the real converting first. So how to design the input operand for NULLIFERROR? I cannot figure out how CAST's return value can be used as 'ERROR' and as input to next NULLIFERROR clause.
For example: we can generate such an expression:
   NULLIFERROR( CAST () )
If translate into pCode, it will generate two blocks of pCode instructions, first is for CAST, during pCode execution of that block, if it meets converting error, it will return error to the caller immediately and no chance to go to next block of NULLIFERROR. If we modify the CAST pCode to continue and save error conditions somewhere so next NULLIFERROR can get as input, it will have unknown side effort. And it will be difficult to figure out 'what value is ERROR value?' returned by the CAST function. That will be input to NULLIFERROR. We have nullIndicator in a row, but no errorIndicator. So I am confused here. 
So it needs a lot more details about how to use NULLIFERROR.

traflm:
redesign and rework according to discussion offline.
This time, make the changes are localized to where conversion takes place.
In generator, if the CQD HIVE_SCAN_SPECIAL_MODE set to 2, disable pCode and set the expressionMode to ERROR_CONTINUE.
In clause evaluation method, when mode is in ERROR_CONTINUE, move null to target and continue.
Revert all other changes.

core/sql/exp/exp_eval.cpp
```diff
@@ -862,7 +862,17 @@ ex_expr::exp_return_type ex_expr::evalClauses(ex_clause *clause,
...
 
       if (retcode == ex_expr::EXPR_ERROR)
-	return retcode;
+      {
+        if((getPCodeMode() & ex_expr::ERROR_CONTINUE) > 0)
+        {
+          ExpTupleDesc::setNullValue( op_data[0],
+                                      clause->getOperand(0)->getNullBitIndex(),
+                                      clause->getOperand(0)->getTupleFormat() );
+          retcode = ex_expr::EXPR_OK;
```
anoopsharma00: -- this code also will return back from line #871 on the first conversion clause error instead
   of continuing to the next clause.
-- it will also not work correctly if getPCodeMode() is not  set.

core/sql/exp/exp_eval.cpp
```diff
@@ -862,7 +862,17 @@ ex_expr::exp_return_type ex_expr::evalClauses(ex_clause *clause,
...
+        }
+        else
+	  return retcode;
+      }
```
anoopsharma00: -- This change should be inside of ex_conv_clause::eval when an error happens and not in the mainline.
-- a new flag bit should be set at code generation time in ex_conv_clause (something like NullOnError).
  This flag bit will be checked in ex_conv_clause::eval method. That way handling of this case of NullOnError 
  will be contained within that method. That method should return ok after moving null value to target in case
  of error.
-- No need to have a new pcode mode. The ERROR_CONTINUE setting could be removed
-- No change will be needed in mainline exp eval code

core/sql/exp/exp_eval.cpp
```diff
@@ -2953,6 +2963,7 @@ ex_expr::exp_return_type ex_expr::evalPCode(PCodeBinary* pCode32,
 	Int32 srcLen = 0;
 	PTR_DEF_ASSIGN(char, dst, 0);
 	BASE_PTR_DEF_ASSIGN(char, src, 2);
+	BASE_PTR_DEF_ASSIGN(char, rowstart, 0);
```
anoopsharma00: what is rowstart used for?

core/sql/exp/exp_eval.cpp
```diff
@@ -3003,7 +3014,7 @@ ex_expr::exp_return_type ex_expr::evalPCode(PCodeBinary* pCode32,
             if (diagsArea != atp1->getDiagsArea())
                    atp1->setDiagsArea(diagsArea);
 	    if (er == ex_expr::EXPR_ERROR) 
-	      return ex_expr::EXPR_ERROR;
+		return ex_expr::EXPR_ERROR;
```
anoopsharma00: indentation has changed at many places with the new code. 
Not sure what editor was used.
Will be better to keep the older indentation.

traflm: yes, will fix this. I use vim, but this line changed a lot in previous modification and changed the indentation, will restore it correctly.

core/sql/exp/exp_eval.cpp
traflm: left over... Will remove it

core/sql/exp/exp_eval.cpp
anoopsharma00: Couple things:
Codegen time: The flag to move null in case of an error should be set only if target is nullable.
Runtime: Expr code should check to make sure that target is nullable before moving
the null value. If target is non-nullable, then original error should be returned.

traflm:
rework again.
Move the nulliferror handling into ex_conv_clause::eval()
remove changes to mainline expression evaluation.

Now, it works as this:
In code gen of HIVE_SCAN, if the CQD HIVE_SCAN_SPECIAL_MODE is set to '2'. Generate a CAST function setting a special flag. In that CAST codegen, it will set runtime flag based on the special flag. This will create a special ex_conv_clause.
Codegen also disable pCode in this mode.
At runtime, ex_conv_caluse::eval() will move null into target if convDoIt() failed. And continue.

Thanks to Anoop patiently give comments and improve the design and implementation.

Please help to code review again.

traflm:
I will abandon this PR, and start from scratch.This one has too much reworks and now contains merge conflict and sql regression failure.

traflm:
I will create a new PR, this one just has too many reworks. And still one file has unneed changes.
It is a very good discussion procedure though.

traflm:
PR 461 created, close this one

PR #442, opened 2016-04-19
TRAFODION-1940 better NATable statistics plus OID, cleanhb fixes

zellerh:
TRAFODION-1939 Remove OID from the list of reserved words
TRAFODION-1940 Improve NATable statistics
                - Add the context of the NATable cache
                  to the natablecache() UDF
                - Add the entry size to the
                  natablecacheentries() UDF

Also changed the ilh_cleanhb script to print a help text
on how to use it and require a parameter 0, 1 or 2.
Still leaving the old, misspelled directory for zookeeper
in the script until everyone has moved to a later version
that puts the zookeeper data under local_hadoop/data/zookeeper.

Sorry for mixing three fixes into one PR.

TRAFODION-1939: 
- ComResWords.cpp
- ParKeyWords.cpp
- EXPECTED037.SB

TRAFODION-1940:
- CmpContext.h
- NATableSt.cpp
- NATable.cpp
- NATable.h
- EXPECTED009
- TEST009


core/sql/regress/hive/TEST009
```diff
@@ -170,7 +170,7 @@ sh sqlci -i "TEST009(select_stmts)";
 select catalog_name, schema_name, object_name from table(natablecacheentries('user','local')) order by 1,2,3;
 drop external table t009t1 for hive.sch_t009.t009t1;
 get tables in schema "_HV_SCH_T009_";
-select * from table(natablecacheentries('user','local')) order by 2,3,4;
+select row_id, catalog_name, schema_name, object_name from table(natablecacheentries('user','local')) order by 2,3,4;
```
DaveBirdsall: Looks good. I'm wondering if it would  be useful to add a test that also shows the new context column?

DaveBirdsall:
Looks good to me.

core/sql/regress/hive/TEST009
zellerh: Thanks, good point, I'll add that.

PR #443, opened 2016-04-20
TRAFODION [1944] Check status of authentication and authorization via a CLI call

robertamarton:
The SQL_EXEC_GetAuthState_Internal CLI request has been changed to 
SQL_EXEC_GetAuthState and moved from SQLCLIdev.h to sqlcli.h so it is available
to external callers.


DaveBirdsall:
Looks good to me.

PR #448, opened 2016-04-22
[TRAFODION-1916] dcscheck fails with cdh5.4.4 and hbase 1.0.0

arvind-narain:
Zookeeper jar file will now get installed in $DCS_INSTALL_DIR/lib
as was done before TRAFODION-1828. This is similar to how it's
currently been done for REST subsystem.


hegdean:
change looks good to me

PR #450, opened 2016-04-22
Removing hardcoding of Apache version

amandamoran:


svarnau:
Looks good to me.

svarnau:
I've verified this fix against 2.1 version on master branch. (This was the problem version that prompted this change.)

DaveBirdsall:
Is this change ready to be merged into apache:master?

svarnau:
Amanda is on vacation, but Yes, I think it is ready.

PR #451, opened 2016-04-25
ddl xns turned on by default (plus couple more fixes)

anoopsharma00:
-- cqd ddl_transactions is ON by default
-- disabled init/drop trafodion and drop schema from
   running in a single transaction.
   This will be changed once couple of dtm issues are fixed
-- error 8616 is retuned in case of commit conflict
-- fixed begin/commit/set stmts to not retry in case of an error


DaveBirdsall:
Looks good to me.

PR #452, opened 2016-04-27
[TRAFODION-533]Fixes for regressive SPJ calling

kevinxu021:


selvaganesang:
Great. Can you please add the tests in the JIRA as part of the developer regressions

core/conn/jdbc_type2/native/SqlInterface.cpp
```diff
@@ -2570,12 +2570,9 @@ SQLRETURN ALLOCSQLMXHDLS_SPJRS(SRVR_STMT_HDL *pSrvrStmt, SQLSTMT_ID *callpStmt,
 	if (pModule->module_name == NULL)
 	{
 		DEBUG_OUT(DEBUG_LEVEL_STMT,("***pModule->module_name == NULL  Call AllocStmtForRs()"));
-/* Commenting out for now - will be looked at when SPJ is supported
- #ifdef NSK_PLATFORM
 		CLI_AllocStmtForRS(callpStmt,
```
DaveBirdsall: Should we capture the return code here? retcode = CLI_AllocStmtForRS...

xwq: @DaveBirdsall Actually, CLI_AllocStmtForRS is a macro. retcode will capture the return code of "SQL_EXEC_AllocStmtForRS" (Debug.h +125). So we don't need capture the return code here.

kevinxu021: @selvaganesang For spj tests, a spj jar package is required while create a library. I'm not sure the path of the JAR. it seems to be a little complicated. The JAR should be existing on server-side(every node). i'm not sure the IP address/host name(where to upload) while the tests running.

DaveBirdsall:
I am happy to commit this change. I assume that adding tests can be done as a separate pull request. Is this OK with you, @selvaganesang?

selvaganesang:
Fine with me. The SPJ tests can be modeled after the regression tests available at sql/regress/udr/TEST*. It is possible to add to an existing test suite. Adding a test would help to detect this issue earlier as part of developer regressions going forward if it gets broken for some reason.

PR #454, opened 2016-04-28
Fix for JIRA 1953

venkat1m:
Include jetty-security jar in REST server packaging.


DaveBirdsall:
Looks good to me.

svarnau:
This looks like it should be under same Apachev2 license as the other jetty jar files that are already bundled.  That is referenced in licenses/lic-rest-bin.  Looks good to me.

PR #455, opened 2016-04-28
[TRAFODION-1952]jdbc can not work without config hostname and ip in /etc/hosts on client machine

kevinxu021:
Variable hostname is not in use


DaveBirdsall:
Looks good to me.

PR #456, opened 2016-04-28
Fix for TRAFODION-1948

prashanth-vasudev:
Drop schema cascade internally selects against objects table.
The fix is to use a distinct clause in the internal select query.


selvaganesang:
Looks good

PR #459, opened 2016-04-29
[TRAFODION-1955] JNI optimization at the time of compilation

selvaganesang:
Removed ByteArrayList class and used array of byte array instead.
Only one JNI to java transistion is done to getRegionStartKeys
and getRegionEndKeys at the time of query compilation

Also fixed an issue with lib_mgmt Makefile.

### DIFF
--- a/core/sql/executor/ExExeUtilGet.cpp
+++ b/core/sql/executor/ExExeUtilGet.cpp
@@ -3080,13 +3080,11 @@ short ExExeUtilGetHbaseObjectsTcb::work()
                 break;
               }
 
-            Int32 len = hbaseTables_->at(currIndex_).len;
-            char *tmp  = hbaseTables_->at(currIndex_).val;
-            len = hbaseTables_->at(currIndex_).len;
-            if (len > ComMAX_3_PART_EXTERNAL_UTF8_NAME_LEN_IN_BYTES+6)
-                len = ComMAX_3_PART_EXTERNAL_UTF8_NAME_LEN_IN_BYTES+6; 
-            strncpy(hbaseNameBuf_, tmp, len);
-            hbaseNameBuf_[len] = 0;
+            HbaseStr *hbaseName = &hbaseTables_->at(currIndex_);
+            if (hbaseName->len > ComMAX_3_PART_EXTERNAL_UTF8_NAME_LEN_IN_BYTES+6)
+                hbaseName->len = ComMAX_3_PART_EXTERNAL_UTF8_NAME_LEN_IN_BYTES+6; 
+            strncpy(hbaseNameBuf_, hbaseName->val, hbaseName->len);
+            hbaseNameBuf_[hbaseName->len] = 0;
             hbaseName_ = hbaseNameBuf_;
 
             Lng32 numParts = 0;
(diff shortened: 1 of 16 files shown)


core/sql/executor/ExExeUtilGet.cpp
```diff
@@ -3074,18 +3074,21 @@ short ExExeUtilGetHbaseObjectsTcb::work()
...
 
-            Int32 len = 0;
-            hbaseName_ = bal_->getEntry(currIndex_, hbaseNameBuf_, 
-                                        ComMAX_3_PART_EXTERNAL_UTF8_NAME_LEN_IN_BYTES+6, 
-                                        len);
-            hbaseName_[len] = 0;
-            
+            Int32 len = hbaseTables_->at(currIndex_).len;
+            char *tmp  = hbaseTables_->at(currIndex_).val;
+            len = hbaseTables_->at(currIndex_).len;
```
DaveBirdsall: I hope the C++ compiler does common subexpression elimination so we don't keep re-evaluating the "at".

core/sql/executor/HBaseClient_JNI.cpp
```diff
@@ -1439,14 +1223,14 @@ ByteArrayList* HBaseClient_JNI::getRegionStats(const char* tblName)
   jstring js_tblName = jenv_->NewStringUTF(tblName);
   if (js_tblName == NULL) 
   {
-    GetCliGlobals()->setJniErrorStr(getErrorText(HBC_ERROR_DROP_PARAM));
+    GetCliGlobals()->setJniErrorStr(getErrorText(HBC_ERROR_REGION_STATS));
```
DaveBirdsall: Good catch

core/sql/executor/HBaseClient_JNI.cpp
```diff
@@ -5619,3 +5394,41 @@ int convertStringObjectArrayToList(NAHeap *heap, jarray j_objArray,
     return arrayLen;
 }
 
+
+jint convertByteArrayObjectArrayToNAArray(NAHeap *heap, jarray j_objArray, NAArray<HbaseStr> **retArray)
+{
```
DaveBirdsall: For safety, might be a good idea to set retArray = 0 first thing in the method. Otherwise a caller might ignore an error and use the uninitialized pointer, with unpredictable results.

DaveBirdsall:
Looks good to me. Nice change. One recommendation concerning initializing an output parameter.

selvaganesang:
Thanks Dave. I will make the suggested changes and push in another commit. This PR was done to validate if regressions seen with these changes is specific to my environment. Please don't merge these changes yet.

selvaganesang:
This PR is now ready to be reviewed and merged

selvaganesang:
ByteArrayList is a mechanism to ship array of byte arrays from Java to JNI in SQL part of Trafodion. It is found that the use of ByteArrayList class is expensive due to JNI to java transitions for every entry. In almost all cases the size of the byte array is known in advance and hence it would perform best if it is used as array of byte arrays. I have made changes already to avoid the use of ByteArrayList in Trafodion code base in all cases. However, I have retained the ByteArrayList JNI interface to ensure that the build is not broken when this change is merged to other feature branches.  After a short time if there are no objections, I will eliminate the ByteArrayList JNI interfaces entirely.
 
So, please if you can plan to use ByteArrayList in near future, you should defer your changes till this is done and consider using array of byte arrays instead.

prashanth-vasudev:
Looks good to me.

DaveBirdsall:
I will merge this shortly.

PR #460, opened 2016-05-03
TRAFODION-1963 Create index with delimited name fails if table has data

sureshsubbiah:
Delimited name was being sent during load complete.


selvaganesang:
Change looks good. Can you please add a test case in developer regressions.

DaveBirdsall:
Looks good to me.

PR #461, opened 2016-05-03
[TRAFODION-1920] suppress SQL error during HIVE_SCAN when conv error

traflm:
A new PR for JIRA 1920. When read from HDFS in HIVE_SCAN and encounter invalid data, continue and set the offending data to null.

Latest implementation details:
In code gen of HIVE_SCAN, if the CQD HIVE_SCAN_SPECIAL_MODE is set to '2'. Generate a CAST function setting a special flag. In that CAST codegen, it will set runtime flag based on the special flag. This will create a special ex_conv_clause.
Codegen also disable pCode in this mode.
At runtime, ex_conv_caluse::eval() will move null into target if convDoIt() failed. And continue.


selvaganesang:
Looks good. I would appreciate if you can add upsert/insert into trafodion table,,, select * from hive table and test case when the target is not nullable but hive producing invalid column values.

traflm:
Sure, I will add a test to insert into trafodion table with not nullable columns.
There is already a load into test. I will add INSERT and UPSERT as well.

core/sql/regress/hive/EXPECTED005
```diff
@@ -534,4 +534,101 @@ C1           C2
...
++>c7 float(54) not null,
++>c8 smallint
++>);
+
+--- SQL operation complete.
+>>cqd HIVE_SCAN_SPECIAL_MODE '2';
+
+--- SQL operation complete.
+>>insert into trafodion.seabase.traf_tbl_bad select * from tbl_bad;
+
```
selvaganesang: Good that we are getting this error. Are we supposed to get the original error? Or what error will be seen is non-deterministic because it depends which node the expression gets evaluated?

selvaganesang:
Thanks @traflm . The change looks good. I am waiting for another contributor to review this code before I can merge this code.

core/sql/generator/GenRelScan.cpp
```diff
@@ -880,6 +884,12 @@ short FileScan::codeGenForHive(Generator * generator)
   } // for (ii = 0; ii < hdfsVals; ii++)
     
 
+  UInt32 hiveScanMode = CmpCommon::getDefaultLong(HIVE_SCAN_SPECIAL_MODE);
+  //enhance pCode to handle this mode in the future
+  //this is for JIRA 1920
+  if((hiveScanMode & 2 ) > 0)   //if HIVE_SCAN_SPECIAL_MODE is 2, disable pCode
+    exp_gen->setPCodeMode(ex_expr::PCODE_NONE);
+
```
anoopsharma00: This will disable pcode for the whole expression and all conversions.
Better place to disable it is in ex_conv_clause::pcodeGenererate in
file exp/ExpPCodeClauseGen.cpp.
That code should check for conv_to_null_when_error flag in ex_conv_clause
and disable pcode just for that clause.

core/sql/regress/hive/EXPECTED005
anoopsharma00: It might be helpful if the original conversion error is added as a warning. Maybe based on some setting(cqd), warnings could be added and returned. One can also check for max number of warnings instead of adding all of them. But that is not a hard requirement for now.

core/sql/generator/GenRelScan.cpp
traflm: I agree that will be much better. I need to understand that part further, a simple change cannot work, I need to debug more on this.

core/sql/regress/hive/EXPECTED005
traflm: Yes, the error message will depends on which node the expression get evaluated, so it is better to add converting error as warnings. I need to better understand details of DiagArea, I didn't remove errors during convDoIt(), it should still in the DiagArea, I need to understand if DiagArea is statement global or per operator. And not sure which DiagArea the final sqlci displays.

DaveBirdsall:
Looks good. Much simpler than the previous attempts. Agree with @anoopsharma00 and @selvaganesang that keeping the original error too is good, and disabling PCode only for the clause is good.

selvaganesang:
Merging this change now because it is ok to do the rest  later

PR #462, opened 2016-05-04
Testware fix for udr/TEST001 failure. 

sandhyasun:
The "length" keyword  for clob/blob is not valid anymore. now the syntax is just blob(10) or clob(8)


zellerh:
Looks good. Did this go through in PR453 because you ran the LOB regression tests and this was in the UDR regression test?

sandhyasun:
Yes - made sure executor and hive passed and actually my UDR test failed due to dcs issues and  didn't rerun. Now I know there is one more test where the datatype will get tested when on by default.

PR #465, opened 2016-05-04
Known diff file hive/TEST018 for recent plan change that is non deterministic.

sandhyasun:
The diff only happens on cdh an not hdp hence the known diff file. The plan is almost similar on both platforms  but with a slight difference.


DaveBirdsall:
Looks good to me.

selvaganesang:
It is strange that it passed on hdp version and  failed on cdh version. While attempting to look at the difference yesterday, I realized that regress directory is not backed up when the test is successful by Jenkins. The console output regress.log is too large to browse through the web interface.

PR #466, opened 2016-05-04
[TRAFODION-1972] Remove docs directories from exclude list

svarnau:
This will allow these directories to be included in source package,
and hence in the official released code.

Note that it also includes the website pages, which provides a snapshot
in time of the project site, for whatever that is worth.


DaveBirdsall:
Looks good to me.

PR #468, opened 2016-05-05
[[TRAFODION-1951]] Fix for JAVA checking

amandamoran:
Adding fix for JAVA checking if JAVA 7 has an update 100 or greater. 

Also deleting two files that are no longer used (and should have been deleted when the Vanilla Apache support was checked in).


svarnau:
Why are we removing the apache config scripts?

amandamoran:
These scripts are not used at all for the "new" "vanilla" apache support. These are very old scripts, and are mostly duplicated of other scripts with slight modifications. They were added months ago for a quick fix.

svarnau:
Okay, looks good to me.
Can someone else take a quick look?  I'd like to get this merged and included in first release candidate.

PR #469, opened 2016-05-06
[TRAFODION-1978] Report a nice error when attempting UPD STATS on LOBs

DaveBirdsall:
Histogram statistics on LOB columns do not make sense at the moment; Trafodion does not support them. However, the UPDATE STATISTICS command would fail in odd and diverse ways when asked to create stats on LOB columns. This has been fixed. Now when LOB columns are explicitly referenced in an UPDATE STATISTICS command, an error 9246 will result. When LOB columns are implicitly referenced (e.g., in an ON EVERY COLUMN clause), they will be silently skipped.


DaveBirdsall:
Don't merge this just yet. I plan to add some tests to regression test executor/TEST130, but that's taking a little time.

sandhyasun:
Changes look fine and thanks for removing the unneeded  "obeys".

DaveBirdsall:
jenkins, retest

sandhyasun:
Looks good merging...

DaveBirdsall:
Thank you, @sandhyasun

PR #470, opened 2016-05-09
[TRAFODION-1874] fix build issues when PDSH installed in build system

traflm:
There is no need to copy gencat file at build time. Installer will copy it to all nodes during installation.
This in fact will break the build if the dev system happen to installed with pdsh.


DaveBirdsall:
Looks good to me.

PR #471, opened 2016-05-10
TRAFODION-1983 error in preCodeGen plus more items

zellerh:
This also constains a fix for TRAFODION-1710 (setString in UDFs does not accept numeric values with scale) as well as misc small fixes. It may make more sense to review the three commits in this PR separately, as they are on different topics.

### DIFF
--- a/core/sql/generator/GenPreCode.cpp
+++ b/core/sql/generator/GenPreCode.cpp
@@ -11343,19 +11343,6 @@ short HbaseAccess::extractHbaseFilterPreds(Generator * generator,
        preds.advance(vid))
     {
       ItemExpr * ie = vid.getItemExpr();
-
-      // if it is AND operation, recurse through left and right children
-      if (ie->getOperatorType() == ITM_AND)
-        {
-          ValueIdSet leftPreds;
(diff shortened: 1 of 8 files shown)


core/sql/generator/GenPreCode.cpp
```diff
@@ -11343,19 +11343,6 @@ short HbaseAccess::extractHbaseFilterPreds(Generator * generator,
        preds.advance(vid))
     {
       ItemExpr * ie = vid.getItemExpr();
```
DaveBirdsall: Should there be an assert if we find an AND?

DaveBirdsall:
Looks good to me.

core/sql/generator/GenPreCode.cpp
zellerh: There are probably hundreds of places in the code where we could add such an assert, I don't think this place needs it more than others. We had a discussion with Eric a while back and decided that we wouldn't need to check for ANDs here; this code is just a left-over from a previous design.

DaveBirdsall: Thanks; looks good to me.

PR #473, opened 2016-05-10
Fix for [TRAFODION-1979] values clause on Lob columns giving garbage output or abends process.

sandhyasun:


core/sql/regress/executor/EXPECTED130
```diff
@@ -124,7 +134,16 @@ inserted row13
 
 --- 3 row(s) selected.
 >>
->>
+>>values (filetolob('./myfile.txt'));
+
+*** ERROR[8442] Unable to access ExpLOB interface. Call to ExpLOB returned error LOB_PTR_ERROR(532). Error detail 0.
```
DaveBirdsall: What is the end user supposed to do with this error?

sandhyasun: This error will be reutrned anytime an invalid LOB handle is provided to any LOB interface. Before this fix (in ExpLOB.cpp) the mxosrvr could abend. There was no LOB handle provided in this case. So this general error is supposed to indicate this function has no LOB to work on .  If this was a SQL statement referring to actual columns, the LOB column would have provided a LOB handle to for this function to work on.I felt it's better to  return the general error rather than a specific error for the values clause .I can add the causes in the error messages manual.

DaveBirdsall: So, the problem in this case is that ./myfile.txt simply doesn't exist? Would it be better to simply say that? The error message as written seems to suggest some sort of internal error which will likely lead to a support call. A customer won't have any idea what an ExpLOB interface is, so we shouldn't tell him about it unless we expect him to do something with it.

I'm wondering if there are two cases here. One might be the user specified a file name that doesn't exist (as in this example). Another might be that Trafodion stored a file name internally and that file was then deleted. That would be an internal error if Trafodion managed the file I suppose. But if it is an external file, then it is a user error. Perhaps we should simply report that the file doesn't exist.

sandhyasun: No we do have an error for the case where a file does not exist. The problem here is that the filetolob() function cannot be called in a values clause.  Same for function stringtolob(). Even if a file is a valid input file, there is no lob column/storage  that the function needs to work on.

DaveBirdsall: That sounds like a static semantic restriction. Shouldn't the binder catch that? I'm wondering if there are other contexts where the function should be forbidden. For example, inside a CAST, or SUBSTRING, or a column reference in a GROUP BY clause?

sandhyasun: I was thinking about that initially but had to fix the invalid lobHandle issue anyway. We do have checks in several functions in the binder to prevent LOB functions but not sure if it's comprehensive. Let me see if it'll be easy to detect this within a values clause. It will be an additional check at compile time  and the invalid lobHandle check at runtime  will still remain.

sandhyasun: On second thought it's not possible to distiguish the function being used in an input values clause Vs the values clause by itself at compiletime. So in line with other functions like queryid_extract, I can improve the error message so it's clear that the function is not being called on a valid LOB column. A new error 8434 can be introduced for this. 
>>values(filetolob('fffff'));

*** ERROR[8434] Invalid target column for LOB function. The column needs to be blob/clob type.
--- 0 row(s) selected.


>>values(stringtolob('gggggg'));
*** ERROR[8434] Invalid target column for LOB function. The column needs to be blob/clob type.
--- 0 row(s) selected.

>>values(buffertolob(LOCATION 6767676,size 5));

*** ERROR[8434] Invalid target column for LOB function. The column needs to be blob/clob type.
--- 0 row(s) selected.
>>

DaveBirdsall: That sounds better. Now a user knows what to do. They need to add some function that converts the LOB to some other datatype (or processes it somehow to a more palatable form). Thanks for considering this.

PR #476, opened 2016-05-10
[TRAFODION-1982] Package server components into single tar

svarnau:
This pull request includes Amanda's changes [TRAFODION-1985] for installer to take a single tar, which also addresses the issue of looking for a build-version.txt file in the tar-of-tars.

This changes the historical tar-of-tars packaging from the days when DCS and REST where built separately from the core server component (even separate repos). This aligns with other Apache project packaging where there is a LICENSE, NOTICE file at the root of each tar-file deliverable. We need a single deliverable for server-side.

The build will also now name tar-files as the Apache artifacts are normally named.


core/conn/Makefile
```diff
@@ -25,16 +25,17 @@ include ../macros.gmk #top level
 
 RM      = /bin/rm
 ifeq ($(SQ_BUILD_TYPE),release)
-  CLIENT_TAR	?= ../../${DISTRIBUTION_DIR}/trafodion_clients-$(TRAFODION_VER).tgz
+  CLIENT_TAR	?= ../../${DISTRIBUTION_DIR}/apache-trafodion_clients-$(TRAFODION_VER)-incubating.tar.gz
```
hegdean: Instead of hard coding(apache_trafodion) can we use the environment variable TRAFODION_VER_PROD

core/conn/Makefile
```diff
@@ -25,16 +25,17 @@ include ../macros.gmk #top level
...
 else
-  CLIENT_TAR	?= ../../${DISTRIBUTION_DIR}/trafodion_clients-$(TRAFODION_VER)-debug.tgz
+  CLIENT_TAR	?= ../../${DISTRIBUTION_DIR}/apache-trafodion_clients-$(TRAFODION_VER)-debug.tar.gz
```
hegdean: can we use TRAFODION_VER_PROD

core/sqf/Makefile
```diff
@@ -260,18 +260,18 @@ genverhdr: buildinfo
                                         -branch "${PV_BRANCH}" \
                                         -date "${PV_DATE}"
 
-PKG_PROD = trafodion
+PKG_PROD = apache-trafodion
```
hegdean: Can we use the environment variable here (TRAFODION_VER_PROD)

core/conn/Makefile
svarnau: Unfortunately, that variable value is not the right value for a file name. I suppose we could add yet another variable, but that does not seem necessary for packaging name in 2-3 spots.

svarnau:
jenkins, retest -- jenkins test did not account for file suffix change

svarnau:
@amandamoran these failures are stumping me. Both CDH and HDP are failing at the distro version check. Don't see anything in installer changes that would cause that. Changing order of PATH might cause something unexpected, but that's a reach.

amandamoran:
@svarnau I took a look. Did you add the ONE_TAR_INSTALL variable to your config file (and set it to Y)?

I added a cat to the TRAF_CONFIG when install fails on traf_config_check so that maybe we could debug this easier. I committed and pushed the change to my branch. 

Thanks!

svarnau:
Good call. Thanks to Arvind's suggestion, we now archive the config file in the test logs.  It looks like the jenkins script check for the one-tar case was incorrect.  I'll get a fix in.

svarnau:
jenkins, retest -- jenkins side script fix params to installer

PR #479, opened 2016-05-12
[TRAFODION-1988] Better Java exception handling in Trafodion

selvaganesang:


selvaganesang:
jenkins, extra tests

DaveBirdsall:
I did a quick first pass at these changes. Things do look simpler, and that's a good thing. It would help me understand better, though, if you could give an overview of the changes. I notice some classes are deleted, for example. Why are the no longer needed?

selvaganesang:
Yes. The deleted java files are no longer needed. These java classes are used in the earlier versions of Trafodion when Trafodion used thrift interface to connect to hbase. These are the general guidelines to be followed to review this change if it is syntactically correct. For semantics correctness, please feel free to weigh in your expertise. 

a) No method should be declared as throws Exception - The exception should be more specific like IOException
b) Ensure that only specific exceptions are caught - Shouldn't contain catch(Exception e) when needed.
c) Avoid catching a specific exception just to throw it back as the same exception. If it is morphed into a different exception, it could be ok.
 4) Log.Error and throw from the exception handler should not miss the original exception. It should take in the exception as a parameter so that the original exception is not missed out.

DaveBirdsall:
Looks good to me. Thanks for the explanation.

selvaganesang:
Are we ready to merge this PR to master? If you have any other comments please feel free to add it here

DaveBirdsall:
+1 to merge.

PR #480, opened 2016-05-13
[TRAFODION-1971] Avoid redundant reinitialization during bulk load

blfritch-esgyn:
After each flush during the preparation phase of bulk load, HFile
initialization was taking place redundantly. This imposed significant
overhead and slowed down loads.

### DIFF
--- a/core/sql/executor/ExHbaseIUD.cpp
+++ b/core/sql/executor/ExHbaseIUD.cpp
@@ -1064,7 +1064,7 @@ ExHbaseAccessBulkLoadPrepSQTcb::ExHbaseAccessBulkLoadPrepSQTcb(
     hdfsSampleFile_(NULL),
     lastErrorCnd_(NULL)
 {
-   hFileParamsInitialized_ = false;  ////temporary-- need better mechanism later
+   hFileParamsInitialized_ = false;
    //sortedListOfColNames_ = NULL;
    posVec_.clear();
 
--- a/core/sql/executor/ExHbaseIUD.cpp
+++ b/core/sql/executor/ExHbaseIUD.cpp
@@ -1285,7 +1285,6 @@ ExWorkProcRetcode ExHbaseAccessBulkLoadPrepSQTcb::work()
       matches_ = 0;
       currRowNum_ = 0;
       numRetries_ = 0;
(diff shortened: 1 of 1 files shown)


selvaganesang:
jenkins, add user

core/sql/executor/ExHbaseIUD.cpp
```diff
@@ -1285,7 +1285,6 @@ ExWorkProcRetcode ExHbaseAccessBulkLoadPrepSQTcb::work()
       matches_ = 0;
       currRowNum_ = 0;
       numRetries_ = 0;
-      hFileParamsInitialized_ = FALSE;
```
selvaganesang: Is this parameter used to create sampling of the loaded data. If so, can you please consider using another condition too.

Also, this flag is being reset to false in ALL_DONE/DONE step. So, commenting out in NOT_STARTED step alone wouldn't help.

I also see this comment in the constructor. Do you think should we retain this comment still.

 hFileParamsInitialized_ = false;  ////temporary-- need better mechanism later

blfritch-esgyn: I'm not sure what you mean by adding another condition. Sampling is controlled by a syntax option of bulk load.

Resetting the flag for ALL_DONE is not a problem because it is only executed when the load (preparation phase) is done.  On the other hand, removing the one in NOT_STARTED, which is executed after each flush, improved load time for a wide, 52M-row table from 3.5 hours to 35-40 minutes (without Suresh's fix that changed flush frequency).

I'm not sure why Khaled thought a better mechanism was needed. I would be fine with removing the comment.

selvaganesang: It looks like the sampling is controlled by (if samplingRate > 0). I guess you mean flush to add a batch of rows to HFile in one shot. Otherwise, I thought that flushing of hfile should happen when the hfile is closed. Is that correct?

Please remove the comment.  Then this fix is good to go and it is definitely a great find to improve the bulk load further.

blfritch-esgyn: samplingRate will be >0 only if the syntax to do it is used. Yes, the flush refers to our own buffers. The HFile flushing is separate. I removed the comment.

PR #482, opened 2016-05-13
Trafodion 1993

amandamoran:


svarnau:
I might have gone with $(ls -d $SQROOT/dcs*)
But I think what you have here is fine. Thanks for the quick turn-around on this!

DaveBirdsall:
Looks good to me.

install/installer/traf_config_check
```diff
@@ -702,7 +702,7 @@ function checkClouderaVersion {
 if [[ $CDH_5_3_HDP_2_2_SUPPORT == "N" ]]; then
    #Check that Cloudera 5.2 or 5.3 are not installed.
    if [[ "$CDH_5_4_SUPPORT" == "Y" ]] || [[ "$CDH_5_5_SUPPORT" == "Y" ]]; then
-      nameOfVersion=$(ssh -q -n $node grep "Version" $HOME/hbaseVersion.txt | sed 's/,.*//' | sed 's/.*\-//' | grep cdh5.[4-6].*)
+      nameOfVersion=$(ssh -q -n $node grep "Version" $HOME/hbaseVersion.txt | sed 's/,.*//' | sed 's/.*\-//' | grep cdh5.4.*)
```
DaveBirdsall: Not sure I get it. The "IF" includes the condition $CDH_5_5_SUPPORT, yet we don't look for 5.5? Should the "IF" change?

amandamoran: @DaveBirdsall you are right. This still works as is... but doesn't "make sense". Let me fix and check it in.

DaveBirdsall:
Looks good to me.

PR #483, opened 2016-05-13
[TRAFODION-1992] Re-packaging server tar file broke install_traf_components

svarnau:
Change install_traf_components to use built target directory
instead of tar file, now that tar file is not produced for DCS and REST.

Submitting to release2.0 branch. Since this is a developer script, end-user is not affected. Not critical for 2.0.0, but can be picked up in next patch release if it does not make this one.


DaveBirdsall:
Looks good to me.

PR #484, opened 2016-05-14
[TRAFODION-1974] Limit HBase cache size for ustat direct sampling

blfritch-esgyn:
This is not a definitive fix for the issue of timeouts during Update
Statistics processing for very large tables, but should decrease the
frequency of such occurrences. It extends the fix for TRAFODION-1740
(SHA-1 99a143f29910c7d3c8e02fb1bf101214568bf80a, Dave Birdsall), which
modified the CQDs controlling HBase cache size based on the sampling
rate when the ustat sample table is populated. This commit does the
same thing when the sample can be read directly into memory from the
table being sampled.


DaveBirdsall:
Looks good to me. Thank you for doing this, @blfritch-esgyn

PR #485, opened 2016-05-17
[TRAFODION-1853] Fix overflow of long type bit wise shifting in T4 LogicalByteArray

ryzuo:
Both extractLong methods of LogicalByteArray and Bytes classes are using bit wise shifting to extract the long value from byte array, but larger then 32 bit wise shifting will overflow. Change to use java.nio.ByteBuffer to construct and convert the long value from the source byte array.


ryzuo:
Oops, accidentally merged un-willingly commits from my local branch into this PR, close this one, then will open a new one...

PR #486, opened 2016-05-17
[TRAFODION-1853] Fix overflow of long type bit wise shifting in T4 LogicalByteArray

ryzuo:
Both extractLong() methods of LogicalByteArray and Bytes classes are using bit wise shifting to extract the long value from byte array, but larger then 32 bit wise shifting will overflow. Change to use java.nio.ByteBuffer to construct and convert the long value from the source byte array.

### DIFF
--- a/core/conn/jdbcT4/src/main/java/org/trafodion/jdbc/t4/Bytes.java
+++ b/core/conn/jdbcT4/src/main/java/org/trafodion/jdbc/t4/Bytes.java
@@ -21,6 +21,8 @@
 
 package org.trafodion.jdbc.t4;
 
+import java.nio.ByteBuffer;
+
 /**
  * <code>Bytes</code> contains a set of static methods used for byte
  * manipulation. There are three basic types of methods:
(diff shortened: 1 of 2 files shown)


selvaganesang:
Is this for extracting the column data from array of bytes sent from mxosrvr.

ryzuo:
Yes, you can see the Bytes.extractLong() is most used in all getXXXFetchXXX methods in InterfaceResultSet.java, which extract data from the protocol bytes array from mxosrvr.

selvaganesang:
I wonder if we can optimize this implementation as follows:
1) Construct a ByteBuffer for the whole array using array of bytes
2) Use get methods of ByteBuffer to extract the data from byte array to get various columns and other info

I would think this should reduce the pathlength on the driver side and possibly improve the response time for fetching each row.

ryzuo:
This is actually a good idea, but to do this, we need to change the entire protocol parsing code in driver, and some other related methods, and requires lots of test work. I rather to file another new JIRA to track this enhancement work, and just keep this one a bug fix check-in.

DaveBirdsall:
Is this ready for commit?

ryzuo:
From my perspective, I think we have nothing else to do for this bug fix PR.
Hi @selvaganesang,
Any other comment?

core/conn/jdbcT4/src/main/java/org/trafodion/jdbc/t4/Bytes.java
```diff
@@ -101,19 +103,16 @@ static long extractLong(byte[] array, int offset, boolean swap) {
...
-			}
-			
-		}else {
-			for (int shift = 56; shift >= 0; shift -= 8) {
-				value |= ( (long)( array[i] & 0xff ) ) << shift;
-				i++;
-			}
-		}
-		
+        ByteBuffer buffer = ByteBuffer.allocate(8);
```
selvaganesang: Can you please consider using ByteBuffer.wrap(array, offset, 8) to avoid allocation and then copy the data. In addition, it is not clear what was the problem with the original code. A test case or an explanation how it failed would be good to understand the issue. My understanding is that shift operator would work fine with long data type too.

PR #487, opened 2016-05-17
[TRAFODION-994]SQLColumns return incorrect values

xwq:
[TRAFODION-994]Servera1 values returned by SQLColumns are incorrect.
1. ColNullable, change to return the value from metadata table directly.
2. Remark was not support now, return a empty string.


selvaganesang:
Looks good to me.  I believe similar issue is with Type 2 JDBC driver too. Please consider taking care of this issue in type 2 JDBC driver also

xwq:
@selvaganesang Yes, JDBC type2 has. I prepared to create a new JIRA and a new pull request to track it after this one is merged.

DaveBirdsall:
Looks good to me.

PR #488, opened 2016-05-18
[TRAFODION-1956] Timeout for arkcmp process

selvaganesang:
A new CQD COMPILER_IDLE_TIMEOUT is added to control the compiler
timeout duration. Default value is 1800 seconds.

The compiler processes are killed when a SQL session ends at the
time of ODBC/JDBC disconnection. For an active session the compiler
process is killed when a compiler process remains idle more
than the compiler idle timeout.

The process in session needs to do some SQL for this feature to
kick in. For the idle connected process, the connection idle
timeout will enable killing compiler process. The compiler idle
timeout will be honored but it would take at least connection
idle timeout duration.


core/sql/cli/Context.cpp
```diff
@@ -4943,6 +4939,27 @@ Lng32 parse_statsReq(short statsReqType,char *statsReqStr, Lng32 statsReqStrLen,
...
+ 
+  if (arkcmpArray_.entries() == 0)
+     return;
+  if (arkcmpArray_[0]->getServer() == NULL)
+     return;
+  compilerIdleTimeout = getSessionDefaults()->getCompilerIdleTimeout();
+  if (compilerIdleTimeout == 0)
+     return;
+  if (currentTimestamp == -1)
+     currentTimestamp = NA_JulianTimestamp();
```
DaveBirdsall: No need to test currentTimestamp == -1, as it is initialized to -1; the 'if' is always true. Why not just initialize currentTimestamp to NA_JulianTimestamp? (Of course in the release build, the C++ compiler will optimize this out, but perhaps we can make it easier for the human reader.)

core/sql/cli/SessionDefaults.cpp
```diff
@@ -193,6 +194,8 @@ SessionDefaults::SessionDefaults(CollHeap * heap)
   setEspStopIdleTimeout(60);
   // Default is 1800 (idle ESPs time out in 30 minutes)
   setEspIdleTimeout(30*60);
+  // Default is 1800 (Compiler Idle time out in 30 minutes)
+  setCompilerIdleTimeout(30*60);
```
DaveBirdsall: What do we use the compiler for now? SQL compiles are all done in the embedded compiler. I'm guessing DDL and some utilities are the only reasons we use a separate compiler. Since these operations are relatively infrequent, I wonder if we should make the timeout shorter (say, 5 minutes) since the compiler will be idle most the time. It would reduce the process and memory footprint on the system.

DaveBirdsall:
Good change. A couple of minor comments, but OK to merge as-is from my point of view.

core/sql/cli/Context.cpp
selvaganesang: Will do

core/sql/cli/SessionDefaults.cpp
selvaganesang: Yes. DDL and update statistics would start a new compiler. We could. But, I would leave it for 30 minutes for now to avoid the possibility of increasing the regression run time due to arkcmp startup assuming the regression tests wouldn't take more than 30 minutes between consequent DDL statements.

PR #489, opened 2016-05-18
[TRAFODION-1998]JDBC Type2 column information improve

xwq:
JDBC type2 use query to get the columns information from SQL.
1. For Nullable, use the origin value from SQL
2. For remark, return empty string since not support


DaveBirdsall:
Looks good to me.

DaveBirdsall:
This is just the T2 analogy of https://github.com/apache/incubator-trafodion/pull/487, right? @selvaganesang, if you like this change I think it can be merged immediately.

selvaganesang:
Yes. We can merge @DaveBirdsall

DaveBirdsall:
Will merge it now

PR #496, opened 2016-05-20
[TRAFODION-762] support HIVE data type VARCHAR and DATE

traflm:
Add new HIVE data type support for VARCHAR and DATE.
There are more to add, this is the first iteration. 
DATE and VARCHAR are most important, and simplest to add support for.

DATE is used very popular
With VARCHAR instead of string, Trafodion can get rid of HIVE_MAX_STRING_LENGTH CQD
Please help to review.


core/sql/executor/ExExeUtilGet.cpp
```diff
@@ -5171,14 +5212,24 @@ short ExExeUtilHiveMDaccessTcb::work()
 	    str_pad(infoCol->dtQualifier, 28, ' ');
 
 	    if (infoCol->fsDatatype == REC_DATETIME)
-	      {
+	    {
+              if(infoCol->colSize > 10) {
 		// hive currently only supports timestamp
```
DaveBirdsall: Do we want to remove this comment since we support DATE now?

core/sql/executor/ExExeUtilGet.cpp
```diff
@@ -4940,6 +4946,41 @@ Lng32 ExExeUtilHiveMDaccessTcb::getLengthFromHiveColType(const char* hiveType)
...
+      }
+      else if(hiveType[i] == ')')  
+        break;
+      if(copyit == 1 )
+      {
+        maxLen[j] = hiveType[i];
+        j++;
+      }
+    }
+
```
DaveBirdsall: This code assumes that everything between '(' and ')' is a digit. Is it possible for there to be a non-digit there? (Another way to think about the question: Does the datatype info we are scanning here come from Hive? Or might it come from humans?)

core/sql/optimizer/NATable.cpp
```diff
@@ -3553,6 +3553,48 @@ NAType* getSQColTypeForHive(const char* hiveType, NAMemory* heap)
...
+      }
+    }
+    Int32 len = atoi(maxLen);
+
+    if(len == 0) return NULL;  //cannot parse correctly
+
+    NAString hiveCharset =
+        ActiveSchemaDB()->getDefaults().getValue(HIVE_DEFAULT_CHARSET);
+
+    return new (heap) SQLVarChar(CharLenInfo((hiveCharset == CharInfo::UTF8 ? 0 : len),len),
```
zellerh: This creates a type VARCHAR(n BYTES) CHARACTER SET UTF8, if HIVE_DEFAULT_CHARSET is UTF8. Hive, however, does not have this optimization, and it will allow n *characters* in the column. I hate to waste so much space, but I think we need to multiply the second "len" by the max length in bytes of a character.

I just also note that this code compares the string hiveCharset with CharInfo::UTF8, which is a number. You copied this code from the "string" case, which also may be wrong. We should be converting this to upper case and comparing it to SQLCHARSETSTRING_UTF8. It probably works by chance, the comparison is always FALSE, so we always pass "len" to the CharLenInfo constructor, but this is irrelevant for anything other than UTF8.

Here is my suggested code:

```C++
hiveCharset.toUpper();
CharInfo::CharSet hiveCharsetEnum = CharInfo::getCharSetEnum(hiveCharset);
Int32 maxNumChars = 0;
Int32 storageLen = len;
```

The following applies to the VARCHAR case only, but the code above and below should also be changed for the STRING case:

```C++
if (CharInfo::isVariableWidthMultiByteCharSet(hiveCharsetEnum))
  {
    // For Hive VARCHARs, the number specified is the max. number of characters,
    // while we count in bytes when using HIVE_MAX_STRING_LENGTH for Hive STRING
    // columns. Set the max character constraint and also adjust the required storage length.
    maxNumChars = len;
    storageLen = len * CharInfo::maxBytesPerChar(hiveCharsetEnum);
  }
```

Again, this applies to both cases:

```C++
return new (heap) SQLVarChar(CharLenInfo(maxNumChars, storageLen),
                             ...
```

Note that we are now treating STRING and VARCHAR differently. Let's say HIVE_MAX_STRING_LENGTH is set to 32.

STRING becomes VARCHAR(32 BYTES) CHARACTER SET UTF8.

VARCHAR(32) becomes VARCHAR(32) CHARACTER SET UTF8.

Is that ok or do we need still more CQDs or a different default behavior?

zellerh:
Really good to have support for these additional data types. Unfortunately, I just found a problem we need to solve, see comment.

selvaganesang:
Currently, I believe target data type conversion is missing in Hdfs scan operator. @anoopsharma00 is planning to add the target data type conversion. Can we hold this PR till it is done.

core/sql/executor/ExExeUtilGet.cpp
traflm: yes, I will.

core/sql/executor/ExExeUtilGet.cpp
traflm: Valid VARCHAR definition will only contain digit between the parenthesis, and this comes from Hive metastore, should be valid, otherwise Hive not allow the wrong varchar type to be created. And for safety, if the atoi() return 0, the code will treat this as a unknown data type.

traflm: Sorry, I am wrong, it seems ANSI support other token like BYTES in there, I need to do more tests... This is really something need to consider more.
But at least , if it is VARCHAR(10 BYTES), this code cannot support. I need to enhance it to recognize more syntax

core/sql/optimizer/NATable.cpp
traflm: I just copy the code from 'string' case, and assume it is correct :-) Need to modify the code here, yes.

I prefer to change the default behavior of hive HIVE varchar(32) into Trafodion varchar(32 bytes) character set utf8, to make it same as string. I hope to avoid another CQD. What do you think?

zellerh: Yes, I would like that, too, but the problem is that someone could store 32 Chinese or European characters in a Hive column and they would cause a string overflow in Trafodion, if Traf only allows 32 bytes.

traflm:
Hi, Selva, 
I don't know the impact of 'data type conversion in scan' with this change. I totally agree to hold this PR till that job is done. I am not fully understand the whole picture here, so if this change has some dependencies, we need to hold this PR.
But I just don't understand why it is dependent on Anoop's change, is there a JIRA to describe it?
thanks,
Ming

selvaganesang:
@ming The JIRAs are TRAFODION-1961 and TRAFODION-1164. At the time of scanning the hive data is treated like VARCHAR instead of the actual data type and this was causing the empty string not to be treated as NULL when there are numeric columns. So, target data type concept is being introduced.

traflm:
@selvaganesang thanks, I see, so I will wait and rework after that change. Logic maybe different.

core/sql/optimizer/NATable.cpp
traflm: @zellerh Yes, that will be an issue. What is your suggestion we can do? Thanks.

zellerh: One solution is to do this mapping:

Hive STRING ==> Trafodion VARCHAR(n BYTES), with n determined by HIVE_MAX_STRING_LENGTH

Hive VARCHAR(n) ==> Trafodion VARCHAR(n)

This treats STRING and VARCHAR differently, the max size of STRING is in bytes, the max size of VARCHAR is in characters. This is not ideal, users will probably expect string + CQD HIVE_MAX_STRING_LENGTH 'n' to work similar to VARCHAR(n)?

There is no good way to achieve this, though, at least not without wasting space and making incompatible changes.

I wonder what other people think.

DaveBirdsall: @zellerh, your suggestion sounds good to me. To make it a bit more clear we could change the name of the CQD to HIVE_MAX_STRING_LENGTH_IN_BYTES.

zellerh: Yes, that would be good, agree, just would require some application changes, maybe we would have to deprecate the old CQD instead of removing it. I'll also send an email to the user list to ask this question.

core/sql/executor/ExExeUtilGet.cpp
zellerh: Just wanted to mention that the BYTES syntax comes from Oracle, we decided to support it as well. I don't think Hive has it. It is not ANSI standard syntax.

traflm:
the new external table support @anoopsharma00  changed is another good way to support similar issue, this PR still has its own value to support direct Hive access. So let me fix Hans and Dave's comments and commit again.

core/sql/executor/ExExeUtilGet.cpp
```diff
@@ -4940,6 +4946,41 @@ Lng32 ExExeUtilHiveMDaccessTcb::getLengthFromHiveColType(const char* hiveType)
...
+  {
+    //try to get the length
+    char maxLen[32];
+    memset(maxLen, 0, 32);
+    int i=0,j=0;
+    short copyit = 0;
+
+    if(strlen(hiveType) > 39)  return -1;  
+ 
+    for(i = 0; i < strlen(hiveType) ; i++)
```
selvaganesang: Not sure if the compiler would optimize it to avoid calling strlen everytime.

selvaganesang:
Are we changing the HIVE_MAX_STRING_LENGTH to mean characters too?

zellerh:
Looks good to me.

core/sql/executor/ExExeUtilGet.cpp
traflm: Compiler probably not do this optimization, so let me change it. 
This is not in the hot path of sql execution, but still we should improve the code quality. So let me fix it.

traflm:
@selvaganesang the CQD change does not get widely support in the community. HIVE_MAX_STRING_LENGTH is already widely used, we should keep it and add a new CQD for now. And mark HIVE_MAX_STRING_LENGTH as deprecated. Until we make sure it is no longer used, we can remove it.
So let me add a new CQD HIVE_MAX_STRING_LENGTH_IN_BYTES as Dave suggested, and keep the current CQD unchanged.

core/sql/optimizer/NATable.cpp
```diff
@@ -3553,6 +3560,60 @@ NAType* getSQColTypeForHive(const char* hiveType, NAMemory* heap)
...
+  if ( !strncmp(hiveType, "varchar", 7) )
+  {
+    char maxLen[32];
+    memset(maxLen, 0, 32);
+    int i=0,j=0;
+    int copyit = 0;
+
+    //get length
+    for(i = 0; i < strlen(hiveType) ; i++)
+    {
```
selvaganesang: @traflm you have missed out the change to optimize strlen. I am ok if you can take care of it the next time It is not necessary to do another push commit just for this.

selvaganesang:
Is this PR ready to merge?

DaveBirdsall:
I'm happy with it.

core/sql/optimizer/NATable.cpp
traflm: @selvaganesang thanks Selva for patient review. Yes, I must pay more attention to these, and improve the coding quality next time.

DaveBirdsall:
I will merge this.

core/sql/executor/ExExeUtilGet.cpp
```diff
@@ -5171,14 +5227,23 @@ short ExExeUtilHiveMDaccessTcb::work()
...
 		infoCol->dtEndField = 6;
-	      }
+              }
+              else
+              {
+		infoCol->dtCode = SQLDTCODE_DATE;
+		infoCol->colScale = 0;
+	        str_pad(infoCol->dtQualifier, 28, ' ');
+		infoCol->dtStartField = 1;
+		infoCol->dtEndField = 6;
```
sureshsubbiah: The dtEndField should be 3 I think. Please see common/DatetimeType.h. SqlDate goes from REC_DATE_YEAR to REC_DATE_DAY. I don't if this minor error will lead to a noticeable issue. If it could be addressed in a later PR that will be good.

PR #498, opened 2016-05-23
[TRAFODION-1931]add a script to build odb for windows

xwq:
Add a script to build odb for windows. The target will be copied to
C:\Build\odb. User can modify the PACKDIR to other place if needed .

### DIFF
--- a/core/conn/odb/build.bat
+++ b/core/conn/odb/build.bat
@@ -52,5 +52,5 @@ if %ERRORLEVEL% == 0 (
 	copy /Y odb\x64\Release\odb.exe %PACKDIR%
 	echo Build windows ODB success
 ) else (
-	echo Build windows ODB fialed
+	echo Build windows ODB failed
 )
--- a/core/conn/odb/build.bat
+++ b/core/conn/odb/build.bat
@@ -1,0 +1,56 @@
+@echo off
+REM @@@ START COPYRIGHT @@@
+REM
+REM Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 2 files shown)


core/conn/odb/build.bat
```diff
@@ -0,0 +1,56 @@
...
+
+echo=
+echo ===============================
+echo     BUILD WIN64 RELEASE
+echo ===============================
+echo=
+
+echo Building ODB - Win64 Release...
+cd %BUILDDIR%\odb
+msbuild.exe /t:rebuild odb.vcxproj /p:Platform=x64 /p:Configuration=Release /p:ZlibLibDir=%ZLIB_LIB_PATH%
```
DaveBirdsall: Do you want to have an option to build a Debug version?

core/conn/odb/build.bat
```diff
@@ -0,0 +1,56 @@
...
+cd %BUILDDIR%
+if %ERRORLEVEL% == 0 (
+	copy /Y odb\x64\Release\odb.exe %PACKDIR%
+	echo Build windows ODB success
+) else (
+	echo Build windows ODB fialed
```
DaveBirdsall: Spelling error; should be "failed"

core/conn/odb/build.bat
xwq: I think debug version isn't necessary now. 

Actually , in my plan, I still need add something else like version number to ODB tool for windows. I will add the option later. Maybe I can add the option when I commit the version number if I can finish the test.

PR #499, opened 2016-05-24
[TRAFODION-2004] UPDATE STATS not supported on volatile tables

DaveBirdsall:
UPDATE STATISTICS currently does not work on volatile tables in Trafodion. Before this change, it gave confusing and varying error messages when invoked on a volatile table. This code change will cause UPDATE STATISTICS to put out error 9247 "UPDATE STATISTICS is not supported on volatile tables presently" instead.

Ironically, to get this error message to come out in all the right contexts, there were two bugs in CQD logic that needed to be addressed. It is ironic, because we'd have to fix these bugs to make UPDATE STATISTICS work for volatile tables. So, we've taken a small step toward that goal with these changes.

The bugs were in the CQD HOLD and CQD RESTORE mechanism. CQD HOLD essentially allows one to push the current value of a CQD in preparation for setting a different value temporarily. CQD RESTORE pops that value. The first bug was that the current logic implemented a stack of depth one, however there are code paths in Trafodion that do two CQD HOLDs, then two CQD RESTOREs. This caused a stack overflow; the original value going in the bit bucket. The fix was to essentially implement a stack of depth two. The second bug was that CQD RESTORE would only update the currentDefaults_ array in nadefaults.cpp. It failed to update other cached information such as currentTokens_, nor did it perform the semantic actions associated with particular CQDs. The result was that sometimes the CQD didn't really get restored. The fix was to do the normal semantic processing of a CQD at RESTORE time, instead of simply setting currentDefaults_.

These two bugs manifested in UPDATE STATISTICS by giving it an inaccurate impression of when a volatile schema was active. Sometimes UPDATE STATISTICS would think that no volatile schema was in use when in fact one was. Thus, an incorrect error message would be issued in that case if attempting UPDATE STATISTICS on a volatile table.

### DIFF
--- a/core/sql/sqlcomp/nadefaults.cpp
+++ b/core/sql/sqlcomp/nadefaults.cpp
@@ -3775,6 +3775,7 @@ void NADefaults::initCurrentDefaultsWithDefaultDefaults()
   currentTokens_	= new NADHEAP DefaultToken * [numAttrs];
   currentState_		= INIT_DEFAULT_DEFAULTS;
   heldDefaults_	        = new NADHEAP char * [numAttrs];
+  heldHeldDefaults_	= new NADHEAP char * [numAttrs];
 
   // reset all entries
   size_t i = 0;
@@ -3790,6 +3791,7 @@ void NADefaults::initCurrentDefaultsWithDefaultDefaults()
   memset( currentFloats_, 0, sizeof(float *) * numAttrs );
   memset( currentTokens_, 0, sizeof(DefaultToken *) * numAttrs );
   memset( heldDefaults_, 0, sizeof(char *) * numAttrs );
+  memset( heldHeldDefaults_, 0, sizeof(char *) * numAttrs );
 
   #ifndef NDEBUG
     // This env-var turns on consistency checking of default-defaults and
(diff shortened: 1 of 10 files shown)


selvaganesang:
Looks good to me

core/sql/sqlcomp/nadefaults.cpp
```diff
@@ -6004,10 +6013,13 @@ enum DefaultConstants NADefaults::holdOrRestore	(const char *attrName,
...
     {
-      if (heldDefaults_[attrEnum])
-	{
-	  NADELETEBASIC(heldDefaults_[attrEnum], NADHEAP);
-	}
+      if (heldHeldDefaults_[attrEnum])
+        {
+          // Gasp! We've done three successive HOLDs... it's off to
+          // the bit bucket for the deepest value
+          NADELETEBASIC(heldHeldDefaults_[attrEnum], NADHEAP);
```
zellerh: Would it be better to assert here?

zellerh:
Looks good to me, too.

core/sql/sqlcomp/nadefaults.cpp
DaveBirdsall: I considered this when developing the fix and decided against it. A person can enter CQD HOLD commands at a sqlci prompt. With an assert, sqlci would abend when a person did three such CQDs for the same default in succession. Having user input cause an assert didn't seem good to me. Another possibility is to raise an error or a warning. I will consider tihs for a later check-in.

zellerh: My preference as a user would be an assert instead of incorrect behavior. But this is probably a very little-used feature, so I doubt anyone will run into this.

PR #501, opened 2016-05-25
[TRAFODION-2014]Fix JAR conflict issue on DCS website runtime ENV

kevinxu021:


DaveBirdsall:
Looks good to me, but this is outside my area of expertise. Will commit tomorrow if there are no dissenting comments.

arvind-narain:
Thanks Kevin. Looks ok to me. Don't exactly understand all the intricacies of xml but guess we are explicitly excluding the dependencies. Following commit removed the corresponding dependencies (tomcat) earlier:
https://github.com/apache/incubator-trafodion/commit/7fe6dd13600a931986769fb30c886ca6efb7c7ee

PR #503, opened 2016-05-26
TRAFODION-2016

mashengchen:
change APACHE_1_0_X_SUPPORT to APACHE_1_1_X_SUPPORT.
vanilla hbase 1.0 and 1.1 have several different, such as entend class, method return type..
so we can only support one version at one time.
if users' hbase env is 1.0 set APACHE_1_0_X_SUPPORT="Y", not need to set APACHE_1_1_X_SUPPORT="N"
if users' hbase env is 1.1 set APACHE_1_1_X_SUPPORT="Y", not need to set APACHE_1_0_X_SUPPORT="N"

do not set APACHE_1_1_X_SUPPORT & APACHE_1_0_X_SUPPORT at the same time


svarnau:
In other cases, we support multiple versions simultaneously by generating multiple TRX files.  I'm not sure we want to add more TRX files, but I'm also not sure we want to switch how the build works based on content of install_features file.

Previously it works the other way round -- the install_features file indicates what TRX files are available, hence which distros are supported.

mashengchen:
will pull another PR

PR #505, opened 2016-05-26
[TRAFODION-2017] Use a maximum HBase row cache of 50 for USTAT sample…

DaveBirdsall:
… scans

### DIFF
--- a/core/sql/ustat/hs_globals.cpp
+++ b/core/sql/ustat/hs_globals.cpp
@@ -529,8 +529,19 @@ Int32 getMissingCols (const NABitVector &map1, const NABitVector &map2, NABitVec
 // done without a timeout, trading off possible speed improvements
 // by using a smaller cache size.
 //
-// Note that when we move to HBase 1.1, with its heartbeat protocol,
-// this time-out problem goes away and we can remove these CQDs.
+// Another issue is that it's been observed that time-outs also
+// depend on system load. Through experimentation we've discovered
+// that a value of 50 works well in loaded scenarios, assuming the
+// table is not too large. So, we use a maximum of the workable
+// value computed above and 50.
+//
+// Another note: If the user has already set HBASE_NUM_CACHE_ROWS_MAX,
+// then we don't do anything here. We respect the user's choices
+// instead.
+//
+// We had hoped that HBase 1.1, with its heartbeat protocol, would
+// solve this time-out problem for good. But early testing seems
+// to suggest that this is not the case.
 //
 // Input:
 // sampleRatio -- Percentage of rows being sampled.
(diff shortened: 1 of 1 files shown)


core/sql/ustat/hs_globals.cpp
```diff
@@ -546,10 +557,11 @@ NABoolean HSGlobalsClass::setHBaseCacheSize(double sampleRatio)
...
   if (workableCacheSize < 1)
     workableCacheSize = 1;  // can't go below 1 unfortunately
+  else if (workableCacheSize > 50)
+    workableCacheSize = 50; 
 
-  Int32 max = getDefaultAsLong(HBASE_NUM_CACHE_ROWS_MAX);
-  if ((workableCacheSize < 10000) && // don't bother if 10000 works
-      (max == 10000))  // don't do it if user has already set this CQD
+  Int32 maxDefault = getDefaultAsLong(HBASE_NUM_CACHE_ROWS_MAX);
+  if (maxDefault == 10000) // don't do it if user has already set this CQD
```
zellerh: This is existing code, but wanted to comment anyway: The code will break if we ever update HBASE_NUM_CACHE_ROWS_MAX in nadefaults.cpp and fail to update this file to the same value.

There is another way to do this check: NADefaults::getProvenance() will return INIT_DEFAULT_DEFAULTS if a default hasn't been updated from the hard-coded defaults.

I'm also wondering whether a higher value chosen by a user would be right here, given that we may use very small sample ratios. This may be different than the other workloads a user runs? Wouldn't it be better to use MIN(maxDefault, workableCacheSize)?

zellerh:
Looks good, one lower priority question.

core/sql/ustat/hs_globals.cpp
DaveBirdsall: Thanks for the suggestion for NADefaults::getProvenance(). I'll make this change.

In the case that the user sets the CQD him/herself, I wanted to give them complete freedom to experiment as they please. Higher values may make UPDATE STATS run faster (though experiements suggest not by much). The trade-off in using higher values is an increased risk of failure.

zellerh:
Looks good!

core/sql/ustat/hs_globals.cpp
```diff
@@ -546,10 +557,13 @@ NABoolean HSGlobalsClass::setHBaseCacheSize(double sampleRatio)
...
+  else if (workableCacheSize > 50)
+    workableCacheSize = 50; 
 
-  Int32 max = getDefaultAsLong(HBASE_NUM_CACHE_ROWS_MAX);
-  if ((workableCacheSize < 10000) && // don't bother if 10000 works
-      (max == 10000))  // don't do it if user has already set this CQD
+  // if the user himself set the CQD, don't do anything
+  NADefaults &defs = ActiveSchemaDB()->getDefaults();
+  if (defs.getProvenance(HBASE_NUM_CACHE_ROWS_MAX) == 
+      NADefaults::INIT_DEFAULT_DEFAULTS)
```
selvaganesang: Should we override it even if it is set in the default table.

DaveBirdsall: Thanks, @selvaganesang. Will change in the next rework.

selvaganesang:
Thanks @DaveBirdsall  Changes look good to me. Note for committers: Use --squash option to squash all the rework into one commit

PR #507, opened 2016-05-27
External lob support [TRAFODION-2008]

sandhyasun:
Changes to support new syntax to allow an external hdfs file to be specified as a LOB without having to import the lob data into trafodion. Thisis similar to Oracle's BFILE support. 

The table column needs to be specified as external during table creation. Once declared as external the column cannot later hold internal LOB data. Similarly a column declared as a regular BLOB/CLOB column without specifying the external attribute cannot later point to an external file. 
Details in JIRA 2008. 

Changes include support for insert/update, extract and select. Also includes changes to showddl and "get lob stats" command to show that a column contains external file data.


sandhyasun:
Syntax and details have been added to the attached spec.
[Trafodion LOB interface.docx](https://github.com/apache/incubator-trafodion/files/287193/Trafodion.LOB.interface.docx)

core/sql/bin/SqlciErrors.txt
```diff
@@ -409,6 +409,7 @@
 1429 ZZZZZ 99999 BEGINNER MINOR DBADMIN Inserts into _ROW_ format external hbase tables can only use the VALUES clause and must use the column_create function to create values.
 1430 3F000 99999 BEGINNER MAJOR DBADMIN A schema name that starts and ends with an "_"(underscore) is reserved for internal usage. It cannot be used to create a user schema.
 1431 ZZZZZ 99999 BEGINNER MINOR DBADMIN Object $0~String0 exists in HBase. This could be due to a concurrent transactional ddl operation in progress on this table.
+1432 ZZZZZ 99999 BEGINNER MINOR DBADMIN Input LOB type $0~Int0 does not match column's storage type : $1~Int1 Column name : $0~String0 .
```
DaveBirdsall: Nit: It's better English style to put the colons right after the word before them. E.g., "Column name:"

core/sql/comexe/ComTdbExeUtil.cpp
```diff
@@ -2417,6 +2425,15 @@ short ComTdbExeUtilLobShowddl::getLOBnum(short i)
   return lobNum;
 }
 
+NABoolean ComTdbExeUtilLobShowddl::getIsExternalLobCol(short i)
+{
+  if ((i > numLOBs_) || (i <= 0))
+    return -1;
```
DaveBirdsall: Does -1 fit in an NABoolean? Would it be evaluated as "TRUE"? I'm wondering if NABoolean is the right  return type for this method.

core/sql/comexe/ComTdbExeUtil.h
```diff
@@ -2845,35 +2846,36 @@ class ComTdbExeUtilLobShowddl : public ComTdbExeUtil
...
+  Int64 objectUID_;                                  
 
-  NABasicPtr schName_;                               // 32-39
-  short schNameLen_;                                 // 40-41
+  NABasicPtr schName_;                               
+  short schNameLen_;                                
 
-  short sdOptions_;                                  // 42-43
-
-  char fillersComTdbExeUtilLobShowddl_[4];           // 44-47
```
DaveBirdsall: I guess we don't have to count bytes in TDBs anymore since all SQL is dynamic now. :-)

core/sql/exp/ExpLOB.h
```diff
@@ -264,11 +265,11 @@ class ExpLOBoper : public ex_clause {
 
   // Generates LOB handle that is stored in the SQL row.
   // LOB handle max len:  512 bytes
-  // <flags><LOBnum><objectUid><LOBlen><descKey><descTS><chunkNum><schNameLen><schName>
-  // <--4--><--4---><----8----><---8--><---8---><--8---><----2---><---2------><--vc--->
+  // <flags><LOBType><LOBnum><objectUid><LOBlen><descKey><descTS><chunkNum><schNameLen><schName>
+  // <--4--><--4----><--4---><----8----><---8--><---8---><--8---><----2---><---2------><--vc--->
   static void genLOBhandle(Int64 uid,
```
DaveBirdsall: Is this change upward-compatible? What happens if this code is installed on an existing instance with tables that already have LOB data?

core/sql/optimizer/ItemFunc.h
```diff
@@ -2749,6 +2749,11 @@ class LOBoper : public BuiltinFunction
      hdfsPort_((Lng32)CmpCommon::getDefaultNumeric(LOB_HDFS_PORT)),
      hdfsServer_( CmpCommon::getDefaultString(LOB_HDFS_SERVER))
    {
+     if ((obj == STRING_) || (obj == BUFFER_) || (obj == FILE_))
+       lobStorageType_ = Lob_HDFS_File;
+     else if (obj == EXTERNAL_)
+       lobStorageType_ = Lob_External_HDFS_File;
+
```
DaveBirdsall: What if obj is none of these things? Is lobStorageType_ simply uninitialized?

core/sql/parser/sqlparser.y
```diff
@@ -25258,21 +25264,14 @@ optional_loggable : TOK_LOGGABLE
 					}
 
 // pElemDDLL
-optional_lobattrs : TOK_LOB TOK_STORAGE TOK_TYPE QUOTED_STRING
+optional_lobattrs :  TOK_STORAGE  QUOTED_STRING
```
DaveBirdsall: Less wordy syntax is nice.

DaveBirdsall:
Looks good to me. The only potentially serious concern is the upward/backward compatibility concern but maybe that's not a real concern.

core/sql/exp/ExpLOB.h
sandhyasun: Its ok to make changes like this for now sicne the feature isn't officially externalized yet since JDBS and driver changes are not in.

core/sql/comexe/ComTdbExeUtil.h
sandhyasun: Yes

core/sql/comexe/ComTdbExeUtil.cpp
sandhyasun: Removing the check since the calling code (ExeUtilGet.cpp) checks for the input being between 1and max num of lobs

core/sql/bin/SqlciErrors.txt
sandhyasun: Done

sandhyasun:
Investigating one regression test failure. Please hold merge.

PR #508, opened 2016-05-30
[TRAFODION-2018]T4 catalog （getProcedures and getProcedureColumns)

kevinxu021:


kevinxu021:
jenkins, retest

kevinxu021:
jenkins, retest

core/conn/jdbcT4/src/main/java/org/trafodion/jdbc/t4/T4DatabaseMetaData.java
```diff
@@ -3092,27 +3090,40 @@ public boolean dataDefinitionIgnoredInTransactions() throws SQLException {
...
-				(int) 0, // uniqueness
-				(int) 0, // accuracy
-				(short) 0, // sqlType
-				(int) 0, // metadataId
-				"", // fcatalog
-				"", // fschema
-				"" // ftable
-		);
+		return null;
+	}
```
selvaganesang: getSQLCatalogsInfo is used for all catalog APIs. It looks like the change is to do it differently for getProcedures and getProcedureColumns. Can you please provide some context about the need for this change.

arvind-narain:
Hi Kevin, Do please provide context for this change. 

Currently we handle the  catalog API queries in mxosrvr and T2 driver. Are you planning to shift that logic to all the drivers ? Is it possible to encapsulate the logic only in T2 and not make all the other drivers aware of the metadata layout? This should also address the configuration when we move away from mxosrvrs. 

Again just guessing why this change may be needed, but please do provide the reasoning.

core/conn/jdbcT4/src/main/java/org/trafodion/jdbc/t4/T4DatabaseMetaData.java
kevinxu021: As you know, it doesn't support getProcedures and getProcedureColumns and you can have a try by any Trafodion driver. Trafodion has JDBC/ODBC/ADO drivers that they has different description for metadata API. So it's not necessary to implement on server-side(MXOSRVR) by if condition( if JDBC then ... else if ODBC then ... else if ADO then ...)

kevinxu021:
Hi Arvind, Different driver has different definition of metadata(especially, what kinds of columns they want). And yes, I'll move all of them into driver-side exception ODBC. As you see, these two APIs were not supported and one of our customer required them, so started from there.

kevinxu021:
By the way, procedure list can't be shown on DBVisualizer before. Now it's works with this fixes.

core/conn/jdbcT4/src/main/java/org/trafodion/jdbc/t4/T4DatabaseMetaData.java
selvaganesang: I strongly feel that catalog API should be encapsulated in the server side. It would become hard to manage on the client side when there are changes in the metadata. Can you please highlight the expected difference between the drivers. We can handle it on case-by-case basis and possibly map it on the client side.

kevinxu021: For ODBC: https://msdn.microsoft.com/en-us/library/ms715368(v=vs.85).aspx
For JDBC: http://docs.oracle.com/javase/8/docs/api/     java.sql -> DatabaseMetaData -> getProcedures(..)
As you see, it's different between JDK1.7 and ODBC3.0. Do you want to add if-else on server-side? If any changes, I think it's pretty easy to upgrade drivers instead of server for customer, right? It should be nice for server to treat it as a common SELECT command instead of extra maintenance work. My suggestion is to put it into drivers, not sure how many differences they will have between driver in the future.

selvaganesang: I would politely disagree with the statement updating a client driver is easy especially when it involves hundreds of clients. Also, it not clear to me why JDBC/ODBC needs to be serviced by the same statement with if and else etc. Isn't possible to create many different statement like the way it is being done in the client side on the server side. I would wait for the community to provide the input before we push through this PR.

kevinxu021: It's so obvious for getTables API(https://msdn.microsoft.com/en-us/library/ms711831(v=vs.85).aspx), ODBC has 5 columns while JDBC has 10 columns. On server-side you may notice there is an if-else for them. Is it reasonable?

selvaganesang:
The metadata queries shouldn't be influenced by the settings in the user context. The user application could have used CQDs, Control query shape and other settings to influence the behavior of the user query, It can make the catalog APIs to behave differently. Also, in the past we have used control query shape to influence and produce optimal  query plan for the metadata queries. In the days before Trafodion, we had used module file concept to isolate the metadata queries from being influenced by the user settings. In Trafodion, we need to have such isolation, otherwise we can get non-deterministic results/errors from these queries. The best place to do this isolation is the server side. 

It is good that DBVisualizer works with this change. But, we need to look at the change in a holistic way to make Trafodion robust.

PR #509, opened 2016-05-31
[TRAFODION-2004][TRAFODION-1775] Two follow-on fixes

DaveBirdsall:
[TRAFODION-2004] In the review for https://github.com/apache/incubator-trafodion/pull/499, @zellerh suggested adding an assert when CQD HOLDs overflow their stack. This has now been done.
[TRAFODION-1775] Pull request https://github.com/apache/incubator-trafodion/pull/292 (released in release 2.0) added a new CQD, USTAT_YOULL_LIKELY_BE_SORRY. Unfortunately it was added as a 32-bit integer, which meant that if one wanted to set it to 10 billion, then that would fail with a validation error. But 10 billion is a reasonable size for a table. This code change changes the validation rule to be a float greater than or equal to 1. So now 10 billion would be possible.

### DIFF
--- a/core/sql/sqlcomp/NADefaults.h
+++ b/core/sql/sqlcomp/NADefaults.h
@@ -398,14 +398,10 @@ private:
   float            **currentFloats_;
   DefaultToken	   **currentTokens_;
 
+  class HeldDefaults;  // forward reference for helper class used below
+
   // these default values were 'held' through a cqd HOLD stmt.
-  char             **heldDefaults_;
-
-  // and these default values were 'held' through a sequence of two HOLD stmts.
-  char             **heldHeldDefaults_;
-
-  // if there are three HOLD stmts in succession, the first set of values go
-  // into the bit bucket.
+  HeldDefaults     **heldDefaults_;
 
   Provenance       currentState_;
   LIST(NAString)   tablesRead_;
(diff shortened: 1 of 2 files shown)


DaveBirdsall:
jenkins, retest

DaveBirdsall:
Do not merge this fix just yet. It appears the assert is actually hitting in a few of the regression tests (unexpectedly!).

zellerh:
Code looks good to me so far, hope you won't have too much trouble getting the regressions to work (will we need a third level??).

DaveBirdsall:
Yeah, @zellerh, the assert is firing under some DDL operations. So I know I need a three-deep stack at least. Or perhaps we are simply too promiscuous with our HOLDs and RESTOREs. Debugging it now.

DaveBirdsall:
I've pushed a rework that makes it a stack of depth 3. Executor regressions now pass. Will run the other regressions as well while this PR is under review.

zellerh:
Thanks, looks good, and sorry for being the troublemaker that caused you extra work fixing old bugs.

DaveBirdsall:
No problem. The code is better off for it.

selvaganesang:
Looks good to me.

DaveBirdsall:
jenkins, extra tests

selvaganesang:
Looks good to me. Good rework.

DaveBirdsall:
The failures in the extra tests were to executor/TEST015 and hive/TEST003 and hive/TEST018 but only on the Cloudera side. On the Hortonworks side they pass. I will merge the change.

PR #510, opened 2016-05-31
[TRAFODION-2009] fix parallel scanner issue not scanning full regions…

eowhadi:
…. Was linked with cache size handling. Cache was kept same size as original cache size on parallele scanner, now it is divided by the number of threads.

Added a test case in regression test to validate the fix.

### DIFF
--- a/core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/TrafParallelClientScanner.java.tmpl
+++ b/core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/TrafParallelClientScanner.java.tmpl
@@ -116,21 +116,22 @@ public class TrafParallelClientScanner extends AbstractClientScanner implements 
     LOG.debug("Found "+locs.size()+" region(s).");
     if (locs.size() == 0) return;
 
+    int threads = parallelScaling>1.0 ?(int)Math.ceil(parallelScaling) : (int)Math.ceil(locs.size() * parallelScaling);
+
     Map<HRegionLocation, Queue<Scan>> tasks = new HashMap<HRegionLocation, Queue<Scan>>(locs.size());
     int i=0;
     // organize region locations by region server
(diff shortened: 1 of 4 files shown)


core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/TrafParallelClientScanner.java.tmpl
```diff
@@ -116,21 +116,22 @@ public class TrafParallelClientScanner extends AbstractClientScanner implements
     LOG.debug("Found "+locs.size()+" region(s).");
     if (locs.size() == 0) return;
 
+    int threads = parallelScaling>1.0 ?(int)Math.ceil(parallelScaling) : (int)Math.ceil(locs.size() * parallelScaling);
+
```
DaveBirdsall: Not sure I understand the logic behind this. Suppose locs.size() is 3. If parallelScaling is 1.1, then threads gets 2. If parallelScaling is 1.0, then threads gets 3. Seems odd that parallelScaling getting smaller should result in more threads.

eowhadi: there is 2 ways to set DOP: as a % of regions to scan, or define the number of thread manually.
parallelScaling = 0 disabled.
0<parallelScaling<=1.0, DOP expressed as percentage of regions to scan. 1.0 means 100%, .2 means 20%,
parallelScanner > 1.0, parallelScanner should be an integer number representing the number of thread you want. If you define parallelScanner > 1 but not integer, we convert using ceiling...

DaveBirdsall: Got it. Thanks.

DaveBirdsall:
Looks good to me.

PR #511, opened 2016-05-31
JIRA TRAFODION-2006  Hive table structure and data changes are now detected correctly.

anoopsharma00:


core/sql/optimizer/NATable.cpp
```diff
@@ -7308,6 +7308,16 @@ NATable * NATableDB::get(const ExtendedQualName* key, BindWA* bindWA, NABoolean
      }
   }
 
+  // the reload cqd will be set during aqr after compiletime and runtime
+  // timestamp mismatch is detected.
+  // If set, reload hive metadata.
+  if ((cachedNATable->isHiveTable()) &&
+      (CmpCommon::getDefault(HIVE_DATA_MOD_CHECK) == DF_ON) &&
+      (CmpCommon::getDefault(TRAF_RELOAD_NATABLE_CACHE) == DF_ON))
```
DaveBirdsall: Will this cause a reload *all* NATable information? It's too bad we don't have some way to limit the reload to just the tables that have a metadata mismatch.

core/sql/regress/hive/EXPECTED003
```diff
@@ -138,6 +132,10 @@ P_PROMO_SK   P_PROMO_ID                 P_START_DATE_SK  P_END_DATE_SK  P_ITEM_S
 
 *** ERROR[8822] The statement was not prepared.
 
+*** WARNING[8597] Statement was automatically retried 1 time(s). Delay before each retry was 0 seconds. See next entry for the error that caused this retry.
```
DaveBirdsall: Do we want to print out these warnings when the retry failed for the same reason as the original error? Seems like unnecessary repetition.

DaveBirdsall: Oh, sorry... ignore my comment. Looks like this happens only when we set a CQD?

DaveBirdsall:
Looks good to me.

core/sql/optimizer/NATable.cpp
anoopsharma00: For non-hive trafodion tables, query invalidation feature detects when a particular
table structure has changed and invalidates natable for that table only.
 
For hive tables, table modification detection is based on detection of timestamp 
change of the hive dirs since query invalidation feature does not apply to hive
tables as they could be created/populated from outside of trafodion. 
Also unlike traf tables, we need to know when data has changed for a
hive table as optimizer uses that information to generate data access maps.

So aqr is done when timestamp change is detected for hive data/struct dirs.
AQR granularity is at query level and NATables for all tables that are part of the 
query which is being AQR'd will be reloaded. 
Tables that are not part of that query will not be reloaded.

DaveBirdsall: Thanks for the explanation. Good change.

core/sql/executor/ExHdfsScan.cpp
```diff
@@ -410,16 +411,81 @@ ExWorkProcRetcode ExHdfsScanTcb::work()
...
             else
-              step_ = DONE;
+              step_ = CHECK_FOR_DATA_MOD_AND_DONE;
 	  }
 	  break;
 
+        case CHECK_FOR_DATA_MOD:
+        case CHECK_FOR_DATA_MOD_AND_DONE:
+          {
+            char * dirPath = hdfsScanTdb().hdfsRootDir_;
```
zellerh: Is this a single directory for the table? We may or may not completely support that yet, but partitioned tables can be spread over multiple directories outside of the location that's specified in the CREATE TABLE statement.

anoopsharma00: At this time, data modification check for partitioned tables is not being
supported. Once we support partitioned tables, we will need to enhance
code to handle them.

Here is the code in GenRelScan.cpp disabling it:

  // partitioned files. Skip data mod check for them.
 +      if (numOfPartLevels > 0)
 +        hdfsRootDir = NULL;

zellerh: Thanks, makes sense. I also see the comment now that explains how to do this in the future.

zellerh:
Thanks, this plugs a big hole in our design! I didn't look at every nook and cranny, but what I saw looks good to me.

PR #512, opened 2016-05-31
[TRAFODION-2020] Safer APIs to manipulate Sql_ParserFlags

DaveBirdsall:
This code change provides new, safer APIs to manipulate the Sql_ParserFlags global.

Or_SqlParser_Flags is intended to replace Set_SqlParser_Flags. It simply turns on whatever bits are requested. If the input parameter is zero, it does nothing, in contrast to Set_SqlParser_Flags, which turns all the bits off in that case.

UnOr_SqlParser_Flags is intended to replace Reset_SqlParser_Flags. It simply turns off whatever bits are requested. If the input parameter is zero, it does nothing, in contrast to Reset_SqlParser_Flags, which turns off all the bits in that case.

For the common use case of a method that simply wants to turn on a few flags then return them to their entry-time state, the new class PushAndSetSqlParserFlags is provided. By coding a single line, one can turn on some flags, and be guaranteed that the flags will return to their original value at method exit (provided exit is via normal flow or via C++ exceptions). Thanks to @nonstop-qfchen and @selvaganesang for suggesting this idea.

This code change also changes the Update Statistics code to use this new class, replacing calls to the older APIs. Over time all calls to the older APIs can be replaced.

### DIFF
--- a/core/sql/parser/SqlParserGlobalsCmn.h
+++ b/core/sql/parser/SqlParserGlobalsCmn.h
@@ -136,6 +136,12 @@ enum SqlParser_Flags_Enum {
   inline static ULng32 Get_SqlParser_Flags(ULng32 flagbits)
   { return SqlParser_Flags & flagbits; }
 
+  // Deprecated; use Or_SqlParser_Flags or the PushAndSetSqlParserFlags
+  // class instead.
+  // This method tends to be error-prone because callers often are
+  // not aware of the different semantics when flagbits are zero.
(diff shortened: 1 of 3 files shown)


nonstop-qfchen:
Looks good!

zellerh:
Looks good to me, too.

sandhyasun:
This is a neat way of setting the parserflags - much clearer and easier to use. Looks good !

PR #513, opened 2016-06-01
[TRAFODION-1684]ODBC:preliminary support SQLStatistics

xwq:
[TRAFODION-1684]ODBC: trafodion doesn't support API SQLStatistics.
It's preliminary implementation for ODBC API SQLStatistics. Now, it
will ignore the parameter "Unique".
SQLStatistics returns information as a standard result set.
 TABLE_CAT        # support
 TABLE_SCHEM      # support
 TABLE_NAME       # support
 NON_UNIQUE       # support
 INDEX_QUALIFIER  # not support
 INDEX_NAME       # not support
 TYPE             # not support
 ORDINAL_POSITION # support
 ASC_OR_DESC      # not support
 CARDINALITY      # support
 PAGES            # not support
 FILTER_CONDITION # not support

### DIFF
--- a/core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
+++ b/core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
@@ -4811,6 +4811,64 @@ odbc_SQLSvc_GetSQLCatalogs_sme_(
 
 			break;
 
+        case SQL_API_SQLSTATISTICS:
+            if (!checkIfWildCard(catalogNm, catalogNmNoEsc) && !metadataId)
+            {
+                exception_->exception_nr = odbc_SQLSvc_GetSQLCatalogs_ParamError_exn_;
+                exception_->u.ParamError.ParamDesc = SQLSVC_EXCEPTION_WILDCARD_NOT_SUPPORTED;
+            }
+            if (tableNm[0] != '$' && tableNm[0] != '\\')
+            {
+                if (strcmp(catalogNm, "") == 0)
+                    strcpy(tableName1, SEABASE_MD_CATALOG);
+                else
+                    strcpy(tableName1, catalogNm);
+            }
+
+            tableParam[0] = tableName1;
+            convertWildcard(metadataId, TRUE, schemaNm, expSchemaNm);
+            convertWildcardNoEsc(metadataId, TRUE, schemaNm, schemaNmNoEsc);
+            convertWildcard(metadataId, TRUE, tableNm, expTableNm);
+            convertWildcardNoEsc(metadataId, TRUE, tableNmNoEsc, tableNmNoEsc);
+            inputParam[0] = schemaNmNoEsc;
+            inputParam[1] = expSchemaNm;
+            inputParam[2] = tableNmNoEsc;
+            inputParam[3] = expTableNm;
+
+            snprintf(CatalogQuery, sizeof(CatalogQuery),
+                    "select "
+                    "cast('%s' as varchar(128)) TABLE_CAT, "
+                    "cast(trim(ob.SCHEMA_NAME) as varchar(128)) TABLE_SCHEM, "
+                    "cast(trim(ob.OBJECT_NAME) as varchar(128)) TABLE_NAME, "
+                    "cast(0 as smallint) NON_UNIQUE, " // not support
+                    "cast('' as varchar(128)) INDEX_QUALIFIER, " // not support
+                    "cast('' as varchar(128)) INDEX_NAME, "
+                    "cast(0 as smallint) TYPE, " // not support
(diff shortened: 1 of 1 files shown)


xwq:
It's not a fully support for SQLStatistics. I think we can merge it to make the ODBC preliminary support for SQLStatistics. 
I will continue investigating metadata table and make it support more result in the furture.

core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
```diff
@@ -4811,6 +4811,64 @@ odbc_SQLSvc_GetSQLCatalogs_sme_(
...
+                    "TRAFODION.\"_MD_\".OBJECTS ob, "
+                    "TRAFODION.\"_MD_\".COLUMNS co, "
+                    "TRAFODION.%s.sb_histograms sb "
+                    "where "
+                    "ob.OBJECT_UID = co.OBJECT_UID "
+                    "and co.COLUMN_NUMBER = sb.COLUMN_NUMBER "
+                    "and (ob.SCHEMA_NAME = '%s' or trim(ob.SCHEMA_NAME) LIKE '%s' ESCAPE '\\')  "
+                    "and (ob.OBJECT_NAME = '%s' or trim(ob.OBJECT_NAME) LIKE '%s' ESCAPE '\\') "
+                    "and (ob.OBJECT_TYPE in ('BT', 'VI')) "
+                    "and (trim(co.COLUMN_CLASS) not in('S', 'M'));",
```
DaveBirdsall: Earlier in the query there is a CASE expression on co.COLUMN_CLASS that looks for 'S'. However, here we've eliminated rows with 'S'. So, that CASE expression seems not to be needed.

core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
```diff
@@ -4811,6 +4811,64 @@ odbc_SQLSvc_GetSQLCatalogs_sme_(
...
+                    "cast(sb.rowcount as integer) CARDINALITY, "
+                    "cast(0 as integer) PAGES, " // not support
+                    "cast('' as varchar(128)) FILTER_CONDITION " // not support
+                    "from "
+                    "TRAFODION.\"_MD_\".OBJECTS ob, "
+                    "TRAFODION.\"_MD_\".COLUMNS co, "
+                    "TRAFODION.%s.sb_histograms sb "
+                    "where "
+                    "ob.OBJECT_UID = co.OBJECT_UID "
+                    "and co.COLUMN_NUMBER = sb.COLUMN_NUMBER "
```
DaveBirdsall: I'm wondering how this query will behave when there are multi-column histograms. Will we get extra rows?

xwq: Thanks Dave, I miss "co.OBJECT_UID=sb.TABLE_UID", I will add this.

For mutlit-column histograms, I don't known the exactly behavior. I will take some time to investigate it. 

Actually , I can't find the  document for metadata and histograms. I am studying through learning the query for other catalog API.  So this question maybe has many issues, and there are many result it can't support now.  I think your comments will help me studying them.

DaveBirdsall: Unfortunately, I don't know of a document either. I've learned the format of the data in the Histograms table by reverse-engineering it, along with a little bit of code reading. Here's what I know: For a multi-column histogram, there will be n rows in the SB_HISTOGRAMS table, where n is the number of columns in the histogram. One way to see this is to create a table with a primary key, put some rows into it (100 rows, let's say),then do UPDATE STATISTICS ON EVERY COLUMN. After that, do SHOWSTATS ON EVERY COLUMN. That will show you the table_uid for that table. Then do "select histogram_id, count(*) from sb_histograms where table_uid = <the value from SHOWSTATS> group by histogram_id". You'll notice some histogram_id's with counts of one. Those are the single column histograms. You'll also notice a few with counts greater than one. These are the multi-column histograms. If you then look at the rows for that particular histogram_id, you'll see there is a row for each column in the histogram. It is a denormalized representation.

xwq: Thanks from your comments. Now I know what is multi-columns histograms.

From the ODBC standard , I think it only need return the signal-column histogram now. I thinks I can use "colcount=1" to filter signal-column histogram out . 

I want to know whether "COLCOUNT=1" is right or not here?

DaveBirdsall: Yes, COLCOUNT=1 looks right. Sorry for not responding sooner.

DaveBirdsall:
Looks good.

PR #514, opened 2016-06-01
Minor Expected file change

sandhyasun:
The diff was caused by space being removed in an error messaage in front of a  colon. 
 1432 ZZZZZ 99999 BEGINNER MINOR DBADMIN Input LOB type $0~Int0 does not match column's storage type: $1~Int1 Column name: $0~String0 .


svarnau:
looks good.

PR #516, opened 2016-06-02
JIRA TRAFODION-2022 Add support for purgedata of hive tables

anoopsharma00:
In addition to purgedata:
-- showplan correctly shows aqr setting
-- data modification check is skipped during insert if overwrite
   option was used

### DIFF
--- a/core/sql/optimizer/RelExeUtil.cpp
+++ b/core/sql/optimizer/RelExeUtil.cpp
@@ -4981,9 +4981,43 @@ RelExpr * ExeUtilFastDelete::bindNode(BindWA *bindWA)
     {
       // do not do override schema for this
       bindWA->setToOverrideSchema(FALSE);
-
+      
       naTable = bindWA->getNATable(getTableName());
-      if (getTableName().isSeabase())
+      if (getTableName().isHive())
+        {
+          if (! naTable)
+            {
+              *CmpCommon::diags() << DgSqlCode(-4222) << DgString0("PURGEDATA");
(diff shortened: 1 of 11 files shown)


core/sql/optimizer/RelExeUtil.cpp
```diff
@@ -4981,9 +4981,43 @@ RelExpr * ExeUtilFastDelete::bindNode(BindWA *bindWA)
...
+          NAString tableDir;
+
+          NABoolean result = ((HHDFSTableStats* )hTabStats)->splitLocation
+            (hiveTablePath, hostName, hdfsPort, tableDir) ;       
+          if (!result) 
+            {
+              *CmpCommon::diags() << DgSqlCode(-4224)
+                                  << DgString0(hiveTablePath);
+              bindWA->setErrStatus();
+              return this;
```
DaveBirdsall: Shouldn't this be "return NULL"?

DaveBirdsall:
Looks good to me.

core/sql/optimizer/RelExeUtil.cpp
anoopsharma00: Binder component returns 'this' at some places and 'null' at others
in case of an error. At some point it probably started with one or the other
but over time it has become inconsistent.

But it will/should always set bindWA->setErrStatus() before return.
And caller must check for bindWA->getErrStatus() to see 
if an error occurred and if the caller must act on that error.

PR #517, opened 2016-06-02
Fix problems with use of HBase row count estimate

blfritch-esgyn:
The CQD ESTIMATE_HBASE_ROW_COUNT is no longer used anywhere but in
Update Stats, which also uses USTAT_ESTIMATE_HBASE_ROW_COUNT. Both
have to be set before the row count is estimated, and the flawed
logic used in this decision sometimes results in estimation being
performed twice. The former CQD has been removed, the latter set
to ON by default, and the decision logic has been cleaned up.
Also, the estimation is avoided if it is known that it will not
be used (e.g., if sampling is not being used).


zellerh:
Looks good to me.

DaveBirdsall:
Looks good to me also. Thank you for this clean-up.

PR #518, opened 2016-06-02
[TRAFODION-2021] Internal error on upsert with index

zellerh:
In this particular case, the extra nodes added by index maintenance
and the WHERE clause interacted in a way that caused some missing
characteristic inputs, leading to the generator error.  The fix is to
add the required inputs not directly to the nodes, but as "outer
references" to the BindScope, so that they get added automatically to
any nodes created in that BindScope. Since the predicates get created
by a method called at a higher level, I also needed to add a "remote
control" feature for the BindScope of a MergeUpdate node.

Also: [TRAFODION-2026] Upsert fails on table with nullable key column.

When creating the predicate that links the target clustering key with
the corresponding source values, use "special nulls" semantics for
nullable key columns, so a comparesion NULL = NULL is allowed and
yields TRUE.

### DIFF
--- a/core/sql/optimizer/BindRelExpr.cpp
+++ b/core/sql/optimizer/BindRelExpr.cpp
@@ -10218,6 +10218,9 @@ NABoolean Insert::isUpsertThatNeedsMerge(NABoolean isAlignedRowFormat, NABoolean
      return FALSE;
 }
 
+// take an insert(src) node and transform it into
+// tsj_flow(src, merge_update(input_scan))
+// with a newly created input_scan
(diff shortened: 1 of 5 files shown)


DaveBirdsall:
Looks good to me. The added comments are a BIG help to following the code. I don't know the answer to your one comment, "is this OK?" though.

zellerh:
Yes, sorry for just putting that question there. I saw this behavior while debugging, but am not sure whether that's an observable bug (or a bug at all), since I'm not that familiar with the binder design.

PR #519, opened 2016-06-02
[TRAFODION-1585] Allow MDAM on inner child of NJ sometimes

DaveBirdsall:
If it is known that there will be exactly one probe of the inner child of a nested join, this change enables consideration of MDAM for that inner child.

This is a replacement of https://github.com/apache/incubator-trafodion/pull/174. It takes the suggestion from @zellerh to use the input logical properties instead of adding constraint information to the input physical properties. The result is a much simpler code change. I will close the older pull request shortly.

### DIFF
--- a/core/sql/optimizer/ScanOptimizer.cpp
+++ b/core/sql/optimizer/ScanOptimizer.cpp
@@ -3462,6 +3462,12 @@ ScanOptimizer::isMdamEnabled() const
       return TRUE;
     }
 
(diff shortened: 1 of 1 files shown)


zellerh:
Looks good to me, thanks for coding the new solution!

DaveBirdsall:
And thank you for the suggestion. Sorry it took so long to get back to it.

PR #520, opened 2016-06-02
Testing if adding Trafodion user id to HBase group will fix tests

amandamoran:


zellerh:
+1

Good thing to try. Probably somebody pointed it out already, in a secure environment where regular HBase and Trafodion are mixed, the system administrators may not want to give Trafodion this blanket permission. But, it will be good to figure out a solution first and then improve it.

selvaganesang:
+1 Thanks for doing this @amandamoran If Trafodion needs to use snapshot, it might need to have this kind of blanket permission. If this is the solution, Trafodion needs to suppress the snapshot usage if this permission is not set.

amandamoran:
jenkins, extra tests

amandamoran:
jenkins, extra tests

sandhyasun:
The Extra Test failure in hive is due to aanother minor difference in results. So this change looks good.

sureshsubbiah:
+1

PR #523, opened 2016-06-05
TRAFODION [2025] Initialize authorization cannot be run in a DDL transaction

robertamarton:
As part of DDL transaction work, there is a goal to run all DDL and DDL like
operations in a single DDL transaction.  This delivery changes initialize
trafodion and initialize authorization to make this happen.

Prior to DDL transactions, initialize authorization would either add or drop
authorization support.  Part of this required that all compiler processes 
associated with the master process be killed to clear out information stored
in memory.  When DDL transactions were added, killing the compiler processes
caused the DDL transaction to abort.  This means that initialize authorization
could not be run in DDL transations.  Initialize trafodion calls initialize
authorization when security is enabled, so initialize trafodion was not 
enabled to run in DDL transactions.


core/sql/cli/Context.cpp
```diff
@@ -3850,6 +3856,48 @@ void ContextCli::createMxcmpSession()
...
+// Method:  updateMxcmpSession
+//
+// Updates security attributes in child arkcmp
+//
+// Returns: 0 = succeeded; -1 = failed
+// ----------------------------------------------------------------------------
+Int32 ContextCli::updateMxcmpSession()
+{
+  // If no child arkcmp, just return
+  if (getArkcmp()->getServer() == NULL)
```
sandhyasun: Do you need to check if getArkcmp() is NULL as well ?

core/sql/cli/Context.cpp
```diff
@@ -3850,6 +3856,48 @@ void ContextCli::createMxcmpSession()
...
+  //     database user name
+  // See CmpStatement::process (CmpMessageDatabaseUser) for more details
+  Int32 userAsInt = (Int32) databaseUserID_;
+  char userMessage [MAX_AUTHID_AS_STRING_LEN + 1 + MAX_USERNAME_LEN + 1 + 2];
+  str_sprintf(userMessage, "%d,%d,%s", authOn, userAsInt, databaseUserName_);
+  char *pMessage = (char *)&userMessage;
+
+  // Send message to child arkcmp, if one exists
+  ExSqlComp::ReturnStatus cmpStatus;
+  ExSqlComp *exSqlComp = getArkcmp();
```
sandhyasun: I think createMAxcmpSession retrieves the complier by indexing into the indexIntoCompilerArray. You may want to keep it consistent with that although I think the index is usually 0.

core/sql/cli/Context.cpp
```diff
@@ -3850,6 +3856,48 @@ void ContextCli::createMxcmpSession()
...
+  ex_assert(exSqlComp, "CliGlobals::getArkcmp() returned NULL");
+  cmpStatus = exSqlComp->sendRequest(EXSQLCOMP::DATABASE_USER,
+                     (const char *) pMessage,
+                     (ULng32) strlen(pMessage));
+  if (cmpStatus == ExSqlComp::ERROR)
+    return -1;
+
+  return 0;
+}
+
```
sandhyasun: The new method above sends a message to the first level external arkcmp if any. Is there a possibility there may be multiple levels of arkcmps ?

core/sql/cli/Context.cpp
robertamarton: At the time ContextCli is instantiated, the first entry in arkcmpArray_ is created.  So there should always be entry 0.  The method getArkcmp() returns the 0 entry if none is specified.  Therefore, getArkcmp() should not be NULL.

core/sql/cli/Context.cpp
robertamarton: I noticed that the indexIntoCompilerArray is sometimes used. But if there is only one arkcmp associated with the process, there will only be entry 0.  Several other places in the code just call getArkcmp() by itself without using indexIntoCompilerArray.

core/sql/cli/Context.cpp
robertamarton: We are trying to remove the necessity of using arkcmps.  At this time, I cannot think of any reason why there would be more than one.

DaveBirdsall: With UPDATE STATISTICS, I routinely notice that there is a child tdm_arkcmp, a grand-child tdm_arkcmp and a great-grandchild tdm_arkcmp.

core/sql/arkcmp/CmpStatement.cpp
```diff
@@ -1254,17 +1255,34 @@ CmpStatement::process(const CmpMessageDatabaseUser &statement)
 {
   NABoolean doDebug = FALSE;
 
-  NAString message = statement.data();
+  // The message contains the following:
```
nonstop-qfchen: Should we check the length of the message to make sure it covers the various fields checked here?

core/sql/cli/Context.cpp
```diff
@@ -3718,19 +3719,27 @@ void ContextCli::createMxcmpSession()
...
+  // The message contains the following:
+  //   (auth state and user ID are delimited by commas)
+  //     authorization state (0 - off, 1 - on)
+  //     integer user ID
+  //     database user name
+  // See CmpStatement::process (CmpMessageDatabaseUser) for more details
   Int32 userAsInt = (Int32) databaseUserID_;
-  char userMessage [MAX_AUTHID_AS_STRING_LEN + 1 + MAX_USERNAME_LEN + 1];
-  str_sprintf(userMessage, "%d,%s", userAsInt, databaseUserName_);
+  char userMessage [MAX_AUTHID_AS_STRING_LEN + 1 + MAX_USERNAME_LEN + 1 + 2];
```
nonstop-qfchen: The length of userAsInt (12 bytes) should be included as well?

core/sql/cli/Context.cpp
```diff
@@ -3850,6 +3856,48 @@ void ContextCli::createMxcmpSession()
...
+  // See CmpStatement::process (CmpMessageDatabaseUser) for more details
+  Int32 userAsInt = (Int32) databaseUserID_;
+  char userMessage [MAX_AUTHID_AS_STRING_LEN + 1 + MAX_USERNAME_LEN + 1 + 2];
+  str_sprintf(userMessage, "%d,%d,%s", authOn, userAsInt, databaseUserName_);
+  char *pMessage = (char *)&userMessage;
+
+  // Send message to child arkcmp, if one exists
+  ExSqlComp::ReturnStatus cmpStatus;
+  ExSqlComp *exSqlComp = getArkcmp();
+  ex_assert(exSqlComp, "CliGlobals::getArkcmp() returned NULL");
```
nonstop-qfchen: If exSqlComp == null is normal (which I think it is), then we should return 0 here instead of ex_assert().

core/sql/cli/ExSqlComp.cpp
```diff
@@ -677,14 +677,19 @@ ExSqlComp::ReturnStatus ExSqlComp::resendControls(NABoolean ctxSw)   // Genesis
...
+    //     database user name
+    // See CmpStatement::process (CmpMessageDatabaseUser) for more details
     Int32 *userID = ctxt->getDatabaseUserID();
     Int32 userAsInt = *userID;
-    char userMessage [MAX_AUTHID_AS_STRING_LEN + 1 + MAX_USERNAME_LEN + 1];
-    str_sprintf(userMessage, "%d,%s", userAsInt, ctxt->getDatabaseUserName());
+    CmpContext *cmpCntxt = CmpCommon::context();
+    NABoolean authOn = cmpCntxt ? cmpCntxt->isAuthorizationEnabled() : FALSE;
+
+    char userMessage [MAX_AUTHID_AS_STRING_LEN + 1 + MAX_USERNAME_LEN + 1 + 2];
```
nonstop-qfchen: Same comment on the length for userAsInt.  Maybe refactor the two instances of the usage into a method.

core/sql/sqlcomp/CmpSeabaseDDLcommon.cpp
```diff
@@ -7764,52 +7734,103 @@ short CmpSeabaseDDL::initSeabaseAuthorization(
...
+    if (beginXnIfNotInProgress(cliInterface, xnWasStartedHere))
+    {
+      SEABASEDDL_INTERNAL_ERROR("initialize authorization");
+      return -1;
+    }
+  }
 
-  return cliRC;
+  // change authorization status in compiler contexts
+  //CmpCommon::context()->setAuthorizationState (1);
```
nonstop-qfchen: Should we remove this commented out line?

core/sql/sqlcomp/CmpSeabaseDDLcommon.cpp
```diff
@@ -7764,52 +7734,103 @@ short CmpSeabaseDDL::initSeabaseAuthorization(
...
-    cliRC = -1; 
     if (CmpCommon::diags()->getNumber(DgSqlCode::ERROR_) == 0)
-     SEABASEDDL_INTERNAL_ERROR("drop authorization");
+      SEABASEDDL_INTERNAL_ERROR("drop authorization");
+    return;
   }
-  else
+
+  // Turn off authorization in compiler contexts
+  GetCliGlobals()->currContext()->setAuthStateInCmpContexts(FALSE, FALSE);
```
nonstop-qfchen: I wonder if a method to turn off/on authorization in compiler, including all contexts and all remote arkcmps, is worthy it.

nonstop-qfchen:
Looks good to me!

core/sql/cli/Context.cpp
robertamarton: We should be good for the size:
MAX_AUTHID_AS_STRING_LEN is the size for userAsInt converted to a string, MAX_USERNAME_LEN is the username length, and the remaining should cover commas and authorization state.  
and 2 it the length of authorization state.
plus all the commas
we should be good for the size

core/sql/cli/Context.cpp
robertamarton: From what I read in the code, the ContextCli constructor allocates the arkcmpArray_ structure and creates an ExSqlComp as the 0 entry.  Therefore, getArkcmp should not return null.  There is code that kills all the arkcmp processes, but they are always restarted.  The way to tell if a child arkcmp exists, is to look at the getServer() in the returned ExSqlCmp class.

core/sql/cli/Context.cpp
robertamarton: I see the situation:  sqlci ->arkcmp1; arkcmp1 -> arkcmp2; arkcmp2 -> arkcmp3 ...  But I don't see any support for sqlci -> arkcmp1 & arkcmp2; arkcmp1 -> arkcmpa & arkcmpb; arkcmp2 -> arkcmpc & arkcmpd.  I don't check to see of the arkcmpArray_ has more than one entry.  May want to do this ...

core/sql/sqlcomp/CmpSeabaseDDLcommon.cpp
robertamarton: Yes, should remove this line.  Setting context is handled correctly in the setAuthStateInCmpContexts so this line is not needed.

core/sql/sqlcomp/CmpSeabaseDDLcommon.cpp
robertamarton: It might be.  We might want to do more things also, such as reset caches.  But for init, drop authorization, this is a one time task and there is (in progress) a specific installation procedure to handle this.

core/sql/arkcmp/CmpStatement.cpp
robertamarton: Thanks - this is a good suggestion.

core/sql/cli/Context.cpp
sandhyasun: That's correct .  I meant to ask if you need to take care of propagating the auth info to the child, grandchild arkcmps as Dave is pointing out. I think you are saying during this phase of initializing authorization there is never more than 1 level of arkcmp.  And yes the situation of 2 different arkcmps in the same level isn't supported currently - that support was initially done  for versioning that is obsolete and not aware of any other use for that.

sandhyasun:
Change looks good overall !

core/sql/cli/Context.cpp
robertamarton: The code does propagate the change through the child, grandchild, etc. For drop authorization there are 4 generations of arkcmps and I need to propagate this value through them all - there is actually are regression test for this - udr/102.

nonstop-qfchen: Sounds good!

On Tue, Jun 7, 2016 at 9:47 AM, Roberta Marton <notifications@github.com>
wrote:

> In core/sql/cli/Context.cpp
> <https://github.com/apache/incubator-trafodion/pull/523#discussion_r66085087>
> :
>
> > +  str_sprintf(userMessage, "%d,%d,%s", authOn, userAsInt, databaseUserName_);
> > +  char *pMessage = (char *)&userMessage;
> > +
> > +  // Send message to child arkcmp, if one exists
> > +  ExSqlComp::ReturnStatus cmpStatus;
> > +  ExSqlComp *exSqlComp = getArkcmp();
> > +  ex_assert(exSqlComp, "CliGlobals::getArkcmp() returned NULL");
> > +  cmpStatus = exSqlComp->sendRequest(EXSQLCOMP::DATABASE_USER,
> > +                     (const char *) pMessage,
> > +                     (ULng32) strlen(pMessage));
> > +  if (cmpStatus == ExSqlComp::ERROR)
> > +    return -1;
> > +
> > +  return 0;
> > +}
> > +
>
> The code does propagate the change through the child, grandchild, etc. For
> drop authorization there are 4 generations of arkcmps and I need to
> propagate this value through them all - there is actually are regression
> test for this - udr/102.
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/apache/incubator-trafodion/pull/523/files/626370ee584eb5c3c847c01cd8615f204d29c3ff#r66085087>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AHuuAuR4LTvaAzHQ1131fgRGmjwnsmqvks5qJYSZgaJpZM4IubeB>
> .
>



-- 
Regards, --Qifan

PR #526, opened 2016-06-06
[TRAFODION-2031] At times T2 JDBC applications dump core at the time of lo…

selvaganesang:
…gging error

Cores were dumped when SQL tries to log the error message via log4cxx. Log4cxx instance
was not initialized in case of T2 driver

[TRAFODION-1956] session defaults was getting corrupted in mxosrvr. COMPILER_IDLE_TIMEOUT wasn't added
in alphabetical order. Also fixed the possible buffer overrun issue with some of the set
commands in SessionDefaults


zellerh:
Looks good to me, but I just have a question: Usually, a process calls msg_init_attach() to initialize its environment in Trafodion. The sqInit() method seems to do only part of that. Is that ok?

zellerh:
Thanks for the explanation and sorry about my confusion.

sureshsubbiah:
Looks good Selva.

PR #528, opened 2016-06-08
2.0.0 Release Notes

svarnau:
Update website with release notes for 2.0.0 and 2.0.1 (up coming).

Re-worked the download page to point to the provisioning guides, rather than try to summarize installation in a release independent way. 

Use apache download guidelines to point to mirror sites for tar-file downloads, but the main apache site for signatures and checksums.

Place-holder for 2.0.1 release download since it is not official yet.

### DIFF
--- a/docs/src/site/markdown/download.md
+++ b/docs/src/site/markdown/download.md
@@ -38,6 +38,7 @@ To build Trafodion from source code, see the [Trafodion Contributor Guide](https
 * Convenience Binaries
     * [Installer][ins200]  -  [PGP][inpgp200] [MD5][inmd5200] [SHA1][insha200]
     * [Server][ser200]  -  [PGP][sepgp200] [MD5][semd5200] [SHA1][sesha200]
+    * Clients binary package not available, must be built from source code.
 * [Documentation](documentation.html#20x_Releases)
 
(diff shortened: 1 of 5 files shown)


docs/src/site/markdown/download.md
```diff
@@ -12,144 +12,54 @@
...
+* [Release Notes](release-notes-2-0-1.html)
+* [Documentation](documentation.html#20x_Releases)
 
-Perform this step as the **`trafodion`** user.
+## 2.0.0
 
-First, check the status of the Trafodion environment.
+* [Release Notes](release-notes-2-0-0.html)
+* [Source Code Release][src200]  -  [PGP][pgp200] [MD5][md5200] [SHA1][sha200]
+* Convenience Binaries
```
zellerh: Should this mention that there is no client package in 2.0, so people don't search for it unnecessarily? Maybe something like a short note: Release 2.0.0 does not include convenience binaries for the ODBC/JDBC clients. Those must be built from source, if needed.

zellerh:
Looks good to me, just wondering whether we need a note about the missing client package.

docs/src/site/markdown/download.md
svarnau: Good idea. I'll add it.

zellerh:
Thanks for adding the note.

sureshsubbiah:
Looks good. +1

DaveBirdsall:
Looks good to me too.

PR #530, opened 2016-06-08
[TRAFODION-2039] Add support for ALTER LIBRARY FILE '<file-name>'

sureshsubbiah:
Library file (.jar, .dll or .so) can now be altered for an existing library 
object. ALTER command will check for existence of new file. Old library
definition will be removed from compiler cache.

### DIFF
--- a/core/sql/optimizer/RelExeUtil.cpp
+++ b/core/sql/optimizer/RelExeUtil.cpp
@@ -80,6 +80,7 @@
 #include "StmtDDLCreateRoutine.h"
 #include "StmtDDLDropRoutine.h"
 #include "StmtDDLCleanupObjects.h"
+#include "StmtDDLAlterLibrary.h"
 
 #include <cextdecs/cextdecs.h>
 #include "wstr.h"
@@ -3914,6 +3915,7 @@ RelExpr * DDLExpr::bindNode(BindWA *bindWA)
   NABoolean alterIdentityCol = FALSE;
   NABoolean alterColDatatype = FALSE;
   NABoolean alterColRename = FALSE;
+  NABoolean alterLibrary = FALSE;
   NABoolean externalTable = FALSE;
   
   returnStatus_ = FALSE;
@@ -4308,6 +4310,14 @@ RelExpr * DDLExpr::bindNode(BindWA *bindWA)
       qualObjName_ = getExprNode()->castToStmtDDLNode()->
         castToStmtDDLDropLibrary()->getLibraryNameAsQualifiedName();
     }
+    else if (getExprNode()->castToStmtDDLNode()->castToStmtDDLAlterLibrary())
+    {
+      isAlter_ = TRUE;
+      isLibrary_ = TRUE;
+      alterLibrary = TRUE ;
+      qualObjName_ = getExprNode()->castToStmtDDLNode()->
+	castToStmtDDLAlterLibrary()->getLibraryNameAsQualifiedName();
+    }
     else if (getExprNode()->castToStmtDDLNode()->castToStmtDDLCreateRoutine())
     {
       isCreate_ = TRUE;
@@ -4342,7 +4352,7 @@ RelExpr * DDLExpr::bindNode(BindWA *bindWA)
           (isAlter_ && (alterAddCol || alterDropCol || alterDisableIndex || alterEnableIndex || 
 			alterAddConstr || alterDropConstr || alterRenameTable ||
                         alterIdentityCol || alterColDatatype || alterColRename ||
-                        alterHBaseOptions || otherAlters)))))
+                        alterHBaseOptions || alterLibrary || otherAlters)))))
       {
 	if (NOT isNative_)
 	  {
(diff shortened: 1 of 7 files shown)


zellerh:
Looks good.

DaveBirdsall:
How do other compiler processes learn that the library has changed?

sureshsubbiah:
That will eventually be through the query invalidation feature. Right now I don't think ALTER of UDF or LIBRARY will cause query invalidation to kick in. We may need to add invalidation key for LIBRARY object and include library in the list of objects used by a query. I updated the checkin comment to say compiler cache instead of caches.

DaveBirdsall:
So, the behavior with this change is that the altered library is immediately visible in the session where ALTER LIBRARY is done, and in any new sessions that start afterward. However, for existing sessions it depends on whether something has been referenced from that library yet. Is this correct?

selvaganesang:
Is the user expected to move the jar and ddl files prior to issuing this command.

sureshsubbiah:
Hi Dave. Yes. We do not cache a library definition per se in the NARoutine cache. We do cache naroutines for Routine objects which will be using the library. 

Hi Selva, The ALTER command will check for existence of the file referred to in the FILE clause. If not found it will raise an error.

selvaganesang:
In that case, will the alter command remove the files from the old location. If there is no query invalidation or if it isn't removed from old location, then the user might see unexpected behavior from UDF/UDR calls till DCS is restarted.  Won't it be confusing then?

DaveBirdsall:
If a process in Linux has a file open, then that open survives a "rm" or "mv" operation. So, I suppose the time at which a new routine is bound to a session depends on when its library file is opened and its NARoutine info is cached. So, imagine this scenario: A library contains two UDFs, F1 and F2. A session opens, references F1. That library is now open, and F1's NARoutine information is cached. Now suppose another session does an ALTER LIBRARY. And after that, suppose the first session references F2. I think what happens is we'll open the new library (because library definitions are not cached, per @sureshsubbiah), and F2 will come from the new library. I'm guessing this is OK, but I wonder what happens if the implementations of F1 and F2 share code within the library. The answers might depend on whether it is C++ or Java (dlopen might have different rules than JVM class paths). It's probably a good idea to think this through a bit and update the documentation with any conclusions you might have.

sureshsubbiah:
Hi Selva, ALTER will not remove files. I think we have two options now

a) Document that ALTER LIBRARY will require a DCS restart or not document this feature at all until we have query invalidation for UDFs ready. i,e, document the restart issue in a wiki and not document the statement in the manual
b) Close this PR for now and revisit it when we have query validation for Routine objects.

Both options seem reasonable to me. Thanks for your input. I agree the feature is partial at best now.

DaveBirdsall:
+1 for a). It is perfectly normal and OK to do features incrementally. In the JIRA, might be good to describe what work remains.

selvaganesang:
I am ok with either of the options. However, it may be good to send an warning to the user about this because it is unique to this DDL command alone.  There is no easy way for the user to detect this anomaly if the user doesn't restart DCS,

DaveBirdsall:
Is this ready to merge?

sureshsubbiah:
Hi Dave, Give me 2 more days please. Let me find out what ALTER statements are currently covered by Query Invalidation. If ALTER LIBRARY is indeed an outlier, then I would hesitate to merge this change as is. Either a warning or a change to include QI for this ALTER seems better. I will report back with my finding on QI for ALTER in a couple of days.

DaveBirdsall:
By the way, will there be a change to the manuals? (Is ALTER LIBRARY new syntax?)

sureshsubbiah:
I made an incorrect statement above. The good news is that stale NARoutine objects will be cleared from NARoutine caches **for all users**. This is due to a change by @robertamarton which I had not considered in my previous responses. Sorry.
https://github.com/apache/incubator-trafodion/commit/8e97ff43a7e5324227704cad4ec3e770d521473f

When we clear NARoutine object from the local NARoutine cache, it causes through RMS shared memory for the object to be cleared from other caches on the cluster. Please see the method removeNARoutine2 in @robertamarton 's PR. The CLI call causes an SQL_QIKEY to be placed in shared memory, eventually accessible by other executors.

With this item cleared I think the ALTER LIBRARY statement can be included in the manual.

I checked what type of DDL statements will cause NATable/NARoutine caches to be cleared.
This was added by @mikehanlonathpdotcom in https://github.com/apache/incubator-trafodion/commit/1bd19cba4b2c4d43fd78a5a56172ad8bcb153e73 (and 2 subsequent commits). The list has grown since.

I see that these cases are covered. There could be more. 
alter table add column
alter table add check constraint
alter table add unique constraint
alter table add primary key
alter table add foreign key (RI)
alter table rename
alter table hbase options
alter table rename column
create index

drop column
drop check constraint
drop unique constraint
drop primary key
drop foreign key (RI)
drop index

selvaganesang:
Merging this change to master now

PR #532, opened 2016-06-08
[TRAFODION-2041] Turn off bucket cache in HBaseClient estimateRowCount

DaveBirdsall:
This solves the issue of UPDATE STATISTICS doing select count(*) on large tables. Details are in the JIRA.


selvaganesang:
looks good to me

PR #533, opened 2016-06-09
Fix fir JIRA [TRAFODION-1996] TLOG uses excessive memory when aging old entries.  Better logging for exception stack traces

sbroeder:


DaveBirdsall:
Looks good to me.

PR #536, opened 2016-06-09
[TRAFODION-2043] Bulk load may fail if bucket cache is configured and…

DaveBirdsall:
… large

Details in the JIRA.


sureshsubbiah:
+1. Looks good Dave. Thank you for fixing this important issue.

DaveBirdsall:
Thank you, @sureshsubbiah, for finding this issue. I usually look around for copies of bugs that I fix but didn't this time (regarding JIRA TRAFODION-2041). I'm glad you found this. I did verify that there are no other copies of this bug.

PR #537, opened 2016-06-10
[TRAFODION-2046]change DTM's memoryUsageChore default settings to ena…

traflm:
…ble memory threthold

Change the default behavior of DTM to restrict the memory usage in Region Server to 90%, if above that threshold, stop all new transaction operations until memory released under the threshold.


selvaganesang:
Please add the information about how this change was tested?

traflm:
@selvaganesang , sorry, I need to abandon this change, and have another one, I misread the code

traflm:
wrong fix, close and will send out a new PR

PR #541, opened 2016-06-11
Changes for converting commitId<->ascii time string

sbroeder:


core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/regionserver/transactional/IdTm.java
```diff
@@ -122,13 +124,67 @@ public void id(int timeout, IdTmId id) throws IdTmException {
               LOG.error("native_id returned: " + err + " Throwing IdTmException");
               throw new IdTmException("ferr=" + err);
            }
+           if (id.val == 0) {
+              LOG.error("native_id returned id: " + id.val + " err: " + err + ", Throwing IdTmException");
```
DaveBirdsall: Doesn't hurt anything to report err here,, but you already know in this code path that err == 0 due to the preceding if + throw.

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/regionserver/transactional/IdTm.java
```diff
@@ -122,13 +124,67 @@ public void id(int timeout, IdTmId id) throws IdTmException {
...
+              throw new IdTmException("ferr=" + err);
```
DaveBirdsall: So, here you would get "ferr=0" in the IdTmException. I suppose one could deduce that the problem was that id.val was zero?

core/sqf/src/seatrans/tm/hbasetmlib2/idtm.cpp
```diff
@@ -192,6 +192,89 @@ jint Java_org_apache_hadoop_hbase_regionserver_transactional_IdTm_native_1id(JNI
...
+    jfieldID       lv_id_val;
+
+    lv_ferr = do_init(pp_j_env);
+
+    if (lv_ferr == XZFIL_ERR_OK) {
+       len = (int) j_len;
+       if (gv_verbose)
+           printf("cli: string_to_id() len is %d\n", len);
+       input = (pp_j_env)->GetByteArrayElements(j_id_string, NULL);
+       memcpy(la_ascii_time,(char *)input, len);
```
DaveBirdsall: Are we guaranteed there will never be a buffer overflow? Or does this function simply assume we won't?

DaveBirdsall:
I think this is fine to merge. Will wait a little bit in case my questions make you think otherwise.

PR #542, opened 2016-06-11
[TRAFODION-2046]change DTM's memoryUsageChore default settings to ena…

traflm:
…ble memory threthold

Change the default behavior of DTM to restrict the memory usage in Region Server to 90%, if above that threshold, stop all new transaction operations until memory released under the threshold.

### DIFF
--- a/core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/TrxRegionEndpoint.java.tmpl
+++ b/core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/TrxRegionEndpoint.java.tmpl
@@ -358,6 +358,7 @@ CoprocessorService, Coprocessor {
   private static final int LEASE_CHECK_FREQUENCY = 1000;
   private static final int DEFAULT_SLEEP = 60 * 1000;
   private static final int DEFAULT_MEMORY_THRESHOLD = 100; // 100% memory used
+  private static final int DEFAULT_STARTUP_MEMORY_THRESHOLD = 90; // initial value : 90% memory used
   private static final int DEFAULT_MEMORY_SLEEP = 15 * 1000;
   private static final boolean DEFAULT_MEMORY_WARN_ONLY = false;        
   private static final boolean DEFAULT_MEMORY_PERFORM_GC = false;
@@ -3514,7 +3515,7 @@ CoprocessorService, Coprocessor {
         this.scannerThreadWakeFrequency = config.getInt(HConstants.THREAD_WAKE_FREQUENCY, 10 * 1000);
 
         this.cleanTimer = config.getInt(SLEEP_CONF, DEFAULT_SLEEP);
-        this.memoryUsageThreshold = config.getInt(MEMORY_THRESHOLD, 90);
+        this.memoryUsageThreshold = config.getInt(MEMORY_THRESHOLD, DEFAULT_STARTUP_MEMORY_THRESHOLD);
         this.memoryUsagePerformGC = config.getBoolean(MEMORY_PERFORM_GC, DEFAULT_MEMORY_PERFORM_GC);
         this.asyncWal = config.getInt(CONF_ASYNC_WAL, DEFAULT_ASYNC_WAL);
         this.skipWal = config.getBoolean(CONF_SKIP_WAL, DEFAULT_SKIP_WAL);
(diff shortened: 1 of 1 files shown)


traflm:
Test steps:

Here is the test case:

create table testoom (uniq int not null,
 c100k char(200), c10K char(200) , c1K char(200), c100 char(200),
 c10 char(200), c1 int )
 STORE BY (uniq)
  ATTRIBUTES ALIGNED FORMAT
   ;

  upsert using load into testoom select
0 + (100000 * x100000) + (10000 * x10000) + (1000 * x1000) +
 (100 * x100) + (10 * x10) +( 1 * x1),
'111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111',
'111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111',
'111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111',
'111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111',
'111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111',
0
from (values(0)) t
transpose 0,1,2,3,4,5,6,7,8,9 as x100000
transpose 0,1,2,3,4,5,6,7,8,9 as x10000
transpose 0,1,2,3,4,5,6,7,8,9 as x1000
transpose 0,1,2,3,4,5,6,7,8,9 as x100
transpose 0,1,2,3,4,5,6,7,8,9 as x10
transpose 0,1,2,3,4,5,6,7,8,9 as x1;

update testoom set c10 = '211111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111'  where c1 =0;

before test, change hbase-env.sh, change the Heap size to 500M.

Without the change, update will hang forever, and HBase RS OOM killed.
With this change, update will fail with proper message.
	
*** ERROR[8448] Unable to access Hbase interface. Call to ExpHbaseInterface::insertRow returned error HBASE_ACCESS_ERROR(-706). Cause:
java.io.IOException: org.apache.hadoop.hbase.client.transactional.MemoryUsageException: put memory usage exceeds 90 percent, trxId is 4
org.apache.hadoop.hbase.client.transactional.TransactionalTable.put(TransactionalTable.java:353)
org.apache.hadoop.hbase.client.transactional.RMInterface.put(RMInterface.java:313)
org.trafodion.sql.HTableClient.putRow(HTableClient.java:1558)
org.trafodion.sql.HBaseClient.insertRow(HBaseClient.java:1620)

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/TrxRegionEndpoint.java.tmpl
```diff
@@ -359,7 +359,7 @@ CoprocessorService, Coprocessor {
   private static final int DEFAULT_SLEEP = 60 * 1000;
   private static final int DEFAULT_MEMORY_THRESHOLD = 100; // 100% memory used
```
DaveBirdsall: Why not simply change the "static final int DEFAULT_MEMORY_THRESHOLD" to 90?

traflm: In checkMemoryUsage(), it will check the current memory usage with this DEFAULT_MEMORY_THRESHOLD, if I change the DEFAULT_MEMORY_THRESHOLD, that logic will break. Here the change only change the startup value. The name of DEFAULT_MEMORY_THRESHOLD is confusing, it is not only setting a default value, but also setting a low watermark , so I cannot simply change that value.

DaveBirdsall: I see. Perhaps add a new "private static final int" for your new usage? For example, "private static final int DEFAULT_STARTUP_MEMORY_THRESHOLD = 90;". Just a stylistic nit.

traflm: Yes, Dave, that is a good suggestion, much better coding style.

DaveBirdsall:
Looks good to me. Thanks for making this change.

PR #543, opened 2016-06-13
[TRAFODION-1652]SQLDriverConnect can't get catalog name

xwq:
windows ODBC return a default catalog name in the output connection string
if input connection string doesn't give a catalog name.


DaveBirdsall:
Looks good to me.

PR #544, opened 2016-06-13
[TRAFODION-2044] Support SALT clause on CREATE TABLE LIKE

DaveBirdsall:
Formerly, several clauses (examples: SALT and STORE BY) were allowed on CREATE TABLE LIKE, but were ignored. So, for example, one could say, "CREATE TABLE t2 LIKE t1 SALT USING 6 PARTITIONS", but t2 would end up getting the same salting as t1 instead.

With this change, the parser has been changed so that, except for SALT, specifying these clauses will now result in a syntax error.

For the SALT clause, it is now fully supported with CREATE TABLE LIKE. So, in our example above, t2 would get 6 partitions, and would be salted on the primary key.

A new regression test, seabase/TEST040, has been added to cover this support. The manuals have also been updated, and a few unrelated corrections in CREATE TABLE have been made also.

This pull request also updates the expected results for regression test compGeneral/TEST015, which has been failing for a little while. This change is unrelated to the rest of the changes.

### DIFF
--- a/core/sql/parser/sqlparser.y
+++ b/core/sql/parser/sqlparser.y
@@ -24554,6 +24554,43 @@ table_definition : create_table_start_tokens ddl_qualified_name
                      delete $1; /*TableTokens*/
 		     delete $2 /*ddl_qualified_name*/;
 		   }
+
+table_definition : create_table_start_tokens ddl_qualified_name
+                        like_definition
+	  	   {
+                     $1->setOptions(TableTokens::OPT_NONE);
+		     QualifiedName * qn;
+                     if ($1->isVolatile())
+                       qn = processVolatileDDLName($2, TRUE, TRUE);
+		     else
+                       qn = processVolatileDDLName($2, FALSE, FALSE);
+		     if (! qn)
+                       YYABORT;
+
+		     StmtDDLCreateTable *pNode =
+		       new (PARSERHEAP())
+		       StmtDDLCreateTable(
+			    *qn /*ddl_qualified_name*/,
+			    $3 /*like_definition*/,
+			    NULL,
+			    NULL,
+			    NULL,
+			    PARSERHEAP());
+
+                     $1->setTableTokens(pNode);
+		     pNode->synthesize();
+
+		     if (ParNameCTLocListPtr)
+		       {
+			 delete ParNameCTLocListPtr;
+			 ParNameCTLocListPtr = NULL;
+		       }
+
+		     $$ = pNode;
+                     delete $1; /*TableTokens*/
+		     delete $2 /*ddl_qualified_name*/;
+		   }
+
 		 | TOK_CREATE special_table_name
                         table_definition_body
                         optional_create_table_attribute_list
@@ -24909,7 +24946,6 @@ create_table_as_token: TOK_AS
 
 /* type pElemDDL */
 table_definition_body : table_element_list
-                      | like_definition
                       | external_table_definition
 
 /* type pElemDDL */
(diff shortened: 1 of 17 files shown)


core/sql/parser/sqlparser.y
```diff
@@ -24909,7 +24946,6 @@ create_table_as_token: TOK_AS
 
 /* type pElemDDL */
 table_definition_body : table_element_list
-                      | like_definition
                       | external_table_definition
```
robertamarton: It looks like the new like option is no longer part of a table_definition_body but as a separate high-level production.  Would this break some operations?  For example, I see on line 24594 that a "like" option is syntactically correct with <special_table_name> option.  Would this still be allowed with your change?

docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
```diff
@@ -2235,13 +2235,19 @@ session.
...
-    { table-spec | like-spec }
+    table-spec
     [SALT USING num PARTITIONS [ON (column[, column]...)]]
     [STORE BY {PRIMARY KEY | (key-column-list)}]
-    [HBASEOPTIONS (hbase-options-list)]
+    [HBASE_OPTIONS (hbase-options-list)]
     [LOAD IF EXISTS | NO LOAD]
     [AS select-query]
 
+CREATE [VOLATILE] TABLE IF NOT EXISTS table
```
robertamarton: Not your issue but happened to notice it.  Isn't the IF NOT EXISTS clause optional?

core/sql/parser/sqlparser.y
DaveBirdsall: This is intentional. The problem is that including like_definition in table_definition_body allows things like STORE BY and several other clauses to follow the LIKE option. But the LIKE code just ignores them all. That is, the original implementation was incomplete. I wanted to make the unsupported syntax invalid.

DaveBirdsall: Incidentally, I did a full run of all regression tests having CREATE TABLE LIKE over the weekend to make sure I didn't miss anything. They all passed.

docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
DaveBirdsall: Good catch! Will fix, either in a rework or a later check-in.

core/sql/regress/seabase/TEST040
```diff
@@ -0,0 +1,139 @@
...
+-- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+-- KIND, either express or implied.  See the License for the
+-- specific language governing permissions and limitations
+-- under the License.
+--
+-- @@@ END COPYRIGHT @@@
+
+set schema $$TEST_SCHEMA$$;
+log LOG040 clear;
+
```
robertamarton: May want to add a "cleanup" first before starting the test.  This way, if this test failed previously, you are starting with a clean slate.

robertamarton: As an alternative, you could just create a new schema and drop it with the cascade option.

DaveBirdsall: Thanks; will fix. I copied this from another test.

robertamarton:
Looks good

DaveBirdsall:
Thanks for the careful review. Will submit a rework addressing the items you mentioned.

core/sql/parser/sqlparser.y
robertamarton: This may be correct.  Not sure if the like options should be allowed for other create table productions.

robertamarton:
Thanks Dave, looks good.

PR #545, opened 2016-06-14
JIRA TRAFODION-2060  Add support for tinyint datatype

anoopsharma00:
-- added support for tinyint datatype to be used in create
   and cast statements
-- added support for returning tinyint in select list and
   input params
-- currently tinyint select list and input params is only enabled for
  queries from sqlci. Some support for it has been added to jdbc T4
  but need to be completed. Until then, tinyint select/param support
  is not enabled for t4/t2/odbc interfaces
-- removed obsolete tandem floating point datatypes


core/conn/jdbcT4/src/main/java/org/trafodion/jdbc/t4/InterfaceResultSet.java
```diff
@@ -263,6 +267,14 @@ Object getFetchString(TrafT4Connection conn, int scale, int SQLDataType, int SQL
 				break;
 			}
 			break;
+		case SQLTYPECODE_TINYINT:
+                        short sValue1 = (byte) ibuffer[byteIndex];
+			retObj = new Short(sValue1);
+			break;
+		case SQLTYPECODE_TINYINT_UNSIGNED:
```
DaveBirdsall: Do we have to twiddle a sign bit in one of these cases?

core/conn/jdbcT4/src/main/java/org/trafodion/jdbc/t4/InterfaceResultSet.java
```diff
@@ -503,6 +515,11 @@ static Object getExecute2FetchString(TrafT4Connection conn, TrafT4Desc desc, byt
 				break;
 			}
 			break;
+		case SQLTYPECODE_TINYINT:
+		case SQLTYPECODE_TINYINT_UNSIGNED:
+                        short sValue1 = (byte) values[noNullValue];
+			retObj = new Short(sValue1);
+			break;
```
DaveBirdsall: Do we have to twiddle a sign bit for one of these?

core/sql/common/BaseTypes.cpp
```diff
@@ -412,6 +419,13 @@ short convertTypeToText_basic(char * text,	   // OUTPUT
 	str_sprintf(text, "NUMERIC(%d, %d) UNSIGNED", precision, scale);
       break;
 
+    case REC_BIN8_UNSIGNED:
+      if (!precision)
+        str_sprintf(text, "TINYINT UNSIGNED");
+      else
+	str_sprintf(text, "NUMERIC(%d, %d)", precision, scale);
```
DaveBirdsall: Should this be NUMERIC UNSIGNED?

core/sql/common/NumericType.h
```diff
@@ -445,6 +443,99 @@ class SQLBPInt : public NumericType
...
+    {
+      if (isUnsigned())
+	return REC_BIN8_UNSIGNED;
+      else
+	return REC_BIN8_SIGNED;
+    }
+
+  NABoolean roundTripConversionToDouble() const { return TRUE; };
+
+  virtual Lng32 getMagnitude() const { return isUnsigned() ? 28 : 25; }
```
DaveBirdsall: What's "magnitude"?

DaveBirdsall:
Nice set of changes. I might have found a bug or two. It's been a long time since we added a new data type to the engine. Nice work.

core/conn/jdbcT4/src/main/java/org/trafodion/jdbc/t4/InterfaceResultSet.java
anoopsharma00: We dont have full support for tinyint IO from odbc/jdbc interface.
This code is part of jdbc and will not be reached until that support is in.
Once that support is in, these paths will be fixed and tested.
Until then, tinyint will be converted to smallint in the engine.

Based on the review comment, have updated this code to handle unsigned tinyint.
But it will not be tested and enabled until full jdbc/odbc support has been added.

core/sql/common/BaseTypes.cpp
anoopsharma00: At current time, NUMERIC datatypes are handled as smallint, int or largeint.
They are not converted to tinyint. That could be done at some point.
So this code will not be reached.

But review comment is correct about this code being incorrect and I have fixed it.
If we do enhance NUMERIC with precision <= 2 to be treated as tinyint, then
this code will be exercised.

core/sql/common/NumericType.h
anoopsharma00: Magnitude is the number of digits to the left of decimal point.
NUMERIC(5,2) as magnitude of 3.
It is used during expression set up to correctly determine what an
expression result should be. 

Datatypes with binary precision and no scale (int, smallint, etc) do not have exact
precision. Tinyint has precision of 3 digits (max val
255 for unsigned or 127 for signed). 
The getMagnitude method returns an approx magnitude that
is used during expression set up. 28 means magnitude between 2 and 3
(multiplied by 10).  Unsigned magnitude is larger value than signed
magnitude.

selvaganesang:
The catalog APIs for ODBC/JDBC need to expose that tinyint is supported in database via SQLTypeInfo and getTypeInfo calls. It can be done later too.

anoopsharma00:
Response to Selva's comment. Yes, catalog odbc/jdbc apis will need to be enhanced when full support for jdbc/odbc tinyint is externalized. Thanks for pointing it out.

PR #547, opened 2016-06-16
[TRAFODION-1745] show more related info when TRANSLATE run into SQL E…

traflm:
…rror

Show what is the translation: from which charset to what charset, and showing the offending data in HEX mode.
This will help troubleshooting.

### DIFF
--- a/core/sql/exp/exp_conv.cpp
+++ b/core/sql/exp/exp_conv.cpp
@@ -136,7 +136,68 @@ void double_varchar_length(
    }
 }
 
+//////////////////////////////////////////////////////////////////
+//
+// A helper function to show buffer in HEX 
+//
+// ///////////////////////////////////////////////////////////////
 
+static char *stringToHex(char * out, Int32 outLen, char * in, Int32 inLen)
+{
+  //clear out buffer first
+  memset(out,0,outLen);
+
+  outLen = (outLen / 2) ;
+
+  if(inLen < outLen) outLen = inLen;
+
+  char hex[3];
+  for(int i = 0; i < outLen; i++)
+  {
+    sprintf(hex, "%02x", in[i]);
+    strcat(out,hex);
+  }
+  return out;
+}
+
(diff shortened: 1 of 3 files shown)


selvaganesang:
Change looks good

core/sql/exp/exp_conv.cpp
```diff
@@ -4280,6 +4341,9 @@ unicodeToSByteTarget(
        if ( allowInvalidCodePoint == FALSE )
        {
           ExRaiseSqlError(heap, diagsArea, EXE_INVALID_CHAR_IN_TRANSLATE_FUNC);
+          char hexstr[256];
+          memset(hexstr,0,256);
```
DaveBirdsall: The memset is unneeded since stringToHex also does a memset, but it doesn't hurt any.

core/sql/exp/exp_conv.cpp
```diff
@@ -4280,6 +4341,9 @@ unicodeToSByteTarget(
...
+          *(*diagsArea) << DgString0("UNICODE") << DgString1("ISO88591") << DgString2(stringToHex(hexstr,256,source,sourceLen));
```
DaveBirdsall: It's a bit more robust to say, sizeof(hexstr) rather than hard-coding 256.

DaveBirdsall:
Looks good to me.

sureshsubbiah:
+1. Looks good.
@traflm please let us know if you would like to merge the change as is or would prefer to address the issues @DaveBirdsall raised now.

traflm:
@sureshsubbiah sorry that I was on a trip and not able to reply to this PR. 
I was planning to change the code according to Dave's comments. But i will work on another JIRA which will enhance more Error message displaying, so it is highly related, and I can change code with that PR. The next PR.

DaveBirdsall:
@traflm, sorry, looks like I was too fast with the merge. Yes, feel free to make further changes in another pull request.

PR #549, opened 2016-06-17
[TRAFODION-1128]Remove current catalog& schema check

kevinxu021:
Have check the code context that it's useless to check current catalog and schema. As the comments, it's used for alter table but it has nothing about current catalog and schema against my testing.


sureshsubbiah:
+1. Looks good.

PR #550, opened 2016-06-17
[TRAFODION-2074] Create index should avoid populating the index withi…

selvaganesang:
…n a transaction

Create index is now done in multiple transactions when there is no
active transaction.
1) creating the index table in transaction 1.
2) populate the index without any transaction. If it fails,
   cleanup the index table.
3) If it succeeds complete the rest of the index creation in transaction 2.
4) If there are any failures, cleanup the index table.

When there is an active transaction while the create index is issued, then
all the above steps are done within the user transaction.

Populate index command is also done without any transaction

Also did some cleanup to return proper errors instead of dumping core


robertamarton:
Changes look good to me.

core/sql/executor/ExHbaseAccess.cpp
```diff
@@ -1153,6 +1153,32 @@ Lng32 ExHbaseAccessTcb::createSQRowFromHbaseFormat(Int64 *latestRowTimestamp)
...
+
+              Text colFam, colName;
+              extractColFamilyAndName(colVal, colFam, colName);
+
+              Int64 v = 0;
+              if (colName.length() == sizeof(char))
+                v = *(char*)colName.data();
+              else if (colName.length() == sizeof(unsigned short))
+                v = *(UInt16*)colName.data();
+              else if (colName.length() == sizeof(Lng32))
```
DaveBirdsall: Should be sizeof(ULng32), but since that is the same as sizeof(Lng32), this is an unobservable bug

core/sql/executor/ExHbaseAccess.cpp
```diff
@@ -1153,6 +1153,32 @@ Lng32 ExHbaseAccessTcb::createSQRowFromHbaseFormat(Int64 *latestRowTimestamp)
...
+                v = *(ULng32*)colName.data();
+
+              char buf[10];
```
DaveBirdsall: 2^30 = 1073741824, which is one byte too big to fit in buf (since you need room for a null terminator). We are unlikely to see that many columns, but still would be nice to make buf 11 bytes at least.

DaveBirdsall:
Looks good to me, though the conflicts have to be resolved.

core/sql/exp/ExpHbaseInterface.h
```diff
@@ -62,22 +62,6 @@ class ExHbaseAccessStats;
...
 
-using namespace apache::thrift;
-using namespace apache::thrift::protocol;
-using namespace apache::thrift::transport;
-
-using namespace apache::hadoop::hbase::thrift;
-
-namespace {
-
-  typedef std::vector<std::string> StrVec;
```
sureshsubbiah: Do we have any other Thrift dependencies? I remember seeing some mention of thrift in the makefiles. It will be an added bonus if those could be removed too.

core/sql/executor/ExHbaseAccess.cpp
selvaganesang: Done

core/sql/executor/ExHbaseAccess.cpp
selvaganesang: Done. Changed it buf[20]

core/sql/exp/ExpHbaseInterface.h
selvaganesang: Yes. We can do this as a different PR because I suspect it would involve more code changes.

core/sql/generator/GenPreCode.cpp
```diff
@@ -2824,44 +2823,33 @@ short DDLExpr::ddlXnsInfo(NABoolean &isDDLxn, NABoolean &xnCanBeStarted)
...
+     isDDLxn = TRUE;
+
+     if (purgedataHbase() || upgradeRepos())
+        // transaction will be started and commited in called methods.
+        xnCanBeStarted = FALSE;
+     if ((ddlNode && ddlNode->castToStmtDDLNode() &&
+          ddlNode->castToStmtDDLNode()->ddlXns()) &&
+            ((ddlNode->getOperatorType() == DDL_CLEANUP_OBJECTS) ||
+             (ddlNode->getOperatorType() == DDL_ALTER_TABLE_DROP_COLUMN) ||
+             (ddlNode->getOperatorType() == DDL_CREATE_INDEX) ||
```
selvaganesang: Good catch. I will check and fix it.

PR #552, opened 2016-06-20
JIRA TRAFODION-2072 plus few more

anoopsharma00:
-- long varchars (> 32K) are handled for traf (regular and aligned)
   and hive tables
-- showddl for hive tables shows traf and hive definitions
-- traf external tables can optionally take columns attrs that would
   overwrite corresponding hive col attrs
-- metadata is updated with info no whether col attrs were specified
-- 'repeat' function can optionally take a max length parameter
-- error is returned if an external table is created for non-hive
-- tests added to regress/seabase/TEST031 and hive/TEST009


core/sql/common/CharType.cpp
```diff
@@ -942,7 +942,8 @@ SQLVarChar::SQLVarChar(Lng32 maxLen,
 	       FALSE, allowSQLnull, isUpShifted, isCaseInsensitive,
 	       TRUE, cs, co, ce,
 	       encoding, vcIndLen),
-      clientDataType_(collHeap())  // Get heap from NABasicObject. Can't allocate on stack.
+      clientDataType_(collHeap()),  // Get heap from NABasicObject. Can't allocate on stack.
+      wasHiveString_(FALSE)
```
DaveBirdsall: I'm confused by the name. The string was a Hive string but it isn't any longer? Why should we care? (I imagine I'll find the answer later... there is some different semantic that we want downstream...)

core/sql/common/NAType.cpp
```diff
@@ -670,6 +671,84 @@ short NAType::convertTypeToText(char * text,	   // OUTPUT
 				 displayCaseSpecific);
 }
 
+short NAType::getMyTypeAsHiveText(NAString * outputStr)  // output
```
DaveBirdsall: So, what we are really getting is the Hive type associated with a Trafodion data type?

core/sql/generator/GenRelScan.cpp
```diff
@@ -1131,16 +1139,29 @@ short FileScan::codeGenForHive(Generator * generator)
...
   ULng32 cbuffersize = 
     SqlBufferNeededSize((upqueuelength * 2 / numBuffers),
 			returnedRowlen);
   // But use at least the default buffer size.
   //
   buffersize = buffersize > cbuffersize ? buffersize : cbuffersize;
 
+  // Cap the buffer size at 5M and adjust upqueue entries.
+  // Do this only if returnrowlen is not > 5M
+  if ((returnedRowlen <= FiveM) && (buffersize > FiveM))
```
DaveBirdsall: What happens if the returned row length is 3Mb. Do we waste 2Mb in buffer space?

core/sql/parser/StmtDDLCreate.cpp
```diff
@@ -5007,7 +5020,31 @@ StmtDDLCreateTable_visitTableDefElement(ElemDDLNode * pCreateTableNode,
...
+        ->getDDLLikeNameAsCorrName();
+      pCreateTable->likeOptions_ = 
+        pElement->castToElemDDLLikeCreateTable()
+        ->getLikeOptions();
+
+      if ((NOT pCreateTable->isExternal()) &&
+          (pElement->castToElemDDLLikeCreateTable()->forExtTable()))
+        {
+          *SqlParser_Diags << DgSqlCode(-3242)
+                           << DgString0("'for' clause can only be specified when creating an 'external' table.");
```
DaveBirdsall: It looks like the style of SqlciErrors.txt is to put syntax elements in upper case in error messages, unquoted. So, for example, "FOR clause can only be specified when creating an external table."

core/sql/parser/StmtDDLCreate.cpp
```diff
@@ -5007,7 +5020,31 @@ StmtDDLCreateTable_visitTableDefElement(ElemDDLNode * pCreateTableNode,
...
+          return;
+        }
+      else if ((pCreateTable->isExternal()) &&
+               (NOT pElement->castToElemDDLLikeCreateTable()->forExtTable()))
+        {
+          *SqlParser_Diags << DgSqlCode(-3242)
+                           << DgString0("'like' clause cannot be specified when creating an external table.");
```
DaveBirdsall: Similar

core/sql/common/NAType.cpp
```diff
@@ -670,6 +671,84 @@ short NAType::convertTypeToText(char * text,	   // OUTPUT
...
+    {
+    case REC_MIN_F_CHAR_H ... REC_MAX_F_CHAR_H:
+      *outputStr = "string";
+      break;
+
+    case REC_MIN_V_CHAR_H ... REC_MAX_V_CHAR_H:
+      {
+        SQLVarChar * ct = (SQLVarChar*)this;
+        if (ct->wasHiveString())
+          *outputStr = "string";
```
DaveBirdsall: Ah, now I see the point of the "wasHiveString" business. It's to recover the original Hive data type on a describe.

DaveBirdsall:
Looks good to me. My comments are just stylistic nits.

core/sql/common/NAType.cpp
anoopsharma00: yes, that is correct.

Both hive string and varchar datatypes are represented as traf varchar datatypes.
hive varchar has predefined length. hive string length is determined by a session cqd.

The 'wasHiveString' field tells whether it was originally a string datatype. This is
used during showddl command to show the original datatype.

At a later poing, It will also be used to limit the length of data that can be put into a hive column.
For hive varchar, it will be limited by the pre-specified length. For hive string, it will
be limited by the hive length cqd that is set for that session.

PR #560, opened 2016-06-25
JIRA 2086 Support for largeint unsigned

anoopsharma00:
This commit delivers part 1 of the jira changes.
It includes changes in traf compiler, executor and expressions
to handle largeint unsigned datatype.


core/sql/common/ComSmallDefs.h
```diff
@@ -78,7 +78,7 @@ typedef          Int32          ComSInt32;
 #define MAX_HARDWARE_SUPPORTED_SIGNED_NUMERIC_PRECISION 18
 
 //Unsigned NUMERICs upto this range are supported by hardware.
-#define MAX_HARDWARE_SUPPORTED_UNSIGNED_NUMERIC_PRECISION 9
+#define MAX_HARDWARE_SUPPORTED_UNSIGNED_NUMERIC_PRECISION 20
```
DaveBirdsall: Elsewhere we used 19 for the precision of LARGEINT UNSIGNED?

core/sql/common/ComSysUtils.h
```diff
@@ -129,6 +129,27 @@ inline Int64 reversebytes( Int64 xx )
...
+        char c[8];
+    } sink;
+
+    source.xx = xx;
+
+    for (Int32 i=0; i<8; i++)
+        sink.c[i] = source.c[7-i];
+
+    return sink.xx;
+}
```
DaveBirdsall: Endianness issues sometimes give me a sinking feeling too. :-)

core/sql/common/Int64.cpp
```diff
@@ -136,6 +140,22 @@ void convertInt64ToAscii(const Int64 &src, char* tgt)
   strcpy(tgt, s);
 }
 
+void convertUInt64ToAscii(const UInt64 &src, char* tgt)
+{
```
DaveBirdsall: Odd that this would be a reference parameter.

core/sql/common/Int64.cpp
```diff
@@ -136,6 +140,22 @@ void convertInt64ToAscii(const Int64 &src, char* tgt)
...
+  UInt64 temp = src;
+  char buffer[21];
+  char *s = &buffer[21];
+  *--s = '\0';
+  do {
+    char c = (char) (temp % 10);
+    if (c < 0)
+      c = -c;
```
DaveBirdsall: I'm trying to imagine why c would ever be negative. Seems like since temp is unsigned, we'd always see 0, ... 9 as possible values from temp % 10?

DaveBirdsall:
Looks good to me. A very tedious code change. I looked hard for cut-and-paste errors but did not find any. Congratulations for your very careful approach.

core/sql/common/Int64.cpp
anoopsharma00: And there is your cut-and-paste issue :)
This method is a copy of existing method convertInt64ToAscii
which converts a signed number to ascii.

The new convertUInt64ToAscii method doesn't need the 'c < 0' check though it will
never be reached. Will fix that for a later checkin.

Existing convert methods in Int64.cpp pass in input value as reference 
parameter so the new method just followed the same.
Am not sure what the original intent was, maybe some perf improvement.
I will leave it as a reference param.

PR #561, opened 2016-06-27
TRAFODION-1743

mashengchen:
add seesion to each connection.


mashengchen:
jenkins ,retest

core/conn/jdbc_type2/native/SQLMXConnection.cpp
```diff
@@ -590,6 +590,20 @@ JNIEXPORT void JNICALL Java_org_trafodion_jdbc_t2_SQLMXConnection_connectInit
         strcpy(srvrGlobal->CurrentSchema, "SEABASE");
 
     }
+ 
+    odbc_SQLSvc_SetConnectionOption_sme_(NULL, NULL,
+                 &setConnectException,
+                 dialogueId,
+                 BEGIN_SESSION,
+                 0,
+                 NULL,
```
selvaganesang: Consider passing in client_name and  thread_id  so that it is identifiable on the server side. This optional session name has a limit of 24 characters

traflm:
changes looks good to me. Agree with @selvaganesang to set a session name.
Next time, please use proper PR name, so JIRA can be sync . thanks,

mashengchen:
close it first ,i will do another PR for currect title

ryzuo:
T2 has no client, and does not go through network, so it cannot retrieve "client name" from driver at "client side" like T4 does in protocol. We may have to use pid instead of "client name", and plus the thread id can identify the session. Although it does not make sense since the pid does not identify any real "client" process, especially if mtdcs is enabled in the future, then the session name contains no client information. Anyway, we don't have other better option to do this, so we I think this is the only choice we have now.

DaveBirdsall:
So, where are we on this change? Is this ready to merge? @selvaganesang, do you have further issues?

selvaganesang:
. Using [TRAFODION-1743]  instead of jira1743 is suggested so that PR can be automatically linked to the Trafodion jira. Also, please edit the subject line of this PR so that it conveys what it does. When someone looks at git log later, it would be helpful.

Can we pass optionally pass a property to indicate some kind of client information as part of the connection. And this can be used to identify the client of T2 driver and in the session name.

PR #562, opened 2016-06-27
[TRAFODION-1609]

mkby:
Change swstatus to give a simple "up" or "down" answer, instead of the current, confusing display

### DIFF
--- a/core/sqf/sql/scripts/install_local_hadoop
+++ b/core/sqf/sql/scripts/install_local_hadoop
@@ -849,18 +849,29 @@ EOF
 #!/bin/sh
 cd ${MY_SW_ROOT}
 . $MY_SW_SCRIPTS_DIR/sw_env.sh
-NUM_JAVA_PROCS=\`ps -aef | grep \$USER | grep java | grep -v grep | wc -l\`
-NUM_MYSQLD_PROCS=\`ps -aef | grep \$USER | grep mysqld | grep -v grep | wc -l\`
+JPS_OUTPUT=`jps`
+SERVICES='HMaster NodeManager ResourceManager NameNode DataNode SecondaryNameNode'
+for s in $SERVICES; do
+    if [[ ! $JPS_OUTPUT =~ $s ]]; then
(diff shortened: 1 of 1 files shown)


DaveBirdsall:
Looks good to me.

core/sqf/sql/scripts/install_local_hadoop
```diff
@@ -847,20 +847,31 @@ EOF
...
-
+NUM_MYSQLD_PROCS=\`ps -aef | grep \$USER | grep mysqld | grep -v grep | wc -l\`
+if [[ \$NUM_MYSQLD_PROCS -ne 0 ]]; then
+    echo "\$NUM_MYSQLD_PROCS mysqld processes are running"
+    exit 0
+else
+    echo "ERROR: mysqld process is not running!"
+    exit 1
+fi
 EOF
```
selvaganesang: Is there a need to do it the same way for the other generated scripts too

mkby: What are the other generated scripts?

selvaganesang: See Line numbers 777 and above of this file. It has scripts like swstartall, swstopall, swstarthbase, swstophbase etc.

mkby: Thanks, I see. But they're all start/stop scripts, I don't want to modify them.

traflm:
looks good, I think there is no need to change swstartall, swstopall, swstarthbase etc, this JIRA is to improve only status output to show better info.

sureshsubbiah:
+1. If a change is necessary for swstart* / swstop* scripts, that could be a new JIRA.

zellerh:
Looks good to me, too. I also don't think we need changes to other scripts. Thank you, this will make life easier for everyone!

selvaganesang:
I was referring to the change to escape out the variables so that the script would contain the variables itself. My bad, I wasn't clear enough.

PR #564, opened 2016-06-29
[TRAFODION-1676]support better runtime error message when a SQL funct…

traflm:
…ion meet fital error

During normal query execution, the error message of many functions doesn't contain any clue about what data cause the error, making it rather difficult to find the offending source. Especially difficult when doing bulkload where millions of rows are about to be loaded, if there are some bad value, it is extremely hard to find them.

This PR enhance error messages for these SQL Errors:
EXE_CONVERTTIMESTAMP_ERROR	= 8405,
EXE_JULIANTIMESTAMP_ERROR	= 8407,
EXE_CONVERT_STRING_ERROR	= 8413
EXE_CONVERT_DATETIME_ERROR	= 8415,

By displaying the offending source data content, for CONVERT_STRING_ERROR, because it always contains multi-byte character, so display the data in HEX mode.

This was minor issue, but become high recently because it is really very hard to fix source data if error msg cannot tell the user enough info. With this change, the user can search for bad value in source more easier.


core/sql/cli/Statement.cpp
```diff
@@ -2102,6 +2102,30 @@ Statement * Statement::getCurrentOfCursorStatement(char * cursorName)
 
 }
 
+//////////////////////////////////////////////////////////////////
+////
+//// A helper function to show buffer in HEX
+////
+//// ///////////////////////////////////////////////////////////////
+
+static char *stringToHex(char * out, Int32 outLen, char * in, Int32 inLen)
```
sureshsubbiah: Could this function please be declared in some place like exp/ExpError.h (with an implementation in ExpError.cpp)? I see this function implemented twice in this PR and something similar in a previous PR. I think this is not advisable according to our coding guidelines.

core/sql/exp/exp_conv.cpp
```diff
@@ -1445,6 +1445,10 @@ ex_expr::exp_return_type convUnicodeToDatetime(char *target,
    }
 
    ExRaiseSqlError(heap, diagsArea, EXE_CONVERT_DATETIME_ERROR);
+   char hexstr[256];
```
sureshsubbiah: This 256 should be a constant declared in some common header.

core/sql/exp/exp_conv.cpp
```diff
@@ -1445,6 +1445,10 @@ ex_expr::exp_return_type convUnicodeToDatetime(char *target,
...
+   if(*diagsArea)
```
sureshsubbiah: Can heap, or diagsArea be null here (or other places below)? ExRaiseSqlError does allow these 2 possibilities and returns NULL in both cases. We may be able to convince ourselves that this should never happen, I get uneasy seeing a dereference of a pointer.
 One possibility is that we change 1447 to be 
ExRaiseSqlError(heap, diagsArea, EXE_CONVERT_DATETIME_ERROR, NULL, NULL, NULL, NULL, stringToHex(hexstr, sizeof(hexstr) , source, sourceLen ));
Then lines 1449 and 1450 can be removed.

core/sql/exp/exp_datetime.cpp
```diff
@@ -1908,6 +1932,11 @@ scanField(char *&src,
       // An unknown character was encountered in the string.
       //
       ExRaiseSqlError(heap, diagsArea, EXE_CONVERT_STRING_ERROR);
+      char hexstr[256];
+      memset(hexstr, 0 , sizeof(hexstr) );
```
sureshsubbiah: Please consider removing this line.The memset is also present in stringToHex.

core/sql/cli/Statement.cpp
traflm: thanks @sureshsubbiah , I was not able to find a common header file. ExpError.h is a good one. I will move this function into there.

core/sql/exp/exp_conv.cpp
traflm: In some code paths, both heap and **diagsArea can be NULL, but *diagsArea is not NULL, that is the reason I add a safety checking 
if(*diagsArea), but yes, *diagsArea may also be possible to be NULL (although by design that is not possible IMHO). The proposed change is a very good one. i will make the change.

core/sql/exp/exp_datetime.cpp
traflm: yes, thanks Suresh for such a carefully review!

core/sql/exp/exp_datetime.cpp
```diff
@@ -2019,6 +2051,14 @@ ExpDatetime::convAsciiToDatetime(char *srcData,
   //
   if (format == DATETIME_FORMAT_ERROR) {
     ExRaiseSqlError(heap, diagsArea, EXE_CONVERT_DATETIME_ERROR);
+    char hexstr[256];
+    memset(hexstr, 0 , sizeof(hexstr) );
```
sureshsubbiah: I wonder if lines 2055 to 2059 is equivalent to the following two lines
strncpy(hexstr,originalSrcData,255);  // (using constant declaration-1 instead of 255 of course :))
hexstr[255] = 0;

I guess I am trying to say that strncpy is capable of handling source strings larger than destination and 2057 and 2058 seem unnecessary.

core/sql/exp/exp_conv.cpp
sureshsubbiah: Also what do you think about making this length longer than 256? We are dealing with strings that are much longer quote often. I think ComCondition is quite flexible, though I don't know what the max length of stringParams are. Since this is the error case, we don't need to be too concerned about performance (though for loads we may continue after error). Some balance will have to be struck. Maybe this should be configurable with a session setting or env var? I am happy with 256 too, if you think that is sufficient for now.

sureshsubbiah:
This is an excellent change. I had wanted to do this so many times but never found the courage to dig in. Thanks!

core/sql/cli/Statement.cpp
```diff
@@ -2298,6 +2298,10 @@ RETCODE Statement::resolveNames(LateNameInfoList * lnil,
 	      if (retcode != ex_expr::EXPR_OK)
 		{
 		  diagsArea << DgSqlCode(-EXE_CONVERT_STRING_ERROR);
+                  char hexstr[MAX_OFFENDING_SOURCE_DATA_DISPLAY_LEN];
```
DaveBirdsall: Need to allow room for a trailing null. You'll have to declare one more byte here than you pass as the second parameter to stringToHex. Alternatively, change the stringToHex function to allow for room (for example, change the statement "outLen = (outLen / 2)" to "outLen = (outLen / 2) - 1"

DaveBirdsall:
It's a good change. There is one potential buffer overrun problem though that needs to be looked at. See in-line comments.

core/sql/cli/Statement.cpp
traflm: thanks Dave, I will change it. This may core dump the program.

DaveBirdsall:
Looks like a commit from yesterday conflicted with this change. Could you please merge in the latest master branch into yours and resolve the conflicts?

traflm:
@DaveBirdsall conflict resolved, only the error message file is conflicting.

DaveBirdsall:
Looks good. I will commit once tests pass.

DaveBirdsall:
jenkins, retest

traflm:
jenkins, retest

traflm:
A build issue reporting jar not found, this PR doesn't change any java code, so retest, debug build OK, so it should not be a syntax issue. Probably a temp issue.

PR #565, opened 2016-06-29
[TRAFODION-2037] Improve DDL concurrency + repository and metadata upgrades

DaveBirdsall:
This change includes all the planned metadata changes for Trafodion release 2.1. The following changes are included:

1. Fix for JIRA TRAFODION-2037: This adds a new index, TABLE_CONSTRAINTS_IDX, to the metadata TABLE_CONSTRAINTS table, along with a CONTROL QUERY SHAPE in one of the metadata code paths to force its use. This change improves DDL concurrency.

2. A FLAGS column (LARGEINT) for future use has been added to the DEFAULTS, ROUTINES, LIBRARIES and LIBRARY_USAGES metadata tables. For now, the value of this column will always be zero. On upgrade of existing metadata, these columns are set to zero.

3. The repository tables have been upgraded. On upgrade of existing metadata, new columns are set to NULL. Specifically:
a. The following columns added to “_REPOS_”.METRIC_QUERY_AGGR_TABLE table:
AVERAGE_RESPONSE_TIME  (LARGEINT) and THROUGHPUT_PER_SECOND (LARGEINT)
b. The following columns added to “_REPOS_”.METRIC_QUERY_TABLE table:
QUERY_SIGNATURE_ID (CHAR(160) CHARACTER SET ISO88591) and LAST_UPDATED_TIME                (TIMESTAMP(6))
c. The length of the EXPLAIN_PLAN column in METRIC_QUERY_TABLE was changed from 200000 to 1000000.
d. The following columns added to “_REPOS_”.METRIC_SESSION_TABLE table:
PROFILE_NAME (VARCHAR(130 CHARS) CHARACTER SET UTF8) and SLA_NAME                         (VARCHAR(130 CHARS) CHARACTER SET UTF8).

The version of the metadata changes from 1.0.1 to 2.1.0 with this change.

When you include this patch in your Trafodion 2.1 development instance, if you have existing metadata you will find that any SQL operation will fail with error 1395, which asks you to upgrade your metadata. You may find it useful to perform a backup of your instance before including this patch.

To upgrade your metadata, perform the command "initialize trafodion, upgrade" in a fresh sqlci session. After successful completion, exit that session, and start a new one. It would be helpful if you could forward any experiences you have with this procedure to the Trafodion dev list or to file JIRAs if the experience was negative.

Users of older releases of Trafodion will need to perform these same steps when installing Trafodion 2.1.


DaveBirdsall:
Thanks go to @hedean for providing the Repository code changes which have been merged into the pull request.

DaveBirdsall:
With further testing, I've noticed that "initialize trafodion, upgrade" fails with these changes if privileges were previously enabled. Please don't merge this code until I've addressed that issue.

DaveBirdsall:
The problem with upgrade when privileges were enabled has been fixed. As part of the fix, I needed to refactor how subsystems are upgraded, separating the final drop and the undo steps from the main upgrade step. I did a full refactoring for the repository subsystem, and put in place the framework for refactoring the privilege manager subsystem. The privilege manager subsystem upgrade logic is not fully developed yet (because we haven't made changes to those tables yet); at that time the privilege manager logic refactoring should be completed as well.

core/sql/parser/sqlparser.y
```diff
@@ -16026,13 +16027,30 @@ exe_util_init_hbase : TOK_INITIALIZE TOK_TRAFODION
 
 		 DDLExpr * de = new(PARSERHEAP()) DDLExpr(TRUE, FALSE, TRUE, FALSE,
                                                           FALSE, FALSE,
-							  FALSE, FALSE, FALSE,
+							  FALSE, FALSE, FALSE, FALSE,
```
robertamarton: Lately, when I have been adding new options to existing INITIALIZE commands, instead of add a new parameter to the DDLExpr constructor, I have been adding a new bit to "Flags".  Not sure when a new parameter should be added or when a new option should be defined through the "Flags" bitmap. Adding a new flag does not require changing all calls to DDLExpr.

core/sql/sqlcomp/CmpSeabaseDDLcommon.cpp
```diff
@@ -437,6 +437,10 @@ short CmpSeabaseDDL::convertColAndKeyInfoArrays(
           ComTdbVirtTableKeyInfo &ki = btKeyInfoArray[ii];
           if (strcmp(ci.colName, ki.colName) == 0)
             {
+              if (ki.ordering == 0)
+                nac->setClusteringKey(ASCENDING);
+              else // ki.ordering should be 1
+                nac->setClusteringKey(DESCENDING);
               naKeyArr->insert(nac);
```
robertamarton: Should I add the comment you always do - should there be a literal for ordering?

core/sql/sqlcomp/CmpSeabaseDDLcommon.cpp
```diff
@@ -891,6 +901,10 @@ short CmpSeabaseDDL::createMDdescs(MDDescsInfo *&trafMDDescsInfo)
   return 0;
 
  label_error:
+
+  // When debugging, you can look at the breadCrumb variable to figure out
+  // why you got here.
+
```
robertamarton: Could you somehow include the breadcrumb variable in an error message, or log the information to the masterexe log in the "logs" directory?

core/sql/sqlcomp/CmpSeabaseDDLmd.h
```diff
@@ -661,6 +690,452 @@ static const MDTableInfo allMDHistInfo[] = {
...
+//// END_OLD_MD_v11: 
+////        OLD metadata Version 1.1  (Major version = 1, Minor version = 1).
+//////////////////////////////////////////////////////////////////////////
+
+// Note: The older metadata definitions below were from before we
+// reset the version to 1.1 with the early Trafodion releases. (We
+// went backwards to match the Trafodion release number.)
+
+////////////////////////////////////////////////////////////////////////
 //// START_OLD_MD_v23:
```
robertamarton: Is there any plan to remove old versions or do we plan to just keep tacking on new versions?

core/sql/sqlcomp/CmpSeabaseDDLrepos.cpp
```diff
@@ -47,8 +47,17 @@ short CmpSeabaseDDL::createRepos(ExeCliInterface * cliInterface)
 
   NABoolean xnWasStartedHere = FALSE;
 
+  cliRC = cliInterface->holdAndSetCQD
+    ("TRAF_MAX_CHARACTER_COL_LENGTH", "1000000");
+
```
robertamarton: Why was 1000000 chosen?

core/sql/sqlcomp/CmpSeabaseDDLrepos.cpp
```diff
@@ -114,43 +123,72 @@ short CmpSeabaseDDL::createRepos(ExeCliInterface * cliInterface)
...
                   getSystemCatalog(), SEABASE_REPOS_SCHEMA,
                   (oldRepos ? rti.oldName : rti.newName));
-      
+    
+      if (beginXnIfNotInProgress(cliInterface, xnWasStartedHere))
+        {
+          cliInterface->retrieveSQLDiagnostics(CmpCommon::diags());
+          return -1;
+        }    
+
```
robertamarton: What are the plans for transaction management for upgrade?  I recently made a bunch of changed to initialize authorization to allow it to run in a single transaction.  Are we going to run upgrade trafodion in multiple transactions?

robertamarton:
Thanks for doing this refactoring.  It looks good and the comments are useful to describe our upgrade plans.

core/sql/sqlcomp/CmpSeabaseDDLmd.h
DaveBirdsall: Anoop suggested leaving the old versions there. Perhaps this can be revisited in the future.

core/sql/sqlcomp/CmpSeabaseDDLrepos.cpp
DaveBirdsall: I do not know of any such plans.

core/sql/sqlcomp/CmpSeabaseDDLrepos.cpp
DaveBirdsall: The EXPLAIN_PLAN column in the METRIC_QUERY_TABLE was increased in size from 200000 to 1000000 as part of these changes.

core/sql/parser/sqlparser.y
anoopsharma00: In the beginning there were very limited variation to DDLExpr class so param approach was ok.
But now that there are many more options and variations to DDLExpr, it will be better to
set the bit flags instead of adding new params.

Maybe someone can clean this up at some point and change the callers to set bit flags 
when DDLExpr class is created in parser.

DaveBirdsall: I agree. Cleaning this up is a good idea. I will do it under a separate JIRA. I have written JIRA TRAFODION-2111 for this purpose.

core/sql/sqlcomp/CmpSeabaseDDLcommon.cpp
DaveBirdsall: Will do.

core/sql/sqlcomp/CmpSeabaseDDLcommon.cpp
DaveBirdsall: Will do.

DaveBirdsall:
@robertamarton, I've made two of the changes you suggested (adding logging, and adding enums). For one of your other suggestions, I've created a JIRA TRAFODION-2111 and will address that matter seperately.

PR #566, opened 2016-06-29
TRAFODION 2034

amandamoran:


amandamoran:
jenkins, extra tests

install/installer/traf_cloudera_mods
```diff
@@ -367,27 +370,37 @@ ssh -q -n $HDFS_NODE 'sudo su hdfs --command "' "$HADOOP_BIN_PATH"'/hdfs dfsadmi
 #       enabled and HDFS has been restarted
 echo "***INFO: Setting HDFS ACLs for snapshot scan support"
 ssh -q -n $HDFS_NODE 'sudo su hdfs --command "' "$HADOOP_BIN_PATH"'/hdfs dfs -mkdir -p /hbase/archive"'
-if [ $? != 0 ]; then
+errorFound=$?
+echo "ERROR Code: $errorFound"
```
DaveBirdsall: Do you want this echo here? Users will see a series of "ERROR Code: 0" messages in normal circumstances and wonder what all the fuss is about. Perhaps the value of $errorFound should be displayed inside the 'if' following?

amandamoran:
jenkins, extra tests

sandhyasun:
The extra tests are going to fail since there is a hang issue during a TM rollback. We're still waiting for analysis on that rollback issue to get the regressions clean before anyone can run 'extra tests'.

install/installer/traf_cloudera_mods
sandhyasun: The "echo" was added earlier on to help figure out the error if any from the command. SO that's correct, the echo shodul be within the "if" condition and only if it's non zero.

amandamoran: Fixed.

sureshsubbiah:
+1. Looks good.

PR #567, opened 2016-06-30
[TRAFODION-2089] JDBC_T2 Makefile not add INCLUDE path for log4cxx

traflm:
Following the build instruction on wiki in from a bare metal system, JDBCT2 build fail if one doesn't install log4cxx into system location.
Change JDBCT2 Makefile same to all other components to contain log4cxx referenced by LOG4CXX_INC_DIR


DaveBirdsall:
Looks good to me.

PR #568, opened 2016-06-30
TRAFODION-2016

mashengchen:
support hbase 1.1 & cdh 5.5
for developer pls do set HBASE_DISTRO before source sqenv.sh
export HBASE_DISTRO=APACHE1.0 (APACHE HBASE1.0)
export HBASE_DISTRO=APACHE1.1 (APACHE HBASE1.1)
export HBASE_DISTRO=CDH5.4    (cloudera 1.0.0-cdh5.4.4) 
export HBASE_DISTRO=CDH5.5    (cloudera 1.0.0-cdh5.5.1) 
export HBASE_DISTRO=HDP       (hortonworks 2.3)


traflm:
looks good to me

traflm:
Maybe we should get @prashanth-vasudev to review this PR? I feel it is good, this will be very important, and we still need to support more Hbase versions, especially HBase 1.2. So it will be better that we can decide to merge this or point out problems still need to fix in this PR.

amandamoran:
Where is HBASE_DISTRO set? If this variable needs to be used and set we need to set it in the installer.

mashengchen:
hi Amanda, i suggest HBASE_DISTRO set in .bash_profile, and this is only used for developers.  It doesn't need to set in the installer, because while installation , the installer will get the hbase version by user input 1: Cloudera, 2: Hortonworks, 3: Other.

amandamoran:
@mashengchen That's true, but the installer is not setting HBASE_DISTRO. It sets "HADOOP_TYPE". The installer can set this variable if you need, but it isn't right now.

mashengchen:
@amandamoran ooh, yes ,you are right , it might set by HADOOP_TYPE&HBASE VERSION, i am now not sure where it put this , do you have some suggestions?

amandamoran:
@mashengchen Where HADOOP_TYPE (traf_config_setup) is set I would add HBASE_DISTRO there, or change HBASE_DISTRO to just use HADOOP_TYPE.

mashengchen:
@amandamoran i think add hbase_distro to traf_conf_setup is good

mashengchen:
@amandamoran So , you will do the 'add HBASE_DISTRO' in another PR? or I just add it in this PR ?

selvaganesang:
Please edit the subject line to convey what this PR does. It would be helpful when someone uses git log to determine what are the changes that went in

sandhyasun:
Probably a good idea to run extra tests with this change. I will trigger that.

sandhyasun:
jenkins, extra tests

core/sqf/src/seatrans/hbase-trx/Makefile
```diff
@@ -54,35 +65,49 @@ jdk_1_7_mapr: build_chk_mapr
...
+	$(UNCOMMENT_STRING) HBASE1.0
+	set -o pipefail && $(MAVEN) -f pom.xml.apache10 package -DskipTests | tee build_trx.log
 	mkdir -p $(MY_SQROOT)/export/lib
-	cp -pf target/$(BLD_HBASE_APACHE_TRX_JARNAME) $(MY_SQROOT)/export/lib
+	cp -pf target/$(BLD_HBASE_APACHE10_TRX_JARNAME) $(MY_SQROOT)/export/lib
 	$(RM) $(VFILE)
+
+jdk_1_7_apache11: build_chk_apache11
 	$(REINSTATE_ORIG)
+	$(UNCOMMENT_STRING) HBASE1.1
```
prashanth-vasudev: apache 11 and hbase 1.1 is used interchangeably here. Can we make this consistent so that there is no ambuiguity?
May be change the tag to Apache11?

prashanth-vasudev:
Changes look good to me. Please have HBASE_DISTRO set by installer in trafodion user bashrc so that correct settings are picked up from sqenvcom.sh when trafodion user logs in.  Also please do basic tests with different distros using the installer.  In the developer environment, please run full developer regression tests.

mashengchen:
@prashanth-vasudev i have do the basic tests with different distros using the installer, also do sql regression in develop env.
so can this be merge?

sandhyasun:
Looks ready to merge now.

PR #574, opened 2016-07-05
Fix for [TRAFODION-2091] Lobglobals and hdfsFS handle  fix

sandhyasun:
hdfsFS handling is now done at the Context level and is maintained as  hashqueue for each hdfsPort/hdfsSever combination.
The first time a component connects, it checks thae global list, if the handle is not found a new entry is created and added to the list.
All calls to individual callers making hdfs Disconnect calls are removed since they affect other components usage of hdfs API. 
Cleanup in how LobGlobals are allocated and maintained. They are part of ex_globals and TCBs get a pointer to the ExLobGlobals when they need to use the LOB interface.
Fixed dataModCheck and emptyDirectory to change how lobGlobals initialization happens.


core/sql/src/main/java/org/trafodion/sql/HBulkLoadClient.java
```diff
@@ -475,10 +475,12 @@ public boolean release( ) throws IOException {
        writer.close();
        writer = null;
     }
-    if (fileSys !=null)
+    // sss This is one place that is unconditionally closing the
```
sureshsubbiah: Should we remove "sss". It is a minor nit, and I know it is an oversight. Maybe this can be done in a subsequent checkin?

core/sql/regress/hive/TEST005
```diff
@@ -314,7 +314,7 @@ select a from hive.hive.thive;
 select * from hive.hive.thive;
 
 -- truncate of hive data
-cqd query_cache '0';
+--cqd query_cache '0';
```
sureshsubbiah: Glad that the test is being run with qcache ON now.

sureshsubbiah:
+1. Looks great. Thank you for taking on this JIRA.

core/sql/src/main/java/org/trafodion/sql/HBulkLoadClient.java
sandhyasun: WIll change this.

sandhyasun:
jenkins, extra tests

sandhyasun:
jenkins, extra tests

sandhyasun:
Extra tests failed due to 2 reasons  executor/TEST022 - due to PR 564. hive/TEST018 needs investigation . So looking into it.

sandhyasun:
The hive/TEST018 is not because of the globals issue.After any unload statement, if a select  statement from the target table immediately, hdfs seems to return an EAGAIN error. I have added a sleep after every UNLOAD for now. But this needs to be addressed separately in a new JIRA and our hdfs  selects will need a retry mechanism for certain errors like EAGAIN.

anoopsharma00:
Changes look good.

There are multiple places where code remains within "__ignore" def
or with initials ('sss') or commented out code.
Will those be cleaned up before merge or they left in to be enabled
for some future work?

check tests have failed. Is this due to the issue that is being
discussed for a different pull request?

sandhyasun:
They will get cleaned up in a future cleanup work that's in the pipeline. Want to keep the comments in case these globals changes cause any issues in QA/stress testing- they will be easy to identify. The check tests are failing due to the JDBC issue being worked on separately.

sandhyasun:
jenkins, retest

sureshsubbiah:
Is this good to be merged now?

PR #575, opened 2016-07-05
Fix JIRA306

ryzuo:
Add "WHERE" clause to conditioning select the specific data type, which should be returned to ODBC SQLGetTypeInfo() call.


sureshsubbiah:
+1. Looks good. This is a simple fix to an old problem. Thank you.

arvind-narain:
Thanks for the change RuoYu. Sorry couldn't look into this yesterday. 

Do please check on the API. SQLGetTypeInfo() - it can also return back all the types supported by database. Hence the additional where clause should be added based on a condition. 

Currently jdbc tests are failing since expected count of types is 35 and we get back 0.

Java.lang.AssertionError: Number of datatypes expected:&lt;35&gt; but was:&lt;0&gt;
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:743)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:555)
	at TestCat.TestBasic16(TestCat.java:524)

arvind-narain:
Also do please adjust the count in this test. Thanks.

PR #577, opened 2016-07-06
[TRAFODION-1673] support with clause

traflm:
initial support new SQL syntax 'WITH', which required in many use cases, for example, in 35 TPCDS queries.
This change will transform the WITH definition into table typed sub-query, so can be used in later compile phases.
As tested, in 35 TPCDS queries that contains WITH, this change pass 19 of them , others due to other syntax gap, like rollup. The passed TPCDS queries include:
 q1
q2
q4
q11
q23
q24
q30
q31
q33
q51
q54
q56
q58
q59
q60
q64
q81
q83
q97

Since still not widely tested, this syntax is still protected by MODE_SPECIAL_4 CQD.

Please help to code review it.

### DIFF
--- a/core/sql/bin/SqlciErrors.txt
+++ b/core/sql/bin/SqlciErrors.txt
@@ -410,6 +410,7 @@
 1430 3F000 99999 BEGINNER MAJOR DBADMIN A schema name that starts and ends with an "_"(underscore) is reserved for internal usage. It cannot be used to create a user schema.
 1431 ZZZZZ 99999 BEGINNER MINOR DBADMIN Object $0~String0 exists in HBase. This could be due to a concurrent transactional ddl operation in progress on this table.
 1432 ZZZZZ 99999 BEGINNER MINOR DBADMIN Input LOB type $0~Int0 does not match column's storage type: $1~Int1 Column name: $0~String0 .
+1433 ZZZZZ 99999 BEGINNER MINOR DBADMIN WITH clause redefined. WITH name $0~String0 .
 1500 ZZZZZ 99999 ADVANCED CRTCL DIALOUT The CATSYS - CAT_REFERENCES system schema relationship for catalog $0~CatalogName might be corrupt.
 1501 ZZZZZ 99999 UUUUUUUU UUUUU UUUUUUU --- unused as of 5/7/12 ---
 1502 ZZZZZ 99999 ADVANCED CRTCL DIALOUT The OBJECTS - REPLICAS definition schema relationship for $0~String0 $1~TableName might be corrupt.
--- a/core/sql/bin/SqlciErrors.txt
+++ b/core/sql/bin/SqlciErrors.txt
@@ -410,7 +410,6 @@
 1430 3F000 99999 BEGINNER MAJOR DBADMIN A schema name that starts and ends with an "_"(underscore) is reserved for internal usage. It cannot be used to create a user schema.
 1431 ZZZZZ 99999 BEGINNER MINOR DBADMIN Object $0~String0 exists in HBase. This could be due to a concurrent transactional ddl operation in progress on this table.
 1432 ZZZZZ 99999 BEGINNER MINOR DBADMIN Input LOB type $0~Int0 does not match column's storage type: $1~Int1 Column name: $0~String0 .
-1433 ZZZZZ 99999 BEGINNER MINOR DBADMIN WITH clause redefined. WITH name $0~String0 .
 1500 ZZZZZ 99999 ADVANCED CRTCL DIALOUT The CATSYS - CAT_REFERENCES system schema relationship for catalog $0~CatalogName might be corrupt.
 1501 ZZZZZ 99999 UUUUUUUU UUUUU UUUUUUU --- unused as of 5/7/12 ---
 1502 ZZZZZ 99999 ADVANCED CRTCL DIALOUT The OBJECTS - REPLICAS definition schema relationship for $0~String0 $1~TableName might be corrupt.
@@ -1027,6 +1026,7 @@ $1~String1 --------------------------------
 3285 ZZZZZ 99999 BEGINNER MINOR DBADMIN The C interface for this type of routine is deprecated, please consider migrating to a TMUDF with a C++ or Java interface.
 3286 ZZZZZ 99999 BEGINNER MINOR DBADMIN The routine language that was either specified or implied from the library file name is not supported for this type of routine.
 3287 42000 99999 BEGINNER MINOR DBADMIN The $0~string0 option is not supported for this type of routine.
+3288 42000 99999 BEGINNER MINOR DBADMIN A syntax error occurred. WITH clause redefined. WITH name $0~String0 .
 3300 42000 99999 BEGINNER MAJOR DBADMIN --- available ---
 3301 42000 99999 BEGINNER MAJOR DBADMIN $0~string0 is too long (longer than $1~int0 $2~string1).
 
--- a/core/sql/parser/sqlparser.y
+++ b/core/sql/parser/sqlparser.y
@@ -1927,6 +1927,8 @@ static void enableMakeQuotedStringISO88591Mechanism()
 %type <relx>  		        table_name_as_clause_hint_and_col_list
 %type <relx>  		        rel_subquery_and_as_clause
 %type <relx>  		        rel_subquery_as_clause_and_col_list
+%type <relx>                    with_clause
+%type <relx>                    with_clause_list
 %type <hint>      		optimizer_hint
 %type <hint>      		hints
 %type <hint>      		index_hints
@@ -6579,6 +6581,9 @@ table_reference : table_name_and_hint
               | rel_subquery_and_as_clause
                 { $$ = $1; }
 
+              | with_clause_list
+                {  $$ = $1;}
+
               | rel_subquery_as_clause_and_col_list
                 { $$ = $1; }
 
@@ -6647,6 +6652,26 @@ table_as_tmudf_function : TOK_UDF '(' table_mapping_function_invocation ')'
 
 table_name_and_hint : table_name optimizer_hint hbase_access_options
                       {
+                        NAString tmp = ((*$1).getQualifiedNameAsString());
+                        if(SqlParser_CurrentParser->with_clauses_->contains(&tmp) )
+                        {
+                          RelExpr *re = SqlParser_CurrentParser->with_clauses_->getFirstValue(&tmp);
+                          $$=re->copyTree(PARSERHEAP());
+                        }
+                        else
+                        {
+                          $$ = new (PARSERHEAP()) Scan(*$1);
+                          if ($2)
+                            $$->setHint($2);
+
+                          if ($3)
+                            ((Scan*)$$)->setHbaseAccessOptions($3);
+
+                          delete $1;
+                        }
+                      }
+/*
+                      {
                         $$ = new (PARSERHEAP()) Scan(*$1);
                         if ($2) 
                           $$->setHint($2);
@@ -6656,6 +6681,7 @@ table_name_and_hint : table_name optimizer_hint hbase_access_options
                         
                         delete $1;
                       }
+*/
                     | '(' table_name_and_hint ')'
                       {
                         CheckModeSpecial1();
@@ -6953,13 +6979,22 @@ del_stmt_w_acc_type_rtn_list_and_as_clause_col_list : '('  delete_statement acce
 
 table_name_as_clause_and_hint : table_name as_clause optimizer_hint hbase_access_options
                                 {
-                                  $1->setCorrName(*$2);
-                                  $$ = new (PARSERHEAP()) Scan(*$1);
-                                  if ($3) 
-                                    $$->setHint($3);
-                                  if ($4)
-                                    ((Scan*)$$)->setHbaseAccessOptions($4);
-                                   
+                                   NAString tmp = ((*$1).getQualifiedNameAsString());
+                                   if(SqlParser_CurrentParser->with_clauses_->contains(&tmp) )
+                                   {     
+                                     RelExpr *re = SqlParser_CurrentParser->with_clauses_->getFirstValue(&tmp);
+                                     RenameTable *rt = new (PARSERHEAP()) RenameTable(re, *$2);
+                                     $$=rt->copyTree(PARSERHEAP());
+                                  }
+                                  else
+                                  {
+                                    $1->setCorrName(*$2);
+                                    $$ = new (PARSERHEAP()) Scan(*$1);
+                                    if ($3) 
+                                      $$->setHint($3);
+                                    if ($4)
+                                      ((Scan*)$$)->setHbaseAccessOptions($4);
+                                  }  
                                   delete $1;
                                   delete $2;
                                 }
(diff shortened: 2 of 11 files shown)


DaveBirdsall:
Very glad to see this enhancement. Looking forward to reviewing it!

core/sql/parser/sqlparser.y
```diff
@@ -6579,6 +6581,9 @@ table_reference : table_name_and_hint
               | rel_subquery_and_as_clause
                 { $$ = $1; }
 
+              | with_clause_list
```
zellerh: The ANSI SQL standard defines the "with" clause as part of a query_expression, so I would recommend doing the same here (non_join_query_expression in our case). Also, it would be good to accept the WITH RECURSIVE syntax and to give an error that this is not yet supported.

zellerh: Here is the syntax rule from the ANSI standard:

<query expression> ::=
[ <with clause> ] <query expression body>

core/sql/parser/sqlparser.y
```diff
@@ -7001,6 +7036,46 @@ rel_subquery_and_as_clause : rel_subquery as_clause
                                $$ = $2;
                              }
 
+with_clause_list : with_clause
+                     {  $$ = $1 ; }
+
+                   | with_clause_list ',' correlation_name TOK_AS rel_subquery
```
zellerh: Can we do this in a way that doesn't have two different rules for the first and the subsequent parts of the WITH clause? See comment above for how the ANSI standard handles the WITH clause.

core/sql/parser/sqlparser.y
```diff
@@ -14684,6 +14759,11 @@ optional_limit_spec : TOK_LIMIT NUMERIC_LITERAL_EXACT_NO_SCALE
 
 dml_statement : dml_query { $$ = $1; }
 
+               | with_clause_list dml_query
```
zellerh: Same comment as a above about where the parser rule for the WITH clause should be.

zellerh: Related question: When you made this change, did the number of shift/reduce and reduce/reduce conflicts change?

core/sql/sqlcomp/parser.h
```diff
@@ -261,6 +261,8 @@ ItemExpr *get_w_ItemExprTree(const NAWchar * str,
   NABoolean isHQCCacheable()
     { return HQCKey_?HQCKey_->isCacheable():FALSE;  }
 
+  NAHashDictionary<NAString,RelExpr> *with_clauses_;
```
zellerh: Can you make this a private data member?

zellerh:
Amazing that such a simple change can make such a big difference. Very cool!

core/sql/bin/SqlciErrors.txt
```diff
@@ -410,6 +410,7 @@
 1430 3F000 99999 BEGINNER MAJOR DBADMIN A schema name that starts and ends with an "_"(underscore) is reserved for internal usage. It cannot be used to create a user schema.
 1431 ZZZZZ 99999 BEGINNER MINOR DBADMIN Object $0~String0 exists in HBase. This could be due to a concurrent transactional ddl operation in progress on this table.
 1432 ZZZZZ 99999 BEGINNER MINOR DBADMIN Input LOB type $0~Int0 does not match column's storage type: $1~Int1 Column name: $0~String0 .
+1433 ZZZZZ 99999 BEGINNER MINOR DBADMIN WITH clause redefined. WITH name $0~String0 .
```
zellerh: Parser errors should be in the range 3000-3999. Also, since this is a syntax error, use 42000 (the SQLSTATE value for syntax errors) instead of "ZZZZZ".

traflm:
thanks Hans for valuable comments. I will submit a new version including all changes asap. Some comments, I need your help to clarify, I don't quite understand, I will add reply to those at above.

core/sql/parser/sqlparser.y
traflm: @zellerh , I don't understand what is "number of shift/reduce and reduce/reduce conflicts". How to check this number? What is this?

selvaganesang:
I believe JDBC ( and may be ODBC too) parses the first word in the SQL command to determine if the SQL command produces the result set or not. So, you may want to check the 'with' command from trafci to confirm it it works from it. If not, please consider making the change in JDBC  to flag the 'with' command to the one that produces the result set

core/sql/parser/sqlparser.y
zellerh: That is the number of conflicts encountered by the bison parser when it processes file core/sql/parser/sqlparser.y. Here is how I get this information:

```
touch core/sql/parser/sqlparser.y
cd core/sql/nskgmake
gmake -j 4 -ks linuxdebug
```

Here is the current output:

```
../parser/sqlparser.y: warning: 73 shift/reduce conflicts [-Wconflicts-sr]
../parser/sqlparser.y: warning: 12 reduce/reduce conflicts [-Wconflicts-rr]
```

If you get a different (higher) number, then we would have to investigate whether there are any potential problems with the new syntax.

sureshsubbiah:
jenkins, retest

traflm:
@selvaganesang , that is a good point, I still not try trafci, will do that later. I will be on biz trip for few days, so will not progress on this for a short while. Will continue when come back.

venkat1m:
We also capture query statistics in the repository table. This include detailed statistics for each query that has run longer than a minute and also aggregate metrics for short running queries based on statement type. For example total selects, total inserts etc. We need to check the related to make sure this WITH statement is mapped accordingly to a SELECT if used with a SELECT statement.

traflm:
@venkat1m yes, that is a very good point too. I will update the JIRA to include these requirements.

core/sql/parser/sqlparser.y
traflm: @zellerh , I cannot figure out how to do this. Could you please give me an example? I am very new to the parser.

sureshsubbiah:
+1. Looks good. I did not imagine this could be done in the parser alone.

core/sql/parser/sqlparser.y
```diff
@@ -6647,6 +6652,26 @@ table_as_tmudf_function : TOK_UDF '(' table_mapping_function_invocation ')'
 
 table_name_and_hint : table_name optimizer_hint hbase_access_options
                       {
+                        NAString tmp = ((*$1).getQualifiedNameAsString());
+                        if(SqlParser_CurrentParser->with_clauses_->contains(&tmp) )
+                        {
+                          RelExpr *re = SqlParser_CurrentParser->with_clauses_->getFirstValue(&tmp);
+                          $$=re->copyTree(PARSERHEAP());
```
DaveBirdsall: Should we delete $1 in this code path as well?

core/sql/parser/sqlparser.y
DaveBirdsall: One way would be:

with_clause : WITH with_clause_elements

with_clause_elements : with_clause element
                                  |
                                     with_clause_elements with_clause_element

with_clause_element : correlation_name TOK_AS rel_subquery

By the way, I notice the way this is coded, the "with_clause" non-terminal is a single correlation_name TOK_AS rel_subquery element. It seems more natural (to me anyway) to think of the with_clause as being everything from the WITH keyword through the entire list of subqueries.

core/sql/parser/sqlparser.y
DaveBirdsall: I think this is incorrect. Making "with_clause_list" an alternative for table_reference would allow things like: "select * from with c1 as (select * from t)".

DaveBirdsall: I agree with Hans; it would be better to move the wish_clause non-terminal to the query_expression production. Right now, that production reads:

query_expression : non_join_query_expression

You could change it to:

query_expression: with_clause non_join_query_expression
                            | non_join_query_expression,

or alternatively,

query_expression : with_clause non_join_query_expression

And then code with_clause to allow an empty production, for example:

with_clause :     { $$ = NULL; /* empty with_clause case */ }
                        |  TOK_WITH with_clause_elements

DaveBirdsall:
A parser-only approach is good for now. It makes WITH entirely a syntactic sugar in Trafodion.

In the future, we'll want to reconsider this. WITH (without RECURSIVE) is a form of manual common sub-expression elimination. Some merchant database engines have multiple possible plans for these queries: The subquery could be evaluated once and materialized, then that result is plugged into the query tree wherever needed. Or the subquery can be "inlined", meaning that it will be re-evaluated each time it is encountered. A syntax-only approach doesn't allow us to do this. Though I could imagine that if we put relational common-subexpression logic into our normalizer, we could then consider it.

A minor issue with the syntax-only approach is that if a subquery in the WITH clause is not referenced, we don't do any error checking on it. So, I could say, WITH T1 AS (SELECT * FROM NON_EXISTENT_TABLE) SELECT * FROM EXISTING_TABLE and not get an error. I don't know what the standard says about this or what other products do. It is good enough for now.

core/sql/parser/sqlparser.y
traflm: hi, Hans, Dave,
the current change doesn't get higher number of conflicts, it is still 73/12, but if I tried to move the rule into query_expression, it will increase the shift/reduce conflicts, I still cannot find a good way to write a correct syntax.  It is rather difficult to find the root cause of shift/reduce conflict

zellerh: Yes, I got a similar result. I tried this:

    query_expression : with_clause non_join_query_expression
                     | non_join_query_expression

That gives me three additional reduce/reduce conflicts which I don't understand. Here are two of them:

    State 1941

    1870 query_expression: with_clause non_join_query_expression .
    1871                 | non_join_query_expression .

      TOK_EXCEPT  reduce using rule 1870 (query_expression)
      TOK_EXCEPT  [reduce using rule 1871 (query_expression)]
      TOK_UNION   reduce using rule 1870 (query_expression)
      TOK_UNION   [reduce using rule 1871 (query_expression)]
      $default    reduce using rule 1870 (query_expression)


I also tried this:

    query_expression : optional_with_clause non_join_query_expression

That doesn't cause additional reduce/reduce conflicts, but we get a very large number of additional shift/reduce conflicts, something in the neighborhood of 50-100.

If we can't get the "correct" parser rules to work, then maybe we can stick with the approach you chose (assuming it accepts all the valid WITH clauses) and handle any illegal syntax in the semantic actions.

traflm:
@DaveBirdsall  , yes, this is only the initial attempt to support with_clause. We need to find better ways to improve this in the later iteration. This is only a syntax sugar as you described. 
I am thinking: this will treat the with_clause just as a subquery, so if it is used only once in a query, all same optimizations Trafodion uses for subquery should be able to apply as well, such as subquery to join, or materialize result once, etc. I cannot see any difference, I will do some experiment after the syntax and all above issues solved. Write a query and check the plan. 
In the JIRA 1673, QiFan proposed a new CTE processing idea, that should be implemented in later iteration.
in sum, this is just an initial effort, and I think it is more about parser, and all parser changes can be reused in later iterations.

core/sql/parser/sqlparser.y
traflm: I found a way to avoid the reduce/shift and reduce/reduce error by add with_clause down to the underlying rules of non_join_query_expression, instead of add it at same level of non_join_query_expression. non_join_query_expression can represent many different styles of query expression, including utilities, which we don't want a with_clause as prefix, so in the underlying rules, only add with_clause before those SELECT syntax, and it solve the problem.

And it seems if use redundant rule explicitly (looks annoying) , we can avoid the error. 
For example:
  A : b  | c
is better to change to 
  A : b 
  A : c
I am still trying to understand this.

I checked with PostgreSQL parser, they are doing similar things, comments from postgresql parser:

 * This rule parses the equivalent of the standard's <query expression>.
 * The duplicative productions are annoying, but hard to get rid of without
 * creating shift/reduce conflicts.

core/sql/sqlcomp/parser.h
traflm: yes, changed

core/sql/bin/SqlciErrors.txt
traflm: yes, changed to 3288

core/sql/parser/sqlparser.y
traflm: yes, I am not quite sure how memory was managed in the parser, I  changed to delete $1

traflm:
@selvaganesang trafci works with this with_clause as well. 
@venkat1m how to check repository compatibility? Is there any hint where should I look at?

core/sql/parser/sqlparser.y
zellerh: Thanks, that's excellent news!

core/sql/parser/sqlparser.y
```diff
@@ -13333,14 +13396,41 @@ query_specification : select_token set_quantifier query_spec_body
...
+
+    }
+
+/* type relx */
+query_specification :with_clause select_token set_quantifier query_spec_body 
+     {
+       if ($3) {
+         $$ = new (PARSERHEAP())
+           RelRoot(new (PARSERHEAP())
+                   GroupByAgg($4
```
DaveBirdsall: I don't understand why we stick a GroupByAgg node here.

core/sql/parser/sqlparser.y
```diff
@@ -13418,6 +13507,55 @@ query_specification : select_token '[' firstn_sorted NUMERIC_LITERAL_EXACT_NO_SC
...
               CmpCommon::getDefault(MVQR_LOG_QUERY_DESCRIPTORS) == DF_DUMP_MV)
             ((RelRoot*)$$)->setAnalyzeOnly();
+           } 
+/* type relx */
+query_specification : with_clause select_token '[' firstn_sorted NUMERIC_LITERAL_EXACT_NO_SCALE ']' set_quantifier query_spec_body
+	{
+	  if ($7) {
+	    $$ = new (PARSERHEAP())
+	      RelRoot(new (PARSERHEAP())
+		      GroupByAgg($8
```
DaveBirdsall: I don't understand why we add a GroupByAgg node here.

core/sql/parser/sqlparser.y
```diff
@@ -7000,6 +7024,46 @@ rel_subquery_and_as_clause : rel_subquery as_clause
...
+                            *SqlParser_Diags << DgSqlCode(-3288)
+                                             << DgString0((*$1).toCharStar());
+                             delete $1;
+                             delete $4;
+                             YYERROR;
+                          }
+
+                          SqlParser_CurrentParser->insertWithDefinition($1 , $$);
+                          delete $1;
+                          delete $4;
```
DaveBirdsall: I don't think we want to delete $1 and $4 here, since the RelRoot node points to $4 and the RenameTable node points to $1.

core/sql/parser/sqlparser.y
traflm: if($7) means there is a set_quantifier like (DISTINCT), so it will add a GroupByAgg node here.

core/sql/parser/sqlparser.y
traflm: if ($3) means there is a set_quantifier token, which is DISTINCT for example, so need a GroupByAgg node here.

core/sql/parser/sqlparser.y
traflm: I agree, will change it.

traflm:
I will remove the wrong 'delete $1; delete $4' and fix the regression failure and make another commit, please help to review again. 

RECURSIVE is now a TOKEN , so I change the core/TEST037 to remove it as a reserved word test.

I plan to make next commit next Monday.

traflm:
jenkins, extra tests

core/sql/parser/sqlparser.y
traflm: sorry, I made mistake, by running TESTOK2, it shows the change still introduce one more reduce/reduce conflict. Still looking at it.

traflm:
modify the yacc file to remove the reduce/reduce conflict. The reduce/reduce seems to be related between no_rollback and query_expression, but very hard to make it correct.

ANSI defined 'with clause' is as Hans pointed out, in Trafodion yacc:
query_expression : non_join_query_expression
				| 	with_clause non_join_query_expression

What I proposed to define is:

dml_statement : dml_query { $$ = $1; }

               | with_clause dml_query
                 {
                   $$ = $2;
                 }

dml_query : query_expression order_by_clause access_type
            optional_lock_mode for_update_spec optional_limit_spec
 
          | query_expression order_by_clause_non_empty
            group_by_clause_non_empty access_type optional_lock_mode
            for_update_spec optional_limit_spec

From syntax point of view, these two definitions are same, and ANSI has no dml_query definition, so we cannot 100% aligned with ANSI grammar IMHO.

The second definition will not introduce new reduce/reduce or shift/reduce conflict. I will run a extra test after normal test finished.

core/sql/parser/sqlparser.y
```diff
@@ -6647,15 +6651,26 @@ table_as_tmudf_function : TOK_UDF '(' table_mapping_function_invocation ')'
...
                       {
-                        $$ = new (PARSERHEAP()) Scan(*$1);
-                        if ($2) 
-                          $$->setHint($2);
+                        NAString tmp = ((*$1).getQualifiedNameAsString());
+                        if(SqlParser_CurrentParser->hasWithDefinition(&tmp) )
+                        {
+                          RelExpr *re = SqlParser_CurrentParser->getWithDefinition(&tmp);
+                          $$=re->copyTree(PARSERHEAP());
+                          delete $1;
```
zellerh: One thing we could do here is to raise an error if we have an optimizer hint or HBase access options at this point. Those are allowed in the grammar, but we wouldn't be able to apply them to a more complicated WITH clause.

core/sql/parser/sqlparser.y
```diff
@@ -6953,13 +6968,22 @@ del_stmt_w_acc_type_rtn_list_and_as_clause_col_list : '('  delete_statement acce
...
-                                    $$->setHint($3);
-                                  if ($4)
-                                    ((Scan*)$$)->setHbaseAccessOptions($4);
-                                   
+                                   NAString tmp = ((*$1).getQualifiedNameAsString());
+                                   if(SqlParser_CurrentParser->hasWithDefinition(&tmp) )
+                                   {     
+                                     RelExpr *re = SqlParser_CurrentParser->getWithDefinition(&tmp);
+                                     RenameTable *rt = new (PARSERHEAP()) RenameTable(re, *$2);
+                                     $$=rt->copyTree(PARSERHEAP());
```
zellerh: Same as above, could give an error if we see a hint or an HBase access option.

zellerh:
Very nice, very elegant, very simple!! I love how these few lines of change enable such a big feature. Two small comments, but I think we should commit this code and address the comments later. One is in line comments, we may need a couple of error messages. Here is a second comment: Putting the WITH clause as part of a DML statement works fine at the top level, but it does not allow WITH clauses in query expressions that are not at the top level. I think the ANSI standard allows those. My suggestion: Let's live with that restriction for now, the WITH clause makes the most sense at the outermost level.

PR #583, opened 2016-07-07
[TRAFODION-1676] rework to solve TEST022 failure

traflm:
Modify the EXPECT file for executor/TEST022 which broke the regression test.
Sorry that I didn't do a full regression before check-in, should do 'Jenkins, extra tests' to trigger more tests.

### DIFF
--- a/core/sql/regress/executor/EXPECTED022.SB
+++ b/core/sql/regress/executor/EXPECTED022.SB
@@ -3066,12 +3066,12 @@ C1                         C2
 --- 0 row(s) selected.
 >>select cast('-. 3' as int) from t022t8;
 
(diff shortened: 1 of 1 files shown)


zellerh:
Looks good, committing the changes now, since they don't require much discussion and since the check test failures are unrelated (not in executor tests).

traflm:
Jenkins, extra tests

traflm:
thanks Hans for merge this, yes, it is to fix the regression test failure, so should be merged asap. I trigger an extra tests to verify again. I did regression on my dev workstation and seems good.

sandhyasun:
Thanks for making the fix so quickly.

PR #585, opened 2016-07-07
[TRAFODION-2096] create all tpcds hive tables from install_local_hadoop

traflm:
create all TPCDS required table in hive, and genreate data. So it will be easier to run and test using TPCDS queries.
This is helpful for future testing, for example, by introducing 'WITH clause' or 'rollup', one can run a same query in Hive and in Trafodion and compare the result, using real TPCDS queries.

By this change, I also want to check the extra Jerkins test.

### DIFF
--- a/core/sqf/sql/scripts/install_hadoop_regr_test_env
+++ b/core/sqf/sql/scripts/install_hadoop_regr_test_env
@@ -122,10 +122,10 @@ done
 # (or most recently added item)
 if [[ "$FORCE" != "true" ]]
 then
-  $MY_HIVE_CMD -e 'describe store_orc;' >/dev/null 2>&1
+  $MY_HIVE_CMD -e 'describe ship_mode;' >/dev/null 2>&1
   if (( $? == 0 ))
   then
-    echo "Hive table store_orc found. Skipping TPC-DS set-up."
+    echo "Hive table ship_mode found. Skipping TPC-DS set-up."
     exit 0
   fi
 fi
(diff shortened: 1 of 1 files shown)


zellerh:
Looks good to me.

svarnau:
The short-circuit check at line 125 needs to be updated to check for one of the new tables.

traflm:
Jenkins, extra tests

traflm:
@svarnau let me fix this, thanks to point it out.

traflm:
jenkins, retest

zellerh:
@traflm, I think Steve wanted the test to check whether the additional tables are there or not. If we test store_orc, existing setups won't create the new tables. Will that be a problem?

sandhyasun:
Your extra test failed because of a hang issue that has now been fixed. You can try them again and ensure hive tests run. They may not pass completely but if the result matches that of the master daily results, it shoudl be fine. I will trigger the tests again now.

sandhyasun:
jenkins, extra tests

sandhyasun:
@traflm . the short circuit check should ideally check the last item created  as you have pointed out. But if that is so, you have to alert the dev list to rerun the whole install_local_hadoop to get the new tpcds tables. Or better yet enhance the check to check for all the tables to be present.

traflm:
@sandhyasun I see, I think I can enhance it to check all tables.

traflm:
I think Steve's proposal is the simplest and best I prefer. I will append all CREATE statement after the previous last table, and keep the checking logic as the last one. Anyone who add new tables, if follow the same rule: append the DDL , and update the checking table. It is simple and seems solve concerns that one already run install_local_hadoop.

DaveBirdsall:
Is this ready to merge now?

zellerh:
Yes, looks good to merge to me.

sandhyasun:
Looks fine ! WIll merge this now.

PR #586, opened 2016-07-07
[TRAFODION-2092] Include "incubating" in installer package name

svarnau:
Portion of this JIRA previously taken care of with Anu's change.

Will likely also backport these changes to release2.0 branch in case we need another 2.0.x patch release.


amandamoran:
+1

PR #588, opened 2016-07-08
Backing out change for JIRA-2046 until exception handling is improved

sbroeder:


sandhyasun:
jenkins, extra tests

sandhyasun:
Starting a full set to see if the memory exception is handled and executor suite doesn't hang anymore.

sandhyasun:
Merging this  now. Rework of the original fix needs to be discussed. Hence JIRA-2046 is now reopened.

PR #589, opened 2016-07-08
[TRAFODION-2097] :Adding retry logic when making hdfsOpenFile calls to read an hdfs file [

sandhyasun:
Added a retry loop for hdfs opens to avoid 8442 errors seen on slower  machines intermittently due  to  potential EAGAIN errors. [TRAFODION-2097]

The "sleep" introduced in hive/TEST018 has now been removed so this retry can handle the EAGAIN errors.

### DIFF
--- a/core/sql/executor/ExHdfsScan.cpp
+++ b/core/sql/executor/ExHdfsScan.cpp
@@ -585,6 +585,17 @@ ExWorkProcRetcode ExHdfsScanTcb::work()
                    openType //
                    );
                 
+                if ((retcode < 0) &&
+                    ((errno == ENOENT) || (errno == EAGAIN)))
+                  {
+                    ComDiagsArea * diagsArea = NULL;
+                    ExRaiseSqlError(getHeap(), &diagsArea, 
+                                    (ExeErrorCode)(EXE_HIVE_DATA_MOD_CHECK_ERROR));
+                    pentry_down->setDiagsArea(diagsArea);
+                    step_ = HANDLE_ERROR_AND_DONE;
+                    break;
+                  }
+
                 // preopen next range. 
                 if ( (currRangeNum_ + 1) < (beginRangeNum_ + numRanges_) ) 
                   {
@@ -737,7 +748,7 @@ ExWorkProcRetcode ExHdfsScanTcb::work()
 		    ExRaiseSqlError(getHeap(), &diagsArea, 
                                     (ExeErrorCode)(EXE_ERROR_FROM_LOB_INTERFACE), NULL, 
                                     &intParam1, 
-                                    &cliError, 
+                                    &errno, 
                                     NULL, 
                                     "HDFS",
                                     (char*)"ExpLOBInterfaceSelectCursor/read",
@@ -1247,7 +1258,7 @@ ExWorkProcRetcode ExHdfsScanTcb::work()
 		    ExRaiseSqlError(getHeap(), &diagsArea, 
                                     (ExeErrorCode)(EXE_ERROR_FROM_LOB_INTERFACE), NULL, 
                                     &intParam1, 
-                                    &cliError, 
+                                    &errno, 
                                     NULL, 
                                     "HDFS",
                                     (char*)"ExpLOBInterfaceSelectCursor/close",
--- a/core/sql/executor/ExHdfsScan.cpp
+++ b/core/sql/executor/ExHdfsScan.cpp
@@ -631,7 +631,7 @@ ExWorkProcRetcode ExHdfsScanTcb::work()
                 ExRaiseSqlError(getHeap(), &diagsArea, 
                                 (ExeErrorCode)(EXE_ERROR_FROM_LOB_INTERFACE), NULL, 
                                 &intParam1, 
-                                &cliError, 
+                                &errno, 
                                 NULL, 
                                 "HDFS",
                                 (char*)"ExpLOBInterfaceSelectCursor/open",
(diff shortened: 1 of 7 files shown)


sandhyasun:
jenkins, extra tests

sandhyasun:
Please don't merge until hive tests have passed. Going to run the extra tests a few times to ensure the EAGAIN error is really handled.

sandhyasun:
jenkins, extra tests

sandhyasun:
jenkins, extra tests

sandhyasun:
jenkins, extra tests

sandhyasun:
jenkins, extra tests

sandhyasun:
jenkins, extra tests

sandhyasun:
jenkins, extra tests

sandhyasun:
Not ready for merge. Still trying a few changes relevant only to Jenkins test environment.

sandhyasun:
jenkins, extra tests

sandhyasun:
jenkins, extra tests

core/sql/executor/ExHdfsScan.cpp
```diff
@@ -585,6 +585,17 @@ ExWorkProcRetcode ExHdfsScanTcb::work()
                    openType //
                    );
                 
+                if ((retcode < 0) &&
+                    ((errno == ENOENT) || (errno == EAGAIN)))
```
DaveBirdsall: errno is a global variable that might be set by any system library call. It is dangerous to rely on it except right after the system call where it is set. But in this code, it looks like the library call is buried inside ExpLOBInterfaceSelectCursor. It would be safer if the latter function saved errno after whatever system library call it calls, and returns that as an "out" parameter, which this code could then check.

core/sql/executor/ExHdfsScan.cpp
```diff
@@ -631,7 +642,7 @@ ExWorkProcRetcode ExHdfsScanTcb::work()
                 ExRaiseSqlError(getHeap(), &diagsArea, 
                                 (ExeErrorCode)(EXE_ERROR_FROM_LOB_INTERFACE), NULL, 
                                 &intParam1, 
-                                &cliError, 
+                                &errno,
```
DaveBirdsall: Again, not a safe usage; errno might have changed by this time (for example, underneath a getHeap() call).

core/sql/exp/ExpLOBaccess.cpp
```diff
@@ -552,8 +552,9 @@ Ex_Lob_Error ExLob::statSourceFile(char *srcfile, Int64 &sourceEOF)
    if (srcType == HDFS_FILE)
      {
        hdfsFile sourceFile = hdfsOpenFile(fs_,srcfile,O_RDONLY,0,0,0);
```
DaveBirdsall: Is this where we are expecting errno to be set? We could capture its value here.

DaveBirdsall:
I think this is a good approach, but I have concerns over the safety of relying on errno to be unchanged across layers of Trafodion code.

core/sql/exp/ExpLOBaccess.cpp
sandhyasun: No not this. There are 3 other places in this file where it's set. But still experimenting with this fix. If this works, a new error number will be generated  right after the call to hdfsOpenFile . If errno is EAGAIN or ENOENT, the new error message will be  returned and we will retry based on that error at the higher layer.  But  the suggestion to return errno and perhaps also the filename as an output param may be a good one too. Will consider it.

sandhyasun:
jenkins, extra tests

sureshsubbiah:
+1. 
Agree with Dave that we should check errno as soon as possible after the call that could potentially raise it. errno seems to be checked in about 30 files in our source (under sql), so we may already be open to the issue that Dave has raised, though most of them could be checking sooner rather than later.

core/sql/exp/ExpLOBaccess.cpp
```diff
@@ -1359,22 +1361,26 @@ Ex_Lob_Error ExLob::openDataCursor(char *file, LobsCursorType type, Int64 range,
...
+      {
+        hdfsCloseFile(fs_, fdData_);
+        fdData_ = NULL;
+        openFlags_ = O_RDONLY;
+        fdData_ = hdfsOpenFile(fs_, lobDataFile_, openFlags_, 0, 0, 0);
+       
+        if (!fdData_)
+          {
+            openFlags_ = -1;
+            *hdfsDetailError = errno;
```
DaveBirdsall: In the function prototype, hdfsDetailError defaults to NULL, So, should we check for null before attempting to store errno in there? (An alternative would be to make hdfsDetailError a reference parameter instead of a pointer; then the caller always has to provide it.)

sandhyasun: Yes will be better to check for NULL. I did make it a reference param initially but there are 4 other files that needed to be touched unnecessarily so made it a pointer. Will make that change to be safe.

sandhyasun:
The hive tests all pass clean now with this change.  Think it's ready for merge once the check tests pass after this small NULL pointer check change. But the merge to master needs to be done by squashing all the commits into one !   The text to be used can be the header of this PR.

sureshsubbiah:
I think this change is ready to be merged though it may not be perfect.
@DaveBirdsall and others please comment if you disagree. 
Upon reading the comments I now see that the retry is done at a higher level (through AQR?) for both EAGAIN and ENOENT.

sandhyasun:
That's correct. Initially for experimenting I had added the retry loop at  the lowest level. It did not help. The retry with this error after further debugging we found was due to a query cache issue and required recompilation.   Hence removed all the local retries and made it an AQR retry. Also the errno is captured immediately after hdfsOpenFile and then passed as an output param all the way to the Scan TCB layer. Is there anything that's missing ?

DaveBirdsall:
It looks good to me. Squash away!

PR #590, opened 2016-07-10
JIRA TRAFODION-2099 Add support for boolean datatype

anoopsharma00:
-- boolean datatype support in create, cast and params
  Ex:
   create table t (a boolean);
   insert into t values (true), (false);
   select * from t where a = true;

-- boolean datatype support for hive columns

-- unarith op support. Currently used for '!' operator that
   negates a boolean value.
   Ex: select * from t where !a = false;

-- support for more tinyint and largeint unsigned conversion/comparison
   operations in expressions and pcode

-- some code cleanup: removed reference to downrev compiler code genration

-- create table if not exists as select...   syntax support

-- regress/seabase/TEST003 updated with additional tests


zellerh:
Wow, this is a huge change. I didn't look at all the details, just at the big picture, and that looks good to me. Let me just ask a few questions, sorry, I should probably have answered those by myself by looking at the code:

We have these type classes, numerics, strings, datetime, interval. Where does boolean fit in here, is it treated as an exact numeric in internal methods? If so, some code has subclasses of numbers with binary and decimal precision, would this be another case that allows only two values? I'm partially thinking of UDFs, and how we should treat boolean values there.

As far as i can tell, this follows the ANSI definition of a boolean data type, is this true or are there differences?

Is this just experimental for now or should we add this type to the SQL Reference Manual?

Thanks, there are only a few people who can pull off a big change like this :-)

core/sql/common/MiscType.cpp
```diff
@@ -40,41 +40,107 @@
 
 // ***********************************************************************
 //
-//  SQLBoolean : The boolean data type
+//  SQLBooleanBase : The boolean data type
```
DaveBirdsall: A SQL version of the famous French stew. :-)

core/sql/common/MiscType.cpp
```diff
@@ -40,41 +40,107 @@
...
+            heap                
+            )
+{ 
 }
 
-short SQLBoolean::getFSDatatype() const 
+// ---------------------------------------------------------------------
+// A method which tells if a conversion error can occur when converting
+// a value of this type to the target type.
+// This method is a stub and so just returns true for now.
```
DaveBirdsall: It's not a stub anymore, it seems.

core/sql/common/MiscType.cpp
```diff
@@ -106,21 +172,51 @@ const NAType* SQLBoolean::synthesizeType(enum NATypeSynthRuleEnum synthRule,
...
+
   return NULL;
 } // synthesizeType()
 
+void SQLBooleanNative::minRepresentableValue(void* bufPtr, Lng32* bufLen,
+                                     NAString ** stringLiteral,
+				     CollHeap* h) const
+{
+  assert(*bufLen >= sizeof(Int8));
+  Lng32 valueBuf;
```
DaveBirdsall: It doesn't hurt anything, but it looks like valueBuf is not used other than setting it to zero three lines down.

core/sql/common/MiscType.cpp
```diff
@@ -106,21 +172,51 @@ const NAType* SQLBoolean::synthesizeType(enum NATypeSynthRuleEnum synthRule,
...
-// a value of this type to the target type.
-// This method is a stub and so just returns true for now.
-// ---------------------------------------------------------------------
-NABoolean SQLBoolean::errorsCanOccur(const NAType& target, NABoolean lax) const
+  if (stringLiteral != NULL)
+    {
+      // Generate a printable string for the minimum value
+      char nameBuf[10];
+      strcpy(nameBuf, "false");
+      *stringLiteral = new (h) NAString(nameBuf, h);
```
DaveBirdsall: Just curious: Can we say, *stringLiteral = new (h) NAString("false",h)?

core/sql/common/MiscType.cpp
```diff
@@ -106,21 +172,51 @@ const NAType* SQLBoolean::synthesizeType(enum NATypeSynthRuleEnum synthRule,
...
+    }
+
+} // SQLBooleanNative::minRepresentableValue()
+
+void SQLBooleanNative::maxRepresentableValue(void* bufPtr, Lng32* bufLen,
+                                     NAString ** stringLiteral,
+				     CollHeap* h) const
```
DaveBirdsall: Similar comments here as for minRepresentableValue

core/sql/common/MiscType.h
```diff
@@ -106,25 +118,63 @@ class SQLBoolean : public NAType
 				       CollHeap* h,
 				       UInt32 *flags = NULL) const;
   
+  // ---------------- Methods not inherited from NAType ----------------
+
+  NABoolean canBeSQLUnknown() const { return sqlUnknownFlag_;}
+
+  private:
+  NABoolean sqlUnknownFlag_;
```
DaveBirdsall: So, the reason SQLBooleanRelat is a different class than SQLBooleanNative is that SQLBooleanRelat has to model the three-valued logic of SQL predicates (that is, true/false/unknown), while SQLBooleanNative has to model a possibly nullable two-value logic (true/false/null).

core/sql/common/dfs2rec.h
```diff
@@ -186,9 +177,9 @@
   #define REC_BLOB                     160      // SQ only: blob datatype
   #define REC_CLOB                     161      // SQ only: clob datatype
 
+  #define REC_BOOLEAN                  170
```
DaveBirdsall: Is Boolean considered a numeric type? I'm assuming not, since 170 is not in the normal range for numeric types.

core/sql/common/str.cpp
```diff
@@ -997,19 +997,40 @@ Lng32 str_decode(void *tgt, Lng32 tgtMaxLen, const char *src, Lng32 srcLen)
...
+                        )
 {
+  if (! src)
+    return NULL;
+
   len = str_len(src)-1;
-  while ((len >= 0) && (src[len] == ' '))
-    len--;
+  if (len <= 0)
+    return src;
```
DaveBirdsall: This doesn't look right to me. Suppose the input string is 'a'. Then wouldn't str_len('a') be 1? And wouldn't len get zero? Seems like the returned len value should be one.

core/sql/exp/exp_attrs.cpp
```diff
@@ -278,6 +278,8 @@ switch (datatype)
   case REC_INT_FRACTION: return extFormat? (char *)"INTERVAL FRACTION":(char *)"REC_INT_FRACTION";
   case REC_BLOB: return extFormat? (char *)"BLOB":(char *)"REC_BLOB";
   case REC_CLOB: return extFormat? (char *)"CLOB":(char *)"REC_CLOB";
+  case REC_BOOLEAN: return extFormat ? (char *)"BOOLEAN" : (char *)"BOOLEAN";
```
DaveBirdsall: Not sure why the ternary expressions are used here, since the true and false results are the same.

core/sql/optimizer/ItemArith.h
```diff
@@ -250,4 +250,36 @@ class BiArithCount : public BiArith
 private:
 };
 
+class UnArith : public ItemExpr
+{
+  // ITM_NEGATE. Used with negation of boolean native datatype
```
DaveBirdsall: Do we support unary minus?

core/sql/optimizer/SynthType.cpp
```diff
@@ -1945,18 +1945,50 @@ const NAType *BiArith::synthesizeType()
...
+      vid1.coerceType(NA_BOOLEAN_TYPE);
+    }
+  
+  const NAType& operand1 = vid1.getType();
+  if (operand1.getFSDatatype() != REC_BOOLEAN)
+    {
+      *CmpCommon::diags() << DgSqlCode(-4034)
+                          << DgString0("!")
+                          << DgString1("")
+                          << DgString2(child(0)->getTextUpper());
```
DaveBirdsall: Why "getTextUpper"? So, if I have an expression !'abc' will it print 'ABC'?

core/sql/optimizer/UdfDllInteraction.cpp
```diff
@@ -1214,6 +1214,29 @@ NABoolean TMUDFInternalSetup::setTypeInfoFromNAType(
           switch (length)
             {
               // SMALLINT, INT, LARGEINT, NUMERIC, signed and unsigned
+            case 1:
+              if (CmpCommon::getDefault(TRAF_TINYINT_SPJ_SUPPORT) == DF_ON)
```
DaveBirdsall: Did you mean DF_OFF? Seems odd that if TRAF_TINYINT_SPJ_SUPPORT is 'ON' then we *wouldn't* support tinyint.

DaveBirdsall:
Looks good. All my comments are minor, except the one in str.cpp which I suspect is a genuine bug. 

A general question: It looks like the only operations allowed on BOOLEAN columns are =, !=, and !. Are there plans to add support for AND and OR? Could I AND a BOOLEAN column with the result of a predicate?

core/sql/common/str.cpp
anoopsharma00: Returned value from this method is a pointer to the string. Length is not returned. 
The check at line 1013 returns the input src if its length is 1.

core/sql/exp/exp_attrs.cpp
anoopsharma00: The second string should be "REC_BOOLEAN" instead of "BOOLEAN". Will fix
that in a subsequent checkin.

core/sql/optimizer/ItemArith.h
anoopsharma00: unary minus is converted to "0 - val" during parser/binder step.

core/sql/optimizer/UdfDllInteraction.cpp
anoopsharma00: Some support still need to be added to SPJ/T2 driver path for tinyint.
If this cqd is turned on, it continues to the path of processing tinyint.
Since tinyint support is not yet in for spj/t2, an error is currently returned. 
This is not an externalized cqd.

core/sql/common/str.cpp
DaveBirdsall: Hmmm... Isn't len a reference parameter? Therefore it is an output as well as an input. Perhaps it should be pass by value instead?

Let's take another example. Suppose the input string is ' ' (that is, one blank). The code as written would calculate len = 0 at line 1012, then the 'if' at line 1013 is true, and we return src without stripping a blank. Seems like we should return the empty string instead?

anoopsharma00:
Response to Hans/Dave's comment regarding boolean datatype.

Boolean datatype is not a numeric datatype.
It cannot be used to compare or in arith opers with other numbers.
It is an additional datatype class to NUMERIC, CHARACTER, DATETIME, INTERVAL, BLOB.
Internally it is stored as a one byte value with 0 for false and 1 for true.

There is still some support for boolean needed in drivers and SPJ/UDF. 
Once that is done, we can document it in manuals. 

We could also enhance it later so relational and boolean datatype operations can be mixed.
Ex:  update t set a = (b > 10)  
   or use boolean datatype in a predicate of the form:  where a AND (b >  10);
(one can still use it as:  where (a = true) AND (b > 10) );

DaveBirdsall:
Looks good!

PR #593, opened 2016-07-14
[TRAFODION-2109] Load with log error rows returns SQL error 2034 at t…

selvaganesang:
…imes

Changed the error at the time of logging into error rows into a warning
so that load can continue.

Also, improved the error reporting to display stack trace when java method call
fails at the time of loading and unloading.

### DIFF
--- a/core/sql/executor/ExHbaseAccess.cpp
+++ b/core/sql/executor/ExHbaseAccess.cpp
@@ -239,6 +239,7 @@ ExHbaseAccessTcb::ExHbaseAccessTcb(
   , colValVec_(NULL)
   , colValVecSize_(0)
   , colValEntry_(0)
+  , loggingErrorDiags_(NULL)
 {
   Space * space = (glob ? glob->getSpace() : NULL);
   CollHeap * heap = (glob ? glob->getDefaultHeap() : NULL);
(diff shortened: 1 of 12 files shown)


core/sql/executor/ExHbaseAccess.cpp
```diff
@@ -2996,6 +3003,14 @@ void ExHbaseAccessTcb::handleException(NAHeap *heap,
      errorMsgLen = strlen("[UNKNOWN EXCEPTION]\n");
```
DaveBirdsall: Looks like errorMsg is uninitialized in this code path. Perhaps we need to add, "errorMsg = "[UNKNOWN EXCEPTION]" here? (And maybe change this line to errorMsgLen = strlen(errorMsg)?)

core/sql/executor/ExHbaseAccess.cpp
```diff
@@ -2967,22 +2968,28 @@ void ExHbaseAccessTcb::handleException(NAHeap *heap,
...
-                                    char *loggingFileName)
+                                    char *loggingFileName,
+                                    ComDiagsArea **loggingErrorDiags)
 {
   Lng32 errorMsgLen = 0;
   charBuf *cBuf = NULL;
   char *errorMsg;
   Lng32 retcode;
 
+  if (*loggingErrorDiags != NULL)
```
DaveBirdsall: Not quite sure I understand this logic. If a ComDiagsArea object has already been allocated, does that mean we have already reported the exception somewhere else? Or perhaps this is not the first exception and we simply don't want to log more than one?

DaveBirdsall:
Looks good. I think I found a bug in an existing code path (not in your changes).

core/sql/executor/ExHbaseAccess.cpp
selvaganesang: Basically, we want to skip the error rows in the source and continue to load. Load command has constructs to log the error row and the error information about it.  We just want to report the error while logging the error rows as warning at the end of the load command. This check is skipping the logging once an error is encountered  in the earlier attempt to log the error row.

core/sql/executor/ExHbaseAccess.cpp
selvaganesang: will do

DaveBirdsall:
Looks good to me. Thanks.

traflm:
Very good to not see Error 201 and instead the real error stack trace! 
One minor question is: if for some reason, HDFS cannot write, and it may affect all error rows, will each write attempt generate a new SQL warning and is it possible to make the diagArea too big? Or you keep just the last SQL error?

selvaganesang:
Once we get the error while attempting to log error row, subsequent error rows are not attempted to log at all. This fix is not taking care of transient errors with HDFS writes. These kind of errors less likely to happen.

traflm:
thanks Selva, It looks good for me.

selvaganesang:
Are there any more review comments?

DaveBirdsall:
None from me. I think you can merge it.

PR #594, opened 2016-07-15
[TRAFODION-2079]Catalog API support get indexes

xwq:
After this comment, both JDBC type4 and ODBC Catalog API will support
get the information about index.
[TRAFODION-2079]([3rd party tool-Squirrel] indexes won't show up on
objects panel, sql command line works fine) which related to Catalog
API of JDBC type, it is fixed also.

### DIFF
--- a/core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
+++ b/core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
@@ -4934,7 +4934,7 @@ odbc_SQLSvc_GetSQLCatalogs_sme_(
             convertWildcard(metadataId, TRUE, schemaNm, expSchemaNm);
             convertWildcardNoEsc(metadataId, TRUE, schemaNm, schemaNmNoEsc);
             convertWildcard(metadataId, TRUE, tableNm, expTableNm);
-            convertWildcardNoEsc(metadataId, TRUE, tableNmNoEsc, tableNmNoEsc);
+            convertWildcardNoEsc(metadataId, TRUE, tableNm, tableNmNoEsc);
             inputParam[0] = schemaNmNoEsc;
             inputParam[1] = expSchemaNm;
             inputParam[2] = tableNmNoEsc;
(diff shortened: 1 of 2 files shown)


core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
```diff
@@ -4967,10 +4967,40 @@ odbc_SQLSvc_GetSQLCatalogs_sme_(
...
+                    "cast('%s' as varchar(128)) TABLE_CAT, "
+                    "cast(trim(ob.SCHEMA_NAME) as varchar(128)) TABLE_SCHEM, "
+                    "cast(trim(ob.OBJECT_NAME) as varchar(128)) TABLE_NAME, "
+                    "cast(idx.is_unique as smallint) NON_UNIQUE, "
+                    "cast('' as varchar(128)) INDEX_QUALIFIER, "
+                    "cast(trim(ob_table.OBJECT_NAME) as varchar(128)) INDEX_NAME, "
+                    "cast(3 as smallint) TYPE, "
+                    "cast(0 as smallint) ORDINAL_POSITION, "
+                    "cast('' as varchar(128)) COLUMN_NAME, "
+                    "cast('' as char(1)) ASC_OR_DES, "
```
selvaganesang: Spec says NULL should be returned when the value is not available. Can you please change it to NULL. Attaching a formatted output of this query when the table has index would be helpful to validate the results of the query by the reviewers. Consider creating a use case with two tables one with update statistics performed and another without.

xwq: I will change it to null. I will add the test to ODBC test. 
I think I still need add some new JDBC test. I will take some days.

ryzuo:
This looks good to me.

dcs/src/test/jdbc_test/src/test/java/org/trafodion/jdbc_test/TestGetIndexInfo.java
```diff
@@ -0,0 +1,180 @@
...
+			assertEquals(methondName + " rowNum " + Integer.toString(rowNum) + " dbIndexName ", indexInfo.dbIndexName, rs.getString("INDEX_NAME"));
+			assertEquals(methondName + " rowNum " + Integer.toString(rowNum) + " dbType ", indexInfo.dbType, rs.getShort("TYPE"));
+			assertEquals(methondName + " rowNum " + Integer.toString(rowNum) + " dbOridinalPosition ", indexInfo.dbOrdinalPosition, rs.getShort("ORDINAL_POSITION"));
+			assertEquals(methondName + " rowNum " + Integer.toString(rowNum) + " dbColumnName ", indexInfo.dbColumnName, rs.getString("COLUMN_NAME"));
+			assertEquals(methondName + " rowNum " + Integer.toString(rowNum) + " dbAscOrDesc ", indexInfo.dbAscOrDesc, rs.getString("ASC_OR_DESC"));
+			assertEquals(methondName + " rowNum " + Integer.toString(rowNum) + " dbCardinality ", indexInfo.dbCardinality, rs.getInt("CARDINALITY"));
+			assertEquals(methondName + " rowNum " + Integer.toString(rowNum) + " dbPages ", indexInfo.dbPages, rs.getInt("PAGES"));
+			assertEquals(methondName + " rowNum " + Integer.toString(rowNum) + " dbFilterCondition ", indexInfo.dbFilterCondition, rs.getString("FILTER_CONDITION"));
```
selvaganesang: It looks like value 0 is being compared with SQL value of null in case of numeric columns. Should we use wasNull() method to check if the column returned was really null.

xwq: From the JDBC document, getInt() return 0 when the column is NULL, So I don't check wasNull().  
However, it good to check more in the test case, I will add some the code that check the wasNull().

dcs/src/test/jdbc_test/src/test/java/org/trafodion/jdbc_test/TestGetIndexInfo.java
```diff
@@ -0,0 +1,196 @@
...
+			assertEquals(methondName + " rowNum " + Integer.toString(rowNum) + " dbCatalog ", indexInfo.dbCatalog, rs.getString("TABLE_CAT"));
+			String currentSchema = _conn.getSchema();
+			
+			// Since we may run the test in different Schema, so compare the result to the current schema of the connection here.
+			assertEquals(methondName + " rowNUm " + Integer.toString(rowNum) + " dbSchem ", currentSchema, rs.getString("TABLE_SCHEM"));
+			
+			assertEquals(methondName + " rowNum " + Integer.toString(rowNum) + " dbTableName ", indexInfo.dbTableName, rs.getString("TABLE_NAME"));
+			assertEquals(methondName + " rowNum " + Integer.toString(rowNum) + " dbNoneUnique ", indexInfo.dbNoneUnique, rs.getBoolean("NON_UNIQUE"));
+			// By the Document, dbNoneUnique will return null if the type is SQL_TABLE_STAT
+			System.out.println(rs.wasNull());
```
selvaganesang: Is it needed to display the wasNull.  Where will the output go in Jenkins environment

selvaganesang:
+1

sureshsubbiah:
+1

sureshsubbiah:
Merging with the assumption that any testware change due to the last comment by @selvaganesang can be addressed later.

PR #595, opened 2016-07-15
[TRAFODION-2110] core/TESTRTS is being incorrectly flagged as PASSED

selvaganesang:
Some of the filter conditions were reporting error when it was
applied on both the expected and log output.


DaveBirdsall:
Looks good.

PR #597, opened 2016-07-16
[TRAFODION-1605] sqstart hangs if ulimits not correct

traflm:
add ulimit checking in sqstart
if not proper setup as Trafodion installer does, print WARNING

This should only happen in dev environment as described in the JIRA. Installer already handle this, so a WARNING should be enough.

### DIFF
--- a/core/sqf/sql/scripts/sqstart
+++ b/core/sqf/sql/scripts/sqstart
@@ -245,6 +245,73 @@ function checkKerberos {
    return 0
 }
 
+function printExpectedUlimt {
+
+    echoLog "desired \"ulimit -a\" output: "
+    echoLog "========================================="
+    echoLog "core file size (blocks, -c) 1000000"
+    echoLog "data seg size (kbytes, -d) unlimited"
+    echoLog "scheduling priority (-e) 0"
+    echoLog "file size (blocks, -f) unlimited"
+    echoLog "pending signals (-i) 515196"
+    echoLog "max locked memory (kbytes, -l) 49595556"
+    echoLog "max memory size (kbytes, -m) unlimited"
+    echoLog "open files (-n) 32000"
+    echoLog "pipe size (512 bytes, -p) 8"
+    echoLog "POSIX message queues (bytes, -q) 819200"
+    echoLog "real-time priority (-r) 0"
+    echoLog "stack size (kbytes, -s) 10240"
+    echoLog "cpu time (seconds, -t) unlimited"
+    echoLog "max user processes (-u) 267263"
+    echoLog "virtual memory (kbytes, -v) unlimited"
+    echoLog "file locks (-x) unlimited"
+    echoLog "========================================="
+ 
+}
+
+function checkUlimit {
+  
+  #check ulimit -n 
+  nrOfFile=`ulimit -n`
+  if [[ $nrOfFile -lt 32000 ]] ; then
+    echoLog "WARNING: detect issue during startup, please make sure your ulimit setup correctly"
+    echoLog "WARNING: open files is smaller than 32000"
+    echoLog ""
+    printExpectedUlimt
+    return
+  fi 
+
+ #check ulimit -u
+ userProc=`ulimit -u`
+ if [[ $userProc -lt 267263 ]] ; then
+    echoLog "WARNING: detect issue during startup, please make sure your ulimit setup correctly"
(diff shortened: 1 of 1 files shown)


selvaganesang:
It is not clear if the checked ulimits need to have these values. For eq, Trafodion can support upto 65535 pids only. /proc/sys/kernel/pid_max should be set to a value less than 65535.  So, max user processes ulimit value has no bearing. Do we know which ulimit that is not properly set got into hang?

traflm:
These four ulimit setting is from Trafodion installer, which set all of them, So this checking is to sync with official Trafodion installer for developers who not run trafodion_install, but install_local_hadoop and build Trafodion from source code. It is common that new developers ignore the ulimit settings and will meet various problems. 
These checkings will print warnings to the developer, so as a reminder for them to correct the ulimit, instead of sending email to ask around.

I checked all four ulimit one by one and the real one that block the sqstart is the 'ulimit -l'. But I assume there must be some reason Trafodion set four ulimit, if 'ulimit -u' is not required, we should also remove it from trafodion_install, and remove it from here.

However, this checking will not block the sqstart, just print some warning, in case the user do meet problems.

I am fine to remove the checking of 'ulimit -u'. But this does not do any harm since it is a WARNING. There are many settings : ulimit settings, hbase settings are not clear anymore, the original author of those changes were not active now. So it is hard to decide if it is required or not.

DaveBirdsall:
Just checking: Is this ready to merge? Any further concerns, @selvaganesang?

selvaganesang:
Thanks @traflm for checking it out.  Currently, Trafodion needs 32 MB of locked memory per node. When you issue ipcs, you would see two segments with 32 MB  locked because there are  2 virtual nodes in developer environment . Trafodion product might lock RMS shared segment in future. RMS shared segment default is 64MB with a provision to expand upto 128MB.  So Trafodion requires  320MB of locked shared memory at most.

I believe ulimit -l is the user level limit. So, it would be good if we report an error when it is less than 64 MB and report an warning if it is less than 320MB.  Reporting warning for 40GB of locked memory per user is way too high.

In case of max user process, it would be good if we report an error if /proc/sys/kernel/pid_max is set to higher value than 65535 and possibly ignore ulimit -u settings

Can  32000 for ulimit -n be considered reasonable? This is per process limit.  I guess there will be resource/performance implications with altering these numbers.  I don't have definite answer.

I am not sure about ulimit -i pending signals.

However in summary, my suggestion is that sqstart should throw a warning or an error only when it is clear what Trafodion needs.  I would think this might need to be fixed in installer too. Otherwise, Trafodion might end up giving false indications to the customer to configure such settings.

traflm:
@selvaganesang 49595556  is 47M , not 40G. I will change it to 64M as you suggested. and remove unclear ulimit checkings, avoid give false indication until we can make sure.

core/sqf/sql/scripts/sqstart
```diff
@@ -245,6 +245,63 @@ function checkKerberos {
...
+    echoLog "========================================="
+ 
+}
+
+function checkUlimit {
+  
+ #check ulimit -u
+ userProc=`cat /proc/sys/kernel/pid_max`
+ if [[ $userProc -gt 65535 ]] ; then
+    echoLog "WARNING: detect issue during startup, please make sure your ulimit setup correctly"
```
selvaganesang: Thanks @traflm for the change. This is an error condition.  Because you have a way to flag it as ERROR, can you please change this to ERROR.

core/sqf/sql/scripts/sqstart
```diff
@@ -245,6 +245,63 @@ function checkKerberos {
...
+    echoLog "WARNING: max user processes is smaller than 267263"
```
selvaganesang: Please change the message accordingly. pid_max shouldn't be greater than 65535

core/sqf/sql/scripts/sqstart
traflm: Yes, I agree.

core/sqf/sql/scripts/sqstart
traflm: I will change it

traflm:
I think I'd better abort this PR and start a new PR with all these changes. Thanks for @selvaganesang carefully review. All these are true, I think it will be good to resubmit with only one commit including all these changes.

traflm:
submit  a new PR for this

PR #599, opened 2016-07-17
Cherry-pick changes for JIRA 2095

sbroeder:


shangshengtung:
The fix looks good to me.

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionManager.java
```diff
@@ -1452,6 +1456,72 @@ public Integer doAbortX(final List<TransactionRegionLocation> locations, final l
...
+         catch (ServiceException se) {
+            if (LOG.isTraceEnabled()) LOG.trace("pushRegionEpochX -- ServiceException ", se);
+            throw new IOException(se);
+         }
+         catch (Throwable t) {
+            if (LOG.isTraceEnabled()) LOG.trace("pushRegionEpochX -- Throwable ", t);
+            throw new IOException(t);
+         }
+
+       } while (loopExit == false);
```
selvaganesang: It is not clear why do...while loop is needed.

core/sqf/src/seatrans/tm/hbasetmlib2/src/main/java/org/trafodion/dtm/HBaseTxClient.java
```diff
@@ -694,7 +694,22 @@ public short callCreateTable(long transactionId, byte[] pv_htbldesc, Object[]  b
...
-      trxManager.createTable(ts, htdesc, beginEndKeys);
+      try {
+         trxManager.createTable(ts, htdesc, beginEndKeys);
+      }
+      catch (IOException cte) {
+         if (LOG.isTraceEnabled()) LOG.trace("HBaseTxClient:callCreateTable exception trxManager.createTable, retval: " +
+            TransReturnCode.RET_EXCEPTION.toString() +" txid: " + transactionId +" Exception: " + cte);
+         StringWriter sw = new StringWriter();
+         PrintWriter pw = new PrintWriter(sw);
+         cte.printStackTrace(pw);
```
selvaganesang: 703-705 can be avoided. All the printStackTrace calls were removed as part of java exception handling cleanup. Instead you can LOG.error(<message>, e) methods

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionManager.java
sbroeder: Yes good point

PR #601, opened 2016-07-18
[TRAFODION-2112] Avoid hard-coding DNS name in local Hadoop install

zellerh:
Also removed three dsdgen calls, the *_returns tables are created
together with the *_sales tables, calling dsdgen for those tables will
give an error.


DaveBirdsall:
Looks good to me.

zellerh:
Am going to ignore the JDBC test failure, the test doesn't exercise the changed code and therefore must be caused by something else.

PR #604, opened 2016-07-18

selvaganesang:
… of Trafodion - Part3

The focus is to ensure that the execptions seen at the JNI of TM libs are logged
via log4cxx.

1)	Removed the duplicate logging codes by deleting the following files
-	 hbasetmlogging.h
-	tmlogginglib.cpp
-	xarmlogging.cpp

2)	Removed the duplicate JNI initialization code. Similar files in hbasetmlib2 directory has the exception stack tracing code
-	tm/javaobjectinterfacetm.h
-	tm/javaobjectinterfacetm.cpp

3)	The logs using the method tm_log_write will be seen in $MY_SQROOT/logs/tm_<nid>.log only.  The output won’t be duplicated in $MY_SQROOT/sql/scripts/stdout_dtm_<nid>.log. 

4)	I simulated the error condition in TM java by deleting a line at the time of initialization.
The log seen at tm_1.log is
2016-07-22 21:17:01,607, FATAL, TM, Node: 1 Pid: 2192 Name: $TM1 TransId: -1 Event: 103005220 Message: Error at CHbaseTM::initConnection() caused by exception 
org.trafodion.dtm.TmAuditTlog.<init>(TmAuditTlog.java:564)
org.trafodion.dtm.HBaseTxClient.init(HBaseTxClient.java:273)


narendragoyal:
Looks good to me. Thanks for taking care of the exceptions so that they are now being logged.

sbroeder:
The code seems fine, however, it seems like it would be good to at least log the transid with the exceptions so we can better track down/correlate issues after the fact.

selvaganesang:
I will add transid to the tm_log_write function.  Transid will be passed if it exists or -1  will be used.

sbroeder:
Thanks Selva.  That will be very helpful


shangshengtung:
Good to me.

PR #606, opened 2016-07-19
[TRAFODION-2115]

mkby:
Better UX for README on GitHub main page.

### DIFF
--- a/README
+++ b/README
@@ -1,0 +1,56 @@
+******************************* Apache Trafodion (incubating) **************************
+
+Apache Trafodion is a webscale SQL-on-Hadoop solution enabling transactional or 
+operational workloads on Hadoop.  The name "Trafodion" (the Welsh word for 
+transactions, pronounced "Tra-vod-eee-on") was chosen specifically to emphasize 
+the differentiation that Trafodion provides in closing a critical gap in the 
+Hadoop ecosystem. Trafodion builds on the scalability, elasticity, and 
+flexibility of Hadoop. Trafodion extends Hadoop to provide guaranteed 
+transactional integrity, enabling new kinds of big data applications to run on 
+Hadoop.
+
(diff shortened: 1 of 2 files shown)


mkby:
jenkins, retest

traflm:
you'd better rename the title to TRAFODION-2115, so the JIRA will be auto link to this PR.
Now the title without the '-' , so not auto linked.

No idea why the test fail, that should be check too.

DaveBirdsall:
@svarnau, any idea why this test is failing? And how to figure it out?

svarnau:
https://jenkins.esgyn.com/job/static-checks/1228/consoleFull 
 Unapproved licenses:
   trafodion/README.md

Likely the readme file was in .rat-excludes file, but renaming README caused the exclusion to not match, hence the RAT check failed.

I'm not crazy about idea of renaming readme file just so it formats better on github.  The official use of the file is in the official code releases. The release policies talk about README files, though they don't specify a format or suffix as far as I can see.  I'm just not sure if reviewers of next release are going to complain. It probably won't hurt, but not my preference. Other opinions are welcome.

mkby:
We can add a new file README.md, but keep the original README file. Is it OK for you?

svarnau:
Hmm, that makes some sense, but then we'll have redundant info and maybe some confusion. I think it is better to move it to the new .md name than have two README files.

mkby:
Makes sense. Let's wait for other opinions, if this change is not acceptable, I can close the PR.

traflm:
I think two file should be the best approach, yes, we need to do sync, but I feel there is few times we need to update them and the cost vs benefit looks good for me. We should make the display of github front page better.

mkby:
jenkins, retest

mkby:
I will create another PR for adding the README.md file.

PR #609, opened 2016-07-20
[TRAFODION-2100]SQLGetTypeInfo Catalog API should return identifiers conforming to ODBC 3.0

ryzuo:
1.Changed the returned type code of DATE, TIME, TIMESTAMP all to ODBC 3.0 compliance for SQLGetTypeInfo.
2.Changed a few places of using SQL_OV_ODBC3 for ODBC standard version compatibility to use pre-compilation definition ODBCVER for conditioning, which it should be, both Linux and Windows.

### DIFF
--- a/core/conn/unixodbc/odbc/odbcclient/unixcli/DrvrManager/drvrmanager.cpp
+++ b/core/conn/unixodbc/odbc/odbcclient/unixcli/DrvrManager/drvrmanager.cpp
@@ -3132,42 +3132,42 @@ SQLRETURN  SQL_API SQLGetTypeInfo_common(SQLHSTMT StatementHandle,
 		RETURNCODE (pstmt, SQL_ERROR);
 		LEAVE_STMT (pstmt, SQL_ERROR);
 	}
-	SQLINTEGER ODBCAppVersion = pstmt->getODBCAppVersion();
-	SQLSMALLINT convDataType = DataType;
-	if (ODBCAppVersion >= SQL_OV_ODBC3)
-	{
-		switch(DataType)
-		{
-		case SQL_DATE:
-			convDataType = SQL_TYPE_DATE;
-			break;
-		case SQL_TIME:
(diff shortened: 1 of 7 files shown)


DaveBirdsall:
Looks good to me. I'm not an expert in this area however.

core/conn/unixodbc/odbc/odbcclient/unixcli/DrvrManager/drvrmanager.cpp
```diff
@@ -3132,42 +3132,38 @@ SQLRETURN  SQL_API SQLGetTypeInfo_common(SQLHSTMT StatementHandle,
 		RETURNCODE (pstmt, SQL_ERROR);
 		LEAVE_STMT (pstmt, SQL_ERROR);
 	}
-	SQLINTEGER ODBCAppVersion = pstmt->getODBCAppVersion();
 	SQLSMALLINT convDataType = DataType;
```
selvaganesang: Isn't DriverManager supposed to do the mapping based on ODBC app version at runtime?  I am thinking that the old code is doing the right thing. A unit test using ODBC 2.0 and ODBC 3.0 application would help to determine which way to go.

selvaganesang:
I am unable to see the difference in cstmt.cpp and cdesc.cpp. How do I view it?

ryzuo:
The change is just the same as cstmt.cpp and cdesc.cpp for unix driver. Looks like the CR/LF all changed when pushing from Windows to github. I'll double check and re-push it.

selvaganesang:
Thanks @ryzuo for changing as per review comments. Change looks good

DaveBirdsall:
@ryzuo, would you like to practice committing this change? The instructions on how to do it are here: https://cwiki.apache.org/confluence/display/TRAFODION/Merge+Changes#MergeChanges-MergePullRequests

PR #612, opened 2016-07-21
Fix for issue where the writeOrdering list was not processed in a con…

sbroeder:
…sistent order and could result in rows being truncated


DaveBirdsall:
I see other places in this file where we call writeOrdering.listIterator (for example, line 546). Why don't we add the writeOrdering.size() parameter there?

sbroeder:
We only use the writeOrder.size() when we are adding a new put/delete to the writeOrder list to ensure that the items are added at the end of the list.  Otherwise, when iterating through the list we see the newest items first rather than the oldest items first.


sbroeder:
jenkins, retest

shangshengtung:
The changes are good to me.

DaveBirdsall:
Looks good.

PR #614, opened 2016-07-21
Fix for  leftover debug code causing several "connection refused" log messages related to [TRAFODION-2091]

sandhyasun:
Removing some leftover debug code causing several "connection refused" messages in mon*.log files during regression runs.Related to [TRAFODION-2091]


DaveBirdsall:
Committing this right away so regression tests will be cleaner.

PR #616, opened 2016-07-22
[TRAFODION-2125] ID request times out to idtmsrv process

sbroeder:


DaveBirdsall:
Looks good to me. One question: Why is the do_link function in a header file?

svarnau:
jenkins, retest -- automation bug

sbroeder:
jenkins, retest

svarnau:
jenkins, retest -- second automation bug

selvaganesang:
Where is the output of printf seen in idtmserv process? If it is a server process, is it possible to use log4cxx which is being enabled in TM via #604

sbroeder:
The output goes to stdout_idtmsrv_0.   It's certainly possible to route the output to a different location, and it's probably a good idea, but I think that should be part of a separate commit since that's not the defect being addressed here.

PR #618, opened 2016-07-24
[TRAFODION-2121] add support of HIVE data type of DECIMAL

traflm:
support HIVE data type 'decimal'

HIVE decimal  =>  Trafodion decimal(10, 0)
HIVE decimal(p,s)  => Trafodion decimal(p,s)

hive data type definitions is from https://cwiki.apache.org/confluence/display/Hive/LanguageManual+Types#LanguageManualTypes-decimal 

Please help to review.


traflm:
Jenkins, extra tests

traflm:
Extra Test failed at Hive/Test003, But I can pass it in my own workspace, test it again.

traflm:
Jenkins, extra tests

core/sql/optimizer/NATable.cpp
```diff
@@ -3676,6 +3676,54 @@ NAType* getSQColTypeForHive(const char* hiveType, NAMemory* heap)
                                    CharInfo::IMPLICIT);
   } 
 
+  if ( !strncmp(hiveType, "decimal", 7) )
+  {
+    Int16 i=0, pstart=0, pend=0, sstart=0, send=0, p=0, s = 0;
+    Int16 hiveTypeLen = strlen(hiveType);
+    char pstr[16], sstr[16];
```
sureshsubbiah: It would be nice to use a constant declaration in lieu of literals like 16.

core/sql/optimizer/NATable.cpp
```diff
@@ -3676,6 +3676,54 @@ NAType* getSQColTypeForHive(const char* hiveType, NAMemory* heap)
...
+      else if(hiveType[i] == ')')
+      {
+        send = i;
+      }
+      else
+       continue;
+    }
+    if(pend - pstart > 0)
+    {
+      strncpy(pstr,hiveType+pstart, pend-pstart);
```
sureshsubbiah: What if (pend - pstart > 16) is true.

core/sql/optimizer/NATable.cpp
```diff
@@ -3676,6 +3676,54 @@ NAType* getSQColTypeForHive(const char* hiveType, NAMemory* heap)
...
+      s=atoi(sstr);
+    }
+
+    if( (p>0) && (s>0) )
+    {
+      return new (heap) SQLDecimal( p, s, TRUE, TRUE);
+    }
+    else
+    {
+      return new (heap) SQLDecimal( 10, 0, TRUE, TRUE);
```
sureshsubbiah: Why should a negative value for p or s lead to a DECIMAL(10,0)? Does Hive use such a default?

core/sql/optimizer/NATable.cpp
```diff
@@ -3676,6 +3676,54 @@ NAType* getSQColTypeForHive(const char* hiveType, NAMemory* heap)
...
+      strncpy(pstr,hiveType+pstart, pend-pstart);
+      p=atoi(pstr);
+    }
+    if(send - sstart > 0)
+    {
+      strncpy(sstr,hiveType+sstart,send-sstart);
+      s=atoi(sstr);
+    }
+
+    if( (p>0) && (s>0) )
```
sureshsubbiah: Are there upper limits for p and s ? What if s > p?

sureshsubbiah:
A much needed change. Thank you.

core/sql/optimizer/NATable.cpp
traflm: Yes, agree, will fix.

core/sql/optimizer/NATable.cpp
traflm: that is a problem, we need to protect this.

core/sql/optimizer/NATable.cpp
traflm: negative value should be an error. 
Hive support 'decimal' , without (p,s), and that is identical to decimal(10,0) as mentioned in Hive wiki.
I should check for p <=0 , and raise error.

core/sql/optimizer/NATable.cpp
traflm: yes, I totally ignore the error checking, need to enhance it here.

traflm:
Yes @sureshsubbiah , with decimal supported, only CHAR is not supported for hive simple SQL type.
Then we will be free of CAST when reading from Hive. 
CHAR should be simple to add, I plan to do it next, or can be a good practice or beginners, 
And Trafodion now has native tinyint support, so that mapping can change also. We currently map Hive tinyint to Trafodion short.
All these type supporting are to simplify the bulkload as the most urgent purpose from my point of view. It is very time consuming to add multiple CAST to load data from Hive now.

core/sql/optimizer/NATable.cpp
zellerh: General question, this is machine-generated code, so we probably don't have to check for errors the same way as if this would be human-generated input? If we want to do a complete check, then maybe also check for syntax like decimal-extra-chars(6,0). The above check pointed out by Suresh is still needed, since we should always check for buffer overruns.

core/sql/optimizer/NATable.cpp
zellerh: What if (p>0 && s == 0), shouldn't we use the "if" path for that as well?

core/sql/optimizer/NATable.cpp
zellerh: If p > 18, should we give a better error here or make the type a BIGNUM? Same question as above, I wonder whether Hive would ever generate a negative value for the precision - maybe an assert would be ok, no fancy error message required.

core/sql/optimizer/NATable.cpp
traflm: yes @zellerh , will add this branch. I didn't realize there is decimal(p) , but it is allowed in Hive.

core/sql/optimizer/NATable.cpp
traflm: p>18 is important check, by now, I just raise error if p>18.
Since I cannot find a correct way to use SQLNumeric, It is always be treated as tinyint for some unknown reason.
Hive allow 'precision' up to 38, so when p>18 it will be great to use SQLBignum, I didn't realize BIGNUM can be used as Decimal type, will make a change.

core/sql/optimizer/NATable.cpp
traflm: agree, add checking for buffer overrun. And also other checks suggested by Suresh. This is machine-generated code, so should not have random string format and errors, otherwise Hive should not allow it. But I will still add some basic checking , in case Hive may have changed or some unknown format now and in the future.

core/sql/optimizer/NATable.cpp
```diff
@@ -3678,9 +3682,9 @@ NAType* getSQColTypeForHive(const char* hiveType, NAMemory* heap)
 
   if ( !strncmp(hiveType, "decimal", 7) )
   {
-    Int16 i=0, pstart=0, pend=0, sstart=0, send=0, p=0, s = 0;
-    Int16 hiveTypeLen = strlen(hiveType);
-    char pstr[16], sstr[16];
+    Lng32 i=0, pstart=0, pend=0, sstart=0, send=0, p=-1, s = -1;
```
zellerh: Just a nit: Since a "long" is a 64 bit number, it is better to use Int32 instead of Lng32. The "Lng32" came from porting 32 bit code, where a "long" meant 32 bits, and the port kept it at 32 bits.

core/sql/optimizer/NATable.cpp
```diff
@@ -86,6 +86,10 @@
 #include "CmpMain.h"
 
 #define MAX_NODE_NAME 9
+#define MAX_PRECISION_ALLOWED  18
+#define HIVE_MAX_PRECISION_ALLOWED  38
+#define MAX_SCALE_ALLOWED     6
```
zellerh: I don't think there is a maximum scale, neither for Hive nor for Trafodion. The only condition right now is that the scale can't exceed the precision. Example of a valid scale: DECIMAL(18,18).

core/sql/optimizer/NATable.cpp
```diff
@@ -3702,25 +3706,47 @@ NAType* getSQColTypeForHive(const char* hiveType, NAMemory* heap)
...
     if(send - sstart > 0)
     {
+      if( (send - sstart) >= MAX_NUM_LEN ) // too long
+        return NULL;
       strncpy(sstr,hiveType+sstart,send-sstart);
       s=atoi(sstr);
     }
 
-    if( (p>0) && (s>0) )
+    if( (p>0) && (p <= MAX_PRECISION_ALLOWED) ) //have precision between 1 - 18
```
zellerh: See comment above. Also, if we have an invalid scale, we shouldn't take the next if statement. Instead, if the scale is outside the range [0, p], assert or give an error.

core/sql/optimizer/NATable.cpp
```diff
@@ -3702,25 +3706,47 @@ NAType* getSQColTypeForHive(const char* hiveType, NAMemory* heap)
...
     {
-      return new (heap) SQLDecimal( p, s, TRUE, TRUE);
+      if( ( s >=0 )  &&  ( s<= MAX_SCALE_ALLOWED) ) //have valid scale
```
zellerh: See comment above about valid scale above.

core/sql/optimizer/NATable.cpp
```diff
@@ -3702,25 +3706,47 @@ NAType* getSQColTypeForHive(const char* hiveType, NAMemory* heap)
...
+        return new (heap) SQLDecimal( p, s, TRUE, TRUE);
+      else
+        return NULL;
     }
-    else
+    else if( (p > MAX_PRECISION_ALLOWED)  && ( p <= HIVE_MAX_PRECISION_ALLOWED) )  
     {
+      if ( (s>=0) && ( s< p ) ) //have valid scale
```
zellerh: I think this could check for s <= p. Making scale and precision the same should be allowed.

core/sql/optimizer/NATable.cpp
```diff
@@ -3676,6 +3680,76 @@ NAType* getSQColTypeForHive(const char* hiveType, NAMemory* heap)
...
+      }
+      else if(hiveType[i] == ')')
+      {
+        send = i;
+      }
+      else
+       continue;
+    }
+
+    if(pend - pstart > 0)
```
zellerh: What about a type DECIMAL(4). There is no comma, therefore pend is not set, and it would not be accepted. Maybe Hive never generates this syntax?

core/sql/optimizer/NATable.cpp
traflm: Sorry I misread the SQL reference manual, read again, yes, there is no limitation about Scale.

core/sql/optimizer/NATable.cpp
traflm: agree

core/sql/optimizer/NATable.cpp
traflm: @zellerh , I found Hive will do a kind of normalization here. If we use DECIMAL(4) in DDL, hive will transform it into decimal(4,0) and that is what Trafodion will get and parse it.
So all the checkings here are not necessary, but I still want to leave it since it doesn't harm and not impact performance  too much, this is not hot path.

svarnau:
jenkins, retest -- hadoop restart glitch

zellerh:
Thanks for the changes, looks good to me now.

DaveBirdsall:
Looks good to me also.

selvaganesang:
This PR might have caused the Trafodion daily build failure  https://jenkins.esgyn.com/job/Check-Daily-master/283/

@traflm Can you please check it out?

PR #620, opened 2016-07-25
TRAFODION 2001 (Elasticity) and small changes

amandamoran:


amandamoran:
Tested these changes on CDH 5.4.5 and HDP 2.4

PR #621, opened 2016-07-25

selvaganesang:
… of TM

The exception logged didn't have the type of exception. The output of
Exception.toString() was concatenated after it was released and hence
it was getting junk characters


sbroeder:
Looks good to me.

DaveBirdsall:
Looks good to me also.

PR #622, opened 2016-07-26
[TRAFODION-2128] update installer gitignore file

svarnau:
Jenkins automation is supposed to catch when we miss one, but the
check was not catching everything.  It will be updated when this fix is in.


DaveBirdsall:
Looks good to me.

PR #624, opened 2016-07-28
[TRAFODION-2130] Incorrect subquery unnesting

zellerh:
A missing ">" character caused us to do an unnesting transformation
without having uniqueness constraint on the main query.

Note that this is a rare case, since most tables have unique columns
and since the code tries to get those unique columns if they are not
already included.

### DIFF
--- a/core/sql/optimizer/NormRelExpr.cpp
+++ b/core/sql/optimizer/NormRelExpr.cpp
@@ -3044,14 +3044,23 @@ Join::pullUpGroupByTransformation()
 //             a copy of join as the child. The original tree has not changed.
 //             The predicates in the new groupBy and the new Join will have
 //             changed according to the comments above.
+//
 ------------------------------------------------------------------------------*/
 GroupByAgg* Join::pullUpGroupByTransformation(NormWA& normWARef)
 {
   CollHeap *stmtHeap = CmpCommon::statementHeap() ;
 
-  // Determine a set of unique columns for the left sub-tree
+  // Determine a set of unique columns for the left sub-tree.
+
+  // Note: Scans and joins synthesize uniqueness constraints even for
+  // columns that are not in the characteristic outputs. Other
+  // operators such as groupby or union don't. We make use of these
+  // extra uniqeness constraints here. Any needed columns not yet
+  // added to the characteristic outputs will be added later, in
+  // method getMoreOutputsIfPossible().
+
   ValueIdSet leftUniqueCols ;
-  if (NOT (child(0)-getGroupAttr()->findUniqueCols(leftUniqueCols)))
+  if (NOT (child(0)->getGroupAttr()->findUniqueCols(leftUniqueCols)))
   {
     // Could not find a set of unique cols.
     // If the left sub-tree contains a UNION/TRANSPOSE/SEQUENCE or SAMPLE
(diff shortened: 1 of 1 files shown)


DaveBirdsall:
Looks good to me.

core/sql/optimizer/NormRelExpr.cpp
```diff
@@ -3044,14 +3044,23 @@ Join::pullUpGroupByTransformation()
...
+  // Note: Scans and joins synthesize uniqueness constraints even for
+  // columns that are not in the characteristic outputs. Other
+  // operators such as groupby or union don't. We make use of these
+  // extra uniqeness constraints here. Any needed columns not yet
+  // added to the characteristic outputs will be added later, in
+  // method getMoreOutputsIfPossible().
+
   ValueIdSet leftUniqueCols ;
-  if (NOT (child(0)-getGroupAttr()->findUniqueCols(leftUniqueCols)))
+  if (NOT (child(0)->getGroupAttr()->findUniqueCols(leftUniqueCols)))
```
selvaganesang: Interesting.  Just curious what did the C++ compiler do earlier?

sureshsubbiah: Good catch Hans. Amazing that this has remained undetected for about 9 years!

core/sql/optimizer/NormRelExpr.cpp
```diff
@@ -3062,6 +3071,35 @@ GroupByAgg* Join::pullUpGroupByTransformation(NormWA& normWARef)
...
+    //             |
+    //             Y {pred1}
+    //
+    // Pros: - The groupby is already at a place where it will likely
+    //         end up in the optimal plan
+    // Cons: - We don't get a nice join backbone with all base tables
+    //
+    // Cases where we could attempt this transformation:
+    //   - We fail to find a unique key for X (i.e. we reach here)
+    //   - pred2 has a very high selectivity, making newJoin (in the picture
```
sureshsubbiah: If pred2 is any thing other than an equality predicate, or an equality predicate involving complex expressions, then the join effectively becomes a cartesian product. In such cases doing the groupby early will help reduce dataflow.

core/sql/optimizer/NormRelExpr.cpp
```diff
@@ -3062,6 +3071,35 @@ GroupByAgg* Join::pullUpGroupByTransformation(NormWA& normWARef)
     if (CmpCommon::getDefault(SUBQUERY_UNNESTING) == DF_DEBUG)
       *CmpCommon::diags() << DgSqlCode(2997)   
 	<< DgString1("Subquery was not unnested. Reason: Left child does not have a unique constraint"); 
+
+    // Things to consider (referring to the picture above): If the all of the 
+    // following are true:
+    //   * {pred2} has only equals/VEG predicates of the form X.col = Y.col
+    //   * {aggr} does not have any outer references
+    //   * {pred3} does not have any outer references
```
sureshsubbiah: Lack of outer references in {aggr} and {pred3} may indicate that this transformation may not apply when we more than one 1 level of correlated subqueries. This set of conditions clearly explains when this transformation can be applied. I hope we can add the change needed for this soon. The description here also indicates that NULL preserving semantics is not affected by this new transformation. So we could use this change for OR subqueries where the Join could be a Left Join.

sureshsubbiah:
Good work Hans. +1

core/sql/optimizer/NormRelExpr.cpp
DaveBirdsall: In answer to @selvaganesang, it called getGroupAttr()->findUniqueCols(leftUniqueCols) on "this", and subtracted that results (an NABoolean?) from the pointer child(0). That would almost always be non-zero. So the NOT would render it zero, and the "if" would not be taken.

PR #626, opened 2016-07-28
[TRAFODION-2131] Fix UPDATE STATISTICS cores + some other issues

DaveBirdsall:
1. Fix cores caused by buffer overruns when doing UPDATE STATISTICS on INTERVAL YEAR(1) TO MONTH, TINYINT and LARGEINT UNSIGNED columns. Thanks to @anoopsharma00 for suggesting the  use of CQDs to fix these issues.

2. Add support for BOOLEAN data type in UPDATE STATISTICS.

3. Fix a core in Cast::synthesizeType concerning INTERVAL and TINYINT data types. Thanks to @anoopsharma00 for the code for this fix.

### DIFF
--- a/core/sql/generator/GenExpGenerator.cpp
+++ b/core/sql/generator/GenExpGenerator.cpp
@@ -848,19 +848,33 @@ short ExpGenerator::handleUnsupportedCast(Cast * castNode)
   if ((srcFsType == REC_BIN8_SIGNED) ||
       (srcFsType == REC_BIN8_UNSIGNED) ||
       (tgtFsType == REC_BIN8_SIGNED) ||
-      (tgtFsType == REC_BIN8_UNSIGNED))
+      (tgtFsType == REC_BIN8_UNSIGNED) ||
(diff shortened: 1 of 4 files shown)


zellerh:
Looks good to me.

svarnau:
jenkins, retest -- automation bug caused build fail

selvaganesang:
Looks good to me. But just wonder, if it is ok to the set these CQDs to 'ON" when it is set to 'OFF' in the master. I am assuming this takes care of scenario when these datatypes were created earlier when these CQDs was 'ON'

DaveBirdsall:
Hi @selvaganesang. If we don't set the CQDs to 'ON', and the master has them 'OFF', then we core. So, UPDATE STATISTICS is overriding anything set in the master unconditionally. UPDATE STATISTICS does this for several other CQDs as well. It wants to have tighter control over its environment. That said, someday I expect these particular CQDs to go away once all clients support the new data types.

sureshsubbiah:
+1

PR #630, opened 2016-07-31
[TRAFODION-2119] DDL warning/error if using store by on utf8 column

traflm:
If store by is used and the column is not set 'not null', there will be an extra null indicator header. So the generated hbase split by string will contain un-initialized character and sometime raise warning or error.


traflm:
Something wrong, I need to abandon this PR.

PR #631, opened 2016-08-01
TRAFODION-1986 add code to setSchema api

mashengchen:
add code to method setSchema.
now can use setSchema(string schema) to set schema.


DaveBirdsall:
Looks good to me.

traflm:
@mashengchen , you need to check the regression test failure

mashengchen:
jenkins retest

selvaganesang:
Though this should work, this should be done efficiently to avoid many process hops between driver and server.  It would be better to send a message to the server and encapsulate this logic in the server.

selvaganesang:
It would be good to add some summary info about JIRA in the same line as the [TRAFODION-1986]. Otherwise it just shows an empty subject.

DaveBirdsall:
@selvaganesang: Just wanted to ask, what's the follow-on action? Should we commit this change, and also write another JIRA to improve the efficiency? Another question I wanted to ask, just for my own education: How many process hops does this have? I see one where we execute the statement. Do the properties changes also imply process hops?

ryzuo:
@selvaganesang : Moving the setting logic to server side is an good idea, and the way of doing so is using SetConnectionOption() method to send the message. However, the problem is we don't have data source management now, so driver is not 100% aware of the current catalog the session is on, especially if a "set catalog" was executed during the session lifetime. So if we want to set the schema via SetConnectionOption() method, we must call getCatalog() to retrieve the catalog name from server first, otherwise server will always use the default catalog name to concatenate with the schema name. thus it won't do any help for improving the efficiency. This is the only choice we have right now.

DaveBirdsall:
Are we in agreement that we can commit this now?

core/conn/jdbcT4/src/main/java/org/trafodion/jdbc/t4/TrafT4Connection.java
```diff
@@ -1918,8 +1918,52 @@ public Struct createStruct(String typeName, Object[] attributes)
...
+                    stmt = createStatement();
+                    stmt.execute("set schema " + schema);
+                } catch (TrafT4Exception se) {
+                    performConnectionErrorChecks(se);
+                    throw se;
+                } finally {
+                    if (stmt != null) {
+                        try {
+                            stmt.close();
+                        } catch (Exception e) {
```
selvaganesang: Consider changing the  Exception to SQLException or TrafT4Exception

selvaganesang:
I am kind of confused with the explanation because of a need for a data source management to do connection.setSchema method call.  To issue SET SCHEMA <schema_name> command there is no need to obtain the catalog.  Even in the current implementation,  SET SCHEMA <schema_name> doesn't prefix the catalog name.  Just found that  getCatalog command is also implemented as a request sent to the server. Hence we can let this implementation to go in.

However, my recommendation would be as follows:

JDBC applications are expected to use setCatalog and setSchema method to set the default catalog and schema.  JDBC applications tend to use getCatalog(), getSchema() as if it is lightweight commands. So, there is real need to improve this mechanism.

Yes. Trafodion doesn't have the mechanism to obtain the current catalog and schema via CLI call.  We can improve this and JDBC driver/server can be improved to obtain this default value as a response to connection request. Then the driver can maintain the current catalog/schema whenever setCatalog/setSchema is issued. Then it can service these methods from the driver itself.

To answer @DaveBirdsall question, there are two process hops with the current implementation for getCatalog(), getSchema(), setCatalog() and setSchema() and longer path length.

With the improved mechanism, getCatalog() and getSchema() will have no process hop. setCatalog() and setSchema() will have one process hop and much shorter path length.

traflm:
I agree with Selva's suggestion, and as Selva already agreed, this PR can be merged.
I suggest we create another JIRA to track Selva's proposal. 
This is a functional fix and the performance issue can be addressed later in the new JIRA. I can file a new JIRA once this is merged and everyone agree on this.

DaveBirdsall:
I am happy with that. Feel free to submit the JIRA; I'll then commit the change.

DaveBirdsall:
@traflm, has a JIRA been created for the performance issue? Perhaps you could record that here? Then I will commit this change.

traflm:
@DaveBirdsall , thanks, TRAFODION-2155 created.

PR #632, opened 2016-08-01
TRAFODION-1743 add connection session

mashengchen:
this jira has been merged in #561, but it is strange the changed code in apache/master disappeared. so i do a PR again.
add seesion to each connection.


traflm:
looks good, thanks

DaveBirdsall:
Looks good to me.

selvaganesang:
Isn't there a need to call end session at the time of disconnect?

traflm:
Oh yes, I agree with selva, need to call end session in Java_org_trafodion_jdbc_t2_SQLMXConnection_close() I guess. Otherwise, it never take effect.

traflm:
it is good now. thanks.

DaveBirdsall:
Merged this; looks like @selvaganesang 's comment was addressed.

PR #633, opened 2016-08-02
[TRAFODION-2119] DDL warning/error if using store by on utf8 column

traflm:
If 'store by' is used in a DDL and the column is not set 'not null', there will be an extra null indicator header. So the generated hbase split by string will contain un-initialized character and sometime raise warning or error.

The calling to type->minRepresentableValue should pass the buf address that take null indicator buffer into account, good examples are in:
-EncodedValue::minMaxValue()
-ConstValue::ConstValue()

but getMinMaxValue() is wrong. 

Although the calling conversion is not clear, this will be a fix to the issue, there are other some places to invoke minRepresentatbleValue, so keep the current semantics is good from my point of view.

### DIFF
--- a/core/sql/optimizer/EncodedKeyValue.cpp
+++ b/core/sql/optimizer/EncodedKeyValue.cpp
@@ -66,7 +66,7 @@ NAString * getMinMaxValue(desc_struct * column,
   if (NAColumn::createNAType(&column->body.columns_desc, NULL, type, NULL))
     return NULL;
   
-  Lng32 buflen = type->getTotalSize() - type->getSQLnullHdrSize();
+  Lng32 buflen = type->getTotalSize(); 
   char * buf = new char[buflen]; // deleted at the end of this method
 
   if (type->supportsSQLnullPhysical())
--- a/core/sql/optimizer/EncodedKeyValue.cpp
+++ b/core/sql/optimizer/EncodedKeyValue.cpp
@@ -66,15 +66,22 @@ NAString * getMinMaxValue(desc_struct * column,
   if (NAColumn::createNAType(&column->body.columns_desc, NULL, type, NULL))
     return NULL;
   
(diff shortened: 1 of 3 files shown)


traflm:
jenkins, retest

traflm:
Jenkins, extra tests

core/sql/optimizer/EncodedKeyValue.cpp
```diff
@@ -66,15 +66,22 @@ NAString * getMinMaxValue(desc_struct * column,
...
     return NULL;
   
-  Lng32 buflen = type->getTotalSize();
+  Lng32 buflen = type->getTotalSize(); 
   char * buf = new char[buflen]; // deleted at the end of this method
 
+  if (type->supportsSQLnullPhysical())
+  {
+      Lng32 nullHdrSize = type->getSQLnullHdrSize();
+      buf[0] = buf[1] = '\0';
```
selvaganesang: Is the nullHdrSize implied to be 2 bytes here?

DaveBirdsall: Seems like the code should zero out nullHdrSize bytes (for example, one can use memset, or alternatively a loop).

traflm: yes, @selvaganesang @DaveBirdsall , very good catch, I will fix this.

traflm:
Jenkins, extra tests

traflm:
the fix not working, I need to do more investigation, same code works well in my dev workstation. I will make a new commit and will abandon this PR again.

traflm:
Just want to pass Jenkins test, and don't merge this PR. I will resubmit after rework on it.

traflm:
Jenkins, extra tests

traflm:
abandon

PR #634, opened 2016-08-02
JIRA TRAFODION-2137 metadata access performance improvement for accessing traf objects

anoopsharma00:
Packed version of descriptor structures is now stored in traf TEXT
table during various DDL operations.
During dml, if packed descriptors exist, then they are read and used.
If they dont exist, then descriptors are generated on the fly.

Descriptor classes are defined in file sqlcat/TrafDDLdesc.h and
are based off the older desc structures from desc.h.
These are simplified and class'ified version of the older structures
and are versioned. Changes to these classes should be done carefully
by either using filler bytes or versioning them.

2 new cqds, traf_store_object_descriptor and traf_read_object_descriptor,
are added to control if packed descs should be stored/read.
These are currently set to OFF by default but are enabled during
regressions run. Once this feature is tested, these cqds will be turned
on by default.

In addition, some helper commands have been added to manipulate descrs.
Syntax:
 alter table t [check|generate|delete|disable|enable] stored descritors

If authorization is enabled, then auth/priv specific info
is generated on the fly and currently not stored in metadata.
In a later checkin, these will also be stored in metadata.


anoopsharma00:
JIRA TRAFODION-2137

core/sql/arkcmp/CmpStoredProc.cpp
```diff
@@ -273,77 +273,75 @@ NABoolean CmpSPOutputFormat::SetFormat(Lng32 nCols,
   nCols_ = nCols;
   // set up tableDesc_, allocate the storage from allocate routine,
   // so the delete of the tableDesc_ could be done correctly. 
-  // As in sqlcat/desc.h readtabledef_allocate_desc(), allocate routine gets 
+  // As in sqlcat/desc.h TrafAllocateDDLdesc(), allocate routine gets
```
DaveBirdsall: Should this comment say TrafDDLdesc.h instead of desc.h?

core/sql/arkcmp/CmpStoredProc.cpp
```diff
@@ -273,77 +273,75 @@ NABoolean CmpSPOutputFormat::SetFormat(Lng32 nCols,
...
+  // As in sqlcat/desc.h TrafAllocateDDLdesc(), allocate routine gets 
   // the storage from HEAP ( i.e. CmpCommon::statementHeap() ). 
   // Likewise, copyString should be the same implementation as in 
   // readRealArk.cpp. It was externalized originally, but being put as
   // static function by someone. Currently the code copyString is cloned
   // in the beginning of the file. ( which will allocate the memory from
   // statementHeap ).
 
-  desc_struct* table_desc = readtabledef_allocate_desc(DESC_TABLE_TYPE);
+  TrafDesc* table_desc = TrafAllocateDDLdesc(DESC_TABLE_TYPE, NULL);
```
DaveBirdsall: Why not just make table_desc an object of type TrafTableDesc * and avoid the repetitive casting below?

core/sql/arkcmp/CmpStoredProc.cpp
```diff
@@ -273,77 +273,75 @@ NABoolean CmpSPOutputFormat::SetFormat(Lng32 nCols,
...
-  table_desc->body.table_desc.files_desc = files_desc;
-  files_desc->body.files_desc.fileorganization = KEY_SEQUENCED_FILE;
+  TrafDesc * files_desc = TrafAllocateDDLdesc(DESC_FILES_TYPE, NULL);
+  table_desc->tableDesc()->files_desc = files_desc;
   
-  // populate the columns_desc_struct
-  desc_struct* prev_col_desc = 0;
-  desc_struct* first_col_desc =  0;
+  // populate the TrafColumnsDesc
+  TrafDesc* prev_col_desc = 0;
```
DaveBirdsall: TrafColumnDesc * perhaps?

core/sql/arkcmp/CmpStoredProc.cpp
```diff
@@ -273,77 +273,75 @@ NABoolean CmpSPOutputFormat::SetFormat(Lng32 nCols,
...
-      column_desc->body.columns_desc.colnumber = i;      
+      column_desc->columnsDesc()->colnumber = i;      
     }
 
-  table_desc->body.table_desc.columns_desc = first_col_desc;
+  table_desc->tableDesc()->columns_desc = first_col_desc;
 
   // populate index_desc and key_desc
-  desc_struct* keys_desc = 0;
+  TrafDesc* keys_desc = 0;
```
DaveBirdsall: TrafKeysDesc * perhaps?

core/sql/arkcmp/CmpStoredProc.cpp
```diff
@@ -273,77 +273,75 @@ NABoolean CmpSPOutputFormat::SetFormat(Lng32 nCols,
...
-  desc_struct * index_desc = readtabledef_allocate_desc(DESC_INDEXES_TYPE);
-  index_desc->body.indexes_desc.tablename = objName;
-  index_desc->body.indexes_desc.indexname = objName;
-  index_desc->body.indexes_desc.keytag = 0; // primary index
-  index_desc->body.indexes_desc.record_length = 
-    table_desc->body.table_desc.record_length;
-  index_desc->body.indexes_desc.colcount =
-    table_desc->body.table_desc.colcount;
-  index_desc->body.indexes_desc.blocksize = 4096; // anything > 0
+  TrafDesc * index_desc = TrafAllocateDDLdesc(DESC_INDEXES_TYPE, NULL);
```
DaveBirdsall: TrafIndexesDesc * perhaps?

core/sql/arkcmp/CmpStoredProc.cpp
```diff
@@ -363,7 +361,7 @@ NABoolean CmpSPOutputFormat::getColumnDesc(SP_FIELDDESC_STRUCT* fDesc,
     {
       // convert ElemDDLColViewDef into CatColumn, then column_desc
       if ( !ElemDDLColDef2ColumnDescStruct
-	   (elemDDL, tableDesc_->body.table_desc.tablename, colsDesc) )
+	   (elemDDL, tableDesc_->tableDesc()->tablename, colsDesc) )
```
DaveBirdsall: Why not make tableDesc_ an object of type TrafTableDesc * and skip the cast?

core/sql/sqlcat/TrafDDLdesc.cpp
```diff
@@ -0,0 +1,617 @@
...
+    }
+
+  return vtblptr;
+}
+
+// pack and unpack methods for various descriptor structs
+
+Long TrafDesc::pack(void * space)
+{
+  next.pack(space);
```
DaveBirdsall: These guys are all recursive. Any worries about stack overflow limits?

core/sql/sqlcat/TrafDDLdesc.h
```diff
@@ -0,0 +1,1120 @@
...
+  NABoolean isUnique() { return (indexesDescFlags & UNIQUE) != 0; };
+
+  ComPartitioningScheme partitioningScheme() 
+  { return (ComPartitioningScheme)partitioningScheme_; }
+  void setPartitioningScheme(ComPartitioningScheme v) 
+  { partitioningScheme_ = (Int16)v; }
+  ComRowFormat rowFormat() { return (ComRowFormat)rowFormat_; }
+  void setRowFormat(ComRowFormat v) { rowFormat_ = (Int16)v; }
+
+  char* tablename;  // name of the base table
```
DaveBirdsall: I was wondering about the char *'s. Do these pointer fields need to be packed and unpacked as well? (In the old pre-64-bit days perhaps we had to, but no longer?) I was wondering why they aren't wrapped in some pointer class like descriptor pointers. Perhaps it is because there is nothing inside the character array that itself needs to be packed/unpacked?

core/sql/sqlcat/TrafDDLdesc.h
```diff
@@ -0,0 +1,1120 @@
...
+
+  virtual TrafRoutineDesc *routineDesc() const { return (TrafRoutineDesc*)this; }
+
+  Int64 objectUID;
+  char* routineName;
+  char* externalName;
+  char* librarySqlName;
+  char* libraryFileName;
+  char* signature;
+  ComSInt32 paramsCount;
```
DaveBirdsall: Why ComSInt32 when just a few lines below we use Int32?

core/sql/sqlcomp/CmpSeabaseDDLschema.cpp
```diff
@@ -971,6 +972,251 @@ void CmpSeabaseDDL::dropSeabaseSchema(StmtDDLDropSchema * dropSchemaNode)
...
+          goto label_error;
+ 
+      // A Trafodion schema does not exist if the schema object row is not
+      // present: CATALOG-NAME.SCHEMA-NAME.__SCHEMA__.
+      *CmpCommon::diags() << DgSqlCode(-CAT_SCHEMA_DOES_NOT_EXIST_ERROR)
+                          << DgSchemaName(schemaName.getExternalName().data());
+      goto label_error;
+   }
+
+   if (!isDDLOperationAuthorized(SQLOperation::DROP_SCHEMA,
```
DaveBirdsall: Is there an ADD_SCHEMA privilege? Just curious: Why did we pick DROP_SCHEMA privilege for an alter schema operation?

DaveBirdsall:
Looks good to me. Most of my questions are matters of style.

core/sql/arkcmp/CmpStoredProc.cpp
anoopsharma00: For this checkin, wanted to keep the code structure somewhat similar to how it was.
That made it easier to make changes (lots of search and replace).
The old code used desc_struct which was replaced by TrafDesc class in the new code.
For a later checkin, we can change to using the underlying class instead of base class.

core/sql/sqlcat/TrafDDLdesc.h
anoopsharma00: Right. The 'char *' are pointing to char arrays and not any classes.
Currently, all classes either have native fields (int, short, etc), or 'char *' pointers
or 'TrafDesc' pointers.
If any pointer  in future points to a class, then it will need to include
pack/unpack of the class it is pointing to.

core/sql/sqlcat/TrafDDLdesc.h
anoopsharma00: leftover from previous code. Can change it to Int32  for a later checkin.

core/sql/sqlcomp/CmpSeabaseDDLcommon.cpp
```diff
@@ -5322,19 +5395,74 @@ short CmpSeabaseDDL::updateObjectRedefTime(
...
+                                (strcmp(objType, COM_BASE_TABLE_OBJECT_LIT) == 0
+                                 ? COM_BASE_TABLE_OBJECT : COM_VIEW_OBJECT),
+                                FALSE, ctlFlags, packedDescLen);
+      if (! ds)
+        {
+          processReturn();
+          return -1;
+        }
+
+      cliRC = updateTextTableWithBinaryData
```
zellerh: We update the stored descriptor here if TRAF_STORE_OBJECT_DESC is set to ON. What about the case when a table already has a stored object descriptor and TRAF_STORE_OBJECT_DESC is OFF? Should we at least delete the outdated stored descriptor? Sorry, I didn't check the details, probably this situation can never happen?

core/sql/sqlcomp/CmpSeabaseDDLcommon.cpp
```diff
@@ -6019,6 +6181,93 @@ short CmpSeabaseDDL::updateTextTable(ExeCliInterface *cliInterface,
...
+
+      str_sprintf(queryBuf, "insert into %s.\"%s\".%s values (%Ld, %d, %d, %d, 0, cast(? as char(%d bytes) character set utf8 not null))",
+                  getSystemCatalog(), SEABASE_MD_SCHEMA, SEABASE_TEXT,
+                  objUID,
+                  textType,
+                  subID,
+                  i,
+                  currDataLen);
+      cliRC = cliInterface->executeImmediateCEFC
+        (queryBuf, &inputData[currPos], currDataLen, NULL, NULL, NULL);
```
zellerh: This could chop a UTF-8 character in half. If that would happen, we would get an error. See method lightValidateUTF8Str() in file sql/common/csconvert.h for finding a valid spot to cut.

core/sql/sqlcomp/CmpSeabaseDDLcommon.cpp
```diff
@@ -6019,6 +6181,93 @@ short CmpSeabaseDDL::updateTextTable(ExeCliInterface *cliInterface,
...
+          return -1;
+        }
+    }
+
+  // convert input data to utf8 first.
+  ComDiagsArea * diagsArea = CmpCommon::diags();
+  char * inputDataUTF8 = new(STMTHEAP) char[inputDataLen*4];
+  Lng32 inputDataLenUTF8 = 0;
+  ex_expr::exp_return_type rc =
+    convDoIt(inputData,
```
zellerh: Treating the binary data as ISO8859-1 and converting that to UTF-8 is one way, but it is not very efficient and it results in unprintable characters. Another option would be to use 64 printable characters like '0' (zero) to 'o' (lower case O) and chop the binary data up into 6 bit pieces and encode each one into one of these characters. The overhead would be 33%. The ISO88591 method uses somewhere between 0 and 100% overhead, probably less than 50% in typical cases. It may not be worth it from the point of saving space, but it would avoid the unprintable characters.

Of course you can blame me for suggesting the ISO88591 method in the first place, so I guess I shouldn't complain about it now :-)

core/sql/sqlcomp/CmpSeabaseDDLcommon.cpp
zellerh: Sorry, a better method is findStartOfChar() in the same file.

zellerh: Hopefully my last comment about this: The original method, CmpSeabaseDDL::updateTextTable() seems to have the same problem of chopping UTF-8 characters in half. Maybe we get away with it. The original method also has a not so great method of dealing with embedded quotes, using a parameter like it's done here is much better. Would be nice if we could fix CmpSeabaseDDL::updateTextTable() and call it from here.

PR #635, opened 2016-08-03
fix regression test compGeneral TEST001

traflm:
Fix regression test case failure. 
Should be more carefully next time.


DaveBirdsall:
This is a follow-on for JIRA Trafodion 2047. I will commit it once tests pass.

traflm:
to be safe, let me run extra test, and check the result of compGeneral TEST001

traflm:
Jenkins, extra tests

DaveBirdsall:
My bad; I already committed it. I should have looked to see if compGeneral was included in the original Jenkins tests. This means you'll need to submit yet another pull request if this change was not satisfactory.

traflm:
I run the test before submit, so it should work. Let me git pull and test again.

PR #637, opened 2016-08-03
[TRAFODION-2138] Hive scan on wide tables can result in lost rows or error

sureshsubbiah:
The fix has three parts
a)	RangeTailIOSize will now default to maximum length of a row, if > 16KB. 
Each esp range will look ahead by this size. Previously this was hard coded to 16KB.
b)	If a whole buffer does not have the start of a row, that is not an error.
c)	We raise an error during compile if maximum row size is greater than size of IO buffer. 
We need to have the whole row (from start to finish) in a contiguous buffer to do the rest of our logic. 
Currently IO buffer defaults to 64 MB.


core/sql/generator/GenRelScan.cpp
```diff
@@ -1182,6 +1181,15 @@ if (hTabStats->isOrcFile())
 
   UInt32 rangeTailIOSize = (UInt32)
       CmpCommon::getDefaultNumeric(HDFS_IO_RANGE_TAIL);
+  if (rangeTailIOSize == 0) 
+    {
+      rangeTailIOSize = getTableDesc()->getNATable()->getRecordLength() +
```
zellerh: This record length is what the compiler thinks, the real length may be more or less. Hive makes it probably more complex, some fields like date/time/timestamp are strings in Hive and encoded binaries in Trafodion. Character sets may be different as well in some cases (e.g. GBK). I don't have a better solution than what the code does, but let's say we have this: create table t(a char(20000) not null, b timestamp(6) not null, c timestamp(6) not null). Trafodion will make the length 20022. In reality, a string may look like "20k chars..., 2015-01-01 00:00:00.000, 2016-01-01 00:00:00.000". That's 2049 characters. Maybe the solution is to add the 16K to the record length, not take the max?

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -7582,6 +7582,24 @@ RelExpr *Scan::bindNode(BindWA *bindWA)
...
 
+   if (naTable->isHiveTable() && 
+       !(naTable->getClusteringIndex()->getHHDFSTableStats()->isOrcFile() ||
+	 naTable->getClusteringIndex()->getHHDFSTableStats()
+	 ->isSequenceFile()) &&
+       (CmpCommon::getDefaultNumeric(HDFS_IO_BUFFERSIZE_BYTES) == 0) && 
+       (naTable->getRecordLength() >
+	CmpCommon::getDefaultNumeric(HDFS_IO_BUFFERSIZE)*1024))
+     {
+       // do not raise error if buffersize is set though buffersize_bytes.
```
zellerh: What if the user users HDFS_IO_BUFFERSIZE_BYTES and that size is too small? Sorry, I don't understand the meaning of the 2 CQDs, are they for the same thing and just use different units?

zellerh:
Fix looks good to me, with a few questions on some details.

core/sql/optimizer/BindRelExpr.cpp
sureshsubbiah: HDFS_IO_BUFFERSIZE_BYTES was added initially in cases where we wanted to specify the buffer size in granularity less than a KB. The comment indicates that this is for testing. I was tempted to remove the cqd, but thought we may still need to test with precisely defined buffer sizes.

I am viewing BUFFERSIZE_BYTES as some kind of internal switch, to be used by expert users to override the BUFFERSIZE setting. In case users have a wide table but for some reason want to avoid the new error they can use BUFFERSIZE_BYTES to skip the new error code. Wide tables can be handled if the actual rows are indeed small enough to fit in a buffer, the user may have other reasons for not wanting to changing defaults such as HIVE_MAX_STRING_LENGTH.

If we prefer, I can also remove this BUFFERSIZE_BYTES backdoor.

core/sql/bin/SqlciErrors.txt
```diff
@@ -1282,6 +1282,7 @@ $1~String1 --------------------------------
 4223 42000 99999 BEGINNER MAJOR DBADMIN $0~String0 not supported in this software version.
 4224 42000 99999 BEGINNER MAJOR DBADMIN The directory $0~String0 is not a valid Hive location.
 4225 42000 99999 BEGINNER MAJOR DBADMIN Number of column families cannot exceed 32.
+4226 42000 99999 BEGINNER MAJOR DBADMIN Table $0~TableName has a maximum record length of $0~Int0 which is greater than HDFS_IO_BUFFERSIZE. Increase buffer size setting, or reduce HIVE_MAX_STRING_LENGTH.
```
sureshsubbiah: Done. Thanks for the suggestion.

core/sql/optimizer/BindRelExpr.cpp
zellerh: Another thing you can do is to make HDFS_IO_BUFFERSIZE a FLOAT, then you can specify sizes of less than one KB as well. That would be less confusing. I'm not sure I understand how you wanted to use this as a backdoor and don't have a strong opinion, you will know better how you want to use it.

sureshsubbiah:
jenkins, retest

PR #638, opened 2016-08-03
JIRA-2137 Addnl change to metadata improvement based on review comments

anoopsharma00:
Code now handles writing binary data into TEXT after encoding it.
And decodes it during read.

Non-binary data is added using input params  instead of literals.


zellerh:
Thanks, looks good, nice to have a single method to update the text table that can now handle binary data. I filed TRAFODION-2139 for the other problem where we may split UTF-8 data the wrong way, but that's of course no longer an issue here, since you are encoding the binary data into all ASCII characters.

PR #640, opened 2016-08-04
[TRAFODION-2142] Add script to restart HBase for developer regressions

DaveBirdsall:
This check-in adds a script, keepHBaseUp.py, that can be used in tandem with developer regressions to keep regressions from getting hung when the local_hadoop HMaster goes away (as it is wont to do on busy workstations). The script periodically checks to see if the HMaster is up. If it isn't, it attempts to start it. It will retry the start, at geometrically longer intervals, until over an hour passes without success, then it gives up.

One way to use this script is to start it running in one shell window, then start the developer regressions in another.

The script logs the times when it does its checks, so the developer can correlate HBase down scenarios with regression test failures.

### DIFF
--- a/core/sqf/sql/scripts/keepHBaseUp.py
+++ b/core/sqf/sql/scripts/keepHBaseUp.py
@@ -1,0 +1,113 @@
+# @@@ START COPYRIGHT @@@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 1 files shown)


amandamoran:
Looks good to me.

sureshsubbiah:
+1

PR #641, opened 2016-08-04
[JIRA TRAFODION-2141] Cluster view of hbase stats

anoopsharma00:
-- added column REGION_SERVER to region stats virtual table

-- added cluster view of hbase stats
   Syntax: select * from table(cluster stats());

-- added support for 'where' pred in stats queries
  Syntax: select * from table(cluster stats()) where schema_name = 'SCH';


anoopsharma00:
jenkins retest

core/sql/src/main/java/org/trafodion/sql/TrafRegionStats.java
```diff
@@ -128,5 +155,90 @@ public TrafRegionStats (HTable table, Admin admin) throws IOException {
...
+
+    public TrafRegionStats () throws IOException {
+        Configuration config = HBaseConfiguration.create();
+
+        hbAdmin = new HBaseAdmin(config);
+    }
+
+    public boolean Open () throws IOException {
+        
+        clusterStatus = hbAdmin.getClusterStatus();
```
selvaganesang: HBaseAdmin is deprecated and its use is avoided. Please change it to use Admin instead. You need to pass the connection object and then use connection.getAdmin()

core/sql/src/main/java/org/trafodion/sql/TrafRegionStats.java
```diff
@@ -128,5 +155,90 @@ public TrafRegionStats (HTable table, Admin admin) throws IOException {
...
+
+        return true;
+    }
+
+    public SizeInfo getCurrRegionSizeInfo() {
+        return currRegionSizeInfo;
+    }
+    
+    public boolean Close () throws IOException {
+
```
selvaganesang: Connection.getAdmin is a  light weight operation. There is no need to make hbadmin as class member and then close it.

PR #642, opened 2016-08-05
[TRAFODION-2119] create table report strange WARNING when using store…

traflm:
… by and allow_nullable_unique_key_constraint CQD

### DIFF
--- a/core/sql/common/CharType.cpp
+++ b/core/sql/common/CharType.cpp
@@ -1217,7 +1217,7 @@ void CharType::minMaxRepresentableValue(void* bufPtr,
   if (stringLiteral)
     {
       NABoolean isNull = FALSE;
-      NABoolean res = createSQLLiteral((const char *) bufPtr, *stringLiteral, isNull, h);
+      NABoolean res = createSQLLiteral((const char *) bufPtr  - getSQLnullHdrSize(), *stringLiteral, isNull, h);
       assert(res);
     }
 }
--- a/core/sql/common/CharType.cpp
+++ b/core/sql/common/CharType.cpp
@@ -1217,6 +1217,8 @@ void CharType::minMaxRepresentableValue(void* bufPtr,
   if (stringLiteral)
     {
       NABoolean isNull = FALSE;
+      //the bufPtr passed in point to the postion after null header
+      //createSQLLiteral requires the input pointer to be the begin of the buffer
       NABoolean res = createSQLLiteral((const char *) bufPtr  - getSQLnullHdrSize(), *stringLiteral, isNull, h);
       assert(res);
     }
--- a/core/sql/optimizer/EncodedKeyValue.cpp
+++ b/core/sql/optimizer/EncodedKeyValue.cpp
@@ -68,10 +68,11 @@ NAString * getMinMaxValue(TrafDesc * column,
     return NULL;
   
   Lng32 buflen = type->getTotalSize();
+  Lng32 nullHdrSize = 0;
   char * buf = new char[buflen]; // deleted at the end of this method
   if (type->supportsSQLnullPhysical())
   {
-      Lng32 nullHdrSize = type->getSQLnullHdrSize();
+      nullHdrSize = type->getSQLnullHdrSize();
       for(int i = 0; i < nullHdrSize; i++)
         buf[i] = '\0';
       buflen-= nullHdrSize;
@@ -82,7 +83,7 @@ NAString * getMinMaxValue(TrafDesc * column,
     {
       // low key needed
       if (NOT key->keysDesc()->isDescending()) // ascending
-	type->minRepresentableValue(buf, &buflen, 
+	type->minRepresentableValue(buf + nullHdrSize, &buflen, 
 				    &minMaxValue,
 				    h) ;
       else
@@ -94,7 +95,7 @@ NAString * getMinMaxValue(TrafDesc * column,
 	    }
 	  else
 	    {
-	      type->maxRepresentableValue(buf, &buflen, 
+	      type->maxRepresentableValue(buf + nullHdrSize, &buflen, 
 					  &minMaxValue,
 					  h) ;
 	    }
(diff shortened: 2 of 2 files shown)


traflm:
This PR must need extra test qat001 to pass

traflm:
Jenkins, extra tests

traflm:
introduce new EXECUTOR/TEST015 failure, will try again and it passed in my workstation, so probably a transient glitch?

traflm:
Jenkins, extra tests

traflm:
Failure test case:
HIVE003
HIVE004
compGeneral 042

So it is identical with main branch failures, not introduced by this RP. From regression point of view, this PR is clean now.

core/sql/optimizer/EncodedKeyValue.cpp
```diff
@@ -69,7 +69,14 @@ NAString * getMinMaxValue(TrafDesc * column,
   
   Lng32 buflen = type->getTotalSize();
   char * buf = new char[buflen]; // deleted at the end of this method
-
+  if (type->supportsSQLnullPhysical())
+  {
+      Lng32 nullHdrSize = type->getSQLnullHdrSize();
+      for(int i = 0; i < nullHdrSize; i++)
```
sureshsubbiah: Prior to this change what was being placed in the null bytes by this method?

core/sql/common/CharType.cpp
```diff
@@ -1217,7 +1217,7 @@ void CharType::minMaxRepresentableValue(void* bufPtr,
   if (stringLiteral)
     {
       NABoolean isNull = FALSE;
-      NABoolean res = createSQLLiteral((const char *) bufPtr, *stringLiteral, isNull, h);
+      NABoolean res = createSQLLiteral((const char *) bufPtr  - getSQLnullHdrSize(), *stringLiteral, isNull, h);
```
sureshsubbiah: Could you please explain this line.
Is the understanding that bufPtr passed in does not include null header, but the previous 2 bytes (say type is nullable and null size is 2 bytes) are available and meant to hold null bytes. I suppose I am nervous about taking a pinter passed in and going a few bytes before. How do we know that is still a valid location? Maybe all this handled in createSQLLiteral ?

traflm: The calling to type->minRepresentableValue should pass the buf address that take null indicator buffer into account, good examples are in:
-EncodedValue::minMaxValue()
-ConstValue::ConstValue()

As you can see in above two functions, the buffer ptr is set to the postion:
buffer_start + null_header_len

a buffer should like
[nullHeader][VCHARLEN]Data  

inside minRepresentableValue , it will treat the VCHARLEN, so it seems the agreement is the nullHeader is already bypassed.

So the starting position of bufPtr is
[nullHeader][VCHARLEN]Data  
                   ^

However, in the createSQLLiteral() , it will again bypass the nullHeader:
  const char *valPtr = buf + getSQLnullHdrSize();

So I did this change. As you correctly pointed out, this is dangerous... 
A better way is to change createSQLLiteral, but there are some other places call   createSQLLiteral() which has different assumption of the starting position of the buffer ptr, ConstValue::getConstStr() for example, so I choose the simplest way, I am more nervous to change many other code which I don't understand where will be invoked. But yes, it is not good. 
A better change is to fix in createSQLLiteral semantics, the input buffer ptr will be the position after nullHeader, but that will involve more changes which I feel not confident, but I will try.

traflm: And I forgot to advance the buf pointer in below getMinMaxValue() changes... Need a rework again.

core/sql/optimizer/EncodedKeyValue.cpp
traflm: Prior to this change, the nullIndictor will be filled by 0 by accident, since the minimum value is 0. So it will fill in N zeros into the buffer starting from the null header.
I don't know if it is required for other type. This is a good question, let me check other type as well.
But in fact, this change is not necessary. Since the buf is never used in this function.

I will rework on this.

core/sql/common/CharType.cpp
traflm: I still think move back the pointer will be the best approach here. I will add a comment to explicitly say the starting pos of the input bufPtr point to the position after the null header. And since the createSQLLiteral need the input buf starting address to be before null header, we need to move back. Otherwise, we have too much places to modify. What do you think?

sureshsubbiah: Thank you for explaining to me. I think you know more about this area than do.  Please go with the change that you think is best.

traflm:
jenkins, extra tests

sureshsubbiah:
+1. Looks good to me. Will wait a day or so to see if there are additional comments before merging. Thank you fixing this intricate problem.

DaveBirdsall:
Looks good to me.

core/sql/optimizer/EncodedKeyValue.cpp
```diff
@@ -68,14 +68,22 @@ NAString * getMinMaxValue(TrafDesc * column,
     return NULL;
   
   Lng32 buflen = type->getTotalSize();
+  Lng32 nullHdrSize = 0;
   char * buf = new char[buflen]; // deleted at the end of this method
-
+  if (type->supportsSQLnullPhysical())
+  {
+      nullHdrSize = type->getSQLnullHdrSize();
+      for(int i = 0; i < nullHdrSize; i++)
```
ryzuo: Not familiar with this part, but would it be more efficient to use memset() to do the initial work rather than using a for loop? Just my opinion...

traflm: there are two issues to use memset() here, please keep in mind the nullHdrSize will be 2 or 4

1. memset is good for large buffer initialization, because it will try to do 8 bytes copy in parallel and in memory alignment, but for small buffer , it is same as a while loop. memset() . It will also try to adjust memory alignment, which is useless for 2 or 4 bytes initialization, but just waste a few CPU cycle.
2. memset will require a function call.

And the most important: this is DDL, which performance of this kind of issue is not my concern.

PR #643, opened 2016-08-05
[TRAFODION-2145]sqlcolumns returned error info for some datatype

xwq:
The root cause is the FS_DATA_TYPE is not right for some datatype in the
query for SQLCOLUMNS.


xwq:
This JIRA is about one of the invterval datatype column didn't return result. However , all the interval datatype can't work since the FS_DATA_TYPE is not right. 
I fix all the interval datatype in this Pull Request.

core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
```diff
@@ -4834,20 +4834,19 @@ odbc_SQLSvc_GetSQLCatalogs_sme_(
...
-  "('INTERVAL', 111, 0, '{INTERVAL ''', ''' HOUR TO MINUTE}', NULL, 1, 0, 2, 0, 0, NULL, 'INTERVAL', 0, 0, 'INTERVAL', 205, NULL, 3, 34, 100, 11, 2, 4, 5, 3, 0), "
+  "('INTERVAL', 111, 0, '{INTERVAL ''', ''' HOUR TO MINUTE}', NULL, 1, 0, 2, 0, 0, NULL, 'INTERVAL', 0, 0, 'INTERVAL', 202, NULL, 3, 34, 100, 11, 2, 4, 5, 3, 0), "
   "('INTERVAL', 112, 0, '{INTERVAL ''', ''' HOUR TO SECOND}', NULL, 1, 0, 2, 0, 0, NULL, 'INTERVAL', 0, 0, 'INTERVAL', 206, NULL, 6, 34, 100, 12, 2, 4, 6, 3, 0), "
-  "('INTERVAL', 110, 0, '{INTERVAL ''', ''' DAY TO SECOND}', NULL, 1, 0, 2, 0, 0, NULL, 'INTERVAL', 0, 0, 'INTERVAL', 204, NULL, 9, 34, 100, 10, 2, 3, 6, 3, 0), "
+  "('INTERVAL', 110, 0, '{INTERVAL ''', ''' DAY TO SECOND}', NULL, 1, 0, 2, 0, 0, NULL, 'INTERVAL', 0, 0, 'INTERVAL', 207, NULL, 9, 34, 100, 10, 2, 3, 6, 3, 0), "
   "('INTERVAL', 109, 0, '{INTERVAL ''', ''' DAY TO MINUTE}', NULL, 1, 0, 2, 0, 0, NULL, 'INTERVAL', 0, 0, 'INTERVAL', 203, NULL, 6, 34, 100, 9, 2, 3, 5, 3, 0), "
   "('INTERVAL', 103, 0, '{INTERVAL ''', ''' DAY}', NULL, 1, 0, 2, 0, 0, NULL, 'INTERVAL', 0, 0, 'INTERVAL', 197, NULL, 0, 34, 100, 3, 2, 3, 3, 3, 0), "
```
sureshsubbiah: DAY to DAY is 198 in sqlcli.h, maybe this line should be changed too ?

core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
```diff
@@ -4834,20 +4834,19 @@ odbc_SQLSvc_GetSQLCatalogs_sme_(
...
   "('NUMERIC', 2, 128, NULL, NULL, 'precision,scale', 1, 0, 2, 0, 0, 0, 'NUMERIC', 0, 128, 'SIGNED NUMERIC', 156, 10, -2, -3, 2, NULL, NULL, 0, 0, 3, 0), "
```
sureshsubbiah: The fs datatype of a signed numeric depends on precision too?
130, 132, 134, or 156. Are we allowed to disregard precision here? Similar comment for the following line on NUMERIC UNSIGNED

core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
```diff
@@ -4834,20 +4834,19 @@ odbc_SQLSvc_GetSQLCatalogs_sme_(
   "('DECIMAL', 3, 18, NULL, NULL, 'precision,scale', 1, 0, 2, 0, 0, 0, 'DECIMAL', 0, 18, 'SIGNED DECIMAL', 152, 10, -2, -3, 3, NULL, NULL, 0, 0, 3, 0), "
   "('DECIMAL UNSIGNED', 3, 18, NULL, NULL, 'precision,scale', 1, 0, 2, 1, 0, 0, 'DECIMAL', 0, 18, 'UNSIGNED DECIMAL', 150, 10, -2, -3, -301, NULL, NULL, 0, 0, 3, 0), "
   "('DOUBLE PRECISION', 8, 15, NULL, NULL, NULL, 1, 0, 2, 0, 0, 0, 'DOUBLE', NULL, NULL, 'DOUBLE', 143, 2, 54, -1, 8, NULL, NULL, 0, 0, 3, 0), "
-  "('FLOAT', 6, 15, NULL, NULL, NULL, 1, 0, 2, 0, 0, 0, 'FLOAT', NULL, NULL, 'FLOAT', 142, 2, -2, -1, 6, NULL, NULL, 0, 0, 3, 0), "
```
sureshsubbiah: Should FLOAT not be present in the new file too? Or is FLOAT represented by REAL below?

core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
sureshsubbiah: I count 13 interval types in this file, while in sqlci.h we have 18 interval types. Do all of them need to be recorded here?
Same question about other missing datatypes such as
DECIMAL has more entries in sqlcli.h, depending in precision.
CLOB, BLOB, INT, UNSIGNED INT all seem to be missing here.

xwq: I get value by check the metadata table. Maybe I miss "DAY to DAY", I will check it and other missing type in with sqlcli.h.

core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
xwq: Here the issues is the FS_DATA_TYPE of FLOAT is same with REAL. 

Actually, If I created a table with FLOAT column, but it represented by "DOUBLE " type. I think it's a bug, it should be REAL instead. I can try to fix it.

xwq:
Hi @sureshsubbiah , 
I think it's better to create a new JIRA to track the misssing datatype. I will create a new Pull request with some new test for the missing datatype. 

For the missing INTERVAL datatype, I can't find them in the sqlext.h. So I can't add them with correct DATA_TYPE now.

sureshsubbiah:
+1 Looks good.  Thank you.

PR #644, opened 2016-08-05
Trafodion 1788

robertamarton:


core/sql/bin/SqlciErrors.txt
```diff
@@ -26,7 +26,7 @@
 1024 ZZZZZ 99999 BEGINNER MINOR DBADMIN File system error $0~NSKCode occurred on $1~String0. $2~String1
 1025 ZZZZZ 99999 BEGINNER MINOR DBADMIN Request failed.  Dependent object $0~string0 exists.
 1026 ZZZZZ 99999 BEGINNER MINOR DBADMIN Only the schema owner or super ID can drop a schema.
-1027 ZZZZZ 99999 BEGINNER MINOR DBADMIN The definition schema $0~SchemaName is dropped when its catalog is dropped.
+1027 ZZZZZ 99999 BEGINNER MINOR DBADMIN The view $0~TableName was created before column privileges was supported. To grant column privileges, please recreate the view.
```
DaveBirdsall: Grammar. Should be "...was created before column privileges were supported.

core/sql/common/ComViewColUsage.h
```diff
@@ -0,0 +1,89 @@
...
+
+  Int32 getRefdColNumber () { return refdColNumber_; }
+  void setRefdColNumber(Int32 refdColNumber) { refdColNumber_ = refdColNumber; }
+
+  Int32 getRefdObjectType () { return refdObjectType_; }
+  void setRefdObjectType(ComObjectType refdObjectType) { refdObjectType_ = refdObjectType; }
+
+  void packUsage (NAString &viewColUsageStr)
+  {
+      char buf[sizeof(Int64)*2 + sizeof(Int32)*3 + 100];
```
DaveBirdsall: sizeof(Int64) would be 8, but I think your intention was to allocate enough space for the character representation, right? A signed 64-bit integer takes up to 20 bytes (counting the sign). A signed 32-bit integer takes up to 11 bytes (counting the sign). That said, your 100 is probably enough to make up the difference here.

core/sql/common/ComViewColUsage.h
```diff
@@ -0,0 +1,89 @@
...
+      str_sprintf(buf, "viewUID: %Ld viewCol: %d refUID: %Ld refCol: %d refType: %d;",
+                  viewUID_, viewColNumber_,
+                  refdUID_, refdColNumber_, refdObjectType_);
+      viewColUsageStr = buf;
+  }
+
+  void unpackUsage (const char *viewColUsageStr)
+  {
+      Int32 theRefdObjectType;
+      sscanf(viewColUsageStr, "viewUID: %Ld viewCol: %d refUID: %Ld refCol: %d refType: %d%*s",
```
DaveBirdsall: I take it we don't care about errors here?

DaveBirdsall:
Any idea why github thinks your file core/sql/regress/privs2/EXPECTED143 is a binary file?

core/sql/sqlcomp/CmpSeabaseDDLview.cpp
```diff
@@ -247,6 +249,65 @@ short CmpSeabaseDDL::buildViewColInfo(StmtDDLCreateView * createViewParseNode,
...
+     ComString usedObjColName(usedColRef.getColName());
+     const NAColumn * naCol = nacolArr.getColumn(usedObjColName);
+     if (naCol == NULL)
+     {
+       *CmpCommon::diags() << DgSqlCode(-CAT_COLUMN_DOES_NOT_EXIST_ERROR)
+                           << DgColumnName(usedObjColName);
+        return -1;
+     }
+
+     ComViewColUsage colUsage;
```
DaveBirdsall: Just curious: Why did you choose to set all the members using setter members rather than simply having a constructor do it all?

core/sql/sqlcomp/CmpSeabaseDDLview.cpp
```diff
@@ -394,8 +458,7 @@ short CmpSeabaseDDL::gatherViewPrivileges (const StmtDDLCreateView * createViewN
   const ParViewColTableColsUsageList &vctcul = vu.getViewColTableColsUsageList();
 
   // If DB__ROOT, no need to gather privileges
```
DaveBirdsall: The DB__ROOT condition no longer seems to apply?

core/sql/sqlcomp/PrivMgrMD.cpp
```diff
@@ -630,6 +630,138 @@ bool PrivMgrMDAdmin::isAuthorized (void)
...
+
+  // Select column details for object 
+  std::string selectStmt = "select c.column_number from ";
+  selectStmt += objectMDTable;
+  selectStmt += ", ";
+  selectStmt += colMDTable;
+  selectStmt += " where o.object_uid = ";
+  selectStmt += UIDToString(objectRef->objectUID);
+  selectStmt += " and o.object_uid = c.object_uid";
+  selectStmt += " order by column_number";
```
DaveBirdsall: Why is the join to the OBJECTS table needed? Looks like the only thing you are retrieving is from the COLUMNS table and you have the key info you need for that?

core/sql/sqlcomp/PrivMgrMD.cpp
```diff
@@ -630,6 +630,138 @@ bool PrivMgrMDAdmin::isAuthorized (void)
...
+
+  // Select text rows describing view <=> object column relationships
+  std::string selectStmt = "select text from ";
+  selectStmt += objectMDTable;
+  selectStmt += ", ";
+  selectStmt += textMDTable;
+  selectStmt += " where o.object_uid = ";
+  selectStmt += UIDToString(viewUsage.viewUID);
+  selectStmt += "and o.object_uid = t.text_uid and t.text_type = 8";
+  selectStmt += " order by seq_num";
```
DaveBirdsall: Again, why are you joining to the OBJECTS table? Doesn't look like you're getting any information from it.

core/sql/sqlcomp/PrivMgrMD.cpp
```diff
@@ -630,6 +630,138 @@ bool PrivMgrMDAdmin::isAuthorized (void)
...
+
+  // the text column length in the TEXT table is 10000
+  char value[10000 + 1];
+  
+  // For each row, add it to the existing viewColUsages string
+  objectsQueue->position();
+  for (int idx = 0; idx < objectsQueue->numEntries(); idx++)
+  { 
+    OutputInfo * pCliRow = (OutputInfo*)objectsQueue->getNext();
+    ColumnReference *columnReference = new ColumnReference;
```
DaveBirdsall: Doesn't look like columnReference is used after this. Is this a memory leak?

core/sql/sqlcomp/PrivMgrMD.h
```diff
@@ -187,6 +202,25 @@ class ObjectUsage
   PrivMgrDesc originalPrivs;
   PrivMgrDesc updatedPrivs;
 
+  void copyColumnReferences(const std::vector<ColumnReference *> *refsToCopy)
+  {
+    if (refsToCopy == NULL)
```
DaveBirdsall: Might there be an old value in columnReferences? Do we need to do an NABASICDELETE (or a delete) on it before assigning a new object pointer to it?

core/sql/sqlcomp/PrivMgrMD.h
```diff
@@ -187,6 +202,25 @@ class ObjectUsage
...
+      columnReferences = NULL;
+    else
+    {
+      columnReferences = new std::vector<ColumnReference *>;
+      for (int i = 0; i < refsToCopy->size(); i++)
+      {
+        ColumnReference *newRef = new ColumnReference;
```
DaveBirdsall: A copy constructor might save you the tedium of member-by-member copy.

core/sql/sqlcomp/PrivMgrPrivileges.cpp
```diff
@@ -2101,17 +2093,114 @@ PrivStatus PrivMgrPrivileges::dealWithViews(
...
+  // Get columns for referenced object if they are  not already present
+  if (objectRef->columnReferences == NULL)
+  {
+    retcode = admin.getColumnReferences(objectRef);
+    if (objectRef->columnReferences)
+    {
+      traceMsg += "getting column references: number references found ";
+      traceMsg += to_string((long long int)objectRef->columnReferences->size());
+    }
+    traceMsg += ", retcode is ";
```
DaveBirdsall: If objectRef->columnReferences is NULL, your trace message will begin with a comma. Will look a little odd perhaps in your trace log.

core/sql/sqlcomp/PrivMgrPrivileges.cpp
```diff
@@ -2101,17 +2093,114 @@ PrivStatus PrivMgrPrivileges::dealWithViews(
...
+    if (retcode == STATUS_ERROR)
+      return retcode;
+  }
+
+  // Extract view <=> object column relationship from the TEXT table 
+  // and add it to the viewUsage
+  if (viewUsage.viewColUsagesStr.empty())
+  {
+    retcode = admin.getViewColUsages(viewUsage);
+    traceMsg += "getting view column usages";
```
DaveBirdsall: Do you want "=" here instead of "+="? (With "+=" you will be repeating the traceMsg info that you already logged above.

core/sql/sqlcomp/PrivMgrPrivileges.cpp
```diff
@@ -4208,43 +4514,59 @@ void PrivMgrPrivileges::summarizeColPrivs(
...
+    {
+      for (int32_t j = 0; j < rowList.size();++j)
       {
-        ObjectUsage *currentObj = listOfAffectedObjects[j];
-        if (currentObj->objectUID == row.objectUID_ &&
-            grantorID_ == row.grantorID_ &&
-            objReference.objectOwner == row.granteeID_ )
+        ColumnPrivsMDRow &row = static_cast<ColumnPrivsMDRow &> (*rowList[j]);
+        PrivMgrCoreDesc originalPrivs(row.privsBitmap_, row.grantableBitmap_);
+        PrivMgrCoreDesc original = originalPrivs;
```
DaveBirdsall: I take it "original" is a copy of "originalPrivs"? Not sure why we make this copy? (originalPrivs doesn't seem to be used again other than to make another copy into "updated")

core/sql/bin/SqlciErrors.txt
robertamarton: Fixed

core/sql/common/ComViewColUsage.h
robertamarton: You are correct, will fix.

core/sql/common/ComViewColUsage.h
robertamarton: I did a search on sscanf on our code base and a majority of the callers to not check to make sure the correct number of parms were setup.  But will add a check to make sure the string is setup up as expected.

core/sql/sqlcomp/CmpSeabaseDDLview.cpp
robertamarton: Good suggestion to add a constructor.

core/sql/sqlcomp/PrivMgrMD.cpp
robertamarton: You are correct, there is no need to include the objects table, will fix.

core/sql/sqlcomp/PrivMgrMD.cpp
robertamarton: You are correct, no need to include objects, will fix.

core/sql/sqlcomp/PrivMgrMD.cpp
robertamarton: Must be left over from a prior change, will remove code.

core/sql/sqlcomp/PrivMgrMD.h
robertamarton: Good suggestion.

core/sql/sqlcomp/PrivMgrPrivileges.cpp
robertamarton: Will fix.

core/sql/sqlcomp/PrivMgrPrivileges.cpp
robertamarton: Good find, will fix.

core/sql/sqlcomp/PrivMgrPrivileges.cpp
robertamarton: You are correct, will remove the original steps.

core/sql/common/ComViewColUsage.h
```diff
@@ -67,7 +88,9 @@ class ComViewColUsage
 
   void packUsage (NAString &viewColUsageStr)
   {
-      char buf[sizeof(Int64)*2 + sizeof(Int32)*3 + 100];
+      // usage contains 2 int64 and 3 int32, 100 chars is big enough to hold 
+      // the string representation
+      char buf[100];
```
DaveBirdsall: I count 47 for the text ("viewUID: ..." etc.), 20 for each of the int64's and 11 for each of the int32s. That adds up to 47 + 40 + 33 = 120, + 1 for the null terminator. Seems like buf needs to be at least 121. Sorry, I should have made this calculation more explicit when I commented earlier.

DaveBirdsall:
Looks good, except I think one buffer is still too short.

PR #645, opened 2016-08-06

selvaganesang:
The PR 629 for the above JIRA exposed the following issues.

1) A transaction remained in HUNGABORTED state when a table is created and
dropped in the same transaction. core/TEST116 exhibited one such hung aborted
transaction.

2) The retry logic was not throwing any exception when the allowed number
of retry attempts are exhausted.

3) When a UnknownTransactionException is encountered in the response to co-processor
request, the commit was getting into endless loop. privs1/TEST133 was returning error 97
on commit transaction.

All the above issues have been fixed.

[TRAFODION-1988] Better java exception handling in the java/JNI layer of TM

Cleaned up the exception logging in TM further and allowing more events to
be logged via Common Logger facility.

### DIFF
--- a/core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionManager.java
+++ b/core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionManager.java
@@ -205,6 +205,15 @@ public class TransactionManager {
     return g_TransactionManager;
   }
 
+  public static int retry(int retrySleep) {
+     try {
+         Thread.sleep(retrySleep);
+     } catch(InterruptedException ex) {
+          Thread.currentThread().interrupt();
+     }
+     return (retrySleep += TM_SLEEP_INCR);
+  }
+
   /* increment/deincrement for positive value */
   /* This method copied from o.a.h.h.utils.Bytes */
   public static byte [] binaryIncrementPos(byte [] value, long amount) {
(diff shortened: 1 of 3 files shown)


prashanth-vasudev:
Looks good to me. Thanks.

selvaganesang:
Jenkins, extra tests

shangshengtung:
Looks good to me. Thanks.

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionManager.java
```diff
@@ -2288,19 +2229,10 @@ public void doCommitDDL(final TransactionState transactionState) throws Unsucces
...
-                            } catch(InterruptedException ex) {
-                                Thread.currentThread().interrupt();
-                            }
-                            retrySleep += TM_SLEEP_INCR;
-                        }
+                        if (retry) 
+                            retrySleep = retry(retrySleep);
                     }
-                }while (retryCount < RETRY_ATTEMPTS && retry == true);
+                } while (retry == true);
```
trinakrug: Why is this while statement different than the rest?

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionManager.java
```diff
@@ -2327,21 +2259,12 @@ public void doCommitDDL(final TransactionState transactionState) throws Unsucces
...
-                } catch(InterruptedException ex) {
-                    Thread.currentThread().interrupt();
-                }
-                retrySleep += TM_SLEEP_INCR;
-            }
+            if (retry) 
+                retrySleep = retry(retrySleep);
         }
-    }while (retryCount < RETRY_ATTEMPTS && retry == true);
+    } while (retry == true);
```
trinakrug: Why is this while statement different than the rest?

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionManager.java
selvaganesang: Thanks Trina. Good catch. This needs to be fixed.

PR #646, opened 2016-08-08
TRAFODION-2148 fix : create drop request error codes never reach SQL

prashanth-vasudev:
SQL calls create and drop requests to transaction manager. However, sometimes in error scenarios, the error codes never reach SQL. This results in sql assuming the operation is successful in failure conditions.
Fix is to propagate the error all the way back to RMInterface and raise exception.

### DIFF
--- a/core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/RMInterface.java
+++ b/core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/RMInterface.java
@@ -96,10 +96,10 @@ public class RMInterface {
     }
 
     private native void registerRegion(int port, byte[] hostname, long startcode, byte[] regionInfo);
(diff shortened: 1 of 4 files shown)


prashanth-vasudev:
Jenkins, retest

selvaganesang:
Please consider adding the error details and pass it all the way to SQL client so that the client application can get to see what caused the error in DDL statements.

sureshsubbiah:
+1 Change looks good. @prashanth-vasudev should we merge now or do you intend to make changes to address comments above? Those comments could be addressed later if you wish.

prashanth-vasudev:
Thanks for all the comments. I intend to enhance the fix to address the comments. Once ready, I shall do a new PR. Thank You.

selvaganesang:
Has this been fixed. If not, it would be good to create a new JIRA to return the exception to the client and then merge this PR. This PR at least makes sure that the error is returned when there is a failure

prashanth-vasudev:
TRAFODION-2231 is created to track error string apart from error code.

PR #647, opened 2016-08-08
[TRAFODION-2146] Use of CQD to set scratch directory locations

sandhyasun:
Scratchfile creation cleanup and new code to use CQD SCRATCH_DISKS to specify scratch disk locations for overflow. Removed references to NSK volume patterns and validation routines that expected volume/subvolume format. Added new validation code for linux path names.

Cleaned up and removed STFS_LOCATION envvar usage from ms.env and STFS layer.
Added a  new method that can be called to tell STFS to honor the CQD specification over the envvar specification.


core/sqf/src/stfs/stfs_stub.cpp
```diff
@@ -115,6 +115,7 @@ namespace STFS_stub {
   } stfs_directories;
 
   stfs_directories stfsDirsHDD_, stfsDirsSSD_;
+  char scratchCQDLocationString_[MAX_STFS_LOCATIONS*STFS_PATH_MAX - STFS_SUFFIX+1]="";
```
DaveBirdsall: This seems to allow space for just under 32*4096 = 131072 bytes, but I notice the DEFAULTS table limits CQD values to 1000 bytes. Just wondering what the maximum CQD value length at the command line is and whether you'll ever even get close to 131072 here.

core/sqf/src/stfs/stfs_stub.cpp
```diff
@@ -328,63 +332,69 @@ copySTFSLocation(int pv_instnum)
...
-
+  
 }
 
+/// \brief Sets the scratchdisks as specified in the CQD SCRATCH_DISKS 
+///        and avoids looking at envvar
+void STFS_set_scratch_dirs(char *pv_scratchloc) 
+{
+   memset(STFS_stub::scratchCQDLocationString_, 0, sizeof(STFS_stub::scratchCQDLocationString_));
+  strcpy((char *)STFS_stub::scratchCQDLocationString_,pv_scratchloc);
```
DaveBirdsall: strncpy would be safer (notwithstanding that scratchCQDLocationString_ is almost 131072 bytes)

core/sql/generator/GenResources.cpp
```diff
@@ -83,12 +62,13 @@ static ExScratchDiskDrive * genScratchDriveList(const NAString &def,
   // end temporary
 
   const char *str = def.data();
-  if (!*str)
+  if (!str || str=="")
```
DaveBirdsall: Doesn't "str == """ mean, compare the address str to the address of the constant, empty string? If so, then this test is always false. I suspect what you intended to test is, "*str == '\0'"

core/sql/generator/GenResources.cpp
```diff
@@ -97,92 +77,74 @@ static ExScratchDiskDrive * genScratchDriveList(const NAString &def,
   LIST(ExScratchDiskDrive *) tempList;
   CollHeap *heap = generator->wHeap();
   Space *space = generator->getSpace();
+  struct stat st;
+  
+ 
+  Lng32 nodeNum;
+  char *token = new(heap) char[PATH_MAX];
```
DaveBirdsall: Do you want PATH_MAX+1 to allow for a trailing null?

core/sql/generator/GenResources.cpp
```diff
@@ -97,92 +77,74 @@ static ExScratchDiskDrive * genScratchDriveList(const NAString &def,
...
-
+        {
+          tempList.insert(new(heap) ExScratchDiskDrive(
+                               token,
+                               strlen(token) ));
+        }
+      token = strtok(NULL,sep);
     }
+      
+  NADELETEBASIC(token, heap);
```
DaveBirdsall: I don't think this is doing what you want. My understanding of strtok is that it uses a static buffer to store intermediate state. It is *not* using your buffer "token" for this purpose. When you call strtok, you are assigning addresses from that static buffer into the pointer "token". And when you exit the loop, "token" will be NULL, and this NADELETEBASIC will do nothing. So in fact you've leaked the buffer that you allocated earlier. Another point: strtok is not thread-safe. Consider using strtok_r instead.

DaveBirdsall:
Nice to see the MP-isms and NSK-isms cleaned out.

core/sqf/src/stfs/stfs_stub.cpp
```diff
@@ -328,63 +332,69 @@ copySTFSLocation(int pv_instnum)
...
   if (STFS_stub::locationEnvChecked_) {
     copySTFSLocation(pv_instnum);
     return;
   }
-
-  char *lp_HDDEnvLocation = getenv("STFS_HDD_LOCATION");
-  char *lp_SSDEnvLocation = getenv("STFS_SSD_LOCATION");
-  char *lp_EnvLocation    = getenv("STFS_LOCATION");
-
+  if ((strlen(STFS_stub::scratchCQDLocationString_)>0) && (STFS_stub::overflowType_ == STFS_HDD))
```
prashanth-vasudev: In general, in this stfs layer, since we are in linux environment, HDD or SSD need not be distinguished. Both will be specified by simple directory path. STFS layer need not do separate handling for each.  Code can be simplified.

core/sql/generator/GenResources.cpp
sandhyasun: Thanks ! Yes...I'll need to save the original pointer and deallocate.

core/sql/generator/GenResources.cpp
sandhyasun: That's right - basic string comparison - will fix it :)

core/sqf/src/stfs/stfs_stub.cpp
sandhyasun: OK will consider this for a future improvement.  We need to have a discussion prior to this to  see if there may be a case where we could temporarily use SSD scratch locations and need the ability to have both types.

core/sql/generator/GenResources.cpp
sandhyasun: Changed it to not allocate the string at all. token would just point to the static array. Also changed to use strtok_r. Thank you for pointing this out !

DaveBirdsall:
Looks good to me.

PR #649, opened 2016-08-10

selvaganesang:
Changes as per the review comments of PR 645

### DIFF
--- a/core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionManager.java
+++ b/core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionManager.java
@@ -2232,7 +2232,7 @@ public class TransactionManager {
                         if (retry) 
                             retrySleep = retry(retrySleep);
                     }
-                } while (retry == true);
+                } while (retry && retryCount++ <= RETRY_ATTEMPTS);
             }//while
     }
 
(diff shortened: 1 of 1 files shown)


core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionManager.java
```diff
@@ -2232,7 +2232,7 @@ public void doCommitDDL(final TransactionState transactionState) throws Unsucces
                         if (retry) 
                             retrySleep = retry(retrySleep);
                     }
-                } while (retry == true);
+                } while (retry && retryCount++ <= RETRY_ATTEMPTS);
```
DaveBirdsall: The code you have is fine. It could be made a bit simpler though perhaps by using just one variable, retryCount, and counting down instead of up. You could then simply test retryCount > 0. In code paths where you decide to not retry, just set retryCount = 0.

DaveBirdsall:
Looks good to me.

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionManager.java
selvaganesang: Thanks @DaveBirdsall. I will consider making this change when I need to change this file again

PR #650, opened 2016-08-10
[TRAFODION-2034] fix bug that odb will crash when extract file as xml…

lanbb:
… format

### DIFF
--- a/core/conn/odb/src/odb.c
+++ b/core/conn/odb/src/odb.c
@@ -9889,8 +9889,9 @@ static void Oextract(int eid)
             if ( etab[eid].fho ) {
                 (*hdfswrite)(hfs, etab[eid].fho, (void *)xbuff, xbuffl);
             } else {
-                etab[eid].fo = stdout;
-                (void)fwrite ( xbuff, 1, xbuffl, etab[eid].fo );
+                if ( etab[eid].fo == stdout ) {
+                    (void)fwrite ( xbuff, 1, xbuffl, etab[eid].fo );
+                }
             }
 #else
             (void)fwrite ( xbuff, 1, xbuffl, etab[eid].fo );
--- a/core/conn/odb/src/odb.c
+++ b/core/conn/odb/src/odb.c
@@ -9889,6 +9889,7 @@ static void Oextract(int eid)
             if ( etab[eid].fho ) {
                 (*hdfswrite)(hfs, etab[eid].fho, (void *)xbuff, xbuffl);
             } else {
+                etab[eid].fo = stdout;
                 (void)fwrite ( xbuff, 1, xbuffl, etab[eid].fo );
             }
 #else
--- a/core/conn/odb/src/odb.c
+++ b/core/conn/odb/src/odb.c
@@ -13134,7 +13134,7 @@ static int Otcol(int eid, SQLHDBC *Ocn)
                             str[j]= '\0';
                         }
                     } else {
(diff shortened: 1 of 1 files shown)


core/conn/odb/src/odb.c
```diff
@@ -9889,6 +9889,7 @@ static void Oextract(int eid)
             if ( etab[eid].fho ) {
                 (*hdfswrite)(hfs, etab[eid].fho, (void *)xbuff, xbuffl);
             } else {
+                etab[eid].fo = stdout;
```
ryzuo: If the file handler is null, then we may don't need to force to write the message to output. If it has to(the message is really important), a better and professional way to ensure it happen is to initialize the fields of etab when it is allocated. I noticed lots of the fields are pointers, to initialize them requires a lots of work, but it is the safe and professional way.

etab is allocated in etabnew() at line 5974, after the call of realloc and tested no NULL, the next thing should be initialize them all, especially the pointer fields.

@xwq You are more familiar with ODB than I am, any comments?

lanbb:
I noticed in function parseopt(), if the value of char is #, than turn it to \0, I don't understand about it. Could someone tell me why to exclude # as a file sperator. 
If it is designed like this, [TRAFODION-2102] is not a bug, and I will close jira as well as this pull request.
Thanks.

svarnau:
jenkins, add user

sureshsubbiah:
I suppose there are 2 issues being fixed in this PR. One is TRAFODION-2102. I could not find a JIRA related to "fix bug that odb will crash when extract file as xml". Could you please explain what the issue is for the XML extract problem? The title for the PR should be changed too as all comments are being recorded in 2034 which is for a different issue that has already been resolved.

core/conn/odb/src/odb.c
```diff
@@ -13133,7 +13134,7 @@ static int Otcol(int eid, SQLHDBC *Ocn)
                             str[j]= '\0';
                         }
                     } else {
-                        if ( isspace((int)str[j]) || str[j]=='#' )
```
sureshsubbiah: In Pg 53 of the ODB manual
( http://trafodion.incubator.apache.org/docs/odb/Trafodion_odb_User_Guide.pdf ) we have

"A generic mapfile contains:
• Comments (line starting with #)
• Mappings to link input file fields to the corresponding target table columns."

With this change can we still support comments in the mapfile?

In Pg 74, # is also used as the comment indicator for a file that lists the tables to be loaded/extracted. Does this still work?

~/Devel/odb $ cat tlist.txt
# List of tables to extract src=TRAFODION.MAURIZIO.ORDERS
src=TRAFODION.MAURIZIO.CUSTOMER src=TRAFODION.MAURIZIO.PART
src=TRAFODION.MAURIZIO.LINEITEM

lanbb: Thanks for your explanation.
With this change odb will not support comment in the pamfile, so [TRAFODION-2102] is not a bug is not a bug, and I will close that jira.

lanbb:
@sureshsubbiah Thanks very much for your advise。
jira [TRAFODION-2102] odb crashed when extract single table using xml format in linux.
The issue is that when extract file as xml format, odb will return a Segmentation fault in the end.

lanbb:
Because have more than on commit within the same PR, I will close the PR and create a new one.

lanbb:
Because have more than on commit within the same PR, I will close the PR and create a new one.

PR #651, opened 2016-08-10
Trafodion 2140

blfritch-esgyn:


core/sql/optimizer/RelSample.cpp
```diff
@@ -925,7 +925,6 @@ RelExpr *RelSample::bindNode(BindWA *bindWA)
   RelExpr* myChild = child(0);
   if (myChild->getOperatorType() == REL_SCAN &&
       (static_cast<Scan*>(myChild))->isHbaseTable() &&
-      myChild->selectionPred().entries() == 0 &&
```
sureshsubbiah: I would have assumed that we still need this check for a query with a sample clause and a predicate ? Not for update stats, but for some other SELECT statement with SAMPLE. Simple predicates can be pushed down only for regular format rows, for aligned format rows predicate evaluation will be in Trafodion.

sureshsubbiah:
The change looks great. Could you please add some documentation to either the JIRA or this PR. We may need a sub-jIRA to ensure that documentation is updated. This is a wonderful feature to have in Trafodion.

core/sql/ustat/hs_cli.cpp
```diff
@@ -524,12 +531,14 @@ Lng32 HSSample::create(NAString& tblName, NABoolean unpartitioned, NABoolean isP
...
         // If a transaction is running, the table needs to be created as audited.
         // Otherwise, create table as non-audited.
+        /* TEMPTEMP. Dave need to validate this change.
         if (TM->InTransaction())
           tableOptions += " ATTRIBUTE AUDIT";
         else
           tableOptions += " ATTRIBUTE NO AUDIT";
 
         tableOptions += getTempTablePartitionInfo(unpartitioned, isPersSample);
+        */
```
DaveBirdsall: Commenting out the "ATTRIBUTE [NO] AUDIT" stuff is fine. (Indeed it appears that Trafodion does not support non-audited tables. I tried creating one and DMLs on it fail with error 4211.) But the getTempTablePartitionInfo method supplies a SALT clause, which is important to get a nice degree of parallelism when we populate the sample table. So I think we want to keep that.

core/sql/optimizer/RelSample.cpp
blfritch-esgyn: It doesn't really matter where the predicate is applied.  Sampling is just like another predicate ANDed to any existing conjuncts. Whether you sample first or last, the estimated cardinality is the same, and all constraints imposed by the predicate will be satisfied by the selected rows.

sureshsubbiah: Thanks. That is an interesting observation. I had not thought of sampling and predicate evaluation as being commutative. I suppose even if sampling ratio is small and/or predicate selectivity is high, doing these two operations in either order will be statistically indistinguishable. Thanks for explaining.

DaveBirdsall:
It looks good, Barry. Very nice work. I just had the one comment in hs_cli.cpp.

core/sql/ustat/hs_cli.cpp
DaveBirdsall: Hold off on changing this just yet. As it happens a related issue has arisen which I'm investigating. It may be that this change is correct as-is.

blfritch-esgyn: I'm not sure why I moved that over -- I don't think it's related specifically to IUS. Maybe I mistook it for being part of one of your previous changes related to persistent samples.  Anyway, I tried making the change you suggested, and now there are test failures like this:

> *** ERROR[9214] Object TRAFODION.COMPGENERAL_TEST023.TRAF_SAMPLE_nn_nnnnnnnnnn_nnnnnn could not be created.
> 
> *** ERROR[15001] A syntax error occurred at or before:
> CREATE TABLE TRAFODION.COMPGENERAL_TEST023.TRAF_SAMPLE_nn_nnnnnnnnnn_nnnnnn LIK
> E TRAFODION.COMPGENERAL_TEST023.STEST WITH PARTITIONS EXTENT(8192, 16384) MAXEX
>  ^ (zzz characters from start of SQL statement)
> 
> *** ERROR[8822] The statement was not prepared.
> 

I think I'll leave this like it was, since it seems to be a separate issue and may require more investigation.

blfritch-esgyn: ok. I think if any change is needed it should probably be a separate JIRA anyway.

DaveBirdsall: That's fine. Actually, the reason that this code had to be commented out is due to a change I made in https://issues.apache.org/jira/browse/TRAFODION-2044. Given your experiment, the change looks correct to me now.

sureshsubbiah:
Merging. Its a good today since Trafodion now has IUS.
.

PR #652, opened 2016-08-11
[TRAFODION-2154] Add documentation for IF [NOT] EXISTS to CREATE/DROP…

DaveBirdsall:
… SCHEMA

Adds documentation of the IF NOT EXISTS clause for CREATE SCHEMA and IF EXISTS for DROP SCHEMA.


sureshsubbiah:
+1. Looks good Dave

PR #653, opened 2016-08-11
[TRAFODION-2104]odb crashed when extract single table using xml forma…

lanbb:
…t in linux

reason: in function Oextract file handler etab[eid].fo is null and then fwite it, so odb return a Segmentation fault in the end.
solution: judge file handler first, only if it is stdout, do fwrite operation.

### DIFF
--- a/core/conn/odb/src/odb.c
+++ b/core/conn/odb/src/odb.c
@@ -9889,7 +9889,9 @@ static void Oextract(int eid)
             if ( etab[eid].fho ) {
                 (*hdfswrite)(hfs, etab[eid].fho, (void *)xbuff, xbuffl);
             } else {
-                (void)fwrite ( xbuff, 1, xbuffl, etab[eid].fo );
+                if ( etab[eid].fo == stdout ) {
+                    (void)fwrite ( xbuff, 1, xbuffl, etab[eid].fo );
+                }
             }
 #else
             (void)fwrite ( xbuff, 1, xbuffl, etab[eid].fo );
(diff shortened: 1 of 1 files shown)


core/conn/odb/src/odb.c
```diff
@@ -9889,7 +9889,9 @@ static void Oextract(int eid)
             if ( etab[eid].fho ) {
                 (*hdfswrite)(hfs, etab[eid].fho, (void *)xbuff, xbuffl);
             } else {
-                (void)fwrite ( xbuff, 1, xbuffl, etab[eid].fo );
+                if ( etab[eid].fo == stdout ) {
```
sureshsubbiah: If the current problem is that  etab[eid].fo was NULL and we tried to write to it, will it not be better if the change was 
if ( etab[eid].fo ) {
fwrite (...)
}
This way we can handle cases where etab[eid].fo is a generic file handle too.
Maybe the concern is on Line 9882 we have already called fclose on this file handle   (if its not stdout) ?
This change is safe. You could think about whether it could be made more general.

sureshsubbiah:
+1. Thanks for including the JIRA number and explaining the change. It looks good to me. Will wait a day or so for more feedback and to to hear from @lanbb whether any revisions are meaningful.

traflm:
agree with Suresh, look at etabout() , it can be stdout or other descriptor. So Suresh's suggestion is correct.

lanbb:
thanks @sureshsubbiah , I have changed as 
if ( etab[eid].fo ) {
fwrite (...)
}
and it worked, no segment fault returned.

sureshsubbiah:
Merging. The test failure in JDBC is not related to this ODB change.

PR #657, opened 2016-08-13
[TRAFODION-2159] Unnest correlated subquery with explicit groupby

sureshsubbiah:
Add support for unnesting subqueries with a groupby in addition to the scalar aggregate. 
Only 1 level of subquery is current supported.
Also unnesting that requires left join is excluded from this change.

### DIFF
--- a/core/sql/optimizer/NormRelExpr.cpp
+++ b/core/sql/optimizer/NormRelExpr.cpp
@@ -3131,7 +3131,6 @@ GroupByAgg* Join::pullUpGroupByTransformation(NormWA& normWARef)
   // method getMoreOutputsIfPossible().
 
   ValueIdSet leftUniqueCols ;
-  child(0)->castToRelExpr()->synthLogProp() ;
   if (NOT (child(0)->getGroupAttr()->findUniqueCols(leftUniqueCols)))
   {
     // Could not find a set of unique cols.
--- a/core/sql/optimizer/NormRelExpr.cpp
+++ b/core/sql/optimizer/NormRelExpr.cpp
@@ -2449,20 +2449,21 @@ void Join::createAFilterGrandChildIfNeeded(NormWA & normWARef)
 
       if ((doNotUnnest == FALSE) && candidateForLeftJoin && 
           (CmpCommon::getDefault(SUBQUERY_UNNESTING_P2) != DF_INTERNAL) &&
-          ((normWARef.getLeftJoinConversionCount() >= 1)||nestedAggInSubQ))
+          ((normWARef.getLeftJoinConversionCount() >= 2)||nestedAggInSubQ))
       {
         doNotUnnest = TRUE;
-        // For phase 2 we only unnest 1 level of subqueries 
-        // containing NonNullRejecting Predicates
+        // For phase 2 we only unnest 2 subqueries 
+        // containing NonNullRejecting Predicates. Later we will ensure
+	// that these 2 subqueries are not nested.
 
         if (CmpCommon::getDefault(SUBQUERY_UNNESTING) == DF_DEBUG) 
         {
           if (!nestedAggInSubQ)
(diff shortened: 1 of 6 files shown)


core/sql/optimizer/NormRelExpr.cpp
```diff
@@ -2404,8 +2412,13 @@ void Join::createAFilterGrandChildIfNeeded(NormWA & normWARef)
   if (doNotUnnest == FALSE)
   {
     nonLocalPreds.clear();
-    if (oldRightGrandChild->selectionPred().getReferencedPredicates
-                                        (outerReferences, nonLocalPreds))
+    oldRightGrandChild->selectionPred().getReferencedPredicates
+      (outerReferences, nonLocalPreds) ;
+    if (nestedAggInSubQ)
+      subQGby->selectionPred().getReferencedPredicates
```
DaveBirdsall: Suppose the GROUP BY being unnested has both a WHERE clause and a HAVING clause. Do predicates for both get picked up here?

core/sql/optimizer/NormRelExpr.cpp
```diff
@@ -2404,8 +2412,13 @@ void Join::createAFilterGrandChildIfNeeded(NormWA & normWARef)
...
+	(outerReferences, nonLocalPreds);
+
+    if (!nonLocalPreds.isEmpty())
```
DaveBirdsall: Ah, reading the comments below I think answered my question. Looks like HAVING predicates are included. (And maybe WHERE predicates don't matter?)

core/sql/optimizer/NormRelExpr.cpp
```diff
@@ -2436,15 +2449,21 @@ void Join::createAFilterGrandChildIfNeeded(NormWA & normWARef)
...
 
-        if (CmpCommon::getDefault(SUBQUERY_UNNESTING) == DF_DEBUG)
+        if (CmpCommon::getDefault(SUBQUERY_UNNESTING) == DF_DEBUG) 
+        {
+          if (!nestedAggInSubQ)
             *CmpCommon::diags() << DgSqlCode(2997)
               << DgString1("Skipping unnesting of Subquery due to NonNullRejecting Predicates in more than one subquery");
+          else
+             *CmpCommon::diags() << DgSqlCode(2997)
+              << DgString1("Skipping unnesting of Subquery since we have both NonNullRejecting predicate and nested nested aggregate in subquery.");
```
DaveBirdsall: Is it both? Or is just one enough? (also typo: "nested nested")

DaveBirdsall:
I'm a newbie to the transformation area, but I think this looks good to me.

core/sql/optimizer/NormRelExpr.cpp
sureshsubbiah: I think for a groupby by the time we get past the binder the having clause is the same as the where clause. In other words the predicate is kept in selectionPred() of RelExpr base class. It is applied at the end of the work method before a result row is sent to parent. From a SQL syntax point of view this works for HAVING, since now the selectionPred() can reference aggregates evaluated in this GB.

core/sql/optimizer/NormRelExpr.cpp
sureshsubbiah: I was trying to make the debug message more detailed. If it is nestedAgg case we get one message, while if we have several LeftJoin requiring subqueries we get a different message. I suppose if both conditions are true we will get the more than one (now two with new commit) LeftJoin subquery message.

sureshsubbiah: Thanks for catching the typo, it is fixed now.

DaveBirdsall:
Is this ready to merge?

sureshsubbiah:
@DaveBirdsall, I think so. I shall merge it now.

PR #660, opened 2016-08-17
TRAFODION [2137] Improve metadata access time during query compilation

robertamarton:
A change was made to return privilege information in the descriptor structure
instead of getting it when the NATable or NARoutine object is instantiated.
For tables, storing privileges in the descriptor structure allows privileges
to be saved with other table attributes in the metadata.  This improves metadata 
access time during initial query compilations.


core/sql/common/ComSecurityKey.cpp
```diff
@@ -73,6 +74,59 @@ NABoolean qiCheckForInvalidObject (const Int32 numInvalidationKeys,
...
+  {
+    if ( privInfo->hasObjectPriv((PrivType)i) )
+      {
+        ComSecurityKey key(granteeID,
+                           objectUID,
+                           PrivType(i),
+                           ComSecurityKey::OBJECT_IS_OBJECT);
+        if (!key.isValid())
+          return false;
+        secKeySet.insert(key);
```
DaveBirdsall: Do we care if we return false and secKeySet is not empty (because maybe the first few keys were OK)? (I'll find out when I find this method's caller.)

core/sql/generator/Generator.cpp
```diff
@@ -1737,6 +1737,82 @@ TrafDesc * Generator::createConstrKeyColsDescs(Int32 numKeys,
...
+// ComTdbVirtTablePrivInfo 
+//
+// see ComTdb.h for a description of the ComTdbVirtTablePrivInfo
+// see TrafDDLdesc.h for a description of TrafDesc for the priv_desc
+// ****************************************************************************
+TrafDesc * Generator::createPrivDescs( const ComTdbVirtTablePrivInfo * privInfo,
+                                       Space * space)
+{
+  // When authorization is enabled, each object must have at least one grantee
+  // - the system grant to the object owner
```
DaveBirdsall: I take it we never call this routine if authorization is not enabled?

core/sql/generator/Generator.cpp
```diff
@@ -1737,6 +1737,82 @@ TrafDesc * Generator::createConstrKeyColsDescs(Int32 numKeys,
...
+ 
+  TrafDesc * priv_desc = TrafAllocateDDLdesc(DESC_PRIV_TYPE, space);
+  TrafDesc * first_grantee_desc = NULL;
+  TrafDesc * prev_grantee_desc = NULL;
+
+  // generate a TrafPrivGranteeDesc for each grantee and
+  // attach to the privileges descriptor (priv_desc)
+  for (int i = 0; i < privGrantees->size(); i++)
+    {
+      PrivMgrDesc granteeDesc = (*privGrantees)[i];
```
DaveBirdsall: Is there a need to make a copy of the object here? Or would it be enough to make this a reference?

core/sql/generator/Generator.cpp
```diff
@@ -1737,6 +1737,82 @@ TrafDesc * Generator::createConstrKeyColsDescs(Int32 numKeys,
...
+      TrafDesc * bitmap_desc = TrafAllocateDDLdesc(DESC_PRIV_BITMAP_TYPE, space);
+      PrivMgrCoreDesc objDesc = granteeDesc.getTablePrivs();
+      bitmap_desc->privBitmapDesc()->columnOrdinal = -1;
+      bitmap_desc->privBitmapDesc()->privBitmap = objDesc.getPrivBitmap().to_ulong();
+      bitmap_desc->privBitmapDesc()->privWGOBitmap = objDesc.getWgoBitmap().to_ulong();
+      curr_grantee_desc->privGranteeDesc()->objectBitmap = bitmap_desc;
+
+      // generate a list of TrafPrivBitmapDesc, one for each column and
+      // attach it to the TrafPrivGranteeDesc
+      std::vector<PrivMgrCoreDesc> colDescList = granteeDesc.getColumnPrivs();
```
DaveBirdsall: Copy constructor. Should this be a reference instead?

core/sql/common/ComSecurityKey.cpp
DaveBirdsall: I found one caller below. It appears not to care; it is storing this into a member and just leaves it be after a failure. I don't know if that is OK.

core/sql/sqlcat/TrafDDLdesc.h
```diff
@@ -1105,6 +1129,114 @@ class TrafViewDesc : public TrafDesc {
...
+  }
+
+  virtual void populateImageVersionIDArray()
+  {
+    setImageVersionID(0,getClassVersionID());
+  }
+
+  virtual short getClassSize()      { return (short)sizeof(TrafPrivDesc); }
+
+  virtual Long pack(void *space);
```
DaveBirdsall: I know it has nothing to do with your code, but I wonder why we use "Long" for the return from pack, and "Lng32" for the return from unpack.

core/sql/sqlcomp/PrivMgrPrivileges.cpp
```diff
@@ -555,6 +555,90 @@ PrivStatus PrivMgrPrivileges::buildSecurityKeys(
...
+  {
+    PRIVMGR_INTERNAL_ERROR("objectUID is 0 for getPrivRowsForObject()");
+    return STATUS_ERROR;
+  }
+
+  // generate the list of privileges granted to the object and store in class
+  if (generateObjectRowList() == STATUS_ERROR)
+    return STATUS_ERROR;
+  
+  // generate the list of privileges granted to the object and store in class
```
DaveBirdsall: should be "to the columns"

DaveBirdsall:
Looks good to me. All my line comments are tiny nits.

core/sql/common/ComSecurityKey.cpp
robertamarton: There is similar code in Privilege Manager and it handles issues in the same manner.  However, looking at the code, if the key is invalid, then an invalid parameter was sent to the constructor. In this case, we should probably assert. We should not run into this issue today but if we happen to add another DML like privilege, this could fail and not report an error.  I wrote JIRA 2167 to handle errors better here and elsewhere.

core/sql/generator/Generator.cpp
robertamarton: Correct.

core/sql/generator/Generator.cpp
robertamarton: Using a reference would be more efficient.  Will test it out.

core/sql/comexe/ComTdb.h
```diff
@@ -1143,6 +1145,33 @@ class ComTdbVirtTableSequenceInfo : public ComTdbVirtTableBase
...
+ public:
+  ComTdbVirtTablePrivInfo()
+    : ComTdbVirtTableBase()
+    {
+      init();
+    }
+
+  virtual Int32 size() { return sizeof(ComTdbVirtTablePrivInfo);}
+
+  std::vector<PrivMgrDesc>     *privmgr_desc_list;
```
sureshsubbiah: Can we use STL classes in a TDB? I was under the impression that we needed tighter control on memory used for storage. I suppose this vector lives in the global heap?

robertamarton: I am not sure on all the memory rules and did wonder if I should include a vector class here. I could convert the std::vector of PrivMgrDesc to an NAList of PrivMgrDescs if that would be better.

core/sql/sqlcat/TrafDDLdesc.h
robertamarton: pack returns a pointer - 8 bytes, unpack returns the status - 4 bytes.

PR #661, opened 2016-08-17
[TRAFODION-2161] Support migration of user views on Repository upgrade

DaveBirdsall:
Today, users can create views that reference Repository tables. However, when they do so, "initialize trafodion, upgrade" fails because under the covers it does ALTER TABLE RENAME on old Repository tables and the latter does not support views.

This change adds support for such views to "initialize trafodion, upgrade".

The design is as follows:

1. An internal-only option, "SKIP VIEW CHECK", has been added to ALTER TABLE RENAME. This causes any views on the object being renamed to be ignored; that is, they remain in the metadata, referring to the object by its old name. Note that if we abort the upgrade, we'll rename the old repository tables back to their original names (again using "SKIP VIEW CHECK"), so the views are then in their original state.

2. When we determine that upgrade has progressed far enough that we will not turn back, we drop the views but replay their view text to create new views of the same name. If this succeeds, the view has been successfully migrated. But it might not succeed, for example if it depends on a column in the old repository table that has been dropped. In this case, we save the view text in the metadata TEXT table, with TEXT_TYPE = 9 and report that fact to the user. This gives the user a way to take that view text and modify it at their leisure after the upgrade has completed.

While making these changes, I also noticed and fixed a bug in CmpSeabaseDDL::dropOneTableorView where the SQL parser flags global was not being handled correctly.

### DIFF
--- a/core/sql/sqlcomp/CmpSeabaseDDLrepos.cpp
+++ b/core/sql/sqlcomp/CmpSeabaseDDLrepos.cpp
@@ -371,7 +371,7 @@ short CmpSeabaseDDL::migrateReposViews(ExeCliInterface * cliInterface,
     {
       const MDUpgradeInfo &rti = allReposUpgradeInfo[i];
 
-      if ((! rti.newName) || (! rti.oldName) || (NOT rti.upgradeNeeded))
+      if ((! rti.oldName) || (NOT rti.upgradeNeeded))
         continue;
 
       Int64 tableUID = getObjectUID(cliInterface,
--- a/core/sql/sqlcomp/CmpSeabaseDDLrepos.cpp
+++ b/core/sql/sqlcomp/CmpSeabaseDDLrepos.cpp
@@ -290,10 +290,10 @@ short CmpSeabaseDDL::alterRenameRepos(ExeCliInterface * cliInterface,
         continue;
 
       if (newToOld)
-        str_sprintf(queryBuf, "alter table %s.\"%s\".%s rename to %s; ",
+        str_sprintf(queryBuf, "alter table %s.\"%s\".%s rename to %s skip view check; ",
                     getSystemCatalog(), SEABASE_REPOS_SCHEMA, rti.newName, rti.oldName);
       else
-        str_sprintf(queryBuf, "alter table %s.\"%s\".%s rename to %s; ",
+        str_sprintf(queryBuf, "alter table %s.\"%s\".%s rename to %s skip view check; ",
                     getSystemCatalog(), SEABASE_REPOS_SCHEMA, rti.oldName, rti.newName);
         
       if (beginXnIfNotInProgress(cliInterface, xnWasStartedHere))
@@ -356,6 +356,75 @@ short CmpSeabaseDDL::copyOldReposToNew(ExeCliInterface * cliInterface)
     } // for
 
   return 0;
+}
+
+short CmpSeabaseDDL::migrateReposViews(ExeCliInterface * cliInterface,
+                                       NABoolean & someViewSaved /* out */)
+{
+  short retcode = 0;  // assume success
+  someViewSaved = FALSE;
+
+  // for each table that has been migrated, save and drop any views
+  // on the old table, and attempt to recreate them on the new
+
+  for (Int32 i = 0; i < sizeof(allReposUpgradeInfo)/sizeof(MDUpgradeInfo); i++)
+    {
+      const MDUpgradeInfo &rti = allReposUpgradeInfo[i];
+
+      if ((! rti.newName) || (! rti.oldName) || (NOT rti.upgradeNeeded))
+        continue;
+
+      Int64 tableUID = getObjectUID(cliInterface,
+        getSystemCatalog(), SEABASE_REPOS_SCHEMA, rti.oldName,
+        COM_BASE_TABLE_OBJECT_LIT, NULL, NULL, FALSE, FALSE /* ignore error */);
+
+      if (tableUID != -1)  // if we got it
+        {
+          NAList<NAString> viewNameList;
+          NAList<NAString> viewDefnList;
+ 
+          short retcode1 = 
+             saveAndDropUsingViews(tableUID, cliInterface,
+                                   viewNameList /* out */,viewDefnList /* out */);
+          if (retcode1)
+            retcode = -1;  // couldn't get views for old repository table
+          else if (viewDefnList.entries() > 0)  // if there are views to migrate
+            {
+              Lng32 firstBadOne = -1;
+              retcode1 = recreateUsingViews(cliInterface, viewNameList, viewDefnList,
+                                            TRUE, &firstBadOne);
+              if (retcode1)
+                {
+                  retcode = -1; 
+                  Lng32 objUID = -2;  // use a fake objectUID since there is no object
+                  
+                  // For any view that could not be migrated, put its definition text
+                  // back in the TEXT table so the user can deal with it later. We
+                  // use a distinctive TEXT_TYPE for this purpose. We have to "make up"
+                  
+                  NABoolean xnWasStartedHere = FALSE;
+                  if (beginXnIfNotInProgress(cliInterface, xnWasStartedHere))
+                    retcode = -1;
+                  else
+                    {                    
+                      for (Lng32 i = firstBadOne; i < viewDefnList.entries(); i++)
+                        {                      
+                          updateTextTable(cliInterface, objUID, COM_BAD_VIEW_TEXT,
+                                          0, viewDefnList[i], NULL, 0, TRUE);
+                          objUID--;  // get another fake object UID
+
+                          someViewSaved = TRUE; 
+                        }
+                      endXnIfStartedHere(cliInterface, xnWasStartedHere, 0);
+                    }                 
+                }                  
+            }
+        }
+      else
+        retcode = -1;  // couldn't get objectUID for old repository table
+    }
+          
+  return retcode;
 }
 
 short CmpSeabaseDDL::upgradeRepos(ExeCliInterface * cliInterface,
@@ -506,7 +575,7 @@ short CmpSeabaseDDL::upgradeReposComplete(ExeCliInterface * cliInterface,
         {
         case 0:
           {
-            mdui->setMsg("Upgrade Repository: Drop Old Repository");
+            mdui->setMsg("Upgrade Repository: Migrate any views on Repository");
             mdui->subStep()++;
             mdui->setEndStep(FALSE);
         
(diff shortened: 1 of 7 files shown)


core/sql/sqlcomp/CmpSeabaseDDLrepos.cpp
```diff
@@ -358,6 +358,75 @@ short CmpSeabaseDDL::copyOldReposToNew(ExeCliInterface * cliInterface)
...
+
+                          someViewSaved = TRUE; 
+                        }
+                      endXnIfStartedHere(cliInterface, xnWasStartedHere, 0);
+                    }                 
+                }                  
+            }
+        }
+      else
+        retcode = -1;  // couldn't get objectUID for old repository table
```
robertamarton: If there is a view defined on a REPOS table and the REPOS table is being dropped, would table UID be -1 at this time?  That is, we would not need to migrate the view.

core/sql/sqlcomp/CmpSeabaseDDLrepos.cpp
```diff
@@ -516,6 +585,62 @@ short CmpSeabaseDDL::upgradeReposComplete(ExeCliInterface * cliInterface,
...
+            // definitions (which will contain the old table name, not the
+            // renamed table name) and replay that against the new Repository
+            // table. If the replay fails, we treat that as an unrecoverable
+            // situation and ignore it. Instead, we'll save the view definition
+            // text in the TEXT table with a different TEXT_TYPE and clue the
+            // user in that it is there. They can then try to create the view
+            // at their leisure. Note that they may have to change the view
+            // definition themselves, e.g. if we dropped a column from the
+            // repository table that their view happened to reference.
+            //
```
robertamarton: In a future delivery, it might be good to have a feature that returns the view text to the user instead of requiring a read of the TEXT table.    For example, an option to showddl that returns "invalid views". The person fixing up the views may not have SELECT privilege on the system metadata tables.

core/sql/sqlcomp/CmpSeabaseDDLrepos.cpp
DaveBirdsall: Yes, the test at line 374 handles that case. There would be no rti.newName causing us to skip the migration logic for that table. Later, when the table is dropped, the code does a CASCADE so the view disappears.

core/sql/sqlcomp/CmpSeabaseDDLrepos.cpp
DaveBirdsall: Agreed!

core/sql/sqlcomp/CmpSeabaseDDLrepos.cpp
```diff
@@ -516,6 +585,62 @@ short CmpSeabaseDDL::upgradeReposComplete(ExeCliInterface * cliInterface,
...
+          }
+          break;
+
+        case 2:
+          {
+            // This state is here in case we want to report to the user that
+            // there was a view that could not be migrated. But we saved the
+            // view definition for that view in the TEXT table.
+            mdui->setMsg("Some user views could not be migrated. Check metadata TEXT table, TEXT_TYPE = 9.");
+            mdui->subStep()++;
```
robertamarton: From the previous code, it looks like text_uid is -2?   The primary key of TEXT (if memory serves correct) is the text_uid, text_subid, seq_num and text_type.   Will there be conflicts if two or more views are required to be saved both with a UID of -2?  If someone does not cleanup view text before running upgrade for a future release, will there be conflicts?

core/sql/sqlcomp/CmpSeabaseDDLrepos.cpp
```diff
@@ -358,6 +358,75 @@ short CmpSeabaseDDL::copyOldReposToNew(ExeCliInterface * cliInterface)
...
+
+      Int64 tableUID = getObjectUID(cliInterface,
+        getSystemCatalog(), SEABASE_REPOS_SCHEMA, rti.oldName,
+        COM_BASE_TABLE_OBJECT_LIT, NULL, NULL, FALSE, FALSE /* ignore error */);
+
+      if (tableUID != -1)  // if we got it
+        {
+          NAList<NAString> viewNameList;
+          NAList<NAString> viewDefnList;
+
```
robertamarton: You are using existing code - what does the viewDefnList contain?  Is it the view text (like showddl)  or some abbreviated form.  Could this information contain a lot of information?  That is, should it be put into a heap?

core/sql/sqlcomp/CmpSeabaseDDLrepos.cpp
```diff
@@ -516,6 +585,62 @@ short CmpSeabaseDDL::upgradeReposComplete(ExeCliInterface * cliInterface,
...
+            // text in the TEXT table with a different TEXT_TYPE and clue the
+            // user in that it is there. They can then try to create the view
+            // at their leisure. Note that they may have to change the view
+            // definition themselves, e.g. if we dropped a column from the
+            // repository table that their view happened to reference.
+            //
+            // Note that this work is done in one step because the only state
+            // we can depend upon across redrives to this method is the substep
+            // number. Any in-memory list would be lost across redrives.
+            NABoolean someViewSaved = FALSE;
```
robertamarton: If you have a view on two different REPOS tables and both tables are being changed.  Would the view be dropped and recreated twice?  Would they have to be recreated in any specific order?  Along with that, if there is a view that references a view on multiple REPOS tables, are there any issues?  For example, view one creates a view on tables REPOS1 and REPOS2, view two creates a view on view one,  table REPOS3 and a system metadata table.    REPOS1, REPOS2, REPOS3, and the system metadata table are all being upgraded.

core/sql/sqlcomp/CmpSeabaseDDLrepos.cpp
DaveBirdsall: Actually, I slept on this overnight and realized that the test at 374 doesn't do the right thing. The way it works now is a view that references a dropped table will just silently be dropped. But better behavior would be to attempt to recreate the view, let that fail, and let the text be written to the TEXT table (as it would for a dropped column, say). I will submit a rework.

core/sql/sqlcomp/CmpSeabaseDDLrepos.cpp
DaveBirdsall: At line 414, we keep decrementing the value we use for TEXT_UID, so we won't have any key clashes. Though assuming this code does not change in the future, the next time we do a metadata upgrade, we'd start at -2 again. The code will delete the earlier text and write new. This should be OK, I think, because if the user hasn't recreated the view after one release cycle, he probably no longer cares about it.

core/sql/sqlcomp/CmpSeabaseDDLrepos.cpp
DaveBirdsall: The viewDefnList contains the text from the TEXT table. These classes will use their default heap (which I think is global new/delete). When the variables go out of scope the destructors should clean everything up. I see this usage elsewhere so I'm guessing it is OK.

core/sql/sqlcomp/CmpSeabaseDDLrepos.cpp
DaveBirdsall: I didn't unit test this case (and I think I should), but I think the code works correctly in this case. What will happen is that when we process the first repository table referenced, the view will get dropped and recreated. When we get to the second repository table, the view will no longer exist since we already dropped it.

DaveBirdsall:
jenkins, retest

core/sql/sqlcomp/CmpSeabaseDDLrepos.cpp
robertamarton: If just the select text is saved, than any grants performed directly on the view would be lost.  For example, role admin_role is granted select on REPOS view one.  After upgrade, admin_role would no longer have the privilege and any user granted admin_role that runs a query referencing the view one will fail.  For a future upgrade, we may want handle privileges.

DaveBirdsall: Oooohhh... good catch. The DDL code uses this technique in several areas; all of them should be examined for this shortcoming. I will treat that as follow-on work. I have written https://issues.apache.org/jira/browse/TRAFODION-2169 to cover this issue.

core/sql/sqlcomp/CmpSeabaseDDLrepos.cpp
anoopsharma00: Metadata tables store static information about objects. They dont usually
contain status or error information about user action.

Looks like TEXT table is being overloaded to contain metadata related information
as well as error log status.

Failed actions are logged in log files. install may be putting them in error logs.
Another example of logging error information is when failed rows during load are logged into error logs that users can query later.

Upgrade is a platform only operation that is done as part of install/upgrade.
Would it be better to put failed operations in an error log that can be queried
or looked at later.

Maybe at some point a repository error log table can be added that contains  information
about failed operations. And this table is viewable by everyone.

DaveBirdsall:
Taking into account @robertamarton's comments about privileges, and @anoopsharma00's preference for putting the view text in log files rather than the metadata TEXT table, I'm wondering if the right solution short-term is to not attempt view recreation, and instead log all views to a log file. Once the privileges issue is resolved for the drop/recreate technique, we can revisit recreating the views. Does this sound like a better approach?

robertamarton:
It might be good to think about how alter, views, other dependencies, and
privileges all work together.  I have worked previously on a project that
kept invalid views in the metadata and then provided support to fix them
and make them valid.  I don’t have an issue with just storing invalid view
in a log somewhere for now.



     Roberta



*Sent:* Thursday, August 18, 2016 12:30 PM
*Cc:* Roberta Marton <roberta.marton@esgyn.com>; Mention <
mention@noreply.github.com>
*Subject:* Re: [apache/incubator-trafodion] [TRAFODION-2161] Support
migration of user views on Repository upgrade (#661)



Taking into account @robertamarton <https://github.com/robertamarton>'s
comments about privileges, and @anoopsharma00
<https://github.com/anoopsharma00>'s preference for putting the view text
in log files rather than the metadata TEXT table, I'm wondering if the
right solution short-term is to not attempt view recreation, and instead
log all views to a log file. Once the privileges issue is resolved for the
drop/recreate technique, we can revisit recreating the views. Does this
sound like a better approach?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/apache/incubator-trafodion/pull/661#issuecomment-240830895>,
or mute the thread

DaveBirdsall:
Please do *not* merge this pull request. I am in the midst of redesign. I will later close this request and submit a new one.

DaveBirdsall:
Closing this pull request in favor of a different solution, submitted separately.

PR #664, opened 2016-08-19
TRAFODION-2163 support MYSQL style function GROUP_CONCAT

traflm:
add new SQL syntax for pivot_group() so to allow same MySQL syntax.
Also fix original pivot_group() issue and tested more.
But still protected by CQD mode_special_4, since we still need more tests. So by add this MySQL syntax, we can test more real cases simpler, without modify the query to use pivot_group()

The original pivot_group() has issue when rows are not sorted, its internal pointers to track each bucket's len and pos is shared. One simple fix is to disallow hash groupby for pivot_group, which used in this PR.
A more complex and final fix will be to use bucket's own var for len and pos, but that needs much more modification to current groupby which never need per bucket vars, and that is only for pivot_group. So I would like to fix that later, and also need to implement the sort direction support. 

But this will make some real case to be supported and will simplify further tests, and can already support most commonly used CONCAT_GROUP scenarios we saw previously.


traflm:
jenkins, extra tests

traflm:
jenkins, extra tests

traflm:
Extra Test failed at HIVE 003 and EXECUTOR 022, both are temp issues which can be seen recently from time to time. Not related to this PR.

traflm:
@anoopsharma00 could you help to comment on this change?
This is an initial support of CONCAT_GROUP syntax, and also a fix to an existing privot_group() issue. Next, I will continue to find a better way to use hash_groupby. This is more about a syntax support.

anoopsharma00:
Looks good.  

A separate effort is ongoing to externalize features that are currently within mode_special_4.
At some point, we should enable both pivot and group_concat for general use as well.

traflm:
thanks Anoop!
I plan to modify the pivot eval() to support hash group by next, and do more tests. We do need to externalize features.

PR #665, opened 2016-08-19
TRAFODION [2171] 'get users' returned invalid character error when trying to ...

robertamarton:
'get users' returned invalid character error on centos7.1 and centos7.2,
running the traf_authentication_setup script that enables security.
The problem occurred while referencing a pointer to a local variable that has 
gone out of scope.  Fixed 'get users' to work the same as 'get roles'.


arvind-narain:
Looks good to me Roberta

PR #671, opened 2016-08-24
[TRAFODION-2118] dynamic add coprocessor using Hbase api

mashengchen:
1. mv coprocessor from org.apache.hbase.coprocessor.transactional to org.apache.trafodion.coprocessor, beacuse org.apache.hbase.coprocessor is build-in namespace and the coprocessors under this namespace will be block.
2. hbase-trx-*.jar will never put under $HBASE_HOME/lib , but put into HDFS which is under /hbase/lib. this will help to dynamic add coprocessor.
3. no need to change hbase-site.xml to add hbase.coprocessor.region.classes in the future
4. change related install scripts.


DaveBirdsall:
jenkins, retest

mashengchen:
seems not compatible with version hbase1.2, i will add some condition to judgment it.

PR #672, opened 2016-08-25
[TRAFODION-2170] create table failure when pk length is too long

traflm:
If the length of the primary key greater than 2048, Trafodion will have memory overrun issue and core dump.

The changed method will alloc a 2048 buffer and do java memory copy later into this buffer, which cause memory overrun.

This is rare, but still need to be fixed.

### DIFF
--- a/core/sqf/src/tm/tmddlrequests.cpp
+++ b/core/sqf/src/tm/tmddlrequests.cpp
@@ -24,6 +24,7 @@
 #include "dtm/tm.h"
 #include <string.h>
 #include <iostream>
+#include <stdlib.h>
 
 using namespace std;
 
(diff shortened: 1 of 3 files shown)


core/sqf/src/tm/tmddlrequests.cpp
```diff
@@ -39,13 +40,24 @@ JNIEXPORT void JNICALL Java_org_apache_hadoop_hbase_client_transactional_RMInter
    char la_tbldesc[TM_MAX_DDLREQUEST_STRING];
    char la_tblname[TM_MAX_DDLREQUEST_STRING];
    char* str_key;
-   str_key = new char[TM_MAX_DDLREQUEST_STRING];
+   if(pv_keyLength <= 0) 
+   {
+     cout << "createTableReq bad input pv_keyLength, abort" << endl;
+     abort();
+   }
+   str_key = new char[pv_keyLength];
```
prashanth-vasudev: tmlibmsg.h defines the message types and length. Please check to make sure pv_keyLength does not exceed the defined message types.

traflm: thanks Prashanth, yes, I should add this check. Will fix it and submit again.

traflm: @prashanth-vasudev , I realize I have to wait for TRAFODION-2148 check in first to do length checking. Currently, even the length is too long, there is no way to return error and abort the changes done before.
So I will wait for 2148 to be merged first.

traflm:
Will rework after TRAFODION-2148, close now

PR #673, opened 2016-08-25
[TRAFODION-2161] Fix issue with upgrade failure when Repository views…

DaveBirdsall:
… exist

Before this change, "initialize trafodion, upgrade" would fail if a view existed that referenced a Repository table. With this change, "initialize trafodion, upgrade" will succeed. It will automatically drop any view on a Repository table that needs upgrading. The view text for the view definition is captured in the log file for that session. Also included are design notes on how to improve this to support automatic migration of such views in the future.

Also, while developing this change, a bug in parser flags handling was noticed in sqlcomp/CmpSeabaseDDLcommon.cpp. That bug has been fixed.

### DIFF
--- a/core/sql/parser/StmtDDLAlterTableRename.h
+++ b/core/sql/parser/StmtDDLAlterTableRename.h
@@ -65,10 +65,12 @@ public:
 
   // constructor
   StmtDDLAlterTableRename(const NAString & newName,
-                          ComBoolean isCascade)
+                          ComBoolean isCascade,
+                          ComBoolean skipViewCheck)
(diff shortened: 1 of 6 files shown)


DaveBirdsall:
Trafodion community: I'll be on vacation the next couple of weeks. If there are no major concerns (that is, no regression failures and no reviewer comments pointing to serious issues), any committer can feel free to commit this in my absence. Otherwise, I'll respond to issues with reworks when I return.

DaveBirdsall:
jenkins, retest

sureshsubbiah:
+1. Looks good.

PR #675, opened 2016-08-26
Trafodion 2034 Adding extra lines to DCS sever file 

amandamoran:


sureshsubbiah:
Change looks great.
JIRA id seems to be incorrect in title of PR. I take it that 2191 is the correct JIRA id?

PR #676, opened 2016-08-26
[TRAFODION-2135] Add protobuf-2.5 to build tools

svarnau:
This adds protobuf to traf_tools_setup.sh, which is a script for setting
up build tools for trafodion developers. It skips building protobuf in the case
the right version is already in the path. That would likely be the case for
RH7/Centos7 systems where protobuf2.5 is available in the linux distro.


sureshsubbiah:
+1. Looks good Steve.

PR #677, opened 2016-08-28
Fix to prevent transaction and table hang when trying to close a region

sbroeder:


sbroeder:
This fix closes JIRA [TRAFODION-2193]

shangshengtung:
Looks good.

sureshsubbiah:
Merged.

PR #678, opened 2016-08-30
[TRAFODION-2190] Problems with CQS and TMUDFs

zellerh:
Control query shape with a Table-mapping UDF caused a core dump.
More generally, there was no syntax for forcing a shape with a
TMUDF. Fixed the core dump and also added a "tmudf" keyword to
CQS. tmudf can be used as a leaf and as a unary operator in CQS
(binary operators may be supported in the future).


sureshsubbiah:
+1. Looks good Hans.  2190 is related to 2147. I have assigned 2147 to you now. Merging.

PR #680, opened 2016-08-30
[TRAFODION-2199] SELECT on Hive partitioned table causes timestamp mismatch error The problem is caused by reading hive-staging directory at runtime.

sureshsubbiah:


sandhyasun:
Change looks good . Need to wait for tests to complete before merging.

PR #681, opened 2016-08-30
[TRAFODION-2147] showshape does not work for TMUDFs

zellerh:
The JIRA talks about both showshape and CQS. The CQS part
was fixed with TRAFODION-2190, this is the fix for showshape.


traflm:
looks good to me.

sureshsubbiah:
+1. Merging

PR #683, opened 2016-09-01
[TRAFODION-2329]New SPJ to create intermediate table for connectby

orrk:
example of connectby
call connectby('nodesn','select \* from "signal".TB_MOMP_BUSINESSINFO where %s','select parentnodesn,nodesn from "signal".tb_momp_businessinfo where parentnodesn=''%s''','XX');


venkat1m:
What is the use case for this SPJ? Can you add the JIRA ID in the comments.

venkat1m:
This particular source file is for the system SPJs for library management. If these SPJs are generic SPJs, they should be moved to a separate file/package and handled separately.

orrk:
Hi this is a workaround to support "start with" in oracle, Kevin will review this.

svarnau:
jenkins, add user

svarnau:
please edit title of pull-request to add jira at beginning: [TRAFODION-XXXX]

core/sql/lib_mgmt/src/main/java/org/trafodion/libmgmt/FileMgmt.java
```diff
@@ -547,6 +558,64 @@ private static void checkFile(String fname, int dataSize) throws SQLException {
...
+		try{
+			ids.addAll(bfsSearch(sql,startWithKey,searchLevel,ps));
+			out[0]= Arrays.toString(ids.toArray());
+		} catch(SQLException e){
+			LOG.error(sql,e);
+			throw e;
+		} 
+	}
+	
+	private static ArrayList<String> bfsSearch(String sql, String startWithKey,int searchLevel, PreparedStatement ps) throws SQLException{
```
zellerh: Now I understand the comment people are making about the lack of parallel execution for this approach.

This is a recursive method that returns an array of strings I see several disadvantages with this approach:

1. We process every key serially, instead of using a parallel query for all keys of a given level.
2. We may get very many strings back, millions, maybe hundreds of millions. This will not scale.
3. Having to convert the start with key to a string is not ideal, what if we have multiple columns?
4. It's a recursive call and therefore could overflow the stack.
5. Cycle detection would require searching the string arrays.

Here would be another solution:

1. Create a volatile table to hold the ids (parent, child, level). Note: This step may be a problem, I remember that there are issues with creating volatile tables in SPJs, but maybe we can do this in the master executor. If we need to produce an error if we encounter a cycle, make the "parent" column(s) a unique key.
2. Delete any existing data from the volatile table.
3. Process the "start with" query by inserting the first set of data into the volatile table.
4. In a loop, populate the volatile table with the data for the subsequent levels.
5. The loop ends when the insert statement into the volatile table inserts 0 rows, or when we exceed the maximum number of levels.
6. The caller of the SPJ can now retrieve the result in the volatile table.

zellerh:
@venkat1m, I propose a single directory, a single Maven project, and a single jar for all the UDRs (SPJs and UDFs) that we ship with the Trafodion SQL engine.

When we added the library management SPJs, I was already concerned about having yet another Maven project file. I would really prefer if we could put all the SPJs built into Trafodion into a single Maven project, with a single jar. 

Note also that we need to move at least two more UDRs into that jar, EVENT_LOG_READER and JDBC. This is for security reasons.

zellerh:
We can rename "libmgmt" and use this directory for all the UDRs.

zellerh:
Sorry, I misunderstood Venkat. Sure, we can use another source file and maybe package, my comment is only about keeping it in the same jar and Maven project.

venkat1m:
I agree Hans. We can rename this libmgmt.jar into something generic like systemspjs.jar and have all java code files for the system SPJs be part of this single jar file. And abstract each group of spjs into their own class and package names.

core/sql/lib_mgmt/src/main/java/org/trafodion/libmgmt/FileMgmt.java
traflm: @zellerh , the SPJ can now create volatile table, that bug is already fixed recently in T2 driver.

Agree with Han's opinion. using a table to hold intermediate result is more generic. And the current solution is still serial, or did I misunderstand? 

After getting all the start seeds, it should be able to spawn multiple threads to do it in parallel, but using volatile table and using SQL's parallel infrastructure seems a better approach.

It is not easy to write a generic one, but this will be a very useful SPJ, so thanks for the effort.

orrk:
The spj has been updated to create an intermediate table for connectby

General steps:
There is an inner class Tuple in FileMgmt.java to instead of using arraylist in code.
1. Everytime running this SPJ it will drop old intermediate table(if exist) and then create new one.
2. SPJ get all nodes from source table and create an arraylist of tuples and convert it to adjacent table.
3. Based on the adjacent table we can using nested loops to get all paths ( so it will take long time if the source table is big)

usage
call qa_jdbc.connectbytable('intermediate_tbl','source_tbl','col_name','parent_col_name');
For instance(left is child and right is parent)
{ "0", "-1" },{ "1", "0" }, { "2", "0" }, { "3", "1" }, { "4", "1" },{ "5", "2" }, { "6", "2" }, { "7", "3" }, { "8", "3" }, { "9", "5" }, { "10", "3" }
the result is 
[(0,2), (0,6), (2,6), (0,5), (0,9), (2,5), (2,9), (5,9), (0,1), (0,4), (1,4), (0,3), (0,8), (1,3), (1,8), (3,8), (0,7), (1,7), (3,7), (0,10), (1,10), (3,10)]

traflm:
This looks good to me. +1
Can we merge it now? @venkat1m

zellerh:
-1

This is a very specialized solution to the connect by problem and I don't think we should put such a narrowly defined SPJ into Trafodion, where people would expect a more general solution.

Please understand that the SPJ shown here is good, it solves a real problem. I just feel that we need to have a more general solution as a built-in function in Trafodion. It would be good if others could comment on that as well.

Here are some of the limitations I think we should try to avoid:

- The column structure for recursing is fixed, only a single VARCHAR(100) not null column is supported. Many use cases will need different data types and/or multiple columns.
- The "root query" is hard-coded, in many cases a different root query will be required. Example: List all the employees of a manager who is not at the root of the hierarchy.
- As far as I can tell, the SPJ discards all the root nodes except the last.
- The SPJ stores an in-memory table instead of using SQL. This is not scalable and does not use parallelism in the way a user would expect of Trafodion.
- The SPJ does not support a maximum level of recursion, cycle detection, or the ability to record the recursion level or other aggregates.
- There are no comments in the SPJ, making it hard to read.

What about a much more general SPJ that takes the following arguments:

- The name of a table, partially populated with an initial set of data.
- An SQL query that iterates over this table, computing one level of recursion. The query has an optional input parameter ('%1$s'), which is the level.
- The maximum number of levels to iterate.
- An optional query that executes a test whether iteration is complete. If not specified, iteration stops when the maximum number of levels is reached, an error occurs, or when the iteration inserts no new records.
- An optional SQL query that is executed after the iteration is complete. If not specified, the caller of the SPJ can issue their own query to retrieve the results.

Such an SPJ would avoid all the problems I listed above. It could scale to very large data sets and it could use parallelism.

zellerh:
What about putting this SPJ into the [Trafodion Code Examples](https://cwiki.apache.org/confluence/display/TRAFODION/Trafodion+Code+Examples) wiki so people can use and modify it?

traflm:
It will be a good choice to put this code into the Wiki. I agree with Hans.

PR #684, opened 2016-09-01
[TRAFODION-1676] rework, fix executor regression test022 failures

traflm:
rework of the JIRA, in previous code, it get display string length using strlen(), but the srcData is not null-terminated for sure, so it will randomly get some unknown content.
And also refactor the previous code for better coding rule.
This should address the sql regression failure of executor/TEST022.

### DIFF
--- a/core/sql/exp/exp_datetime.cpp
+++ b/core/sql/exp/exp_datetime.cpp
@@ -99,10 +99,10 @@ static void
 raiseDateConvErrorWithSrcData(int srcLen, ComDiagsArea** diagsArea, char *srcData, CollHeap *heap)
 {
     char errstr[MAX_OFFENDING_SOURCE_DATA_DISPLAY_LEN];
-    memset(errstr, 0 , sizeof(errstr) );
+    str_pad(errstr, sizeof(errstr), 0 );
     if(srcLen > MAX_OFFENDING_SOURCE_DATA_DISPLAY_LEN -1 )
       srcLen = MAX_OFFENDING_SOURCE_DATA_DISPLAY_LEN -1;
-    strncpy(errstr, srcData, srcLen);
+    str_cpy_all(errstr, srcData, srcLen);
     ExRaiseSqlError(heap, diagsArea, EXE_CONVERT_DATETIME_ERROR,NULL,NULL,NULL,NULL,errstr);
 }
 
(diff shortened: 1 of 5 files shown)


core/sql/exp/exp_datetime.cpp
```diff
@@ -92,6 +92,30 @@ copyDatetimeFields(rec_datetime_field startField,
...
+}
+
+// Helper function to format extra error message text and report error
+// 
+static void 
+raiseDateConvErrorWithSrcDataNumeric(ComDiagsArea** diagsArea, long srcData, CollHeap *heap)
+{
+   char errstr[MAX_OFFENDING_SOURCE_DATA_DISPLAY_LEN];
+   memset(errstr, 0 , sizeof(errstr) );
+   sprintf(errstr,"%ld",srcData);
```
zellerh: Minor comment: I try not to use sprintf anymore and use snprintf instead. Some static analysis tools flag this type of code as a potential buffer overrun, even though the length may be long enough.

zellerh:
Looks good to me, thanks for deleting so much unnecessary code!

core/sql/exp/exp_datetime.cpp
anoopsharma00: we should also try to use the str_* methods from common/str.h, if possible.
At one point, use of local str_* methods was mandatory and code wouldn't compile
without it, but over time we have started to use of native C string functions.

In the code listed above, the C  functions will be replaced by these str_* methods: 
memset  ---> str_pad
strncpy   ---> str_cpy_all
sprintf    --->  str_sprintf

str_* methods may internally call native C functions but it provides a
common entry point for them.

traflm: thanks Hans, Anoop, this is very good guidance, I will use these functions instead of C functions now and in the future.
In fact, there are a lot of Trafodion's own underlying libraries, such as C-style string manipulation like above, and C++-style lib like NAString...
However, there is no description about them, so it is rather difficult for new Trafodion engineer to use them, I will track down them as much as I learned during the process, and add into wiki page, people can enhance it one by one.
Or maybe there is such a reference that I missed.

traflm:
there are multiple regression test failure and coding improvement, let me submit another commit.

traflm:
jenkins, extra tests

traflm:
Extra test failed at HIVE003:
DIFF
192,195d191
< *** WARNING[8597] Statement was automatically retried 1 time(s). Delay before each retry was 0 seconds. See next entry for the error that caused this retry.
< 
< *** WARNING[8436] Mismatch detected between compiletime and runtime hive table definitions. DataModMismatchDetails: removed
< 
216a213,216
> *** WARNING[8597] Statement was automatically retried 1 time(s). Delay before each retry was 0 seconds. See next entry for the error that caused this retry.
> 
> *** WARNING[8436] Mismatch detected between compiletime and runtime hive table definitions. DataModMismatchDetails: removed
> 

So it is not related to this PR. Another known issue. 

So I think this PR is clean and should be ready to be merged if no other comments.

PR #686, opened 2016-09-01
Additional changes to JIRA TRAFODION-2060 plus couple other fixes

anoopsharma00:
-- interval computations involving tinyint are fixed.
-- hive data mod mismatch error now prints the compile and runtime
   timestamps that caused the mismatch.
-- hive test003 updated to truncate data before insert


sandhyasun:
Changes look fine.

PR #689, opened 2016-09-08
JIRA TRAFODION-2200 Vol/Ext opers with hive schema default return errors

anoopsharma00:
-- JIRA TRAFODION-2210 some operations on volatile and external tables
  no longer return errors if catalog/schema default is set to hive.hive
-- updated compGeneral/test042 regressions expected/known diff file
-- redef time for views was not being updated correctly. That has
   been fixed
-- removed an unneeded printf stmt from native expr optimizer


sureshsubbiah:
+1. Change looks good. JIRA is 2210 and not 2200. Changing the title will help ensure that appropriate JIRA is updated with this PR. Can merge tomorrow if there are no more comments.

PR #690, opened 2016-09-08

selvaganesang:
Avoided the caching of Hbase table instances on the client side to avoid
compaction queue size build up while constantly loading the data


sureshsubbiah:
+1. Looks good Selva.

selvaganesang:
This change removes HTable/Table caching on the client side of the Trafodion code.  The code to add coprocessor classes at the time of creating HTable instance is removed because it would be drag to instantiate HTable/Table for every hbase operation.  It is a non-issue with Trafodion tables because the TableDescription should have added coprocessor classes automatically at the time of create.  The vanilla hbase table access via Trafodion engine will be non-transactional unless the Trafodion co-processor classes are added to the HTableDescriptor via hbase shell.

For the existing customers who had created Trafodion tables in the very early release of Trafodion also need to add coprocessor classes via hbase shell.

sbroeder:
Looks good to me

prashanth-vasudev:
looks good.

PR #695, opened 2016-09-11
automatic creation of schemas

anoopsharma00:
couple changes:
1) if dev regressions are being run and regr schema ("trafodion.sch")
  doesn't exist, then it is automatically created as part
  of the first 'create' stmt. This schema is also updated
  as the default schema in DEFAULTS metadata table.
2) if cqd traf_auto_create_schema is set and the schema specified
  in a 'create' stmt doesn't exist, then it is automatically
  created. This helps with migrating scripts from other vendors
  if schema is specified as part of create table stmt.


sureshsubbiah:
TRAFODION-2215 created to track this issue. 
+1.
Please remember to edit PR title to include JIRA number.

PR #697, opened 2016-09-13
[TRAFODION-2218] Memory leak from JVM used for TMUDF compile time interaction

sureshsubbiah:
DeleteLocalRef added. ReturnInfo is no longer static. Please see JIRA for an explanation.

### DIFF
--- a/core/sql/langman/LmRoutineJavaObj.cpp
+++ b/core/sql/langman/LmRoutineJavaObj.cpp
@@ -192,6 +192,7 @@ LmResult LmRoutineJavaObj::invokeRoutineMethod(
           (*emitRowPtr_)(NULL,0,&qs);
         }
     }
+  jni->DeleteLocalRef(jniResult);
 
   if (invocationInfoLenOut)
     *invocationInfoLenOut = static_cast<Int32>(iiLen_);
--- a/core/sql/src/main/java/org/trafodion/sql/udr/LmUDRObjMethodInvoke.java
+++ b/core/sql/src/main/java/org/trafodion/sql/udr/LmUDRObjMethodInvoke.java
@@ -68,7 +68,7 @@ public class LmUDRObjMethodInvoke
         return result;
     }
 
-    public static class ReturnInfo
+    public class ReturnInfo
     {
         // return status:
         // <0: Internal error, check for Java exception
--- a/core/sql/src/main/java/org/trafodion/sql/udr/LmUDRObjMethodInvoke.java
+++ b/core/sql/src/main/java/org/trafodion/sql/udr/LmUDRObjMethodInvoke.java
@@ -68,7 +68,7 @@ public class LmUDRObjMethodInvoke
         return result;
     }
 
-    public class ReturnInfo
+    public static class ReturnInfo
     {
         // return status:
         // <0: Internal error, check for Java exception


selvaganesang:
Looks good. Have you observed the growth in C++ memory when run for a long time.

core/sql/langman/LmRoutineJavaObj.cpp
```diff
@@ -192,6 +192,7 @@ LmResult LmRoutineJavaObj::invokeRoutineMethod(
           (*emitRowPtr_)(NULL,0,&qs);
         }
     }
+  jni->DeleteLocalRef(jniResult);
```
DaveBirdsall: It looks like there is a code path where jniResult could be null. Is DeleteLocalRef a no-op in this case?

sureshsubbiah:
For the comment by @selvaganesang : I have seen this issue in 2 environments. In one case there was a growth in mem on the C++ side too. RSS memory as reported by TOP grew to 3 GB for mxosrvr. I have a gcore and will analyze that next. In my test env, I did not observe a similar growth on C++ side, but that could be because I did not run the query sufficient number of times. I will try it out, and if additional issues are found, I hope it can be addressed in a different PR. Thank you for the review and help in understanding this problem.

For the comment by @DaveBirdsall : This used be a bug in gcc, which has been fixed in the version of gcc we use. Please see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=28178 . I also tested by explicitly adding this line to the method that was changed. There was no untoward behaviour. Thank you for the review, I had the same question as I was adding  that line, but went on without doing these checks.
jni->DeleteLocalRef(jniResult);
jni->DeleteLocalRef(NULL); // new temp line for sanity check. Not part of this PR

DaveBirdsall:
Looks good to me.

core/sql/src/main/java/org/trafodion/sql/udr/LmUDRObjMethodInvoke.java
```diff
@@ -68,7 +68,7 @@ static LmUDRObjMethodInvoke makeNewObj(
         return result;
     }
 
-    public static class ReturnInfo
+    public class ReturnInfo
```
zellerh: Just curious, I don't see a code change that makes this necessary, why are we changing the static class to an inner class? Wouldn't the static class be preferable, especially given that we pass it - without an object from the parent class - via JNI? My knowledge about this difference comes from https://docs.oracle.com/javase/tutorial/java/javaOO/nested.html, am not a real expert in Java nested classes.

core/sql/langman/LmRoutineJavaObj.cpp
zellerh: First, sorry for having caused this memory leak and thank you for fixing it!

One question: The code saves data members of the jniResult object in returnedIIL_ and returnedPIL_. Is it legal to destroy the local ref to the Java ReturnInfo object and to keep using local refs to the strings and byte arrays stored in it? If it's not, or if it's questionable, then we could save this local ref in the LmRoutineJavaObj object and clean it up on the next call or in the destructor, like we do it for returnedIIL_. If you like I'd be happy to code this and send you a patch.

selvaganesang: I think, unless  jniResult object reference is retained in the java space or it is made as global reference in JNI, JVM wouldn't guarantee the correctness of jniResult object across two different invocations of the JNI method.

zellerh: @selvaganesang, we are not using the local ref across two invocations. We are using it, however, after returning from this method, before we invoke the method again.

selvaganesang: That can cause problems too. If there are no references retained, it can be gc-ed by JVM leading to unknown results when these variables are accessed. Is this reason this class was made static.  If so, there is another issue that this is not thread safe.

core/sql/src/main/java/org/trafodion/sql/udr/LmUDRObjMethodInvoke.java
sureshsubbiah: @zellerh, I agree that this change is not strictly necessary for this JIRA. I have tried to explain my motivation for making it in the comment section of https://issues.apache.org/jira/browse/TRAFODION-2218

The way I see it here are the pros and cons of Return being a static class
Pro
1) Can be instantiated without an instance of the outer class. Having it be a nested class provides better encapsulation.

Cons
1) It was not clear to me when I made the change when the static class will be garbage collected. Now I read this link and see that I do not need to be concerned with GC. http://stackoverflow.com/questions/17553218/when-is-static-nested-class-gets-garbage-collected . Static variables could get in the way of GC I suppose? So this is not really a con.

2) The static class is not accessing any static data members of the enclosing class. The static class is also not instantiated outside the enclosing class. There seems to be no programmatic rule broken by having the class be non-static.

I can roll this change back if that is preferable. My main concern with GC has been shown to be a non-issue. At this point I suppose use of static modifier here is programming preference.

zellerh: Yes, what you say in your last sentence is what I'm thinking as well. If a nested class doesn't need an association with an object of its parent class then it is preferable to make it static. In this case, I feel that passing a ReturnInfo object across JNI is a bit strange with an inner class, as we would separate the two related objects.

core/sql/langman/LmRoutineJavaObj.cpp
sureshsubbiah: Sorry for not noticing this. It does seem dangerous to use data members from deleted objects. I will work on coding a fix and push another commit for this PR. Previously we know for certain that the ReturnInfo object on the Java side was never GCed. It was "live" always. So we were not exposed to this problem.

zellerh: The idea of this code was to have two C++ calls that need to be made directly one after another:

invokeRoutineMethod() // makes JNI call, returns length of output byte strings
getRoutineInvocationInfo() // retrieves the output byte strings

The local refs are created by the first call and used by the second. They are cleaned up either by the next call to invokeRoutineMethod() or by the destructor.

You are right that if a caller does other JNI calls between those two calls they might get stale objects. I can rewrite this and add an extra copy of the objects. Right now this is not a problem, but we can change this if you prefer. Alternatively I can add comments to make it clear these calls have to follow each other, without other JNI calls in-between.

sureshsubbiah: I have been thinking about this more and talked to @zellerh.

From JNI specs http://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/functions.html#global_local
we have
"Local References
**Local references are valid for the duration of a native method call. They are freed automatically after the native method returns.** Each local reference costs some amount of Java Virtual Machine resource. Programmers need to make sure that native methods do not excessively allocate local references. Although local references are automatically freed after the native method returns to Java, excessive allocation of local references may cause the VM to run out of memory during the execution of a native method."
However in Trafodion's use of JNI, we know by debugging this JIRA that localRefs are never freed, unless they are explicitly deleted. This must be because the "native" method never returns, as the C++ side inits the JVM. Therefore in this use we delete the local Ref to ReturnInfo, but we have new local refs to returnedIIL and returnedPIL. Those are deleted later after they are used (when this method called next or in the destructor). Those 2 byte arrays will not be GCed right after local ref to ReturnInfo is deleted, since there is one other local ref pointing to each one of them. They will be GCed only when all localRefs pointing to them (including parent class) are deleted.

zellerh: Thanks, @sureshsubbiah, that makes sense to me. My conservative rule is the following: Assume that local refs must be explicitly deleted, don't assume that they are valid after the next (C++ to Java) JNI call.

zellerh:
Change looks good to me. Sorry for the long discussion.

core/sql/langman/LmRoutineJavaObj.cpp
DaveBirdsall: Wait a minute... @sureshsubbiah, I think the spec you're referring to is the case where Java code calls a C++ routine, but in this particular fix scenario, we are doing the opposite, going from C++ to Java. So I'm not sure that the comment about "Local references are valid for the duration of a native method call. They are freed automatically after the native method returns" applies here.

zellerh: @DaveBirdsall, I think the problem is that the text quoted by Suresh is the only place where it talks about local references. Therefore we are left to guess what happens in the case where there is no such return.

PR #698, opened 2016-09-13
[TRAFODION-2187] Fix DROP SCHEMA CASCADE when sample tables are present

DaveBirdsall:
The problem was that when a sample table existed, the DROP SCHEMA code would try dropping it but it had already been dropped when its associated user table was dropped.

The fix is to skip over sample tables when dropping user tables, as the sample tables are dropped automatically when their associated user tables are dropped via DROP TABLE.


zellerh:
Looks good to me.

selvaganesang:
Looks good to me.  I would think we need to document that the tables created by user shouldn't start with 'TRAF_SAMPLE_".

DaveBirdsall:
@selvaganesang, thanks for the suggestion. I will follow up on it when addressing  JIRA TRAFODION-2164.

PR #699, opened 2016-09-13
[TRAFODION-2219] charsets regression failures and retrieved columns

zellerh:
Fixed a problem with occasional failures of regressions, due to
non-deterministic order of results.

Also changed the way we show the number of "retrieved columns" for
Hive and ORC scans. The EXPLAIN now only shows the number of
actually needed columns, not all of them.


zellerh:
The jdbc test failure with an internal error is unrelated to the changes.

PR #700, opened 2016-09-14
jira TRAFODION-2184 groupby/orderby construct extensions and enablement

anoopsharma00:
Following extensions to groupby and order by have been enabled
-- group/order by can be done on expressions
-- order by clause can be specified within a subquery
-- first N clause can be set inside of a subquery

Some of these may not be fully tested and if there are bugs,
they will be handled at that time.


anoopsharma00:
yes, if an expression is specified in group by clause, then it
must match an exact expression in the select list.
Examples of group by clauses listed below:

>>select a from t group by a;

--- 0 row(s) selected.
>>

>>select a+1 from t group by a+1;

--- 0 row(s) selected.
>>select a+1 from t group by a;

--- 0 row(s) selected.
>>select a+1 aa from t group by aa;

--- 0 row(s) selected.
>>select a from tt group by a+1;

*** ERROR[4197] This expression cannot be used in the GROUP BY clause.

PR #701, opened 2016-09-14
[TRAFODION-2222] Trafodion on HDP 2.3 needs yarn client in class path

zellerh:


zellerh:
jenkins, extra test

zellerh:
The extra test failures don't seem to be related to this change, the failures happened before any tests could be run, one is on CDH, which is unaffected by this change.

svarnau:
The extra test problem was on HDP side, but agree that it was a machine set-up issue and unrelated to these changes.

The change looks good to me.

amandamoran:
+1

arvind-narain:
+1 Thanks Hans. 

Do we need /usr/hdp/current/hadoop-yarn-client/lib/* also in our classpath  or those are already included ?

DaveBirdsall:
Looks good to me.

zellerh:
@arvind-narain, good question. I (accidentally) tried without the jars in /usr/hdp/current/hadoop-yarn-client/lib/* and that worked, so I didn't include them. Someday, the code may change and make this a requirement. Let me know if you feel we should add the other jars as a defensive measure. I'm just a bit hesitant, given how rarely we use yarn and how long our CLASSPATH already is.

arvind-narain:
You are correct Hans - less we add to our CLASSPATH the better.
+1

PR #702, opened 2016-09-14
optimize using of udf

mashengchen:
this is an enhancement of udf.
user who want to use udf can extends SimpleUDR , and just archieve execute(TupleInfo tuple). the details are showed in SimpleUDRTest

### DIFF
--- a/core/sql/src/main/java/org/trafodion/sql/udr/SimpleUDR.java
+++ b/core/sql/src/main/java/org/trafodion/sql/udr/SimpleUDR.java
@@ -1,0 +1,105 @@
+// @@@ START COPYRIGHT @@@
+//
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+//
+// @@@ END COPYRIGHT @@@
+
+package org.trafodion.sql.udr;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+
+import org.apache.log4j.Logger;
+import org.trafodion.sql.udr.TypeInfo.SQLTypeCode;
+
+public abstract class SimpleUDR extends UDR {
+    private static Logger log = Logger.getLogger(SimpleUDR.class.getName());
+    private List<ColumnInfo> columns = new ArrayList<ColumnInfo>();
+    UDRInvocationInfo info = null;
+
+
+    public abstract void prepare(UDRInvocationInfo info, UDRPlanInfo plan);
+
+    /**
+ *      * recycle the init value
+ *           */
+    public abstract void close(UDRInvocationInfo info, UDRPlanInfo plan);
+
+    /**
+ *      * @throws UDRException
+ *           * 
+ *                */
+    public abstract void execute(TupleInfo tuple) throws UDRException;
+
+    public List<ColumnInfo> getColumns() {
+        return columns;
+    }
+
+    public int getNumColumns() {
+        return columns.size();
+    }
+
+    @Override
+    public final void processData(UDRInvocationInfo info, UDRPlanInfo plan) throws UDRException {
+        this.info = info;
+        TableInfo tblInfo = info.in();
+        for (int i = 0; i < tblInfo.getNumColumns(); i++) {
+            columns.add(tblInfo.getColumn(i));
+        }
+
+        prepare(info, plan);
+
+        while (getNextRow(info)) {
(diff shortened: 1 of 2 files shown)


core/sql/src/main/java/org/trafodion/sql/udr/SimpleUDR.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.Date;
+import java.util.List;
+
+import org.apache.log4j.Logger;
+import org.trafodion.sql.udr.TypeInfo.SQLTypeCode;
+
+public abstract class SimpleUDR extends UDR {
+    private static Logger log = Logger.getLogger(SimpleUDR.class.getName());
+    private List<ColumnInfo> columns = new ArrayList<ColumnInfo>();
+    UDRInvocationInfo info = null;
```
zellerh: The UDR class can be used for multiple UDR invocations, possibly at the same time, so you can't cache the UDRInvocationInfo in it.

zellerh:
Could we submit this to https://cwiki.apache.org/confluence/display/TRAFODION/Trafodion+Code+Examples instead? I think that would be a better match.

core/sql/src/main/java/org/trafodion/sql/udr/SimpleUDR.java
```diff
@@ -0,0 +1,105 @@
...
+        while (getNextRow(info)) {
+            tblInfo = info.in(0);
+
+            execute(tblInfo);
+        }
+
+        close(info, plan);
+    }
+
+    public void emitRow(Object... values) throws UDRException {
```
zellerh: Nice idea to have such a method, I like it.

zellerh:
I'm opposed to adding the current proposal to Trafodion, here are my reasons:

- Right now there is not a 1:1 relationship between a UDR object and an invocation of a TMUDF. Therefore we cannot store data that is related to an invocation in the UDR object. It would be good to have a discussion on this design question.
- The emitRow method doesn't have access to the result columns, and it is not possible to handle things like passthrough columns. This method should have an "info" and a "plan" parameter, also maybe a way to pass only some of the result columns.
- We should try to support multiple input tables in new code we are adding.
- The proposed interface works only for mappers, not for reducers (it does not detect partition boundaries). Also note that I would like to add a new interface that makes writing reducers easier, by having a method to read the next row within a partition. See [TRAFODION-1129](https://issues.apache.org/jira/browse/TRAFODION-1129).
- I'm sure other people will have different views on what a more convenient TMUDF interface will look like. Should we allow them to add their favorite interface as well? How many interfaces should we allow? For that reason I have a preference to keeping a single interface.

Again, adding it to the [code examples](https://cwiki.apache.org/confluence/display/TRAFODION/Trafodion+Code+Examples) would be very good and useful. I could also imagine adding the emitRow method that is proposed here to the Trafodion interface, after comments are addressed.

@sureshsubbiah, I'm wondering what you think.

sureshsubbiah:
Thank you @mashengchen for this code. I do not fully understand the need for this interface and exactly how it will be used. Do you think it will be better if file a JIRA explaining the purpose and providing examples (with a SQL query, and all java listings and example output).

I do get that the motivation for SimpleUDR is to provide a less complex interface when a TMUDF writer wants to acheive simpler tasks. But how this new interface does that, without sacrificing performance is not clear to me. In addition there are the issue @zellerh raised about having a unified interface and having some of the original principles in mind when new extensions are proposed.

In summary as @zeller suggested I see two alternatives
1) We add this change to code examples, to be used by those already comfortable with it
2) OR, we file a JIRA with the objective behind this change and then discuss this further there.

In the current interface other than processData all methods are optional, so I don't quite get how simplicity can be achieved. Maybe the goal is not simplicity after all, but an alternate JDBC type interface with prepare/execute/close?  

Please don't take these comments to indicate a lack of interest in this change.  However, as it stands it is likely not ready for mainline. New interfaces for TMUDF (and new uses for the existing interfaces) are always fascinating. @DaveBirdsall had suggested recently how it would be nice to support Java UDFs that are part of other Apache projects like HiveMall. So this an exciting and wide ranging area. Hope we can discuss this some more.

mashengchen:
@zellerh about your point 1,is that means it’s thread-unsafe in UDR? The UDR always be a single instance? If so I have to change UDRInvocationInfo as param to execute(), but it seems not elegance. Or is there any suggestion?

mashengchen:
@sureshsubbiah thank you for your comments. and yes, this SimpleUDR is to simplify the usage of TMUDF when a TMUDF writer wants to acheive simpler tasks, and this class is just an enhancement of processData, other method like describeXXX, may will have other derive class like SimpleUDF-2, SimpleUDF-3.
and about code examples, is that ok to add wiki for the codes are not merged in?
i created a JIRA #2232 for discussion.

kevinxu021:
@zellerh @sureshsubbiah  SimpleUDR is not trying to replace original UDR. It's true that if the users want to use original UDR, they can go ahead. SimpleUDR is to cover one of the scenarios and try to simplify the implementation. True, if SimpleUDR does not meet your requirements, please extend UDR directly. Again, SimpleUDR exists for special usage and it does not impact any current functions or performance. It's like that in JAVA it has InputStream, why it needs FileInputStream, ByteArrayInputStream, BufferedInputStream and so on? I think you might know what we're doing. Yes for special usage.

zellerh:
@mashengchen, it's not so much about thread safety as about multiple invocations of a UDR. Here is an example:

`select * from udf(myudf()) u1, udf(myudf()) u2`

UDFs U1 and U2 will have two different UDRInvocationInfo objects, but they may share the same UDR object. Your code is not designed to handle that.

mashengchen:
close it first , there should have some further change

PR #703, opened 2016-09-14
[TRAFODION-2208] function support: log2,IS_IPV4,IS_IPV6 etc

traflm:
add a few new built-in functions (help to migrate application to/from MySQL)
log2( )
inet_aton( )
inet_ntoa( )
is_ipv4
is_ipv6

One goal of this PR is to provide a template as how to add new built-in functions for Trafodion.


core/sql/exp/exp_function.cpp
```diff
@@ -45,6 +45,11 @@
 #include <ctype.h>
 #include <string.h>
 #include <stdio.h>
+/*
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
+*/
```
DaveBirdsall: Should we simply delete these lines since they are commented out?

core/sql/exp/exp_function.cpp
```diff
@@ -7724,5 +7753,354 @@ Lng32 ExFunctionExtractColumns::unpack (void * base, void * reallocator)
...
+
+   if(srcData[0] == ' ')
+   { 
+     leadingspace++;
+     for(i=1; i< slen; i++)
+     {
+       if(srcData[i] == ' ') leadingspace++;
+       else break; 
+     }
+   }
```
DaveBirdsall: This looks correct. A simpler way to do it, though, might be:

char * next = srcData;
while (*next == ' ')
   {
      leadingspace++;
      next++;
   }

core/sql/exp/exp_function.cpp
```diff
@@ -7724,5 +7753,354 @@ Lng32 ExFunctionExtractColumns::unpack (void * base, void * reallocator)
...
+   char buf[16]; //big enough
+   unsigned long addr =  *(unsigned long*)op_data[1];
+   char * resultData = op_data[0];
+   Attributes *resultAttr = getOperand(0);
+   const unsigned char *ipv4_bytes= (const unsigned char *) &addr;
+
+   if( ipv4_bytes[0] > 255 || 
+       ipv4_bytes[1] > 255 ||
+       ipv4_bytes[2] > 255 ||
+       ipv4_bytes[3] > 255 ||
```
DaveBirdsall: These first four conditions can never be true, since ipv4_bytes[i] is an 8-bit quantity, right?

core/sql/exp/exp_function.cpp
```diff
@@ -7724,5 +7753,354 @@ Lng32 ExFunctionExtractColumns::unpack (void * base, void * reallocator)
...
+       addr > 4294967295 )
+   {
+      ExRaiseSqlError(heap, diags, EXE_INVALID_CHARACTER);
```
DaveBirdsall: Would EXE_INVALID_CHARACTER be correct here? The input is not a character string but an integer. Perhaps we need a new error saying the input value is not valid for converting to an ipv4 address?

core/sql/optimizer/SynthType.cpp
```diff
@@ -1010,7 +1010,81 @@ const NAType *BuiltinFunction::synthesizeType()
...
+        //input type must be string
+        const NAType &typ1 = child(0)->getValueId().getType();
+
+        if (typ1.getTypeQualifier() != NA_CHARACTER_TYPE)
+          {
+	    *CmpCommon::diags() << DgSqlCode(-4045) << DgString0("IS_IP");
+	    return NULL;
+          }
+        retType = new HEAP
+           SQLSmall(TRUE, FALSE);
```
DaveBirdsall: SQLSmall is a 2-byte integer. Wouldn't you want a 4-byte integer at least for ipv4 and an 8-byte integer for ipv6?

core/sql/optimizer/SynthType.cpp
```diff
@@ -1010,7 +1010,81 @@ const NAType *BuiltinFunction::synthesizeType()
...
+        const NumericType &ntyp1 = (NumericType &) typ1;
+        if (NOT ntyp1.isExact() || ntyp1.getScale() != 0)
+	  {
+	    *CmpCommon::diags() << DgSqlCode(-4046) << DgString0("INET_NTOA");
+	    return NULL;
+	  }
 
+	retType = new HEAP
+	  SQLVarChar(15, FALSE);
+           SQLInt(FALSE, FALSE);
```
DaveBirdsall: Looks like a stray line got left here and needs to be deleted?

core/sql/exp/exp_math_func.cpp
```diff
@@ -348,7 +349,18 @@ ex_expr::exp_return_type ExFunctionMath::eval(char *op_data[],
 	  return ex_expr::EXPR_ERROR;
 	}
       
-      *(double *)op_data[0] = MathLog10(*(double *)op_data[1], err);
+      *(double *)op_data[0] = MathLog2(*(double *)op_data[1], err);
```
zellerh: Shouldn't this say MathLog10 here? Probably left over from debugging.

core/sql/exp/exp_function.cpp
zellerh: +1

core/sql/exp/exp_math_func.cpp
traflm: oops, yes. Sorry about this bug.

core/sql/exp/exp_function.cpp
traflm: Yes

core/sql/exp/exp_function.cpp
traflm: Much better, I appreciate your nicer code.

core/sql/exp/exp_function.cpp
traflm: yes... I will remove them

core/sql/exp/exp_function.cpp
traflm: Yes, let me check if there is a proper one, or add a new error message here.

core/sql/optimizer/SynthType.cpp
traflm: this return value according to MySQL is an integer either 0 or 1. So I use small int.

core/sql/optimizer/SynthType.cpp
traflm: oops, sorry about this.

core/sql/optimizer/SynthType.cpp
DaveBirdsall: Oh, sorry. I was confused; I was thinking of the inet_ATON function. But this is the ISIPV4 and V6 function. I agree: SQLSmall is a good choice here. You could even use SQLTiny if you wanted to perhaps.

PR #704, opened 2016-09-14
[TRAFODION-1794]: Log authentication Information

robertamarton:
Added code to log LDAP connection errors into the associated log4cxx file.
For authentications that occur in mxosrvr, events are logged into the master
executor log.  For standalone utilities such as ldapcheck, events are logged
into a separate file with names: dbsecurity_<hostname>_<pid>.log.

This code was donated to the Trafodion team and it originally logged events
into a repository table and/or a stdout log.  This change will log
authentication details into the standard log4cxx logs.  Session information is
being logged into repository tables which describe connection attributes but
detailed LDAP errors will be only put in the logs.

The ldapcheck utility has been instrumented to log errors in the logs directory
previously, errors were not logged.  This will be needed when TRAFODION-1787
is implemented.

The files authEvent.h and authEvents.cpp replace existing files called
ld_globals.h and ld_port.cpp.


core/dbsecurity/auth/Makefile
```diff
@@ -70,10 +67,11 @@ INCLUDES	= -I. -I./inc -I ../shared/inc \
...
 	        -I $(MY_SQROOT)/export/include \
+	        -I $(MY_SQROOT)/commonLogger \
 	        -I ../../sql/common
 
 
-LINK_OPTIONS	= -L$(LIBEXPDIR) -lldap -lssl -llber
+LINK_OPTIONS	= -L$(LIBEXPDIR) -lldap -lssl -llber -llog4cxx
 LINK_OPTIONS   += $(LNK_FLGS) 
 
 COMMON_LIBS =  -ltdm_sqlcli -larkcmp_dll
```
svarnau: Makefile changes look fine. You might want to remove this COMMON_LIBS line since it is not actually being used.

core/dbsecurity/auth/inc/authEvents.h
```diff
@@ -0,0 +1,89 @@
...
+#include <string>
+#include <vector>
+
+// From a web search, the log message max length is a bit over 8000 bytes.
+// For DBSecurity don't believe any messages will be more than 1M
+#define MAX_EVENT_MSG_SIZE 1024
+
+// The ported code had the caller sending in the filename and line number
+// for certain events.  This has not been implemented at this time.
+struct AuthEvent
```
DaveBirdsall: This is fine as-is. However, since you've defined accessor methods, you're half-way to making this a class. Maybe go the rest of the way?

core/dbsecurity/auth/src/authEvents.cpp
```diff
@@ -0,0 +1,134 @@
...
+#include <sys/time.h>
+#include <str.h>
+#include <unistd.h> 
+#include <stdio.h>
+#include <iostream>
+#include "seabed/ms.h"
+#include "seabed/fserr.h"
+
+static std::string AUTH_COMPONENT = "DBSECURITY";
+std::vector<AuthEvent> authEvents;
```
DaveBirdsall: These kinds of globals are dangerous. You have no control over when the destructor will be called, but it will happen sometime after the main function exits. If another thread then references the object, you can get cores. Best to avoid these if you can.

DaveBirdsall: An alternative to using a global variable would be to use a pointer to a global variable. Then you can control when the object is destroyed, and you can test for its absence.

core/dbsecurity/auth/src/authEvents.cpp
```diff
@@ -0,0 +1,134 @@
...
+  
+    // Log4cxx logging
+    char buf[MAX_EVENT_MSG_SIZE];
+    snprintf(buf, MAX_EVENT_MSG_SIZE, "Node Number: %u, CPU: %u, PIN: %u ,,,, Message: %s", 
+            my_nid, my_cpu, my_pid, eventText_.c_str());
+    
+    // strip off final new line before logging
+    int32_t lastPos = strlen(buf) -1;
+    if (buf [lastPos] == '\n')
+      buf[lastPos] = '.';
```
DaveBirdsall: Do you want a period here or a null character? If the message already ends with a period, you'll get two here.

core/dbsecurity/auth/src/dbUserAuth.cpp
```diff
@@ -1837,13 +1837,74 @@ static void logAuthenticationErrors(int nodeID)
...
+   // languages.
+   string outcomeDesc;
+   switch (outcome)
+   {
+      case AUTH_OK:
+         outcomeDesc = "Authentication successful";
+         severity = LL_INFO;
+         break;
+      case AUTH_NOT_REGISTERED:
+         outcomeDesc = "User not registered";
```
DaveBirdsall: I notice that we initialize severity to LL_INFO at line 1850, then in all cases but this one, we set it again (sometimes to LL_INFO and sometimes to something else). In this case because we didn't set it, it remains LL_INFO. Just checking: Is this an oversight? Or is LL_INFO intended here? (In which case the code works fine; it's just stylistically inconsistent.)

DaveBirdsall:
Is this code ready to commit?

robertamarton:
I am working on an update for this.  Should deliver this change shortly.



     Roberta



*Sent:* Monday, September 19, 2016 2:42 PM
*Cc:* Roberta Marton <roberta.marton@esgyn.com>; Author <
author@noreply.github.com>
*Subject:* Re: [apache/incubator-trafodion] [TRAFODION-1794]: Log
authentication Information (#704)



Is this code ready to commit?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/apache/incubator-trafodion/pull/704#issuecomment-248136146>,
or mute the thread

core/dbsecurity/auth/Makefile
robertamarton: removed the line

core/dbsecurity/auth/inc/authEvents.h
robertamarton: Changed it to a class

core/dbsecurity/auth/src/authEvents.cpp
robertamarton: Changed it to not be a global, passing the structure to all component that may generate events.

PR #706, opened 2016-09-15
sql regression test executor test022 fix, min largeint

traflm:
On Linux platform, min 64 bit int value is -9223372036854775808. So the validation seems wrong. Try to fix it.

### DIFF
--- a/core/sql/exp/exp_conv.cpp
+++ b/core/sql/exp/exp_conv.cpp
@@ -2091,8 +2091,11 @@ ex_expr::exp_return_type convAsciiToInt64(Int64 &target,
                           heap, diagsArea, flags);
   if (ert == ex_expr::EXPR_ERROR)
     return ert;
-
-  if (tempTgt > LLONG_MAX)
+  UInt64 maxLongPlus1 = LLONG_MAX ;
+  maxLongPlus1 ++; // min 64bit long integer is -9223372036854775808
+                   // max 64bit long integer is 9223372036854775807
+                   // so if negative value convert to non-negative value, should check with LLONG_MAX + 1
+  if ((tempTgt > LLONG_MAX && NOT negative) || (tempTgt > maxLongPlus1) )
     {
       ExRaiseDetailSqlError(heap, diagsArea, EXE_NUMERIC_OVERFLOW,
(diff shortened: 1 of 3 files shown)


traflm:
@anoopsharma00 , @sandhyasun  , please help to review this if it is a correct fixing?

DaveBirdsall:
+1 Also I agree with @zellerh's comment.

traflm:
@zellerh , @DaveBirdsall , I leave the original checking below my change since as I googled, the definition of LLONG_MIN is 'undefined', different platform/libc and C and C++ compilers all may have different definition. But there are two basic choices:
LLONG_MIN as -9223372036854775808
or
LLONG_MIN as -9223372036854775807

So I was hoping in a system which defines the LLONG_MIN as 0x800000000000L or (-LLONG_MAX - 1 ) the first check that I modified will take effect and make sure -9223372036854775808 will be regarded as a valid number: 

For some older platform/compiler, the LLONG_MIN will be defined as -LLONG_MAX.  In this case, -9223372036854775808 should be regarded as invalid. The second check at line 2113 should take effect.

But I agree with you, line 2113 seems dangerous, and may not work as expected, but we cannot test it without using different platform/compiler/libc to figure out. And I think supporting older platform is not something valuable. Most mainstream systems use first LLONG_MIN definition (C++ 11 and later will use first def as well). So we are good to remove line 2113. I will remove those code and add comments to link to this RP discussion.
 
I refer to these:
http://howtoprogram.eu/question/integer-c-what-are-c-standard-guarantees-on-relation-between-min-and-max-values-of-signed-integer-types,12603
http://www.cplusplus.com/reference/climits/
http://stackoverflow.com/questions/18687614/c-why-does-llong-min-llong-min

traflm:
jenkins, extra tests

traflm:
Extra Test failed at Known cases: Hive/TEST003, and compGeneral/TEST042,
This is for executor/TEST022. So this does not cause new failure and should be ready for merge.

And this change will not hide the root cause of TEST022 failure, but avoid it from fail, and this is a required change for Trafodion. I will continue to check why create table will sometimes need to convert -9223372036854775808, sometimes not. This can be done via a bugcatcher in ConvDoIt.

PR #707, opened 2016-09-15
[TRAFODION-2164] Incremental Update Statistics documentation

DaveBirdsall:
Added documentation for incremental UPDATE STATISTICS to the SQL Reference Manual.

Also deleted old change history info (that is, the change history for version 2.0 and 2.0.1).

Also updated the revision text (which goes in all manuals) to have a dummy entry for Trafodion version 2.1.0. A date will need to be supplied when this is released.

Added error messages for incremental UPDATE STATISTICS to the Messages Manual. While I was at it, reworked the error messages for UPDATE STATISTICS in general as it was incomplete and out-of-date. I moved these messages out of the "Sort Messages" chapter (they didn't belong there) to a separate, new chapter, "Update Statistics Messages".


docs/shared/revisions.txt
```diff
@@ -26,6 +26,7 @@
 [cols="2",options="header"]
 |===
 | Version    | Date
+| 2.1.0      | TBD
 | 2.0.1      | July 7, 2016
 | 2.0.0      | June 6, 2016
 | 1.3.0      | January, 2016
```
svarnau: I guess my preference was to remove this section, so we don't try to track release dates in documents.

Though it looks like this file is shared by all the docs?  That is less bad than I thought.  We could keep it if updating it is added to https://cwiki.apache.org/confluence/display/TRAFODION/Create+Release#CreateRelease-UpdateProjectWebSite

DaveBirdsall: Yes, it is shared in all documents. My thought was that perhaps the release manager could supply the date at the tail end of the release cycle. If this is agreeable, I can add a comment to the wiki as you suggest.

svarnau: Yes, that sounds good. Just add revisions.txt to the list of those files that needs to be updated and docs published. Thanks.

DaveBirdsall:
Please do not merge this just yet. I forgot to add the error messages to the Messages Manual. Will submit a rework shortly.

PR #708, opened 2016-09-15
jira TRAFODION-2181 Enable incompatible operations in sql queries

anoopsharma00:
Incompatible operations are defined as opers between datatypes
that belong to different categories.
For ex, comparison of numeric to string. In these cases, one side
need to be converted to the other before doing the comparison.

This checkin enables following incompatible operations:
  -- assignment between incompatible datatypes
  -- conversion between incompatible datatypes
  -- comparison between incompatible datatypes
  -- arith between incompatible datatypes
  -- string operation on incompatible datatypes

The infrastructure changes to support was added as part of
a previous jira. Dev regressions have since then been running with
incomp opers enabled.

These operations are on by default now and some operations that
previously failed will no longer fail.

Changes are been tested as much as possible and any issues that
show up during runs will be fixed as a regular bug fix.


DaveBirdsall:
Looks good to me. What's the impact on the documentation?

anoopsharma00:
will update documentation as part of a separate task.

PR #713, opened 2016-09-20
Second set of changes for JIRA [TRAFODION-1435]

sbroeder:
Region operations performed transactionally for delete, chackAndDelete
Improved endpoint coprocessor exception logging

### DIFF
--- a/core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/SsccTransactionalTable.java
+++ b/core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/SsccTransactionalTable.java
@@ -88,10 +88,14 @@ import org.apache.hadoop.hbase.coprocessor.transactional.generated.SsccRegionPro
 import org.apache.hadoop.hbase.coprocessor.transactional.generated.SsccRegionProtos.SsccPutRegionTxRequest;
 import org.apache.hadoop.hbase.coprocessor.transactional.generated.SsccRegionProtos.SsccPutTransactionalResponse;
 import org.apache.hadoop.hbase.coprocessor.transactional.generated.SsccRegionProtos.SsccPutTransactionalRequest;
+import org.apache.hadoop.hbase.coprocessor.transactional.generated.SsccRegionProtos.SsccDeleteRegionTxRequest;
+import org.apache.hadoop.hbase.coprocessor.transactional.generated.SsccRegionProtos.SsccDeleteRegionTxResponse;
 import org.apache.hadoop.hbase.coprocessor.transactional.generated.SsccRegionProtos.SsccDeleteTransactionalRequest;
 import org.apache.hadoop.hbase.coprocessor.transactional.generated.SsccRegionProtos.SsccDeleteTransactionalResponse;
 import org.apache.hadoop.hbase.coprocessor.transactional.generated.SsccRegionProtos.SsccCloseScannerRequest;
 import org.apache.hadoop.hbase.coprocessor.transactional.generated.SsccRegionProtos.SsccCloseScannerResponse;
+import org.apache.hadoop.hbase.coprocessor.transactional.generated.SsccRegionProtos.SsccCheckAndDeleteRegionTxRequest;
+import org.apache.hadoop.hbase.coprocessor.transactional.generated.SsccRegionProtos.SsccCheckAndDeleteRegionTxResponse;
 import org.apache.hadoop.hbase.coprocessor.transactional.generated.SsccRegionProtos.SsccCheckAndDeleteRequest;
 import org.apache.hadoop.hbase.coprocessor.transactional.generated.SsccRegionProtos.SsccCheckAndDeleteResponse;
 import org.apache.hadoop.hbase.coprocessor.transactional.generated.SsccRegionProtos.SsccCheckAndPutRequest;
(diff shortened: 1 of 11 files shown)


core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/SsccTransactionalTable.java
```diff
@@ -274,6 +278,48 @@ public SsccDeleteTransactionalResponse call(SsccRegionService instance) throws I
...
+    public void deleteRegionTx(final long tid, final Delete delete, final boolean autoCommit) throws IOException {
+      SingleVersionDeleteNotSupported.validateDelete(delete);
+      final String regionName = super.getRegionLocation(delete.getRow()).getRegionInfo().getRegionNameAsString();
+         Batch.Call<SsccRegionService, SsccDeleteRegionTxResponse> callable =
+                new Batch.Call<SsccRegionService, SsccDeleteRegionTxResponse>() {
+            ServerRpcController controller = new ServerRpcController();
+            BlockingRpcCallback<SsccDeleteRegionTxResponse> rpcCallback =
+            new BlockingRpcCallback<SsccDeleteRegionTxResponse>();
+
+            public SsccDeleteRegionTxResponse call(SsccRegionService instance) throws IOException {
```
DaveBirdsall: Just curious: Why no @Override here? (I saw one in a similar method down below.)

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/SsccRegionEndpoint.java.tmpl
```diff
@@ -2283,6 +2578,110 @@ CoprocessorService, Coprocessor {
...
+      {
+        if (request.hasRow()) {
+          row = request.getRow();
+
+          if (!Bytes.equals(rowArray, request.getRow().toByteArray()))
+            t = new org.apache.hadoop.hbase.DoNotRetryIOException("Action's " +
+                 "Delete row must match the passed row");
+        }
+
+        if (t == null) {
```
DaveBirdsall: I know this code is presently commented out (and so, presumably a work in progress), but I'm confused by the logic that extracts the row etc. then doesn't do anything more with it. Is this just saving these values in the members of the object? If so, why do we not care about their previous state?

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/SsccRegionEndpoint.java.tmpl
```diff
@@ -2283,6 +2578,110 @@ CoprocessorService, Coprocessor {
...
+          if (request.hasFamily())
+            family = request.getFamily();
+          if (request.hasQualifier())
+            qualifier = request.getQualifier();
+          if (request.hasValue())
+            value = request.getValue();
+
+          try {
+           result = checkAndDeleteRegionTx(tid, startId,
+               request.getRow().toByteArray(),
```
DaveBirdsall: I assume if request.hasRow() had returned false above, we'd throw an exception here?

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/SsccRegionEndpoint.java.tmpl
```diff
@@ -2594,7 +2994,76 @@ CoprocessorService, Coprocessor {
...
+    SsccTransactionState state = this.beginTransIfNotExist(tid, startId, true); // This is a regionTX
+    boolean result = false;
+    byte[] rsValue = null;
+
+    Get get = new Get(row);
+    get.addColumn(family, qualifier);
+
+    Result rs = this.get(tid, startId, get);
+
+    boolean valueIsNull = value == null ||
```
DaveBirdsall: I assume this is an HBase-ism. I wonder how zero-length varchar strings are represented?

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/SsccRegionEndpoint.java.tmpl
sbroeder: There is no point since the comparison later pulls the row directly from the request.

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/SsccRegionEndpoint.java.tmpl
sbroeder: Yes, it should be caught by the throwable in the catch

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/SsccTransactionalTable.java
sbroeder: Good comment.  I don't think @Override is needed on any of the region operations since none are present in the super class

PR #714, opened 2016-09-20
[TRAFODION-1988] Better java exception handling in the java/JNI layer

selvaganesang:
The stack trace of the exception displayed from JNI side now contains the
cause of the exception displayed in a nested way.

### DIFF
--- a/core/sqf/src/seatrans/tm/hbasetmlib2/javaobjectinterfacetm.cpp
+++ b/core/sqf/src/seatrans/tm/hbasetmlib2/javaobjectinterfacetm.cpp
@@ -269,6 +269,9 @@ JOI_RetCode JavaObjectInterfaceTM::initJVM()
         gThrowableToStringMethodID = _tlp_jenv->GetMethodID(gThrowableClass,
                       "toString",
                       "()Ljava/lang/String;");
+        gGetCauseMethodID = _tlp_jenv->GetMethodID(gThrowableClass,
+                      "getCause",
+                      "()Ljava/lang/Throwable;");
      }
   }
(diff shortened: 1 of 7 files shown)


core/sqf/src/seatrans/tm/hbasetmlib2/javaobjectinterfacetm.cpp
```diff
@@ -421,21 +422,32 @@ bool  JavaObjectInterfaceTM::getExceptionDetails(JNIEnv *jenv)
...
        return false;
    }
+   error_msg = new std::string("");
+   appendExceptionMessages(jenv, a_exception, error_msg);
+   *error_msg += "\n";
+   _tlp_error_msg = error_msg;
+   return true;
+}
+
+void JavaObjectInterfaceTM::appendExceptionMessages(JNIEnv *jenv, jthrowable a_exception, std::string *error_msg)
```
DaveBirdsall: If we don't expect error_msg to be null, it might be better to code this as a reference parameter rather than a pointer.

selvaganesang: @Dave fixed as per your comments

sureshsubbiah:
+1. Executor portion looks very nice. I am curious to see what kind of output the getCause method produces. If you have a sample handy, please add it to  the JIRA.

prashanth-vasudev:
Changes look good. This change will also help during startup if TM encounters exceptions connecting to Hbase.

trinakrug:
Looks good to me.

PR #715, opened 2016-09-20
[TRAFDODION-2170] create table with too long key causing core dump

traflm:
when the given table DDL have key length greater than 2048, it will cause buffer overflow.
This fix add a length checking to return error in this case.

### DIFF
--- a/core/sqf/src/tm/tmddlrequests.cpp
+++ b/core/sqf/src/tm/tmddlrequests.cpp
@@ -40,14 +40,21 @@ JNIEXPORT jint JNICALL Java_org_apache_hadoop_hbase_client_transactional_RMInter
    char la_tbldesc[TM_MAX_DDLREQUEST_STRING];
    char la_tblname[TM_MAX_DDLREQUEST_STRING];
    char* str_key;
-   str_key = new char[TM_MAX_DDLREQUEST_STRING];
    char** la_keys;
    la_keys = new char *[TM_MAX_DDLREQUEST_STRING];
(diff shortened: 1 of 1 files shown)


traflm:
something wrong, I will abandon this PR.

traflm:
jenkins, retest

core/sqf/src/tm/tmddlrequests.cpp
```diff
@@ -40,14 +40,21 @@ JNIEXPORT jint JNICALL Java_org_apache_hadoop_hbase_client_transactional_RMInter
...
+   {
+     cout << "Table key length is larger than max allowed" << endl;
+     return FEBADKEYDESC;
+   }
+   str_key = new char[pv_keyLength];
+
    int lv_tblname_len = pp_env->GetArrayLength(pv_tblname);
    if(lv_tblname_len > TM_MAX_DDLREQUEST_STRING) {
       cout << "Table name length is larger than max allowed" << endl;
+      return FEBADNAME;
```
zellerh: A leak of str_key here, could delete it before returning.

traflm: Yes, will fix it. I will abandon this RP and create another one. This PR cannot pass basic regression test.
All the metadata table created with key length as 0, so 'initialize trafodion' will not success.
I will create a new PR and with this leak fix as well.
thanks Hans for the careful review.

traflm:
This change has a severe issue, so will abandon and resubmit with a new PR.

PR #716, opened 2016-09-20
[TRAFDODION-2170] create table with too long key causing core dump

traflm:
If the create table DDL have key length (total len of all PK field) is greater than 2048, the buffer created in the function will be overrun, and cause memory corruption and core dump.
Adding a length check. The error message will be handled by TRAFODION-2231.

Also fix memory leak issues pointed by Hans in previous PR review.

### DIFF
--- a/core/sqf/src/tm/tmddlrequests.cpp
+++ b/core/sqf/src/tm/tmddlrequests.cpp
@@ -40,14 +40,24 @@ JNIEXPORT jint JNICALL Java_org_apache_hadoop_hbase_client_transactional_RMInter
    char la_tbldesc[TM_MAX_DDLREQUEST_STRING];
    char la_tblname[TM_MAX_DDLREQUEST_STRING];
    char* str_key;
-   str_key = new char[TM_MAX_DDLREQUEST_STRING];
    char** la_keys;
    la_keys = new char *[TM_MAX_DDLREQUEST_STRING];
    int lv_error = FEOK;
 
+   if( pv_keyLength > TM_MAX_DDLREQUEST_STRING)
+   {
+     cout << "Table key length is larger than max allowed " << pv_keyLength << endl;
+     delete [] la_keys;
+     return FEBADKEYDESC;
+   }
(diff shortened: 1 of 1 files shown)


traflm:
jenkins, retest

traflm:
pyodbc test failed, not related to this change, retest

traflm:
jenkins, retest

core/sqf/src/tm/tmddlrequests.cpp
```diff
@@ -40,14 +40,24 @@ JNIEXPORT jint JNICALL Java_org_apache_hadoop_hbase_client_transactional_RMInter
...
    char la_tblname[TM_MAX_DDLREQUEST_STRING];
    char* str_key;
-   str_key = new char[TM_MAX_DDLREQUEST_STRING];
    char** la_keys;
    la_keys = new char *[TM_MAX_DDLREQUEST_STRING];
    int lv_error = FEOK;
 
+   if( pv_keyLength > TM_MAX_DDLREQUEST_STRING)
+   {
+     cout << "Table key length is larger than max allowed " << pv_keyLength << endl;
```
DaveBirdsall: A native English speaker will see "Table key length is larger than max allowed 5123" and think that 5123 is the max, instead of the actual key length value. It might be a bit clearer if it were "Table key length 5123 is larger than max allowed".

traflm: thanks Dave to provide good error message.
There will be another JIRA that will rework all the error msg handling in this module , back to the end user.

selvaganesang: ditto. It is good to change the error message to avoid misinterpretation by the end user

traflm: OK, I need to abandon this PR for now, TRAFODION-2231 is on the way, this will conflict, I would like TRAFODION-2231 to get merged first, and I will do modification as neded.

traflm:
Wait for TRAFODION-2231 to merged first. Abandon for now, rework after that change merged.

PR #718, opened 2016-09-21
jira TRAFODION-2234 turn aligned format on, phase 1

anoopsharma00:
This is phase 1 of aligned format change.

Dev regressions now run with aligned format tables.

To test both hbase and aligned format, some tests have explicit
specification to create hbase format tables.

Tests that test for features that are currently only available
with hbase format (like pushdown sel expr) create hbase format tables.

Many expected files have been updated to reflect aligned format,
mostly in showddl and explain output.


anoopsharma00:
yes, in phase 2, the default will be turned on. It is discussed as part of jira 2234

attribute could be specified explicitly during ddl or based the default value of the cqd.
Once specified, it becomes part of table attribute irrespective of what the default value
is when the table is used or showddl is issued.

Currently, showddl shows aligned attribute explicitly and nothing for hbase attribute.
So if nothing is displayed, then the table has hbase attr.

What we can do is to show explicit attribute for both hbase and aligned format.
This will be similar to how serialization property of a column is explicitly displayed
irrespective of how it was declared (explicitly or through a cqd default value).
That way if showddl will correctly represent how a table was originally created.

We can make that change for phase e.

PR #719, opened 2016-09-21
[TRAFODION-2223] Improve error reporting on bad incremental WHERE clause

DaveBirdsall:
When an invalid WHERE clause was given on an incremental UPDATE STATISTICS command, the error raised was misleading. This has been fixed. As part of this change, a new error message, 9219, was added; the Messages Guide has been updated.

The prototype code for incremental UPDATE STATISTICS contained an alternative syntax, termed the "simple" syntax. A CQD controlled which syntax was in force. This alternative syntax and its associated CQD have been removed.

Some error messages that are no longer used have been deleted.

Incremental update statistics includes several warning conditions that are uninteresting to the average user. They detail whether incremental update statistics simply adjusted existing histograms or computed new ones from the sample data. There is no practical difference from a user perspective, so it seems wise from a human factors point of view to turn these off. However, if update statistics logging is on, the warnings will be raised as this may be of interest to a developer or support person.


traflm:
changes look good to me. +1

PR #720, opened 2016-09-22
Changes for JIRA [TRAFODION-1435]

sbroeder:
Second set of region transaction operations - delete, checkAndDelete
Merges from HEAD
changes from previous review comments


shangshengtung:
Looks good. I assume Observer code to handle edit replay will be checked into in the next round.

sbroeder:
Yes.  Those changes are still pending.

PR #722, opened 2016-09-22
[TRAFODION-2233] Increased the number of cores limit to 256

zcorrea:
The installer also has changes to allow the limits increase.


amandamoran:
+1

PR #723, opened 2016-09-22
JIRA [TRAFODION-2236]

sbroeder:
TM crashes during startup

### DIFF
--- a/core/sqf/sql/scripts/cleanZKNodes
+++ b/core/sqf/sql/scripts/cleanZKNodes
@@ -1,0 +1,33 @@
+!/bin/bash
+#
+# @@@ START COPYRIGHT @@@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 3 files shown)


shangshengtung:
Looks good.

selvaganesang:
Looks good if there is no harm in dropping these zk nodes. Some explanation here why is it not harmful to delete these zk nodes would be helpful. Also, adding subject line to the PR request would help the reviewers when looking for what to review. He/she doesn't need to click to get the details and then decide to review.

sbroeder:
This comment is taken from the JIRA - THis is safe to do because any region server needing recovery will repost to zookeeper and the TM will have no issues connecting to this RS.  My assumption is that all the discussion should be there, but I can duplicate it here.  I can add a more descriptive title for future PRs

svarnau:
FYI, Pull-Request authors can edit title of the PR after the fact.

PR #724, opened 2016-09-23
[TRAFODION-2235] Added sync thread epoll timeout logic and Zookeeper …

zcorrea:
…client thread to the monitor


core/sqf/monitor/linux/monitor.cxx
```diff
@@ -1294,7 +1441,31 @@ int main (int argc, char *argv[])
             }
         }
 
+        if ( IsRealCluster )
```
amandamoran: How is "IsRealCluster" set?

sureshsubbiah:
+1.

PR #726, opened 2016-09-24
[TRAFODION-2245] Multiple sqcheck and jps processes running when moni…

selvaganesang:
…tor is downed and up

as dcsserver checks if trafodion is up

A new option sqcheck -n <nid> is introduced to determine if the node is up and running
DCS server is changed to use this option when it detects the server is gone
rather than at the time of starting the processes

Also fixed the failure of hive/TEST018 due to expected file difference

### DIFF
--- a/dcs/src/main/java/org/trafodion/dcs/server/ServerManager.java
+++ b/dcs/src/main/java/org/trafodion/dcs/server/ServerManager.java
@@ -85,10 +85,11 @@ public final class ServerManager implements Callable {
     private static String sqlplanEnable;
     private static int userProgPortMapToSecs;
     private static int userProgPortBindToSecs;
-    private static ServerHandler[] serverHandlers;
+    private ServerHandler[] serverHandlers;
     private int maxRestartAttempts;
     private int retryIntervalMillis;
     private RetryCounterFactory retryCounterFactory;
+    private String nid = null;
 
     class RegisteredWatcher implements Watcher {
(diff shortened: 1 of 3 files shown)


selvaganesang:
jenkins, retest

hegdean:
Changes look good to me. Thanks for taking care of this

sureshsubbiah:
+1. Looks good @selvaganesang .

dcs/src/main/java/org/trafodion/dcs/server/ServerManager.java
```diff
@@ -367,6 +335,36 @@ public ServerManager(Configuration conf, ZkClient zkc,
...
+    private static boolean isTrafodionRunning(String nid) {
+
+        // Check if the given Node is up and running
+        // return true else return false.
+        // Invoke sqcheck -n <nid> to check Node status. 
+        // 
+        //   sqcheck returns:
+        //    -1 - Not up ($?=255) or node down
+        //     0 - Fully up and operational or node up
+        //     1 - Partially up and operational
```
DaveBirdsall: I didn't see a code path in sqcheck that returned 1 or 2?

selvaganesang: This is the old comment that got moved here with changes to include node down and node up.  I will check and update this comment when I merge something else next time

PR #727, opened 2016-09-25
[TRAFODION-2238] Compiler Internal Error: RelExpr with unknown arity …

traflm:
…encountered with combination of WITH clause and Intersect

WITH will use copyTree() try to do a deep copy of the binded node. But before INTERSECT doesn't implement the CopyTopNode() method, so cannot be copied.
This fix the issue and let the Q14 from TPCDS compile success.


zellerh:
+1
Looks good to me.

PR #729, opened 2016-09-26
TRAFODION-2203 - a user can grant privileges that he doesn’t have ... ... to other users/roles successfully

robertamarton:
In this case, the user/role did not get the privilege requested even though the
operation successfully completed.  So the requester is lead to believe that the
privilege was granted.

ANSI states that: "warning <privilege not granted>" should be displayed for 
each combination of grantee<=>privilege that was not granted. However, 
privileges that can be successfully granted should be granted. The grant code 
does not grant any privileges it cannot grant but is not reporting warnings if
the privilege is not granted. Ditto for revoke.

The code now reports warnings if not all privileges were granted or revoked for
both object and column privileges.


core/sql/sqlcomp/PrivMgrDesc.cpp
```diff
@@ -550,14 +554,28 @@ void PrivMgrCoreDesc::interpretChanges( const bool before,       // in
...
+    int index = other.getColumnPriv(columnLevel_[i].getColumnOrdinal());
+    if (index >= 0)
+    {
+      temp.unionOfPrivs(other.columnLevel_[index]);
+      if (columnLevel_[i].limitToGrantable(temp))
+        result = true;
+    }
+    else
+    {
+      if (columnLevel_[i].limitToGrantable(temp))
```
DaveBirdsall: Did you intend the "if (columnLevel_[i].limitToGrantable(temp)) etc." to be the same in both the "if" and "else" code paths? (Just curious why you bothered with the "else".)

core/sql/sqlcomp/PrivMgrPrivileges.cpp
```diff
@@ -3182,34 +3042,107 @@ PrivStatus PrivMgrPrivileges::revokeColumnPrivileges(
...
+            {
+              if ( !grantedColPriv->getPriv(type))
+                printWarning = true;
+            }
+            else
+            {
+              if (colPrivToRevoke.getWgo(type) && !grantedColPriv->getWgo(type))
+              {
+                printWarning = true;
+                printWgo = true;
```
DaveBirdsall: I didn't see printWgo used anywhere. It's only a problem though if there is some logic you intended but forgot.

core/sql/sqlcomp/PrivMgrPrivileges.cpp
```diff
@@ -6134,6 +5870,79 @@ static bool isDelimited( const std::string &strToScan)
...
+
+     if (!colPrivsNotApplied.isNull())
+     {
+       for ( size_t j = FIRST_DML_PRIV; j <= LAST_DML_PRIV; j++ )
+       {
+         PrivType privType = PrivType(j);
+         if (colPrivsNotApplied.getPriv(privType))
+         {
+           // would be better to add column name instead of number
+           // would require an I/O to read COLUMNS to get the name
```
DaveBirdsall: Might be available in the NATables cache, but I know you don't want that dependency in this code.

core/sql/sqlcomp/PrivMgrDesc.cpp
robertamarton: You are correct, the else is not needed, will fix it in the next checkin.

core/sql/sqlcomp/PrivMgrPrivileges.cpp
robertamarton: printWgo can be removed, will fix it in the next checkin.

PR #730, opened 2016-09-27
Fix to handle bad TM ZK node entries during recovery

sbroeder:
This is the second fix for JIRA [TRAFODION-2236]


core/sqf/src/seatrans/tm/hbasetmlib2/src/main/java/org/trafodion/dtm/HBaseTxClient.java
```diff
@@ -1062,6 +1062,14 @@ public void run() {
...
                                       zookeeper.deleteRegionEntry(regionEntry);
                                 }// DeserializationException
+                                catch (IOException ioe) {
+                                    // It's possible we received a spurious entry from a stale region.
+                                       // It's safe to delete the zookeeper entry and wait for any new posts
+                                       LOG.warn("TRAF RCOV THREAD:IOException calling txnManager.recoveryRequest. " + "TM: " +
+                                               tmID + " regionBytes: [" + regionBytes
+                                               + "].  Deleting zookeeper region entry. exception: " + ioe);
+                                       zookeeper.deleteRegionEntry(regionEntry);
+                                 }// IOException
```
selvaganesang: IOException is more generic. I wonder if it is ok to assume all the IOExceptions are the result of spurious recovery entry.

sbroeder: In the TM any throwable exception gets mapped to an IOException.  Additionally, if the TM gets zero results back from any regions it throws an IOException.  In both cases there is nothing the client can do to facilitate recovery, but we must delete the ZKNode.  Any necessary recovery from the region would post a new ZKNode

shangshengtung:
Looks good. Good fix to remove the "shift" of transaction id and use more generic methods.

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionManager.java
```diff
@@ -3010,12 +3010,12 @@ public RecoveryRequestResponse call(TrxRegionService instance) throws IOExceptio
               rresult = table.coprocessorService(TrxRegionService.class, startKey, endKey, callable);
             }
             catch (ServiceException se) {
-                LOG.error("Exception thrown when calling coprocessor: ", se);
-                throw new IOException("Problem with calling coprocessor, no regions returned result", se);
+                LOG.error("Service exception thrown when recoveryRequest: ", se);
```
DaveBirdsall: Reading the text, I see, "Service exception thrown when recoveryRequest". I'm left wondering, "when recoveryRequest what?" Is that information packaged inside the se object?

DaveBirdsall: Reading the text, I see, "Service exception thrown when recoveryRequest". I'm left wondering, "when recoveryRequest what?" Is that information packaged inside the se object?

core/sqf/src/seatrans/tm/hbasetmlib2/src/main/java/org/trafodion/dtm/HBaseTxClient.java
selvaganesang: As per of cleaning up the exception handling on the client , I attempted to do the following for the coprocessorService request.

 CoprocessorService request throws either ServiceException or Throwable.  ServiceException  are flagged as RetryTransactionException and Throwable are flagged as DoNotRetryException.  

When there is a response, the exceptionString will be looked at and converted into an exception derived from IOException.  If some of these exceptions can be retried, it would be flagged as RetryTransactionException. The RetryTransactionException is retried 'n' numbers of time, before it is thrown as an DoNotRetryException.   

This concept is missed out for recoveryRequest.  It would be good if we can add this concept for recoveryRequest. 

By ignoring the IOException here, I am wondering if we ignoring some exceptions which should not be ignored.  The other exceptions like TableNotFoundException are getting masked here as IOException because recoveryRequest throws only IOException. The tableNotFoundException could be the cause of the IOException.  In that case, we should catch IOException and get the cause of the exception and handle the causes here.

sbroeder: Retrying the coprocessor call inline is not necessary.  The only thing required is deleting the ZKNode that caused the exception.  If a TableNotFoundException was the the issue, we should not retry, because the table doesn't exist and it won't ever exist.  We should just delete the entry and move on.  It was the fact that we were leaving the the ZKNode in place that caused the incessant retries and exceptions that brought the TM down.

If the region really does require recovery, it will post a new ZKNode and that will get picked up by the TM.  That is the intended design of the recovery mechanism.

sbroeder: TableNotFoundException is a non-issue, however, in the case of NotServingRegionException, you are correct.  I will look at addressing this.

sbroeder:
Not yet.  I'm working on an additional set of changes, but thanks for pinging.

sbroeder:
jenkins retest

core/sqf/src/seatrans/tm/hbasetmlib2/src/main/java/org/trafodion/dtm/HBaseTxClient.java
```diff
@@ -1034,34 +1034,23 @@ public void run() {
...
+                                catch (IOException e) {
+                                   // For all cases of Exception, we rely on the region to redrive the request.
+                                   // Likely there is nothing to recover, due to a stale region entry, but it is always safe to redrive.
+                                   // We log a warning event and delete the ZKNode entry.
+                                   LOG.warn("TRAF RCOV THREAD:Exception calling txnManager.recoveryRequest. " + "TM: " +
+                                              tmID + " regionBytes: [" + regionBytes + "].  Deleting zookeeper region entry. \n exception: ", e);
+                                   zookeeper.deleteRegionEntry(regionEntry);
+
+                                   // In the case of NotServingRegionException we will repost the ZKNode after refreshing the table.
+                                   if (e instanceof NotServingRegionException){
```
selvaganesang: Do you think we should consider checking e.getCause() instanceof NotServingRegionException too. Rationale: Assuming recoveryRequest throws NotServingRegionException, it would be caught in 3018 and transformed into IOException with cause as NotServingRegionException.

sbroeder:
That is correct.  If HBase has some underlying issue that prevents us from recovering we must fix the HBase issue.  We should never give up recovery or there would be database inconsistency

PR #732, opened 2016-09-28
TRAFODION-2234 Updated phx tsts to work around an upsert on aligned table pbm

anoopsharma00:
aligned format and upsert with missing cols issue being tracked by jira trafodion-2247.
When that jira is fixed, phoenix testware changes will be reverted back.


anoopsharma00:
jenkins retest

PR #733, opened 2016-09-29
[TRAFODION-2246] group by rollup feature

anoopsharma00:
See comments listed below as part of commit.


core/sql/optimizer/RelCache.cpp
```diff
@@ -418,6 +418,10 @@ void GroupByAgg::generateCacheKey(CacheWA &cwa) const
     groupExpr_.rebuildExprTree(ITM_ITEM_LIST);
   if (grpExpr) { 
     cwa += " gBy:"; 
+
+    if (isRollup())
+      cwa += " roll:";
+
```
zellerh: I think we also need to add the rollupGroupExprList() to the cache key. If we rebuild the list above from a ValueIdSet on line 418 above, it is probably going to be in the same order, regardless whether it was ROLLUP(a,b) or ROLLUP(b,a).

zellerh:
+1

Didn't look at all the details, especially not in the executor, but the overall idea looks really good to me. I'm glad we are able to extend the sort groupby operator, our earlier idea of using a UDF would have been more messy. This also adds a few useful features we can use elsewhere, like an expression that can tell us where an error occurred. This could be useful when converting Hive records to Trafodion format, so we can handle errors better.

core/sql/optimizer/RelCache.cpp
anoopsharma00: yes, good point. Tried it and it does incorrectly caches them.
Will fix it.

PR #734, opened 2016-09-29
[TRAFODION-2252] Fix memory leak during certain insert

sureshsubbiah:


zellerh:
+1

Thanks, good find. I suspect there are other places where we allocate a collection from the heap without passing a heap to its constructor. Maybe we can all be on the lookout for those cases.

sureshsubbiah:
Thanks @zellerh. @selvaganesang suggested to me that we comment out the default constructor for the common collection classes and fix as many compile errors as possible. I have started with NAArray. Will submit those changes in one or more subsequent PRs. Being on the lookout for this mistake that I have often made in the past is a good approach too.

selvaganesang:
Thanks @sureshsubbiah for taking up this task. It would help a long way to ensure that the memory allocated within SQL engine are accounted within the heap management infrastructure of Trafodion.

PR #737, opened 2016-10-01
[TRAFODION-2247] upsert on altered aligned format table with missing …

selvaganesang:
…columns inserts incorrect values

While generating mergeInsertExpr or updateExpr, the added columns were not considered. This resulted
in incorrect relative offset value in the expressions in case of aligned row format and the row got
corrupted.


core/sql/generator/GenRelUpdate.cpp
```diff
@@ -539,7 +415,9 @@ static short genHbaseUpdOrInsertExpr(
      updateRowLen,      // (OUT) Destination tuple length
      updateExpr,  // (OUT) Generated expression
      updateTupleDesc, // (OUT) Tuple descriptor for destination tuple
-     ExpTupleDesc::LONG_FORMAT);
+     ExpTupleDesc::LONG_FORMAT,
+     NULL, NULL, 0, NULL, NULL,
```
DaveBirdsall: Should this be NULL, NULL, 0, NULL, FALSE? (the last of these is an NABoolean, disableConstFolding). Though the code still works since NULL would map to FALSE.

PR #743, opened 2016-10-04
[TRAFODION-2259] TopN sort changes.

prashanth-vasudev:
This includes first of changes related to sort implementation in executor.
cqd gen_sort_topn_size 'N' forces sort to use topn.
Subsequent changes in compiler will be able to push down topn to sort.

Additional cleanup , error handling will be checked in subsequent changes once compiler changes are in.

1. Sort would initially maintain Top N array of elements to being with.
2. Read records into TopN array. 
3. Once TopN array is full, heapify the array into max heap. Top node in the heap is always the highest node.
4. Subsequent record read either gets discarded( if greater than top node) or replace top node( if lesser then top node) . if replaced top node, re-balance the heap.
5. Repeat steps 4 until last record is read.
6. sort the final heap using heap sort.


prashanth-vasudev:
Do not merge. Few additional minor changes being added.

core/sql/sort/Topn.cpp
```diff
@@ -0,0 +1,337 @@
...
+		  
+		 buildHeap();
+	  }
+	  return SORT_SUCCESS;
+  }
+  
+  //Reaching here means, heap is already build. 
+  //Check if the new rec belongs to this heap by comparing the
+  //new rec key with the root node of the heap ( root node is always the greatest).
+  insertRec(rec, len, tupp);
```
DaveBirdsall: Odd that this method name has the same name as the variable declared at line 121. Doesn't hurt anything but does make the code a bit more complicated to understand. And it might break if you reorder the code later.

core/sql/sort/Topn.cpp
```diff
@@ -0,0 +1,337 @@
...
+
+	int root = 0; //Always, root node is the zero'th element in array.
+	
+	Record * insertRec = new Record(rec, len, tupp, heap_, sortError_);
+	insertRecKey_.key_ = insertRec->extractKey(keySize_, sortUtil_->config()->numberOfBytesForRecordSize());
+	insertRecKey_.rec_ = insertRec;
+
+	if (compare(topNKeys_[root].key_ ,insertRecKey_.key_) == KEY1_IS_GREATER)
+    {
+      swap( &topNKeys_[root],&insertRecKey_);
```
DaveBirdsall: It's places like this where the mixture of tabs and non-tabs makes the code hard to read due to odd indentation.

core/sql/sort/Topn.cpp
```diff
@@ -0,0 +1,337 @@
...
+}
+
+
+void TopN::insertRec(void *rec, ULng32 len, void* tupp) 
+{
+	ex_assert(isHeapified_, "TopN::insertRec: not heapified");
+
+	int root = 0; //Always, root node is the zero'th element in array.
+	
+	Record * insertRec = new Record(rec, len, tupp, heap_, sortError_);
```
DaveBirdsall: Do you want global new here?

core/sql/sort/Topn.cpp
```diff
@@ -0,0 +1,337 @@
...
+      satisfyHeap();
+      
+    }
+	
+	//Once swapped, or not swapped, delete the unneeded record.
+	//This step is very important to discard the unwanted record.
+	//Note releaseTupp() also deletes the tupp allocated in sql 
+	//buffer. This makes space for new records read in.
+	insertRecKey_.rec_->releaseTupp();
+    delete insertRecKey_.rec_;
```
DaveBirdsall: Do you want global delete here?

core/sql/sort/Topn.cpp
```diff
@@ -0,0 +1,337 @@
...
+    recNum_++;
+  }
+  else {
+    len = 0;
+  }
+
+  return SORT_SUCCESS;
+}
+/*
+//----------------------------------------------------------------------
```
DaveBirdsall: I guess this commented out code could be removed?

core/sql/sort/Record.cpp
```diff
@@ -55,11 +55,23 @@ Record::Record(ULng32 size, NABoolean doNotallocRec, CollHeap* heap)
   }
 }
 
+Record::Record(void *rec, ULng32 reclen, void* tupp, CollHeap* heap, SortError* sorterror)
+{
+	 recSize_   = reclen;
```
sandhyasun: Tab seems to be 4 instead of 2.

core/sql/sort/Topn.cpp
```diff
@@ -0,0 +1,337 @@
...
+{
+  //if heap not built means, TopN array has more slots 
+  //available to fill. 
+  if(! isHeapified_)
+  {
+	  ex_assert(loopIndex_ >= 0, "TopN::sortSend: loopIndex_ is < 0");
+	  ex_assert(loopIndex_ < allocRunSize_, "TopN::sortSend: loopIndex_ > allocRunSize_");
+	  ex_assert(rec != NULL, "TopN::sortSend: rec is NULL");
+	
+	  Record * insertRec = new Record(rec, len, tupp, heap_, sortError_);
```
sandhyasun: A heap pointer needs to be passed in and used here.

core/sql/sort/Topn.cpp
sandhyasun: Heap pointer needs to be propagated here too .

prashanth-vasudev:
Thank you for thorough review and valuable feedback. I have addressed all the feedback in the additional commit and also included changes to pick GETN request in executor parent down queue request to drive TopN sort. If the request is GETALL, then TopN sort is not involved.

Please review and merge.

DaveBirdsall:
Would be good for @sandhyasun to take another look before merging. She can merge if she likes, otherwise I'll do so after her review (might be Monday).

prashanth-vasudev:
jenkins, extra tests

prashanth-vasudev:
jenkins, extra tests

DaveBirdsall:
Looks like compGeneral/TEST042 is giving different results?

sandhyasun:
Will wait for the updated known diff file for TEST042 and then merge this. The differences are in line numbers only.

DaveBirdsall:
Something strange has happened to this pull request. It is as if a zillion other changes have been added to it. Perhaps we should close this and submit a fresh set of changes?

prashanth-vasudev:
jenkins, extra tests

selvaganesang:
Is there a way to see only your changes?

prashanth-vasudev:
These are the two commits that will show diffs.
https://github.com/apache/incubator-trafodion/commit/3cd201a6eb98ec412e33af44f106b69b215e2dcf

https://github.com/apache/incubator-trafodion/commit/08081bd70956bfe5d8edbd35dd87ec75d3de70e5


As my workspace was rebased as opposed to a merge (after the above two push commits) , update to known diff file (regress/compGeneral/Test042) brings in rest of the history.

selvaganesang:
It is becoming difficult to review this code because I can't view the final changes that are being merged in one view.  If others have reviewed this code already, please go ahead.  However, I have a comment because I am unable to decipher from the change.

Are the topN records to be in memory or the topN record keys to be in memory?

prashanth-vasudev:
TopN records will be in memory as tupps in sql buffer. Record keys is the data structure to perform sort. Once sort is complete followed by all records submitted to parent queue, record keys structure is destroyed.

PR #745, opened 2016-10-04
[TRAFODION-2261] Mxosrvr or java core with starting from Java_org_tra…

selvaganesang:
…fodion_jdbc_t2_SQLMXConnection_close

[TRAFODION-2262] Mxosrvr or Java core with the stack trace pointing to log4Cxx functions

IpcServerClass was not getting destructed correctly. When the CLI context is deleted, the ESP
server class object an IpcServerClass instance that manages the ESP server started
in the CLI context are also destroyed.  Fixed the IpcServerClass and IpcServer destructors
so that there is no memory corruption.

Log4Cxx infrastructure in SQL was recursively attempting to log the messages when there is an
issue with it. This was causing stack corruption.

Also added a code to initialize the log4cxx infrastructure if it is not done already upon
the first CLI call.


sureshsubbiah:
+1.

PR #746, opened 2016-10-05
jira TRAFODION-2263 Similarity check changes and obsolete code removal.

anoopsharma00:


core/sql/cli/Statement.cpp
```diff
@@ -4517,61 +3343,58 @@ RETCODE Statement::execute(CliGlobals * cliGlobals, Descriptor * input_desc,
...
+                setState(OPEN_);
+                readyToReturn = TRUE;
+                break;
+              }
+          }
+        break;
+          
+	case RE_EXECUTE_:                
 	  {
 	    // save lnil & stuff
```
DaveBirdsall: I'm guessing "lnil" refers to late name input list? Perhaps the comment should be updated.

core/sql/comexe/CmpMessage.h
```diff
@@ -419,8 +415,7 @@ class CmpCompileInfo
 
   char * sqltext_;    // 00-03
   Lng32 sqlTextLen_;   // 04-07
-  RecompLateNameInfoList * rlnil_;  // 08-11
-  Lng32 rlnilLen_;  // 12-15
+  Lng32 unused_;       // 12-15
```
DaveBirdsall: Should this be an Int64? and should the comment say, // 08-15? (Though I don't think it matters anymore since we don't do static SQL.)

core/sql/comexe/LateBindInfo.cpp
```diff
@@ -1319,37 +1049,102 @@ Int16 AnsiOrNskName::quoteNSKExtName()
...
+                                                 Int64 modTS, Int32 numPartnLevels,
+                                                 Queue * hdfsDirsToCheck,
+                                                 char * hdfsHostName,
+                                                 Int32 hdfsPort)
+     : NAVersionedObject(-1),
+       tableName_(tableName),
+       hdfsRootDir_(hdfsRootDir),
+       modTS_(modTS), numPartnLevels_(numPartnLevels),
+       hdfsDirsToCheck_(hdfsDirsToCheck),
+       hdfsHostName_(hdfsHostName), hdfsPort_(hdfsPort)
```
DaveBirdsall: Looks like we need an initializer to set flags_ = 0. If we don't do this, and isHive is false, the HIVE flag will be uninitialized (and therefore random).

core/sql/comexe/LateBindInfo.cpp
```diff
@@ -1319,37 +1049,102 @@ Int16 AnsiOrNskName::quoteNSKExtName()
...
 }
 
-void ResolvedNameList::translateFromOldVersion(ResolvedNameListPre1800 *oldrnl)
+TrafSimilarityTableInfo::TrafSimilarityTableInfo()
+     : NAVersionedObject(-1),
+       tableName_(NULL),
+       hdfsRootDir_(NULL),
+       modTS_(-1), numPartnLevels_(-1),
+       hdfsDirsToCheck_(NULL),
+       hdfsHostName_(NULL), hdfsPort_(NULL)
```
DaveBirdsall: Need a flags_ inintializer here too.

PR #748, opened 2016-10-07
[TRAFODION-2266] Fix for a few memory leaks

sureshsubbiah:
Constructor for NACollection derived class now require an explicit heap.
This helps catch a few leaks. It also allows for better accounting of 
how memory is being used using the standard NAMemory infrastructure.


sureshsubbiah:
Testing done on this change : 
(a) Release and debug builds 
(b) Debug regression testing, with an all PASS
(c) Debug regression testing with MEMDEBUG=1 envvar set, all PASS

core/sql/common/Collections.h
```diff
@@ -2217,15 +2217,15 @@ template <class T> class NAArray : public NACollection<T>
 public :
 
   // default constructor
-  NAArray(CollIndex initialElements = 0) : 
-       NACollection<T>(initialElements) {}
+  //NAArray(CollIndex initialElements = 0) : 
+  //NACollection<T>(initialElements) {}
```
selvaganesang: Can you please remove the commented out lines

core/sql/common/Collections.h
```diff
@@ -2389,10 +2389,15 @@ template <class T> class NASimpleArray : private NAArray<T>
   
   // operator s new and delete get inherited from private base class,
   // make them publicly available by the following inline functions
-  static inline void * operator new(size_t size, CollHeap* h=0)
+  static inline void * operator new(size_t size, CollHeap* h)
                                 { return NAArray<T>::operator new(size, h); }
   static inline void operator delete(void *buffer)
                                   { NAArray<T>::operator delete(buffer); }
+
+  inline void setHeap(CollHeap *heap)
```
selvaganesang: May not be good idea to do setHeap after it is constructed. This can possibly result in different heaps to be used for different elements and cause core dumps

core/sql/common/Ipc.h
```diff
@@ -572,7 +572,9 @@ class IpcConnection : public NABasicObject
 
   static const char *getConnectionStateString(IpcConnectionState s);
 
-  IpcConnection() : fileNumForIOCompletion_(InvalidGuaFileNumber) {}
+ IpcConnection() 
+   : fileNumForIOCompletion_(InvalidGuaFileNumber),
+    recvStreams_(NULL) {} // recvStream on C++ heap
```
selvaganesang: I think you can use IpcEnvironment heap

core/sql/common/Collections.h
sureshsubbiah: Please see comment from before.

sureshsubbiah:
Response to comments by @zellerh 
1. The code adds NULL to the constructor of many collection objects. On one hand, this is good, as it makes people think, on the other hand I'm not sure it really adds much.
Yes, passing in NULL is the same using the default constructor before. Objects are allocated in the global heap. Currently code calls NACollection constructors with NULL for class variables only if no suitable NAHeap object is available. I would claim this is about a dozen or so instances. In these cases the change does no harm but makes no improvement. It will be a longer term project to create a NAHeap object that could be used in these instances. Mostly these classes are shared across many processes or are required early in the life of a process when NAHeap objects have not been created. For stack variables there are many more instances of using NULL. These are harmless as they would get deallocated when the local variable goes out of scope. I view them as equivalent to numerous NAString stack variables that are created without a heap and are therefore allocated in the global heap. Those are not leaks as I understand that memory is reclaimed when the destructor is called on those objects

3. This is the case discussed in item 1. I don't think they cause problems any worse than now.

I don't quite follow the line of argument in the conclusion, where we state that this makes the situation worse. In these these comments there are already diametrically opposite views on what to do with the default constructors (remove the commented out lines vs. put those lines back and use them). I think this change makes some improvement without causing any regression.

Please using your voting options to indicate if this should not go in. I will explicitly state which suggested changes suggested I am unable to make. At this point I won't have time to revisit most of these files and make a systematic change. If that is better, we could defer this for later. After all, these leaks have not hurt us explicitly and we could (a) decide on  consensus approach and then start making changes or (b) fix issues in a smaller scale with a real problem at hand.

core/sql/export/ComDiags.cpp
```diff
@@ -1671,8 +1671,8 @@ ComDiagsArea::ComDiagsArea (CollHeap* ptr): IpcMessageObj(IPC_SQL_DIAG_AREA,0),
 
 ComDiagsArea::ComDiagsArea () :             IpcMessageObj(IPC_SQL_DIAG_AREA,0),
                                             collHeapPtr_(NULL),
-                                            errors_(),
-                                            warnings_(),
+                                            errors_(NULL),
```
sureshsubbiah: Yes. 
This is the ComDiagsArea constructor without a heap pointer. There is another one with a heap pointer. I see two options to avoid this leak
a) try to get rid of this constructor OR
b)  Make sure that these NACollection objections are cleared at the end of their lifecycle.

Either of them would take some effort. I don't have a fix that can be included now. I would claim that this code does no worse than before here, while making marginal improvements in other places. It also ensures that no one unintentionally creates objects on the system heap.

core/sql/arkcmp/CmpStoredProc.cpp
```diff
@@ -733,7 +733,7 @@ CmpStoredProc::ExecStatus CmpStoredProc::close(CmpSPDataObject&)
 // Methods of CmpISPFuncs
 // -----------------------------------------------------------------------
 
-NAList<CmpISPFuncs::ProcFuncsStruct> CmpISPFuncs::procFuncsArray_(256);
+NAList<CmpISPFuncs::ProcFuncsStruct> CmpISPFuncs::procFuncsArray_(NULL,256);
```
sureshsubbiah: I tried this change. Had to include Globals.h in this file. Build went fine, but RMS would not come up on sqstart. Did not investigate further. Did try clean build a couple of times, so I am reasonably sure it is not an env issue.

core/sql/common/Collections.h
sureshsubbiah: Done for now.

core/sql/common/Ipc.h
sureshsubbiah: I don't think that heap is available here. I am way of making this change any bigger in scope by changing how IpcConnection objects are constructed. So this too will have to be deferred. We could make a list of all classes with NACollection data members on the system heap and then come back of them one by one.

PR #751, opened 2016-10-10
[TRAFODION-2264] Add binder messages to message guide

DaveBirdsall:
The Binder messages (4000-4999) were missing from the Messages Guide. Actually, they were present in the asciidoc source, but they were not included in the index document. This has been fixed.

While I was at it, I spent some time proofing and researching messages 4000-4100. In a few places I made editorial changes. I removed message 4018 from the guide, as it is obsolete. (The work to remove it from the Binder code will wait for later.) Several other messages refer to features as not being supported but support has since been added. The messages can now only occur when certain CQDs are set. I updated those messages accordingly. Eventually those CQDs and messages should be removed from both code and documentation. Message 4065 was missing; I added it.


sureshsubbiah:
+1

PR #753, opened 2016-10-10
[TRAFODION-2272] Insure adequate hbase.client.scanner.timeout.period

DaveBirdsall:
UPDATE STATISTICS on large tables fails on some cluster installations because the hbase.client.scanner.timeout.period setting is too low. The symptom is that a sample scan fails with a SocketTimeoutException. This change adds code to insure the setting is at least 1 hour.

This is viewed as a short-term fix. The setting is hard-coded in the Trafodion HBase client code. Very soon there will be another change that will abstract this out into some sort of properties file mechanism that is easier to manipulate.

Too, as soon as we migrate to HBase 1.1-based distributions, the HBase heartbeat protocol should eliminate the need for this particular change.


DaveBirdsall:
jenkins, retest

sureshsubbiah:
+1

PR #754, opened 2016-10-11
[TRAFODION-2262] Mxosrvr or java core with the stack trace pointing t…

selvaganesang:
…o log4Cxx functions

The Logger repository needs to be initialized. The method QRLogger::initLog4cxx
initialized the commonLogger infrastructure also.

PR #745 merged to fix this issue exposed a problem in SSMP. This
caused SSMP to dump core when a node is brought down.
SSMP should have create the connection to SSCP on that node.

### DIFF
--- a/core/sql/qmscommon/QRLogger.cpp
+++ b/core/sql/qmscommon/QRLogger.cpp
@@ -38,6 +38,8 @@
 #include "seabed/ms.h"
 #include "seabed/fserr.h"
 
+BOOL gv_QRLoggerInitialized_ = FALSE;
+
 using namespace log4cxx;
 using namespace log4cxx::helpers;
 
@@ -186,6 +188,9 @@ NABoolean QRLogger::initLog4cxx(const char* configFileName)
 {
   NAString logFileName;
 
+  if (gv_QRLoggerInitialized_)
+     return TRUE;
+ 
   // get the log directory
   logFileName = "";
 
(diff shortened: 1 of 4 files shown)


sureshsubbiah:
Looks good @selvaganesang. +1

core/sql/qmscommon/QRLogger.cpp
```diff
@@ -186,6 +188,9 @@ NABoolean QRLogger::initLog4cxx(const char* configFileName)
 {
   NAString logFileName;
 
+  if (gv_QRLoggerInitialized_)
+     return TRUE;
+ 
   // get the log directory
```
DaveBirdsall: I cannot figure out if this code is thread-safe. What if two threads concurrently enter this function and both see gv_QRLoggerInitialized_ as FALSE. Both would then call CommonLogger::initLog4cxx and possibly introduceSelf(). Is there any harm in this?

selvaganesang: The caller sqInit is protected by a semaphore.  sqInit is called by the first CLI call in a process . See the function below in the same file CliExtern.cpp

908 static Lng32 CliNonPrivPrologue()
    909 {
    910 
    911   Lng32 retcode;
    912 
    913   if (cli_globals == NULL || cli_globals->getIsInitialized() == FALSE)
    914   {
    915     globalSemaphore.get();
    916     if (cli_globals != NULL)
    917     {
    918       globalSemaphore.release();
    919       return 0;
    920     }
    921     sqInit();
    922     CliGlobals::createCliGlobals(FALSE); // this call will create the globals.
    923     ex_assert(SQL_EXEC_SetEnviron_Internal(1) == 0, "Unable to set the Environment");
    924     globalSemaphore.release();
    925   }
    926   return 0;
    927 }

selvaganesang:
I will merge this because it takes care of SSMP core. However, we continue to see java cores at log4cxx at times.

PR #757, opened 2016-10-12
[TRAFODION-2167]: Invalid query invalidation keys not working properly

robertamarton:
When a user is revoked from a role, invalidation keys are not being
processed correctly. Therefore, users can still run queries even though
privileges have been removed.  Query invalidation is complicated when
table descriptors are stored in metadata.

Changes:

--> The list of priv_descs created (and stored) was changed to include an entry
    for each user and each role accumulated across all grantors. (Today, each
    priv_desc entry includes the users' direct grants plus grants on their
    active roles.)
--> When an object is loaded into NATable or NARoutine cache, the priv_desc is
    accessed and the privilege bitmap is now generated by combining the users'
    privileges with privileges of their active roles. Correct invalidation keys
    are now being created and stored with the object. In the first code drop,
    the users' active roles are read from the role_usage table. In the next
    code drop, the active roles will be stored and maintained in executor
    globals.
--> When a plan is compiled, the correct invalidation keys for users, roles,
    and the public authorization are added to the plan.
--> Changes in the compiler were required to handle the invalidation keys for
    revoke role and revoke privilege from "PUBLIC".
--> Cleaned up the code that manages invalidation keys in privilege manager.
--> Included the correct create and redef times (if available) in the stored
    object descriptor - today they are always set to 0.
--> Added new regression test to test all the revoke options.


selvaganesang:
@robertamarton Can you please resolve these conflicts

robertamarton:
jenkins, retest

PR #760, opened 2016-10-12
[TRAFODION-2267] Fix ALTER TABLE RENAME documentation

DaveBirdsall:
Problem: The SQL Reference manual describes a CASCADE keyword for ALTER TABLE RENAME, however if you try to use it, you get error message 1427, "Table cannot be renamed. Reason: Cascade option not supported."

Supporting this option seems like a fair amount of work, so for the moment, we've simply removed mention of CASCADE from the SQL reference manual.

While working on this issue, I noticed that the DDL messages section of the Messages Guide is woefully out-of-date. Among other things, it lacks several messages including the message 1427 mentioned above. Fixing all of that is a big job, but with this set of changes I did a small part of it. I added documentation for messages 1420-1429.

While doing this I noticed that ALTER TABLE RENAME to an invalid HBase name produced a very ugly 8448 error (unlike CREATE TABLE and DROP TABLE which produce a nice tidy error 1422 in this case). I fixed the error reporting code in ALTER TABLE RENAME so that we now get the nice error 1422 instead.

### DIFF
--- a/docs/messages_guide/src/asciidoc/_chapters/ddl_msgs.adoc
+++ b/docs/messages_guide/src/asciidoc/_chapters/ddl_msgs.adoc
@@ -3044,3 +3044,146 @@ statement.
 *Recovery:* Remove the creation of triggers from the CREATE SCHEMA
 statement and resubmit. Create triggers in separate statements.
 
+<<<
+[[SQL-1420]]
+== SQL 1420
+
+```
+Column <column-name> cannot be dropped or altered as it is part of the table's primary key.
+```
+
+*Cause:* You attempted to drop a column that is a part of a primary key.
+
+*Effect:* The operation fails.
+
+*Recovery:* Correct the column name or table name as appropriate and resubmit.
+
+[[SQL-1421]]
+== SQL 1421 
+
+```
+Column <column-name> cannot be dropped or altered as it is used in the secondary index <index-name>.
+```
+
+*Cause:* You attempted to drop a column, but that column is used by the secondary index named.
+
+*Effect:* The operation fails.
+
+*Recovery:* If the secondary index is no longer needed, drop it and resubmit. If the column
+name or table name are incorrect, correct them and resubmit.
+
+[[SQL-1422]]
+== SQL 1422 
+
+```
+An invalid HBase name was specified in this DDL statement. A valid name can only contain these characters: [a-zA-Z_0-9-.]
+```
+
+*Cause:* You attempted to use a name containing an invalid character for a {project-name} object that is
+stored in HBase.
+
+*Effect:* The operation fails.
+
+*Recovery:* Correct the name and resubmit.
+
+<<<
+[[SQL-1423]]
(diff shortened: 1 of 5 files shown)


docs/messages_guide/src/asciidoc/_chapters/ddl_msgs.adoc
```diff
@@ -3044,3 +3044,146 @@ statement.
...
+[[SQL-1422]]
+== SQL 1422 
+
+```
+An invalid HBase name was specified in this DDL statement. A valid name can only contain these characters: [a-zA-Z_0-9-.]
+```
+
+*Cause:* You attempted to use a name containing an invalid character for a {project-name} object that is
+stored in HBase.
+
```
robertamarton: I assume {project-name} gets changed to Trafodion during document rendering.

docs/messages_guide/src/asciidoc/_chapters/ddl_msgs.adoc
```diff
@@ -3044,3 +3044,146 @@ statement.
...
+*Recovery:* None. Contact the {project-name} Developer Distribution List.
+
+<<<
+[[SQL-1429]]
+== SQL 1429
+
+```
+Inserts into _ROW_ format external hbase tables can only use the VALUES clause and must use the column_create function to create values.
+```
```
robertamarton: hbase ->HBase?

DaveBirdsall: Oops... I'll fix this.

docs/messages_guide/src/asciidoc/_chapters/ddl_msgs.adoc
```diff
@@ -3044,3 +3044,146 @@ statement.
...
+[[SQL-1427]]
+== SQL 1427 
+
+```
+Table cannot be renamed. <reason>
+```
+
+*Cause:* You attempted to perform ALTER TABLE RENAME but the rename could not be performed.
+The <reason> explains why.
+
```
robertamarton: nie - In this Cause - "The <reason> explains why." is used.  In message 1425 "<reason> gives more details", may want to make both the same.

robertamarton:
I personally, forget to change the messages manually when a new message is added/updated.  Perhaps as part of reviews, we should verify that this step is taken.
Is there a JIRA to support the CASCADE option at some point in time; that is, is this something that would be needed to make names more meaningful?

Changes look fine, just a couple of nits.

DaveBirdsall:
@robertamarton, there's no JIRA to support CASCADE at the moment. It looks like intent behind CASCADE is for the rename to succeed even if dependent objects exist. (If you try a rename now and a view exists, the operation fails with error 1427, "Table cannot be renamed. Reason: Operation not allowed if dependent views exist. Drop the views and recreate them after rename." I can't be completely sure that this is the whole intent, so some definitional work needs to be done before such a JIRA can be written.

PR #762, opened 2016-10-13
[TRAFODION-1698]ODBC: trafodon doesn't support SQLForeginKeys

xwq:
Fix the error of the old implementation and add test cases for it.

### DIFF
--- a/core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
+++ b/core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
@@ -4963,17 +4963,17 @@ odbc_SQLSvc_GetSQLCatalogs_sme_(
             char fkexpSchemaNm[MAX_ANSI_NAME_LEN + 1];
             char fkexpTableNm[MAX_ANSI_NAME_LEN + 1];
 
-            if (!checkIfWildCard(fkcatalogNm, fkcatalogNmNoEsc) ||
-                !checkIfWildCard(fkschemaNm, fkschemaNmNoEsc)   ||
-                !checkIfWildCard(fktableNm, fktableNmNoEsc))
-            {
-                exception_->exception_nr = odbc_SQLSvc_GetSQLCatalogs_ParamError_exn_;
-                exception_->u.ParamError.ParamDesc = SQLSVC_EXCEPTION_WILDCARD_NOT_SUPPORTED;
-                goto MapException;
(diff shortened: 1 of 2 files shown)


core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
```diff
@@ -4982,40 +4982,39 @@ odbc_SQLSvc_GetSQLCatalogs_sme_(
...
+                    "and PKKV.CONSTRAINT_NAME = rcv.UNIQUE_CONSTRAINT_NAME "
                     "and PKCO.TABLE_NAME = PKKV.TABLE_NAME "
                     "and FKCO.TABLE_NAME = FKKV.TABLE_NAME "
                     "and PKCO.COLUMN_NAME = PKKV.COLUMN_NAME "
                     "and FKCO.COLUMN_NAME = FKKV.COLUMN_NAME "
-                    "and (rcv.TABLE_NAME = '%s' or trim(rcv.TABLE_NAME) LIKE '%s' ESCAPE '\\') "
                     "and (PKCO.SCHEMA_NAME = '%s' or trim(PKCO.SCHEMA_NAME) LIKE '%s' ESCAPE '\\') "
                     "and (PKCO.TABLE_NAME = '%s' or trim(PKCO.TABLE_NAME) LIKE '%s' ESCAPE '\\') "
                     "and (FKCO.SCHEMA_NAME = '%s' or trim(FKCO.SCHEMA_NAME) LIKE '%s' ESCAPE '\\') "
-                    "and (FKCO.TABLE_NAME = '%s' or trim(FKCO.TABLE_NAME) LIKE '%s' ESCAPE '\\');",
```
selvaganesang: Is there a need to add  a condition on CONSTRAINT_TYPE in the where clause? You can try to enhance your test script to create the foreign key table with a primary key and check the output. BTW, thanks for adding the test script

xwq: It need not to add this condition here. The information of foreign keys were read from REF_CONSTRAINTS_VIEW . 
I have tried the case you mentioned, it can pass.

selvaganesang: Great. In that case this change looks good. It would be good to add the test case that you tried out to also to the repository

PR #766, opened 2016-10-19
[TRAFODION-2282][TRAFODION-882] New scheme to invalidate histogram cache

DaveBirdsall:
This set of code changes leverages the query invalidation and security key invalidation infrastructure to invalidate the histogram cache. The design is as follows:

1.	At UPDATE STATISTICS time, after statistics have been written, a new stats invalidation key is generated. The key contains the table UID of the table whose stats have been updated.
2.	RMS propagates this key to all compilers in the cluster. Note that the master executors need not know, as changing histograms does not invalidate query plans. (Hence SscpNewIncomingConnectionStream::processSecInvReq ignores stats invalidation keys.)
3.	The compiler main, when checking for invalidation keys it has not seen before, processes the new stats invalidation keys. It removes from the HistogramCache any HistogramCacheEntry that has the same table UID.
4.	Fail safe logic to invalidate the entire histogram cache was added in the case that the compiler has been idle so long that it may have missed seeing a stats invalidation key.

Older code in HistogramCache::getHistograms (optimizer/NATable.cpp) that used a refresh interval timing scheme to invalidate cache entries has been completely removed. As part of this cleanup, two CQDs are no longer needed and are removed. Stubbed routines in ustats/hs_read.cpp have been removed. Too, it was noticed that the modified time stamp was stubbed (and cannot be un-stubbed), and so that was removed. Note that the redefinition time stamp is also stubbed but it is so pervasive in so many data structures I chose to leave it alone for now.

Several regression test expected files have been updated, as UPDATE STATISTICS now takes effect immediately in a session resulting in plan changes and related data.


core/sql/optimizer/NATable.cpp
```diff
@@ -275,75 +282,28 @@ void HistogramCache::getHistograms(NATable& table)
...
+    // (Possibly useless) sanity tests
+
+    // Check to see if the redefinition timestamp has changed. This seems
+    // to be always stubbed to zero today on Trafodion, so this check
+    // seems to never fail.
 
     if (cachedHistograms && (cachedHistograms->getRedefTime() != redefTime))
     {
       deCache(&cachedHistograms);
     }
```
robertamarton: I actually changed DDL code to use the non-zero redeftime in descriptors as part of my role invalidation change, but it caused some interesting regression failures.  So I removed the change before delivering the code; hopefully, I can get back to it sometime.

core/sql/optimizer/opt.cpp
```diff
@@ -5034,6 +5029,20 @@ void OptDefaults::initialize(RelExpr* rootExpr)
...
+  char *sct = getenv("RMS_SIK_GC_INTERVAL_SECONDS");  // in seconds
+  if (sct)
+    {
+      siKeyGCinterval_ = ((Int64) str_atoi(sct, str_len(sct)));
+      if (siKeyGCinterval_ < 10)
+        siKeyGCinterval_ = 10;
+    }
+  else
+    siKeyGCinterval_ = (Int64)24 * 60 * 60; // 24 hours
+
```
robertamarton: The sik key interval is being stored as a class member.  Could someone change the envvar between the time you set the variable and the time it is used?  Would it be better to calculate it on the fly?

core/sql/optimizer/NATable.cpp
DaveBirdsall: Good thing I didn't rip out all the redeftime stuff :-)

core/sql/optimizer/opt.cpp
DaveBirdsall: Environment variable values are inherited at process create time. I suppose it is possible that we could write code in our processes to change it but that seems a devilish thing to do. I've followed the same design used for caching CQD values in opt.cpp.

robertamarton:
Looks good, glad invalidation keys fit in so well with histogram cache.  Makes the code simpler and more consistent with all the other caches.

PR #767, opened 2016-10-19
[TRAFODION-2292] Add RPM packaging

svarnau:
Make rpm package to contain same things as server tarball.
Only additional pre-install action is to create the trafodion user.
This mimics the regular installer default action to locate files in
/home/trafodion, owned by trafodion user.

File locations and ownership perhaps ought to be changed, but that is
a separate effort from this initial RPM packaging.


sureshsubbiah:
+1

PR #772, opened 2016-10-20
Do not merge! Preview of TRAFODION-2127 common subexpression changes.

zellerh:
[TRAFODION-2127] Initial changes for materializing common subexpressions

This is a first set of changes to allow us to make use of CTEs
(Common Table Expressions) declared in WITH clauses and to create
a temp table for them that is then read multiple times in the query.

Summary of changes:

- Adding a unique statement number in CmpContext
- Moving the execHiveSQL method from the OSIM code to CmpContext
- Adding a list of common subexpressions and their references
  to CmpStatement
- Adding the ability to the Hive Truncate operator to drop the
  table when the TCB gets deallocated
- Adding the ability to the HDFS scan to compute scan ranges at
  runtime. Those are usually determined in the compiler. This is
  only supported for simple, non-partitioned, delimited tables.
  We need this because we populate the temp table and read from
  in in the same statement, without the option of compiling
  after we inserted into the temp table.
- Special handling in the MapValueIds node of common subexpressions.
  See the comment in MapValueId::preCodeGen().
- Moved the binder code to create a FastExtract node into a new
  method FastExtract::makeFastExtractTree(), to be able to call
  it from another place.
- MapValueIds no longer looks at the "used by MVQR flag" to determine
  the method for VEGRewrite. Instead it checks whether a list of
  values has been provided to do this.
- Added a new method, RelExpr::prepareMeForCSESharing, that is
  kind of an "unnormalizer", undoing some of the normalizer
  transformations.
- Implemented the steps for common subexpression materializations
  described below.
- Added the ability to suppress the Hive timestamp modification
  check when truncating a Hive table
- Added an optimizer rule to eliminate CommonSubExprRef nodes.
  These nodes should not normally survive past the SQO phase, but
  if the SQO phase gets interrupted by an exception, that could
  happen, since we then fall back to a copy of the tree before
  SQO. In the future, we can consider cost-based decision on
  what to do with common subexpressions.
- Add CommonSubExprRef nodes in the parser whenever we expand
  a CTE reference.
- Got rid of the now meaningless "seamonster" display in EXPLAIN.

Other changes contained in this change set (file OptLogRelExpr.cpp):

- Optimization for empty scans, like select * from t where 1=0
  This now generates a cardinality constraint with 0 rows, which
  can be used later to eliminate parts of a tree.
- [TRAFODION-2280] Need to remove salt columns from uniqueness
  constraints generated on salted tables.

Generated Plans
---------------

The resulting query plan for a query Q with n common
subexpressions CSE1 ... CSEn looks like this:

                         Root
                           |
                      MapValueIds
                           |
                      BlockedUnion
                       /        \
                    Union        Q
                    /   \
                  ...    CTn
                  /
               Union
               /   \
             CT1   CT2

Each of the CTi variables looks like the following, an
INSERT OVERWRITE TABLE tempi ...

         BlockedUnion
          /        \
      Truncate  FastExtract TEMPi
        TEMPi        |
                    CSEi

The original query Q has the common subexpressions replaced
with the following:

           MapValueIds
               |
             scan TEMPi

Here is a simple query and its explain:

```
prepare s from
with cse1 as (select d_date_sk, d_date, d_year, d_dow, d_moy from date_dim)
select x.d_year, y.d_date
from cse1 x join cse1 y on x.d_date_sk = y.d_date_sk
where x.d_moy = 3;

```
```
>>explain options 'f' s;

11   .    12   root                                                  1.46E+005
5    10   11   blocked_union                                         1.46E+005
7    9    10   merge_join                                            7.30E+004
8    .    9    sort                                                  1.00E+002
.    .    8    hive_scan                       CSE_TEMP_CSE1_MXID11  1.00E+002
6    .    7    sort                                                  5.00E+001
.    .    6    hive_scan                       CSE_TEMP_CSE1_MXID11  5.00E+001
1    4    5    blocked_union                                         7.30E+004
2    .    4    hive_insert                     CSE_TEMP_CSE1_MXID11  7.30E+004
.    .    2    hive_scan                       DATE_DIM              7.30E+004
.    .    1    hive_truncate                                         1.00E+000

--- SQL operation complete.
>>

```
CQDs to control common subexpressions
-------------------------------------

```
CQD                    Value      Default  Behavior
---------------------  ---------  -------  ---------------------------------------
CSE_DEBUG_WARNINGS       OFF        Y      No change
                         ON                Emit diagnostic warnings that show why
                                           we didn't create temp tables for
                                           common subexpressions

CSE_FOR_WITH             OFF        Y      Keep current behavior
                         ON                Insert a CommonSubExprRef node in the
                                           tree whenever we reference a CTE
                                           (table defined in a WITH clause)

CSE_HIVE_TEMP_TABLE      OFF               Meaningless at this point
                         ON         Y      Create Hive temp tables

CSE_USE_TEMP             OFF        Y      Disable creation of temp tables
                                           for common subexpressions
                         SYSTEM            Same as OFF for now
                         ON                Always create a temp table for
                                           common subexpressions where possible

```
CommonSubExprRef relational operators
-------------------------------------

This is a new RelExpr class that is introduced. It marks the common
subexpressions in a RelExpr tree. This operator remembers the name of
a common subexpression (e.g. the name used in the WITH clause).
Multiple such operators can reference to the same name. Each of
these references has a copy of the tree.

Right now, these operators are created in the parser when we expand a
CTE (Common Table Expression), declared in a WITH clause.  If the CTE
is referenced only once, then the CommonSubExprRef operator is removed
in the binder - it also doesn't live up to its name in this case.

The remaining CommonSubExprRef operators keep track of various changes
to their child trees, during the binder and normalizer phases.  In
particular, it tracks which predicates are pushed down into the child
tree and which outputs are eliminated.

The CmpStatement object keeps a global list of all the
CommonSubExprRef operators in a statement, so the individual operators
have a way to communicate with their siblings:

- A statement can have zero or more named common subexpressions.
- Each reference to a common subexpression is marked in the RelExpr
  tree with a CommonSubExprRef node.
- In the binder and normalizer, common subexpressions are expanded,
  meaning that multiple copies of them exist, one copy per
  CommonSubExprRef.
- Common subexpressions can reference other common subexpressions,
  so they, together with the main query, for a DAG (directed
  acyclic graph) of dependencies.
- Note that CTEs declared in a WITH clause but not referenced are
  ignored and are not part of the query tree.

In the semantic query optimization phase (SQO), the current code makes
a heuristic decision what to do with common subexpressions - to
evaluate them multiple times (expand) or to create a temporary table
once and read that table multiple times.

If we decide to expand, the action is simple: Remove the
CommonSubExprRef operator from the tree and replace it with its child.

If we decide to create a temp table, things become much more difficult.
We need to do several steps:

- Pick one of the child trees of the CommonSubExprRefs as the one to
  materialize.
- Undo any normalization steps that aren't compatible with the other
  CommonSubExprRefs. That means pulling out predicates that are not
  common among the references and adding back outputs that are
  required by other references. If that process fails, we fall back
  and expand the expressions.
- Create a temp table tmp.
- Prepare an INSERT OVERWRITE TABLE tmp SELECT * FROM cse tree
  that materializes the common subexpression in a table.
- Replace the CommonSubExprRef nodes with scans of the temp table.
- Hook up the insert query tree with the main query, such that it
  is executed before the main query starts.

Temporary tables
----------------

At this time, temporary tables are created as Hive tables, with a
fabricated, unique name, including the session id, a unique statement
number within the session, and a unique identifier of the common
subexpression within the statement. The temporary table is created at
compile time. The query plan contains an operator to truncate the
table before populating it. The "temporary" Hive table is dropped when
the executor TCB is deallocated.

Several issues are remaining with this approach:

- If the process exits before executing and deallocating the statement,
  the Hive table is not cleaned up.
  Solution (TBD): Clean up these tables like we clean up left-over
  volatile tables. Both are identified by the session id.
- If the executor runs into memory pressure and deallocates the TCB,
  then allocates it again at a later time, the temp table is no longer
  there.
  Solution (TBD): Use AQR to recompile the query and create a new table.
- Query cache: If we cache a query, multiple queries may end up with
  the same temporary table. This works ok as long as these queries are
  executed serially, but it fails if both queries are executed at the
  same time (e.g. open two cursors and fetch from both, alternating).
  Solution (TBD): Add a CQD that disables caching queries with temp tables
  for common subexpressions.

In the future we also want to support volatile tables. However, those also
have issues:

- Volatile tables aren't cleaned up until the session ends. If we run
  many statements with common subexpressions, that is undesirable. So,
  we have a similar cleanup problem as with Hive tables.
- Volatile tables take a relatively long time to create.
- Insert and scan rates on volatile Trafodion tables are slower than
  those on Hive tables.

To-do items are marked with "Todo: CSE: " in the code.


DaveBirdsall:
A question (maybe I'll find the answer as I read the code): Is the replacement of WITH by a blockedUnion always done? Or is it a costed plan choice? I'm wondering what would happen if the WITH query was a simple scan with a highly selective key predicate for example. It might be faster to just scan it twice rather than scan it, write it to a temp table, then scan the temp table twice.

core/sql/generator/GenPreCode.cpp
```diff
@@ -6058,6 +6059,98 @@ RelExpr * MapValueIds::preCodeGen(Generator * generator,
...
+                // augment the base columns with their index columns,
+                // the bridge value is likely an index column
+                for (ValueId v=subtractions.init();
+                     subtractions.next(v);
+                     subtractions.advance(v))
+                  if (v.getItemExpr()->getOperatorType() == ITM_BASECOLUMN)
+                    {
+                      subtractions +=
+                        static_cast<BaseColumn *>(v.getItemExpr())->getEIC();
+                      baseCol = v;
```
DaveBirdsall: baseCol doesn't seem to be used anywhere?

core/sql/generator/GenPreCode.cpp
```diff
@@ -10715,7 +10808,10 @@ RelExpr * PhysicalFastExtract::preCodeGen (Generator * generator,
   if (nodeIsPreCodeGenned())
     return this;
 
-  generator->setIsFastExtract(TRUE);  
+  if (getIsMainQueryOperator())
+    generator->setIsFastExtract(TRUE);
+  else
+    generator->setContainsFastExtract(TRUE);
```
DaveBirdsall: I would have thought that "contains" would include the "is" case... but perhaps mathematically speaking we are talking about proper containment?

core/sql/generator/GenRelScan.cpp
```diff
@@ -1119,7 +1124,8 @@ short FileScan::codeGenForHive(Generator * generator)
   if (expirationTimestamp > 0)
     expirationTimestamp += 1000000 *
       (Int64) CmpCommon::getDefaultLong(HIVE_METADATA_REFRESH_INTERVAL);
-  generator->setPlanExpirationTimestamp(expirationTimestamp);
+  if (!getCommonSubExpr())
+    generator->setPlanExpirationTimestamp(expirationTimestamp);
```
DaveBirdsall: Could you explain what a plan expiration timestamp is? (not in the code necessarily... but in a reply comment maybe?)

core/sql/optimizer/ItemExpr.cpp
```diff
@@ -10441,6 +10441,20 @@ void ConstValue::changeStringConstant(const NAString* strval)
...
 
+NABoolean ConstValue::isAFalseConstant() const
+{
+  NABoolean result = FALSE;
+
+  if (type_->getTypeQualifier() == NA_BOOLEAN_TYPE && !isNull())
+    {
+      CMPASSERT(storageSize_ == sizeof(Int32));
+      if (*(reinterpret_cast<Int32 *>(value_)) == 0)
+        result = TRUE; // that means the constant is FALSE!!
```
DaveBirdsall: :-)

zellerh:
Good question about when the transformation is done. The plan is to have three values for CQD CSE_USE_TEMP: OFF, SYSTEM and ON. OFF and ON would do this transformation never or whenever possible. SYSTEM should decide based on heuristics. One of the reasons why the transformation happens in the semantic query optimization phase is that we can get some information on statistics to help us decide. I would like to leave the work on a heuristic decision to a future JIRA and deliver the code with just ON and OFF as choices.

In the even longer term, it would be good if we would make the decision in the optimizer, based on rules. Once we have that, I would hope that the SQO phase would heuristically decide the easy cases, and would leave some more difficult decisions to the optimizer.

core/sql/arkcmp/CmpContext.cpp
```diff
@@ -588,6 +644,7 @@ void CmpContext::init()
   // initSchemaDB();		-- This was done in the ctor.
   // diags()->clear();		-- This loses any initialization errors;
   //				-- clear() is done in unsetStatement above.
+  statementNum_++;
```
zellerh: Thanks, fixed.

core/sql/arkcmp/CmpStatement.cpp
```diff
@@ -1688,3 +1690,26 @@ void CmpStatement::setTMUDFRefusedRequirements(const char *details)
...
   detailsOnRefusedRequirements_->insert(new(heap_) NAString(details, heap_));
 }
+
+void CmpStatement::addCSEInfo(CSEInfo *info)
+{
+  if (cses_ == NULL)
+    cses_ = new(CmpCommon::statementHeap())
+      LIST(CSEInfo *)(CmpCommon::statementHeap());
+
+  info->setCSEId(cses_->entries());
```
zellerh: No, this list is ever-increasing for the life of one compilation.

core/sql/arkcmp/CmpStatement.cpp
```diff
@@ -1688,3 +1690,26 @@ void CmpStatement::setTMUDFRefusedRequirements(const char *details)
...
+  cses_->insert(info);
+}
+
+CSEInfo * CmpStatement::getCSEInfo(const char *cseName)
+{
+  if (cses_)
+    for (CollIndex i=0; i<cses_->entries(); i++)
+      {
+        if ((*cses_)[i]->getName() == cseName)
```
zellerh: As far as I understand, the WITH clause can appear only at the beginning of a statement and the names must be unique. The scope of these names extends across the entire statement.

core/sql/bin/SqlciErrors.txt
```diff
@@ -212,6 +212,9 @@
 1210 ZZZZZ 99999 UUUUUUUU UUUUU UUUUUUU Column $0~String0 is not allowed at position $1~Int0 in the SPLIT BY clause, because SPLIT BY must specify a prefix of the clustering key columns and the next clustering key column is $2~String1.
 1211 ZZZZZ 99999 UUUUUUUU UUUUU UUUUUUU The column values of the SPLIT BY first key in position $0~Int0 (one-based) are not greater (for ascending columns) or less (for descending columns) than its predecessor.
 1212 ZZZZZ 99999 UUUUUUUU UUUUU UUUUUUU An HBase region start key provided only part of the value for Trafodion key column $0~Int0. This could in some unusual situations cause errors.
+1213 ZZZZZ 99999 UUUUUUUU UUUUU UUUUUUU Unable to create or initialize a connection to Apache Hive.
```
zellerh: Yes, very good point. This is a security hole. A safer way is to do it is to use volatile tables, which are not yet implemented. I filed [TRAFODION-2316](https://issues.apache.org/jira/browse/TRAFODION-2316) for this issue. I also should have made that very clear in the description of this checkin - will fix that for the "real" pull request.

core/sql/comexe/ComTdbExeUtil.cpp
```diff
@@ -1302,6 +1302,7 @@ void ComTdbExeUtilFastDelete::displayContents(Space * space,
 ComTdbExeUtilHiveTruncate::ComTdbExeUtilHiveTruncate(
      char * tableName,
      ULng32 tableNameLen,
+     char * hiveTableName,
```
zellerh: Could you explain what you mean by "bound"? This is just added here to include the table name in the TDB, so we have access to it at runtime. Maybe you are saying that we could use a host variable for the name instead? Yes, that would be doable, but it's not currently done for the truncate command and it probably wouldn't be easy for most other commands such as load, etc. In this particular case, I'm just adding the Hive table name, the SQL statement uses the ANSI SQL name, e.g. HIVE.HIVE.T, and that name is also stored in the TDB.

core/sql/generator/GenPreCode.cpp
zellerh: The "is" and the "contains" refer to the entire statement. The Generator::isFastExtract() method is used to determine the overall statement type. What I needed was a flag indicating that there is some fast extract somewhere in the statement (e.g. the one used to populate a temp table), while the overall statement type is something else, like a select. I added a comment to make that more clear.

core/sql/comexe/ComTdbExeUtil.cpp
DaveBirdsall: By "bound at compile time" I mean the Hive table name is chosen at compile time rather than execution time. If the table name were chosen at execution time, we would put it in the TCB (or some other executor data structure) but not the TDB. If we chose the name at execution time, then the query plan could be shared in the query cache without worries.

DaveBirdsall: I think I see the reason for the confusion. We are using an existing operator that until now referred to Hive tables specified in the SQL statement itself. For those purposes, binding at compile time makes sense. I'm wondering though if for temporary tables whose data is used only within the scope of one execution (such as the ones you're introducing with CSE) should have their names bound at run time instead. So the ComTdbExeUtilHiveTruncate tdb would have to be extended to have two modes: One where the table name is passed to it by the generator and one where the table name is missing and some other token or mechanism to identify it is used instead (e.g. "temporary table number 3"). In the latter case the corresponding TCB would perhaps consult an executor global data structure to learn the table name.

PR #776, opened 2016-10-22
Fix for JIRA [TRAFODION-2305]

sbroeder:
After a region split the transactions to check against list is not fully populated


shangshengtung:
Looks good to me. Good change.

PR #777, opened 2016-10-23
[TRAFODION-2306] Trafodion customization using its own configuration …

selvaganesang:
…file

See JIRA for details


arvind-narain:
Looks good Selva. For the build error do please add the copyright information in the new traf-site.xml

svarnau:
Yes, traf-site.xml needs license statement. Also, wouldn't it be better to put it in the conf directory?

core/sql/src/main/java/org/trafodion/sql/TrafConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Adds Trafodion configuration files to a Configuration
+ */
+public class TrafConfiguration extends HBaseConfiguration {
+
+  static Logger logger = Logger.getLogger(TrafConfiguration.class.getName());
+
+  public static Configuration addTrafResources(Configuration conf) {
+    Configuration lv_conf = new Configuration();
+    String trafSiteXml = new String(System.getenv("MY_SQROOT") + "/etc/trafodion-site.xml");
```
zellerh: Just curious, shouldn't we search for the trafodion-site.xml file in the CLASSPATH?

selvaganesang:
I chose $MY_SQROOT/etc directory because the earlier way of configuring table coprocessors had a file traf_coprocessor.properties here. This file is removed in PR instead we use a file trafodion-site.xml. In addition, trrafodion environment variable file ms.env is also in the same directory. Currently, $MY_SQROOT/conf directory contains the configuration files of lo4j and log4cxx only

prashanth-vasudev:
+ 1

core/sql/src/main/java/org/trafodion/sql/TrafConfiguration.java
zellerh: Thanks.

sureshsubbiah:
Looks good.  +1.

selvaganesang:
This PR change was attributed to DamagedWalException (DEW) seen in region server (RS) logs in Esgyn internal testing. This is primarily a client side change except that co-processors are always added as the table attribute when a trafodion table is created.  Prior to this change, the coprocessors are added when the attribute hbase.coprocessor.region.classes is not visible on the client side. This is the case with CDH distro. However, the service side configuration should continue to see the attribute hbase.coprocessor.region.classes set to regionObserver, endpoint coprocessor and AggregateCoprocesor classes both in CDH and HDP distros. It is possible that RS can be influenced in some way due to double loading of these classes. But, it was confirmed by @narendragoyal  that the DWE is seen even when the coprocessor is not added as the table attribute. 

I read the change again to check if something is amiss with this change. I couldn't find anything.  Reviewers/Contributors, can you please review this change again to determine what can possibly DWE due to the change.

PR #782, opened 2016-10-25
[TRAFODION-2310] Changed soft down node processing to propagate node …

zcorrea:
…state change to remote monitor prior to killing processes.


sureshsubbiah:
+1

PR #783, opened 2016-10-25
[TRAFODION-2309] Memory leak observed in Repository context

arvind-narain:
Following changes:

1. Use REALLOCSQLMXHDLS() after each execute.
2. Avoid missing stats message in repository context
3. Handle memory leak in reprepare (delete of tmpSqlString)


sureshsubbiah:
Looks good @arvind-narain. Thank you. +1

arvind-narain:
Thanks Suresh, Selva. I'll check on T4/ODBC concern raised and create a separate JIRA if it is an issue.

PR #786, opened 2016-10-26
[TRAFODION-2313] dcscheck doesn't return in REST

mkby:


core/sqf/sql/scripts/dcscheck
```diff
@@ -138,15 +138,15 @@ if ( [ $sq_stat == 0 ] || [ $sq_stat == 1 ] ); then
   fi 
   echo
   if ( [ "$actual_dcsmaster_cnt" '!=' 0 ] && [ "$actual_dcsserver_cnt" '!=' 0 ] ); then
-   echo "ls $dcsznode"|$DCS_INSTALL_DIR/bin/dcs zkcli > $dcstmp
```
arvind-narain: Looks good. I remember there was some issue with a particular version of zookeeper where the command line argument to zkcli was not being processed, hence the echo approach was used instead. This should work.

PR #788, opened 2016-10-26
Traf properties

selvaganesang:


DaveBirdsall:
The changes look just right. Not sure why all those other commits show up in this pull request though. Probably harmless.

PR #790, opened 2016-10-26
[TRAFODION-2298] Update expected results for test compGeneral/TEST023

DaveBirdsall:
I neglected to notice that the expected results for this test needed to be updated with https://github.com/apache/incubator-trafodion/pull/770.


sureshsubbiah:
+1

PR #795, opened 2016-10-28
[TRAFODION-2270] support ORDER BY in PIVOT_GROUP/GROUP_CONCAT function

traflm:
to support ORDER BY in PIVOT_GROUP function, so as in GROUP_CONCAT function.

The idea is to introduce a new attribute in GroupBy, named as extraOrderExpr(). Which record the extra sort requirement for this groupby. 
In the optimizer, generate sort requirement for this groupby with the new extraOrderExpr() if available. 
This can be reused in some other functions as well, who need special order so can add the itemexpr into this extraOrderExpr().


core/sql/optimizer/RelGrby.h
```diff
@@ -552,7 +560,9 @@ class GroupByAgg : public RelExpr
   // ---------------------------------------------------------------------
   ItemExpr    * groupExprTree_;
   ValueIdSet  groupExpr_;
-
+  
+  ItemExpr  * extraGrpOrderby_;
+  ValueIdSet  extraOrderExpr_;
```
sureshsubbiah: I am confused how an order expression can be stored in a ValueIdSet. Should this be a ValueIdList instead?

The order (A,B) is different from (B,A) but both would be stored in the same ValueIdSet.

This same question applies to change in OptPhysRelExpr.cpp. There we take a ValueIdSet, convert it to a List and make a order requirement out of it. If order is important maybe it should start as List. If order is not important then we could add an arrangement requirement instead of a order requirement.

core/sql/optimizer/OptPhysRelExpr.cpp
```diff
@@ -11814,10 +11814,23 @@ void SortGroupBy::addArrangementAndOrderRequirements(
...
     // if we are in DP2
-    if (rg.getStartRequirements()->executeInDP2())
-      rg.addArrangement(groupExpr(),NO_SOT);
-    else
-      rg.addArrangement(groupExpr(),ESP_SOT);
+
+   if( NOT extraOrderExpr().isEmpty())
+   {
+     ValueIdList groupExprCpy(groupExpr());
+     for (ValueId vid = extraOrderExpr().init(); extraOrderExpr().next(vid);
```
sureshsubbiah: At this point does extraOrderExpr contain (studentname, test_score) or just (test_score). This statement uses the query in TEST002 that is part of this PR. I thought it would contain only test_score, if yes, then how we skip adding the arrangement requirement with GroupExpr? 

Also, is there a check somewhere that if such a clause is used we will always come though sort groupby and not hash groupby?

Do we also know that partial groupby rule and shortcut groupby rule will be prevented from firing? I assume we cannot get everything to work with partial groups or shortcut groupby.

core/sql/regress/executor/TEST002
```diff
@@ -1102,7 +1102,7 @@ SELECT student_name,
 
 SELECT student_name,
           GROUP_CONCAT(DISTINCT test_score
-                    ORDER BY test_score SEPARATOR '-')
+                    ORDER BY test_score desc SEPARATOR '-')
```
sureshsubbiah: Are we thinking that previously this ORDER BY was actually not doing its job (it was after all guarded by MODE_SPECIAL_4).  It would also be nice if we had a more complex example, just so that we can understand this feature and see it working. This is just a suggestion, please ignore if you do not agree.

core/sql/optimizer/OptPhysRelExpr.cpp
traflm: At this point, extraOrderExpr contains (test_score) only. The groupExpr is required by the SortGroupby, here we just add the "extra " sort key, groupExpr is required to do the sort groupby.
PIVOT_GROUP will force the optimizer to use sort groupby instead of hash groupby. As you can see, above code is in fact in SortGroupBy method. It is in the rule's match() method.
I ignore partial groupby and shortcut groupby. That must be an important issue, one possibility is to prevent them just as we prevent in hash groupby. This is a very critical question, thanks Suresh to remind me this. I need to study these two groupby methods.

core/sql/regress/executor/TEST002
traflm: Previous ORDER BY do nothing, it is ignored. So there will be regression failure , since the order is random. 
I can insert more rows into the table, above syntax is the most complex syntax for this feature. But I can design a more complex query to use join for example?

ORDER BY in an aggregate function is regarded as extraOrder requirement, that is the basic idea of this change. Because function is ItemExpr, so it cannot have its own sort requirement as I understand.
Let me first check the shortcut groupby and partial groupby issue

core/sql/optimizer/RelGrby.h
traflm: You are right Suresh. I will change it into ValueIdList, and avoid the ugly conversion in the OptPhysRelExpr.cpp

core/sql/optimizer/OptPhysRelExpr.cpp
traflm: It seems it works with partial groupby, 
As far as I understand, the shortcut groupby can only be fired if the aggreation function is min/max/anytrue, so it should not be fired as well.

traflm:
jenkins, extra tests

traflm:
@sureshsubbiah would you please help to review again?

core/sql/optimizer/OptPhysRelExpr.cpp
```diff
@@ -11814,10 +11814,22 @@ void SortGroupBy::addArrangementAndOrderRequirements(
   {
     // Shouldn't/Can't add a sort order type requirement
```
selvaganesang: Please remove these 2 comments too

traflm: I can make that change Selva.

core/sql/optimizer/OptPhysRelExpr.cpp
```diff
@@ -11814,10 +11814,22 @@ void SortGroupBy::addArrangementAndOrderRequirements(
...
     // if we are in DP2
-    if (rg.getStartRequirements()->executeInDP2())
-      rg.addArrangement(groupExpr(),NO_SOT);
-    else
-      rg.addArrangement(groupExpr(),ESP_SOT);
+
+   if( NOT extraOrderExpr().isEmpty())
+   {
+     ValueIdList groupExprCpy(groupExpr());
+     for (CollIndex i=0; i< extraOrderExpr().entries(); i++)
```
zellerh: Instead of this for loop, I think you can just do a single statement:
`groupExprCpy.insert(extraOrderExpr());`

core/sql/optimizer/OptPhysRelExpr.cpp
```diff
@@ -11856,6 +11868,7 @@ SortGroupBy::synthPhysicalProperty(const Context* myContext,
                                                              planNumber,
                                                              pws);
 
+      //PhysicalProperty(((isRollup() ||NOT extraOrderExpr().isEmpty()) ? ValueIdList() : sppOfChild->getSortKey()),
```
zellerh: Could you remove this dead code?

core/sql/optimizer/RelGrby.h
```diff
@@ -152,6 +154,9 @@ class GroupByAgg : public RelExpr
   inline void setGroupExpr(ValueIdSet &expr) { groupExpr_ = expr;}
   inline void addGroupExpr(ValueIdSet &expr) { groupExpr_ += expr;}
 
+  inline ValueIdList & extraOrderExpr() { return extraOrderExpr_; }
```
zellerh: Returning a writable reference to a data member isn't really a good idea. I know it's done in many other places, but it's still not good. Do you really need this?

core/sql/optimizer/RelGrby.h
```diff
@@ -552,7 +560,9 @@ class GroupByAgg : public RelExpr
   // ---------------------------------------------------------------------
   ItemExpr    * groupExprTree_;
   ValueIdSet  groupExpr_;
-
+  
+  ItemExpr  * extraGrpOrderby_;
+  ValueIdList  extraOrderExpr_;
```
zellerh: Anoop added the rollup feature recently, see https://github.com/apache/incubator-trafodion/pull/733. This also introduces an extra order by expression for a groupby. Maybe you and Anoop can decide whether you want to combine your code, for example by using this extra order for the ROLLUP groupby as well.

core/sql/optimizer/RelGrby.h
traflm: Yes, it is not a good practice. I need this since this valueID list will be populated (write) by convertToValueIdList(). Maybe I can rename this to extraOrderExprWritable(), So the extraOrderExpr() will not be accidentally written.

zellerh: No problem, only if it would have been easy... A better solution, if possible, could be a setExtraOrderExpr(const ValueIdList &newExtraOrder) method. But again, just wanted to share a thought.

core/sql/optimizer/OptPhysRelExpr.cpp
traflm: This is correct. Will fix it.

core/sql/optimizer/RelGrby.h
traflm: Yes, let me check Anoop's code and think about how to combine the code.

traflm: It is possible to combine, but not very simple. Since there are different purpose to add extra order, and in some cases, for example display, it needs to tell which orderby is used for rollup and which is for group_concat, it is possible but will complicated the code. So I prefer to keep it as for now.

traflm:
jenkins, extra tests

traflm:
no new error in full regression, known issues. so it is ok to be merged if no other comments.

PR #802, opened 2016-10-31
[TRAFODION-2324] Sequence operators are not parallel starting with th…

traflm:
…e WITH clause support

Important fields not copied, so sequence cannot work well together with current WITH clause implementation.


sureshsubbiah:
+1

PR #804, opened 2016-10-31
Testware change for compGeneral/TEST042

sureshsubbiah:


zellerh:
+1

This test is often a challenge to me, I question whether it provides any value for the tons of trouble it creates.

The test bundles several queries into a unit, then when we look at the aggregates we have no clue which query caused a change and whether that is a problem. It might be better to have a test that flags individual queries that change in their query cache status. Some day when we have lots of time to do this...

PR #805, opened 2016-10-31
[TRAFODION-1077] AQR info missing from GET STATISTICS

selvaganesang:
Automatic query retry info was populated, but it got reset as part
of successful retry prepare.


sureshsubbiah:
+1

PR #806, opened 2016-11-01
TRAFODION-2330 Using trafci, a select from a table succeeds even if t…

robertamarton:
…he user

               does not have the priv

There is a problem when the session user changes in a mxosrvr process.  The
existing compiler caches are not getting cleared so the new user will be
accessing the previous users' caches.  This could lead to allowing someone
that does not have privileges to gain access to an object.

The change is to clear all caches during a session user change operation.

### DIFF
--- a/core/sql/arkcmp/CmpContext.cpp
+++ b/core/sql/arkcmp/CmpContext.cpp
@@ -1157,4 +1157,16 @@ void CmpContext::resetLogmxEventSqlText()
    delete sqlTextBuf_ ;
    sqlTextBuf_ = NULL ;
 }
+
+void CmpContext::clearAllCaches()
+{
+   qcache_->makeEmpty();
+   schemaDB_->getNATableDB()->setCachingOFF();
+   schemaDB_->getNATableDB()->setCachingON();
+   schemaDB_->getNARoutineDB()->setCachingOFF();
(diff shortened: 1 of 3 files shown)


core/sql/arkcmp/CmpContext.cpp
```diff
@@ -1157,4 +1157,16 @@ void CmpContext::resetLogmxEventSqlText()
    delete sqlTextBuf_ ;
    sqlTextBuf_ = NULL ;
 }
+
+void CmpContext::clearAllCaches()
+{
+   qcache_->makeEmpty();
```
sureshsubbiah: The constructor for CmpContext seems to initialize histogramCache_ the same as qcache_. Is it possible qcache_ can be NULL at some point? It may be safer to guard it, this is a really minor nit to be handled later if necessary.

There is one more cache, optPCodeCache_ . Now this cache may not cause any priv violations I suppose. So a comment later here may be all we need, that PCode cache is not addressed by this method.

robertamarton:
I will add the change to check for NULL cache pointer and optPCodeCache comment with my next Trafodion delivery.  I did use the code from OptimizerSimulator  (OptimizerSimulator::cleanupSimulator) as a starting point for cleanAllCaches – should probably replace code in cleanupSinulator with a call to cleanupAllCaches.

PR #807, opened 2016-11-01
[TRAFODION-1839] Trafodion Installer Evolution

mkby:
new python version installer

### DIFF
--- a/install/python-installer/traf_dep.py
+++ b/install/python-installer/traf_dep.py
@@ -1,0 +1,119 @@
+#!/usr/bin/env python
+
+# @@@ START COPYRIGHT @@@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+# @@@ END COPYRIGHT @@@
+
+### this script should be run on all nodes with sudo user ###
+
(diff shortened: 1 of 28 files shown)


install/python-installer/traf_dep.py
```diff
@@ -0,0 +1,119 @@
...
+
+import re
+import os
+import sys
+import json
+import platform
+from common import run_cmd, cmd_output, err
+
+# not used
+EPEL_REPO = """
```
svarnau: This should not need to be hard-coded. Doing a yum install of "epel-release" package will get you the epel repo file.

install/python-installer/traf_package.py
```diff
@@ -0,0 +1,47 @@
...
+    dbcfgs = json.loads(dbcfgs_json)
+
+    TRAF_DIR = '%s-%s' % (dbcfgs['traf_basename'], dbcfgs['traf_version'])
+
+    # untar traf package
+    TRAF_PACKAGE_FILE = '/tmp/' + dbcfgs['traf_package'].split('/')[-1]
+    run_cmd('mkdir -p %s' % TRAF_DIR)
+    run_cmd('tar xf %s -C %s' % (TRAF_PACKAGE_FILE, TRAF_DIR))
+
+    print 'Trafodion package uncompressed successfully!'
```
svarnau: I think "extracted" would make more sense to me than "uncompressed". But opinions may vary.

install/python-installer/traf_dep.py
mkby: we only need 'pdsh' dependency from epel repository, so I'm thinking to install 'pdsh' rpm directly, but not use epel repository. Is it acceptable?

svarnau: Directly? I'm not sure what is meant, but I think it is best to install from epel repo.

gtapper: I was wondering the same. Directly seems to imply that we're distributing the pdsh package?

traflm:
A huge change, need sometime to review it. Is previous installer still work together with this new one? I mean, one can still use the old installer as well this new installer? Or it is a replacement abruptly?

mkby:
The two installer will coexist for a short of time, both of them can be used. Once the new one is stable and acceptable by users, we can remove the old one. There're still features need to be implemented for the new installer, for example, elasticity.

amandamoran:
+1 to Ming. Would also like a chance to review but if we are not making switch yet that's different.

traflm:
is this good to be merged? 
I cannot have any comments, would let it be tested.

PR #810, opened 2016-11-02
[TRAFODION-2293] remove minor version for the hbase-trx-apache_*.jar

mashengchen:


core/sqf/sql/scripts/install_local_hadoop
```diff
@@ -1554,7 +1554,7 @@ echo "$MY_LOCAL_SW_DIST/${HBASE_TAR}"
       org.apache.hadoop.hbase.security.visibility.VisibilityController"
   fi
 
-  if [[ "$HBASE_DISTRO" != "APACHE1.2" || "$HBASE_DISTRO" != "CDH5.7"]]; then 
+  if [[ "$HBASE_DISTRO" != "APACHE1.2" && "$HBASE_DISTRO" != "CDH5.7"]]; then
```
traflm: not able to determine if this work correctly or not, @mashengchen , please describe the UT you did after this change.
Since this only change '||' to '&&', but from what others reported, there are other issues like 'space' and the local_hadoop directory. So it is a must to run it yourself before I can approve this. 
So please describe how you did the UT.

DaveBirdsall: There needs to be a space added after "CDH5.7". Without that some shells give a syntax error.

mashengchen:
sorry, i was thinking it's an easy modification, and not test carefully..

PR #812, opened 2016-11-03
revert the bad commit to install_local_hadoop

traflm:
This is a temp fix to the install_local_hadoop issue,
only consider to merge it if an urgent fix is desired. A followup fix for it is in PR list, but still have issue to solve, and need more tests.


traflm:
no need for this PR anymore

PR #813, opened 2016-11-03
[TRAFODION-2229] add Hashing functions to Trafodion

traflm:
Compatibility enhancement 
add new functions for MySQL:
CRC32()	Compute a cyclic redundancy check value
MD5()	Calculate MD5 checksum
SHA1(), SHA()	Calculate an SHA-1 160-bit checksum
SHA2()	Calculate an SHA-2 checksum

Unit test is to test several input value, and invoke same query in MySQL, and compare the result. Pass the test since results are same.


zellerh:
+1
Looks good to me. Did you follow the guide that Dave Birdsall put together a while ago for adding new functions?

core/sql/parser/sqlparser.y
```diff
@@ -9671,6 +9676,62 @@ misc_function :
...
+     | TOK_CRC32 '(' value_expression ')'
+                {
+                    $$ = new (PARSERHEAP())
+                    BuiltinFunction(ITM_CRC32,
+                            CmpCommon::statementHeap(),
+                            1, $3);
+                }
+
+
+     | TOK_SHA1 '(' value_expression ')'
```
DaveBirdsall: Looks like this production is in there twice?

core/sql/parser/sqlparser.y
```diff
@@ -9671,6 +9676,62 @@ misc_function :
...
+                {
+                    $$ = new (PARSERHEAP())
+                    BuiltinFunction(ITM_SHA1,
+                            CmpCommon::statementHeap(),
+                            1, $3);
+                }
+
+     | TOK_SHA2 '(' value_expression ')'
```
DaveBirdsall: Looks like this production is here twice?

core/sql/common/OperTypeEnum.h
```diff
@@ -787,6 +787,11 @@ enum OperatorTypeEnum {
                         ITM_ISIPV6 = 2631,
                         ITM_INET_ATON = 2632,
                         ITM_INET_NTOA = 2633,
+                        ITM_CRC32     = 2634,
+                        ITM_MD5       = 2635,
+                        ITM_SHA1      = 2636,
+                        ITM_SHA2      = 2637,
+                        ITM_SHA       = 2638,
```
DaveBirdsall: ITM_SHA is never used. Looking at the parser changes, it appears that SHA is simply a syntactic sugar meaning the same as SHA1, so the difference between SHA and SHA1 disappears after parsing. So ITM_SHA is not needed.

core/sql/exp/exp_clause.h
```diff
@@ -203,7 +203,12 @@ class SQLEXP_LIB_FUNC  ex_clause : public NAVersionedObject {
     FUNC_ISIP_ID             = 118,
     FUNC_INETATON_ID         = 119,
     FUNC_INETNTOA_ID         = 120,
-    AGGR_GROUPING_ID         = 121
+    AGGR_GROUPING_ID         = 121,
+    FUNC_CRC32_ID            = 122,
+    FUNC_MD5_ID              = 123,
+    FUNC_SHA1_ID             = 124,
+    FUNC_SHA2_ID             = 125,
+    FUNC_SHA_ID              = 126
```
DaveBirdsall: FUNC_SHA_ID is never used and should be deleted.

core/sql/exp/exp_function.cpp
```diff
@@ -40,6 +40,10 @@
 
 
 #include <math.h>
+#include <zlib.h>
+#include "openssl/md5.h"
```
DaveBirdsall: I'm wondering why this isn't <openssl/md5.h>. Seems like one would have to copy the md5.h file into the source tree somewhere for this to compile cleanly. Maybe you meant to use angle brackets instead of quotes?

core/sql/exp/exp_function.cpp
```diff
@@ -7900,6 +7936,114 @@ ex_expr::exp_return_type ExFunctionInetNtoa::eval(char * op_data[],
...
+  Lng32 rlen = resultAttr->getLength();
+
+  memset(op_data[0], 0, rlen);
+
+  SHA256_CTX  sha_ctx;
+
+  SHA256_Init(&sha_ctx);  
+  SHA256_Update(&sha_ctx, op_data[1], slen);
+  SHA256_Final((unsigned char*) sha,&sha_ctx); 
+  char tmp[2];
```
DaveBirdsall: Looks too short. The sprintf will write three bytes (two digits + a null terminator), so it will overrun this buffer.

core/sql/exp/exp_function.cpp
```diff
@@ -7900,6 +7936,114 @@ ex_expr::exp_return_type ExFunctionInetNtoa::eval(char * op_data[],
...
+  Lng32 slen = srcAttr->getLength(op_data[-MAX_OPERANDS+1]);
+  Lng32 rlen = resultAttr->getLength();
+  memset(op_data[0], 0, rlen);
+
+  SHA_CTX  sha_ctx;
+
+  SHA1_Init(&sha_ctx);  
+  SHA1_Update(&sha_ctx, op_data[1], slen);
+  SHA1_Final((unsigned char*) sha,&sha_ctx); 
+  char tmp[2];
```
DaveBirdsall: This buffer is too short.

core/sql/exp/exp_function.cpp
```diff
@@ -7900,6 +7936,114 @@ ex_expr::exp_return_type ExFunctionInetNtoa::eval(char * op_data[],
...
+  Lng32 slen = srcAttr->getLength(op_data[-MAX_OPERANDS+1]);
+  Lng32 rlen = resultAttr->getLength();
+
+  memset(op_data[0], 0, rlen);
+  MD5_CTX  md5_ctx;
+
+  MD5_Init(&md5_ctx);  
+  MD5_Update(&md5_ctx, op_data[1], slen);
+  MD5_Final((unsigned char*) md5,&md5_ctx); 
+  char tmp[2];
```
DaveBirdsall: This buffer is too short.

core/sql/parser/sqlparser.y
traflm: thanks Dave for catching this, this is a bug, will fix.

core/sql/exp/exp_clause.h
traflm: will fix

core/sql/exp/exp_function.cpp
traflm: agree, will fix

traflm:
@DaveBirdsall please help to review again, I update according to your comments. And add a few regression tests.

selvaganesang:
+1 Nice change

traflm:
Thanks all for the review, I will merge this if no other comments. I fixed all previous comments in last commit.

DaveBirdsall:
I will take a quick look now...

PR #814, opened 2016-11-03
[TRAFODION-2331] UPD STATS sometimes fails silently after some other …

DaveBirdsall:
…failure

There are three changes in this check-in:

1. UPDATE STATISTICS does a ROLLBACK after detecting a SQL error. Sometimes the ROLLBACK fails, because the Executor has already rolled back the transaction. This caused UPDATE STATISTICS to get confused about transaction state. This has been fixed; now after an error on ROLLBACK, UPDATE STATISTICS simply checks to see if a transaction is still active, and if not, just ignores the error.

2. When executing DDL statements, UPDATE STATISTICS tries to start a transaction. If it is confused about transaction state (as in item 1 above), it thinks this attempt to start a transaction has failed. In that particular case, no error was being reported. This has been fixed.

3. Some obsolete code that used static SQL for BEGIN WORK, COMMIT and ROLLBACK has been removed.

### DIFF
--- a/core/sql/ustat/hs_cli.cpp
+++ b/core/sql/ustat/hs_cli.cpp
@@ -391,8 +391,17 @@ Lng32 HSFuncExecDDL( const char *dml
   HSTranMan *TM;
   NABoolean startedTrans = FALSE;
   TM = HSTranMan::Instance();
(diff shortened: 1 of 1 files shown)


zellerh:
+1

PR #815, opened 2016-11-03
[TRAFODION-2326] TopN sort reuse of released tuples

prashanth-vasudev:


core/sql/executor/ex_sort.cpp
```diff
@@ -96,6 +96,72 @@ ExOperStats * ExSortTcb::doAllocateStatsEntry(CollHeap *heap,
...
+  // maximum of GEN_SORT_MAX_BUFFER_SIZE. The memory quota system will
+  // come into force for additional buffers following this initial buffer.
+  Lng32 numBuffs = (sortTdb().sortOptions_->memoryQuotaMB()<= 0)?sortTdb().numBuffers_:2;
+  Lng32 numSortBuffs = 0;
+
+  // need separate pools for sorting and saving result rows
+  if (sortTdb().partialSort())
+  {
+     // give result pool and sort pool each half of the space
+     numSortBuffs = numBuffs = (numBuffs + 1)/2;
```
DaveBirdsall: So, if numBuffs is set to 3 at line 111, then line 118 will set numSortBuffs to 2 and numBuffs to 2, exceeding the limit of 3. Was that your intent?

core/sql/executor/ex_sort.cpp
```diff
@@ -183,6 +212,11 @@ ExSortTcb::ExSortTcb(const ExSortTdb & sort_tdb,
   processedInputs_ = qparent_.down->getTailIndex();
   workAtp_ = allocateAtp(sort_tdb.workCriDesc_, space);
   workAtp_->getTupp(2) = new(space) tupp_descriptor();
+  
+  //buffer pools are allocated in SORT_PREP work phase.
+  topNSortPool_ = NULL;
+  regularSortPool_ = NULL;
+  partialSortPool_ = NULL;
```
DaveBirdsall: Should we set receivePool_ to NULL here also?

core/sql/executor/ex_sort.cpp
```diff
@@ -330,21 +364,39 @@ ExSortTcb::~ExSortTcb()
...
+  {
+    delete regularSortPool_;
+    regularSortPool_ = NULL;
+  }
+  if (topNSortPool_)
+  {
+    delete topNSortPool_;
+    topNSortPool_ = NULL;
+  }
+  if (sortPool_ && (sortPool_ != receivePool_))
```
DaveBirdsall: What if sortPool_ is not null and  sortPool_ == receivePool_? Shouldn't we set it to NULL in this case (without doing a delete)?

core/sql/executor/ex_sort.cpp
DaveBirdsall: Should we set sortPool_ to NULL here also?

core/sql/executor/ex_sort.h
```diff
@@ -356,4 +363,119 @@ class ExSortFromTopPrivateState : public ex_tcb_private_state
   ~ExSortFromTopPrivateState();       // destructor
 };
 
+class ExSortBufferPool : public NABasicObject
```
DaveBirdsall: A more object-oriented way to do this is to create a class hierarchy, with ExSortBufferPool as the abstract base class, and two specializations, one for simple buffers and the other for SQL buffers. Then all your methods are virtual and you get rid of all the "if (sqlBufferPool_)" checks. The code will be slightly more efficient too though probably not significantly so.

core/sql/executor/ex_sort.cpp
prashanth-vasudev: numBuffs is the initial number to start with. There is another constant max_num_buffs that determine the max buffers that can be used, controlled by cqd.  In the current case, the thinking is to divide the numbuffs to two halfs. This is done only for partialSort feature which is not enabled by default.

core/sql/executor/ex_sort.cpp
prashanth-vasudev: I will make this change. The thinking was to delete the memory once, either by sortPool_ or receivePool_.  setting them to null seems to be a no op since it is called only in the destructor.

core/sql/executor/ex_sort.h
prashanth-vasudev: Creating a abstract class here was considered as the first option. However, making this change would have ripple effect all over the code since sql_buffer_pool is packed and unpacked when hopping many operators. If a major change is made to sql buffer management in future, it would be nice to consider having a abstract class. For now, to keep the change concise, this approach was taken.

prashanth-vasudev:
There is a test failure related to CIF and TopN sort, I am looking into this issue, will commit a change to this PR.

prashanth-vasudev:
jenkins, extra tests

prashanth-vasudev:
jenkins, extra tests

prashanth-vasudev:
The following two tests in compGeneral has failed not due to this change.  Rest of the extra tests pass.
2016-11-07 19:54:37  TEST023  ### FAIL (6 lines)     ###
2016-11-07 19:55:35  TEST042  ### FAIL (78 lines (vs. 88 known))     ###

sureshsubbiah:
+1. Looks good.

selvaganesang:
There might be memory leak if the buffers are initialized in the workUp again and again. Please check if there is a memory leak with prepare and execute many times of a SQL command that involves sort.

prashanth-vasudev:
Very good catch selvaganesang. I have push a new change that addresses reuse of allocated pools.

prashanth-vasudev:
jenkins, extra tests

prashanth-vasudev:
jenkins, extra tests

prashanth-vasudev:
test.java.org.trafodion.phoenix.end2end.QueryExecTest passes when run in workstation. Will do one last extra tests to validate again. 
CompGeneral tests below fail due to known issues unrelated to this change.
TEST023 ### FAIL (6 lines) ###
TEST042 ### FAIL (78 lines (vs. 88 known)) ###

prashanth-vasudev:
jenkins, extra tests

DaveBirdsall:
jenkins, extra tests

prashanth-vasudev:
This PR is ready for merge. The only failures are compGeneral failures unrelated to this PR change.
2016-11-16 02:51:09  TEST023  ### FAIL (6 lines)     ###
2016-11-16 02:52:04  TEST042  ### FAIL (78 lines (vs. 88 known))     ###

PR #816, opened 2016-11-03
[TRAFODION-2332] SequenceFileReader cleanup.

prashanth-vasudev:
Fixes cleanup of SequenceFileReader instance. Without this fix, we see this error where jvm is out of memory,

2016-10-28 17:39:00,079, ERROR, SQL, Node Number: 0, CPU: 0, PIN: 55991, Process Name: $Z001APR, SQLCODE: 8447, QID: MXID11000020372212344459045477158000000000106U3333300_4993_HS_CLI_DYNSTMT, *** ERROR[8447] An error occurred during hdfs access. Error Detail: JNI NewObject() failed


zellerh:
+1

PR #817, opened 2016-11-04
[TRAFODION-2328]There is no active DcsMaster in the output of the dcscheck when stop the lead zookeeper that is the active DcsMaster

orrk:
The root cause is the parser only get one quorum
Using the apache.commcon.cli to enhance it and can input arguments to get more information
-lq :list all quorums in of zookeeper
-la :list all configuration of zookeeper
-key [val] : get specified configuration value of zookeeper

### DIFF
--- a/dcs/src/main/java/org/trafodion/dcs/zookeeper/ZooKeeperMainServerArg.java
+++ b/dcs/src/main/java/org/trafodion/dcs/zookeeper/ZooKeeperMainServerArg.java
@@ -24,8 +24,14 @@ under the License.
 package org.trafodion.dcs.zookeeper;
 
 import java.util.Properties;
+import java.util.Iterator;
 import java.util.Map.Entry;
 
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.CommandLineParser;
+import org.apache.commons.cli.Options;
+import org.apache.commons.cli.ParseException;
+import org.apache.commons.cli.PosixParser;
 import org.apache.hadoop.conf.Configuration;
 import org.trafodion.dcs.util.DcsConfiguration;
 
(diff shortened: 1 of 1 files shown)


traflm:
As commented by Arvind in the JIRA 2328, this single change in command line parser seems not a full solution? Introducing new parameter, but I don't see where it is used. Do you need to update dcscheck script as well?
Could you explain how unit test of this is done? So maybe I can better understand the solution. 
Also, I paste Arvind's comments from JIRA to here, you may need to take a look and feedback:

Thanks for working on this defect. Could you share how these new options ( la, lq, key ) will be used ?
For the problem seen when one of the zookeeper nodes/process among the quorum is down, I think we need to pass in the quorum information to the "-server" option. Currently we pick one server and port and pass that. With the change if we invoke "dcs zkcli" it should default to showing all servers of the quorum like hbase zkcli does:
> hbase zkcli
Connecting to wmstesting-1.novalocal:2181,wmstesting-3.novalocal:2181,wmstesting-2.novalocal:2181...
>bin/dcs zkcli
Connecting to wmstesting-1.novalocal:2181
If you do make this change to -server, then note that dcscheck and dcsstart will need to be modified to pick the correct row/column.
Other suggestion - we could eliminate the configuration in dcs-site.xml by utilizing the following env variables that get set in a cluster environment:
[trafodion@wmstesting-1 dcs-2.2.0]$ env|grep ZOO
ZOOKEEPER=zookeeper
ZOO_PORT_NODES=wmstesting-2.novalocal:2181,wmstesting-1.novalocal:2181,wmstesting-3.novalocal:2181
ZOOKEEPER_PORT=2181
ZOOKEEPER_NODES=wmstesting-3.novalocal,wmstesting-1.novalocal,wmstesting-2.novalocal

selvaganesang:
@orrk Is this PR still active?

PR #818, opened 2016-11-04
[TRAFODION-2338]Try to close connection while got exception

kevinxu021:
Even this is a packing issue, it should not take up so many connection.
While got this error, Trafci will retry to setup a new session and  doesn't close existing connection.
The fix is to close previous connection in case that it take up more connections.

### DIFF
--- a/core/conn/trafci/.gitignore
+++ b/core/conn/trafci/.gitignore
@@ -2,3 +2,4 @@ buildId
 temp/
 utils/trafci.cmd
 utils/trafci.sh
+/target/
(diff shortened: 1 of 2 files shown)


selvaganesang:
Looks good. But the connection would be closed for successful case too. If that is the intention, then it is good to go. But the description doesn't match with the changed code

kevinxu021:
@selvaganesang Yes, as you see, it's in finally block, then no worries about success or not. for the description, it matched, if you see the changes is in a retry-block, you would realize that if there is an exception after connection created, the connection should be closed. Otherwise, the connection would be wild.

ryzuo:
The finally block is at the end of main function, right before exit. Looks good to me.
@selvaganesang Is it good to go?

selvaganesang:
Yes. My comment was referring to the fact that connection will be closed for bothe success and exception paths. If it is the intent, then please go ahead and merge it in

PR #819, opened 2016-11-04
[TRAFODION-2334] Installer should not set namenode heap size

svarnau:
This was only being done for CDH. This change removes that HDFS setting.


amandamoran:
I agree with @traflm 
+1

PR #821, opened 2016-11-07
[TRAFODION-2340]need a installer for windows ODB

xwq:


xwq:
@hegdean 
Hi, Anu. Please check the details like application like application name, version number before merging it. I think we need be careful with these message since the user can see them.

hegdean:
Changes look good to me.

PR #826, opened 2016-11-09
{TRAFODION-1562] Index maintenence performance improvements

sandhyasun:
Changes to transform an upsert plan into an efficient plan when indexes are present and allow for rowset type operations.  The plan makes use of existing operators to eliminate duplicates at the source so we don't need to perform a row at a time check and put operation. The code is currently turned off with a CD traf_upsert_to_eff_tree.

### DIFF
--- a/core/sql/optimizer/BindRelExpr.cpp
+++ b/core/sql/optimizer/BindRelExpr.cpp
@@ -2424,6 +2424,7 @@ Scan *RelExpr::getScanNode(NABoolean assertExactlyOneScanNode) const
   return (Scan *)result;
 }
 
+
 Scan *RelExpr::getLeftmostScanNode() const
 {
   RelExpr *result = (RelExpr *)this;   // cast away constness, big whoop
@@ -2434,6 +2435,32 @@ Scan *RelExpr::getLeftmostScanNode() const
   }
 
   return (Scan *)result;
+}
+
+
+
+Join * RelExpr::getLeftJoinChild() const
+{
+  RelExpr *result = (RelExpr *)this;
+  
+  while(result)
+    {
+      if (result->getOperatorType() == REL_LEFT_JOIN) break;
+      result = result->child(0);
+    }
+  return (Join *)result;
+}
+
+RelSequence* RelExpr::getOlapChild() const
+{
+  RelExpr *result = (RelExpr *)this;
+  
+  while(result)
+    {
+      if (result->getOperatorType() == REL_SEQUENCE) break;
+      result = result->child(0);
+    }
+  return (RelSequence *)result;
 }
 
 // QSTUFF
@@ -9672,6 +9699,10 @@ RelExpr *Insert::bindNode(BindWA *bindWA)
       return boundExpr;
     }
   }
+
+ 
+
+
    
   bindChildren(bindWA);
   if (bindWA->errStatus()) return this;
@@ -9806,6 +9837,8 @@ RelExpr *Insert::bindNode(BindWA *bindWA)
      if (bulkLoadIndex) {
        setRETDesc(child(0)->getRETDesc());
       } 
+
+
 
     for (i = 0; i < tgtColList.entries() && i2 < newTgtColList.entries(); i++) {
       if(tgtColList[i] != newTgtColList[i2])
@@ -10298,10 +10331,17 @@ RelExpr *Insert::bindNode(BindWA *bindWA)
   }
 
   if (isUpsertThatNeedsMerge(isAlignedRowFormat, omittedDefaultCols, omittedCurrentDefaultClassCols)) {
-    boundExpr = xformUpsertToMerge(bindWA);
-    return boundExpr;
+    if (CmpCommon::getDefault(TRAF_UPSERT_TO_EFF_TREE) == DF_OFF) 	
+      {
+	boundExpr = xformUpsertToMerge(bindWA);  
+	return boundExpr;
+      }
+      else if( CmpCommon::getDefault(TRAF_UPSERT_TO_EFF_TREE) == DF_ON)
+	boundExpr = xformUpsertToEfficientTree(bindWA);
+    
+    
   }
-  else if (NOT (isMerge() || noIMneeded()))
+   if (NOT (isMerge() || noIMneeded()))
     boundExpr = handleInlining(bindWA, boundExpr);
 
   // turn OFF Non-atomic Inserts for ODBC if we have detected that Inlining is needed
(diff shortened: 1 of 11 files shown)


core/sql/optimizer/Inlining.cpp
```diff
@@ -618,25 +619,58 @@ RETDesc *GenericUpdate::createOldAndNewCorrelationNames(BindWA *bindWA, NABoolea
...
+			  bindWA->wHeap(),
+			  OLDCorr);
+	
+        //	ColumnDescList *colList = (olapChild->getRETDesc())->getColumnList();
+	for (short i = 0; i< olapChild->getRETDesc()->getDegree();i++)
+	  {
+	    // we remembered if the original columns was from the right side of
+	    // this olap node so add those to the RetDesc since those are the 
+	    //ones we want to delete from the dependent indexes.
+	    if ((olapChild->getRETDesc()->getValueId(i)).getItemExpr()->origOpType() == ITM_INSTANTIATE_NULL)
```
zellerh: The left side of the left join could also produce some ITM_INSTANTIATE_NULL expressions, couldn't it? Wouldn't that happen if the select part of the upsert is itself a left join? Maybe get to the left join somehow and use its nullInstantiatedOutput() method, then map that through the sequence node's map?

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -10383,6 +10423,310 @@ NABoolean Insert::isUpsertThatNeedsMerge(NABoolean isAlignedRowFormat, NABoolean
...
+  ValueIdList tablePKeyVals = NULL;
+  ValueIdList sourcePKeyVals = NULL;
+  
+  pkeyVals->convertToValueIdList(tablePKeyVals,bindWA,ITM_ITEM_LIST);
+  updateToSelectMap().mapValueIdListDown(tablePKeyVals,sourcePKeyVals);
+  
+
+
+  Join *lj = new(bindWA->wHeap()) Join(child(0),targetTableScan,REL_LEFT_JOIN,keyPred);
+  lj->doNotTransformToTSJ();
```
zellerh: I assume this is because we want to use the old state of the target table, before we applied any updates. But, what if I do an upsert of 1000 rows into a table with 10 billion rows? Do we really want to read 10 billion rows with all their columns into a hash join or merge join? The merge would be vastly more efficient for this case - or am I missing something?

Another way to see the old state of the table would be to add a sort node above the join, so in this case we would only need to sort 1000 rows and we may need to do a sort anyway for the sequence. So, maybe just add code that forces a sort operator for the required order of the sequence node (at least forces it if we use a nested join or a merge join w/o a sort).

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -2436,6 +2437,32 @@ Scan *RelExpr::getLeftmostScanNode() const
...
+
+
+Join * RelExpr::getLeftJoinChild() const
+{
+  RelExpr *result = (RelExpr *)this;
+  
+  while(result)
+    {
+      if (result->getOperatorType() == REL_LEFT_JOIN) break;
+      result = result->child(0);
```
sandhyasun: This is cloned from the existing method getLeftMostScanNode(). They can atmost return NULL which is ok.

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -2436,6 +2437,32 @@ Scan *RelExpr::getLeftmostScanNode() const
...
+    }
+  return (Join *)result;
+}
+
+RelSequence* RelExpr::getOlapChild() const
```
sandhyasun: getOlapChild() is the only new method corrently used. The LeftJoin access method isn't used for now. I am going to comment it out. But if I do, will make generic one.

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -10298,10 +10331,17 @@ RelExpr *Insert::bindNode(BindWA *bindWA)
...
 
   if (isUpsertThatNeedsMerge(isAlignedRowFormat, omittedDefaultCols, omittedCurrentDefaultClassCols)) {
-    boundExpr = xformUpsertToMerge(bindWA);
-    return boundExpr;
+    if (CmpCommon::getDefault(TRAF_UPSERT_TO_EFF_TREE) == DF_OFF) 	
+      {
+	boundExpr = xformUpsertToMerge(bindWA);  
+	return boundExpr;
+      }
+      else if( CmpCommon::getDefault(TRAF_UPSERT_TO_EFF_TREE) == DF_ON)
```
sandhyasun: Will fix al indentations.

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -10298,10 +10331,17 @@ RelExpr *Insert::bindNode(BindWA *bindWA)
...
+	boundExpr = xformUpsertToEfficientTree(bindWA);
+
```
sandhyasun: Yes removed the second "if" so it can default to one of the cases.

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -10383,6 +10423,310 @@ NABoolean Insert::isUpsertThatNeedsMerge(NABoolean isAlignedRowFormat, NABoolean
      return FALSE;
 }
 
+#ifdef __ignore
```
sandhyasun: Put it in comments with a clear comment that we could try out a perf enhancement in the future.

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -10383,6 +10423,310 @@ NABoolean Insert::isUpsertThatNeedsMerge(NABoolean isAlignedRowFormat, NABoolean
...
+*/
+         
+RelExpr* Insert::xformUpsertToEfficientTree(BindWA *bindWA) 
+{
+  NATable *naTable = bindWA->getNATable(getTableName());
+  if (bindWA->errStatus())
+    return NULL;
+  if ((naTable->getViewText() != NULL) && (naTable->getViewCheck()))		
+    {		
+      *CmpCommon::diags() << DgSqlCode(-3241)
```
sandhyasun: Yes if the CQD is off it calls the old method to revert to merge.

core/sql/optimizer/BindRelExpr.cpp
zellerh: Thanks, I checked and it seems like we do not blow up if we call child(0) on a leaf node.

core/sql/optimizer/BindRelExpr.cpp
zellerh: I think commented-out code is something people don't like. I would leave it or delete it.

core/sql/optimizer/BindRelExpr.cpp
zellerh: What about the case where the CQD is ON and we reach like 10437 here. I would assume that this would cause an error? Could we go back to the old method instead of giving an error?

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -10383,6 +10423,310 @@ NABoolean Insert::isUpsertThatNeedsMerge(NABoolean isAlignedRowFormat, NABoolean
...
+      baseCol = (BaseColumn *)(tableCols[i].getItemExpr()) ;
+      if (baseCol->getNAColumn()->isSystemColumn())
+	continue;
+
+      targetColRef = new(bindWA->wHeap()) ColReference(
+						       new(bindWA->wHeap()) ColRefName(
+										       baseCol->getNAColumn()->getFullColRefName(), bindWA->wHeap()));
+    
+
+      if (baseCol->getNAColumn()->isClusteringKey())
```
sandhyasun: The system columns like SALT and DIVISION don't need to be part of the join key, I suppose. Also I checked that we don't do this transformation in case the table has a syskey. Method Insert::isUpsertThatNeedsMerge(() ensures this. So I think we are ok with this ..

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -10383,6 +10423,310 @@ NABoolean Insert::isUpsertThatNeedsMerge(NABoolean isAlignedRowFormat, NABoolean
...
+	    {
+	      keyPred = new(bindWA->wHeap()) BiLogic(ITM_AND,
+						     keyPredPrev,
+						     keyPred);  
+	    }
+	  pkeyValPrev = pkeyVals;
+    
+	  pkeyVals = tableCols[i].getItemExpr();
+	  
+	  if (i > 0)
```
sandhyasun: done

core/sql/optimizer/BindRelExpr.cpp
sandhyasun: Not sure how exactly to address this now. Seems like there is some choices of the JOIN we need to do here . Can we treat it as a performance enhancement and I keep it as a subtask TBD in this JIRA?

core/sql/optimizer/Inlining.cpp
sandhyasun: We are not remembering the origOpType in any other operator. So I think we may be ok with this. Only in the RelSequence operator, we remeber what the original opType was.

core/sql/optimizer/BindRelExpr.cpp
zellerh: Sorry, I did not check the old method. Please ignore my comment, I should have checked.

core/sql/optimizer/BindRelExpr.cpp
zellerh: Yes, sorry again, I messed up on this one as well. Please ignore this comment, too, it is wrong. Good to know that we don't come here in the one case where it would be an issue, with SYSKEY.

core/sql/optimizer/BindRelExpr.cpp
zellerh: Wouldn't you agree that this could be a really serious performance issue, making this "optimization" 1000s of times worse than the original method?

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -10383,6 +10423,310 @@ NABoolean Insert::isUpsertThatNeedsMerge(NABoolean isAlignedRowFormat, NABoolean
...
+ 
+  // Map the table's primary key values to the source lists key values
+  ValueIdList tablePKeyVals = NULL;
+  ValueIdList sourcePKeyVals = NULL;
+  
+  pkeyVals->convertToValueIdList(tablePKeyVals,bindWA,ITM_ITEM_LIST);
+  updateToSelectMap().mapValueIdListDown(tablePKeyVals,sourcePKeyVals);
+  
+  Join *lj = new(bindWA->wHeap()) Join(child(0),targetTableScan,REL_LEFT_JOIN,keyPred);
+  lj->doNotTransformToTSJ();
```
sureshsubbiah: As @zellerh and @sandhyasun discussed in person, lines 10526 and 10527 seem unnecessary. We would like this LJ to be implemented as a TSJ in certain cases. Also this join is definitely not for WRITE. So the flag in 10527 seems incorrect.

sandhyasun: Removing this restriction

core/sql/optimizer/Inlining.cpp
```diff
@@ -618,25 +619,58 @@ RETDesc *GenericUpdate::createOldAndNewCorrelationNames(BindWA *bindWA, NABoolea
...
-	getUpdateCKorUniqueIndexKey())
+      getUpdateCKorUniqueIndexKey() ||
+      ((getOperatorType() == REL_UNARY_INSERT) &&((Insert *)this)->isMerge()) ||
+      ((getOperatorType() == REL_UNARY_INSERT) && ((Insert *)this)->isUpsert() && (CmpCommon::getDefault(TRAF_UPSERT_TO_EFF_TREE) == DF_ON )))  
   {
     // DELETE or UPDATE --
     // Now merge the old/target/before valueid's (the Scan child RETDesc)
     // into this RETDesc such that these cols are all named "OLD@.<col>"
-    //
+    //f
```
sandhyasun: removed

core/sql/optimizer/RelSequence.cpp
```diff
@@ -1179,20 +1179,21 @@ RelExpr *RelSequence::bindNode(BindWA *bindWA)
...
     newColumn->synthTypeAndValueId();
+    newColumn->setOrigOpType(columnValueId.getItemExpr()->origOpType());
     
     resultTable->addColumn(bindWA,
       colList->at(i)->getColRefNameObj(),
       newColumn->getValueId(),
       USER_COLUMN,
       colList->at(i)->getHeading());
-
+    ncToOldMap.addMapEntry(columnValueId,newColumn->getValueId());
```
sandhyasun: Done

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -10383,6 +10423,310 @@ NABoolean Insert::isUpsertThatNeedsMerge(NABoolean isAlignedRowFormat, NABoolean
...
+  ((ItmLeadOlapFunction *)leadItem)->setIsOLAP(TRUE);
+  boundLeadItem = leadItem->bindNode(bindWA);
+  if (bindWA->errStatus()) return this;
+  boundLeadItem->convertToValueIdSet(sequenceFunction);
+  seqNode->setSequenceFunctions(sequenceFunction);
+  
+  // Add a selection predicate (post predicate) to check if the LEAD item is NULL
+  ItemExpr *selPredOnLead = NULL;
+  selPredOnLead = new (bindWA->wHeap()) UnLogic(ITM_IS_NULL,leadItem);
+  selPredOnLead->bindNode(bindWA);
```
sandhyasun: Done

core/sql/optimizer/BindRelExpr.cpp
sandhyasun: Removed the restriction of the Join so we allow NJ. We do have a blocking sort operator on the left side which will disallow upserts from happening before the suplicates are removed on the right.. It isn't clear which option to go with for this performance. So will open a sub-JIRA for this case  to be addressed separately since it needs a little more work in the optimizer.

selvaganesang:
Though PR is already merged, please consider adding the following test:

Create a non-aligned table with columns having a default value as current_date as default value with index on this column.
Issue upsert command with missing default value column.

Consider doing the above with a non-date default columns also,

PR #827, opened 2016-11-09
[TRAFODION-2256] installer not detecting true init errors

svarnau:
Using Amanda's recommended fix.

### DIFF
--- a/install/installer/traf_start
+++ b/install/installer/traf_start
@@ -58,7 +58,7 @@ else
        error1392=$(grep "1392" $SQ_ROOT/initializeErrors.txt | wc -l)
        error1395=$(grep "1395" $SQ_ROOT/initializeErrors.txt | wc -l)
        if [ "$error1395" -ge "1" ] || [ "$error1392" -ge "1" ]; then
-          ssh $STARTING_NODE "cd $SQ_ROOT/sql/scripts;  echo \"get version of metadata;\" | sqlci" | tee -a $SQ_ROOT/initializeErrors.txt
+           ssh $STARTING_NODE "cd $SQ_ROOT/sql/scripts;  echo \"get version of metadata;\" | sqlci" | tee $SQ_ROOT/initializeErrors.txt
            metadataCurrent=$(grep "Metadata is current" $SQ_ROOT/initializeErrors.txt | wc -l)
            if [[ "$metadataCurrent" -ne "1" ]]; then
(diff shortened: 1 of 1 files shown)


install/installer/traf_start
```diff
@@ -58,19 +58,19 @@ else
...
            if [[ "$metadataCurrent" -ne "1" ]]; then
               ssh $STARTING_NODE "cd $SQ_ROOT/sql/scripts;  echo \"initialize Trafodion, upgrade;\" | sqlci" | tee -a $SQ_ROOT/initializeErrors.txt
            fi
-        fi
-
-        if [ "$?" != "0" ]; then
-           echo "***ERROR: initialize Trafodion failed.  Check install log files for details." | tee -a $INSTALL_LOG
-           exit -1;
-        fi
+       else
```
zellerh: What if the initialize trafodion, upgrade fails, shouldn't we check that for errors as well? This else would prevent that as far as I can see. Maybe better to remove the "-a" from the tee command on line 64 and overwrite file initializeErrors.txt. Then remove this else statement on line 66 and continue unconditionally?

install/installer/traf_start
```diff
@@ -58,19 +58,18 @@ else
...
            metadataCurrent=$(grep "Metadata is current" $SQ_ROOT/initializeErrors.txt | wc -l)
            if [[ "$metadataCurrent" -ne "1" ]]; then
-              ssh $STARTING_NODE "cd $SQ_ROOT/sql/scripts;  echo \"initialize Trafodion, upgrade;\" | sqlci" | tee -a $SQ_ROOT/initializeErrors.txt
+              ssh $STARTING_NODE "cd $SQ_ROOT/sql/scripts;  echo \"initialize Trafodion, upgrade;\" | sqlci" | tee $SQ_ROOT/initializeErrors.txt
            fi
-        fi
-
-        if [ "$?" != "0" ]; then
+       fi
+       anyError=$(grep "ERROR" $SQ_ROOT/initializeErrors.txt | wc -l)
```
hegdean: Since the error file is being appended for upgrade, the errors txt file will contain 1395 failures. We want the install to proceed and not exit

svarnau: Agreed. If initialization is already done and upgrade is not needed, then there is a problem of flagging 1395 as error.  One more change coming to overwrite the error file when checking metadata version.

PR #828, opened 2016-11-10
[TRAFODION-2342] hdfs access threads init/cleanup changes

prashanth-vasudev:
In the effort to identify zombie threads , there were few places in code that require init and cleanup calls be made that create and destroy the hdfs access threads. This set of changes is initial set. More investigation to identify root cause of zombie threads in progress.

### DIFF
--- a/core/sql/executor/ExExeUtilMisc.cpp
+++ b/core/sql/executor/ExExeUtilMisc.cpp
@@ -2275,13 +2275,6 @@ void ExExeUtilHiveTruncateTcb::freeResources()
 Int32 ExExeUtilHiveTruncateTcb::fixup()
 {
   lobGlob_ = NULL;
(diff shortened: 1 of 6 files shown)


selvaganesang:
It is not clear what is the intent of moving ExpLobInterfaceInit method from fixup to work method. By moving to the work method, we would be setting up the threads for every execution. This increases the probability of having more zombie threads

prashanth-vasudev:
Removed init calls from fixup since there is chance of re-fixups. There were cleanup calls in some cases in work method, in these cases, matched the init call in work method. In case of hdfsscan.cpp, the cleanup call is in destructor, in this case moved the init to constructor of the tcb.  
I shall move the init and cleanup to tcb constructor/destructor.

PR #829, opened 2016-11-10

selvaganesang:
[Trafodion 2325] Reduce path length by avoiding expensive APIs

Load with log error rows missing some error rows being logged.
The hdfs file where the error rows are not logged wasn't closed

Avoided unnecessary copy of the transactional put list


prashanth-vasudev:
+ 1. Looks good to me.

sureshsubbiah:
+1. Looks good.

PR #831, opened 2016-11-11
Fix for [Trafodion-2348]

sbroeder:
TransactionState.hasConflict gets a null pointer exception


shangshengtung:
Looks good.

prashanth-vasudev:
+1

PR #832, opened 2016-11-14
jira 2354 : TMUDF doesn't support "double" as a output column type.

andyyangcn:
add code to support "double" as a output column type.


traflm:
jenkins, test

svarnau:
jenkins, add user

zellerh:
Just another quick comment: It's not true that Trafodion didn't support double as an output type, these "addColumn" methods are just convenience functions to avoid the more complicated way of adding a column. So, if you need to add a more complex data type, you can just create a TypeInfo and a ColumnInfo and then call the AddColumn() method.

PR #834, opened 2016-11-15
[TRAFODION-2170] certain create table DDL will fail with core dump

traflm:
If the DDL has defined the total cluster key length greater than 2048, it will corrupt the memory and cause various core dump.
the fix is to allocate enough memory for the given key. It is not recommended to use very long clustering key, but should be allowed.

### DIFF
--- a/core/sqf/src/tm/tmddlrequests.cpp
+++ b/core/sqf/src/tm/tmddlrequests.cpp
@@ -40,7 +40,7 @@ JNIEXPORT jint JNICALL Java_org_apache_hadoop_hbase_client_transactional_RMInter
    char la_tbldesc[TM_MAX_DDLREQUEST_STRING];
    char la_tblname[TM_MAX_DDLREQUEST_STRING];
    char* str_key;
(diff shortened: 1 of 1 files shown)


selvaganesang:
It is not clear why we need to impose the restriction on the length of the DDL string.  The length of the DDL string is known in advance. Isn't possible to allocate the required length instead of allocating the pre-defined length.

traflm:
@selvaganesang , there is no reason to restrict the length of the DDL string from my point of view as well. This change is to remove one of the fixed length buffer allocation instead use dynamic allocation. 
I will study the rest of the code here in subsequent fixes. @prashanth-vasudev is also refactoring this same file, so I would like to further fix others after his refactoringis done. This PR is submitted now, because in some other forks, same change already done, so this will help the code merging. And also required to get rid of the DDL core dump which we saw several times.

selvaganesang:
@traflm Please go ahead and merge if needed because it avoids the core dump and fixes a memory leak in TM.

traflm:
thanks all for review.

PR #835, opened 2016-11-15
[TRAFODION-2342] hdfs access threads init/cleanup changes

selvaganesang:
The worker threads to read hdfs directly was being started
unnecessarily in many cases. Also, ensured that these worker
threads are cleanly exited when the statement is deallocated.


core/sql/exp/ExpLOBaccess.cpp
```diff
@@ -2280,8 +2280,10 @@ Ex_Lob_Error ExLobsOper (
 	  return LOB_GLOB_PTR_ERROR;
 	}
     }
-
-  if ((globPtr != NULL) && (operation != Lob_Init))
+  if (globPtr != NULL)
+  {
+    lobGlobals = (ExLobGlobals *)globPtr;
+    if ((operation != Lob_Init) && (operation != Lob_Cleanup))
```
DaveBirdsall: It doesn't hurt anything (and an optimizing compiler will optimize it out) but it looks like line 2288 duplicates line 2285.

selvaganesang: Thanks @DaveBirdsall.  Line 2285 is needed. I will remove line 2288 when I get a chance to submit something else.

prashanth-vasudev:
Looks good to me. + 1

PR #838, opened 2016-11-17
[TRAFODION-1262]ODBC: Insert data into large varchar column(>32K) failed with error 8402

rwbzx:
Issue Causes:
1. Dealing with Varchar-oriented data input, if length of **input string(end with '\0')** is larger than 32767, then there will be 4-byte memory assigned to indicate data length, otherwise, 2-byte space will do the job. While odbc driver only allocates _varchar-col-length_+2 bytes to store input data as well as its length, and when length of input string equals 199999, the insert will return SQL_SUCCESS but input data will be truncated.
2. Say that input data type is Char, and db column type is Varchar, while odbc driver decides how many bytes in memory to indicate length of input data by length of **input string**, mxosrvr decides how many bytes to retrieve the length of input data by length of **column type**. Thus, when length of input string is smaller than 32768, odbc assigns 2 bytes for the length but mxosrvr retrieves 4 bytes including first two bytes of input data, and this causes error 8402-string overflow.

Code Modification:
1. Calculate memory space according to _varchar-col-length_: if _varchar-col-length_ is bigger than 32767, then _varchar-col-length_+4 bytes of space will be allocated when building value list.
2. Odbc driver decides how many bytes in memory to indicate length of input data by length of **column type**, thus the assign&retrieve principal will be identical.


rwbzx:
jenkins, retest

PR #839, opened 2016-11-17
[TRAFODION-2228]Add AES_ENCRYPT/AES_DECRYPT functions

xwq:
Add two functions which used to encrypt and decrypt the data using the
    official AES (Advanced Encryption Standard) algorithm.
AES_ENCRYPT(str, key_str[, init_vector])
AES_DECRYPT(crypt_str, key_str[, init_vector])

Both these two function have optional init_vector argument which
provides an initalization vector for block encryption modes that require
it. For modes that require the optional init_vector argument, an error
occurs if init_vector is missing. For modes that don't need init_vector,
       it will raise a unused argument waring if init_vector is
       provided.

The new CQD block_encryption_mode controls the mode for block-based
encryption algorithm. The default value is 0, which use aes-128-ecb.

Here is the list for diffferent values for block_encryption_mode

CQD value          algorithm
   0             aes-128-ecb
   1             aes_192_ecb
   2             aes_256_ecb
   3             aes_128_cbc
   4             aes_192_cbc
   5             aes_256_cbc
   6             aes_128_cfb1
   7             aes_192_cfb1
   8             aes_256_cfb1
   9             aes_128_cfb8
   10            aes_192_cfb8
   11            aes_256_cfb8
   12            aes_128_cfb128
   13            aes_192_cfb128
   14            aes_256_cfb128
   15            aes_128_ofb
   16            aes_192_ofb
   17            aes_256_ofb


core/sql/exp/exp_function.cpp
```diff
@@ -8235,5 +8249,165 @@ ex_expr::exp_return_type ExFunctionIsIP::eval(char * op_data[],
...
+      goto aes_encrypt_error;
+
+  if (!EVP_EncryptUpdate(&ctx, result, &u_len, (const unsigned char *)source, source_len))
+      goto aes_encrypt_error;
+
+  if (!EVP_EncryptFinal(&ctx, result + u_len, &f_len))
+      goto aes_encrypt_error;
+
+  if (!EVP_CIPHER_CTX_cleanup(&ctx))
+      goto aes_encrypt_error;
```
traflm: If this error raised, the error handling logic will do it again. Hope it will not have bad side effect.

xwq: You are right. It don't need check the return value of EVP_CIPHER_CTX_cleanup, I will fix it.

xwq:
@traflm As usually, user can use SHA2 function to generate a key when using AES_ENCRYPT. But now, the SHA2 function only support 256 bits .
I'd like to enhance it first and then update the both test case and the document.

gtapper:
Has anyone looked into what this means from a license perspective? There are usually strict regulations related to encryption so I'd like to know a bit more about the origins of this code, the licenses related to it, and whether there are export restrictions involved.

xwq:
@gtapper Thanks for the reminder. I think it it a exemption for open source. 

MySQL also providers these encrypt algorithms. From the code level, I use OpenSSL lib.

However I found a doc in DOC, https://www.bis.doc.gov/index.php/forms-documents/pdfs/1441-encryption-webinar-transcript-2861771-introduction-to-encryption-export-controls/file

I am totally confused with the open source part. Maybe we need legal expert to review it.

svarnau:
Here is the ASF info on crypto code: https://www.apache.org/dev/crypto.html

--Steve

xwq:
@svarnau Thanks for you info. It seems we have many things to do before we can merge the code.

PR #840, opened 2016-11-18
[TRAFODION-2235] Changed EPOLL logic and ZClient to handle transient

zcorrea:
errors with Zookeeper quorum.


zcorrea:
jenkins, retest

PR #841, opened 2016-11-18
[TRAFODION-2356] Trafodion process can dump core at times because JNI…

selvaganesang:
…Env is not initialized


sureshsubbiah:
+1. Very nice.

PR #842, opened 2016-11-19
[TRAFODION-2353] support of the regexp operator

traflm:
This is a required feature to support basic Regular Expression in Trafodion.
This work is following the design at https://cwiki.apache.org/confluence/display/TRAFODION/Example+Enhancement%3A+Add+a+New+Built-in+Function
The difference is in this version, I didn't use Google's RE2, but use GNU LibC's POSIX Regexp modules. The benefit is GUN libc is already installed and it is very common, so no need to introduce a new dependency. GNU regexp follows the POSIX regular expression, although others add more features on POSIX, but this already support all basic regular expression syntax. If there are strong requirement to support more Regular Expression syntax that GNU cannot support, we can enhance it later.


robertamarton:
GNU licenses are not supported by Apache.  Do we require binding in any code that had a GNU license?  If not, does the customer that builds Trafodion require any new software to be downloaded?

svarnau:
Good catch, Roberta.  The /usr/include/regex.h file on my system has a GNU license. 

I think it is okay to use g++ compiler as long as we're not distributing that code. We are pulling in regex functions from libc, but it is shared library we are not distributing even with convenience binaries. But this is just part of the libc system shared lib, which I think is a standard part of the linux OS. So I'm thinking this is okay.

traflm:
I agree with Steve, this is part of libc function. It is same thing as printf()/malloc() from my point of view. But Google's RE2 is BSD-License, so if needed, we can use RE2, but that involve a new third-party software, and need to do more work to setup dev env, and also the distribution will require Google's RE2 as a dependency as well. I am trying to avoid that.

core/sql/common/OperTypeEnum.h
```diff
@@ -470,6 +470,10 @@ enum OperatorTypeEnum {
                         ITM_OLAP_LEAD = 2176,
                         ITM_OLAP_LAG = 2177,
 
+                        // Regular Expression
+                        ITM_PATTERNMATCHINGFUNCTION = 2178,
```
DaveBirdsall: Is ITM_PATTERNMATCHINGFUNCTION used anywhere? I didn't see it.

core/sql/exp/exp_like.cpp
```diff
@@ -477,6 +501,51 @@ ex_like_clause_base::processNulls(char *op_data[], CollHeap *heap,
...
+  char * pattern;
+  char *srcStr= new (exHeap) char[len1+1];
+
+  cflags = REG_EXTENDED|REG_NEWLINE;
+  pattern = new (exHeap) char[len2+1];
+  pattern[len2]=0;
+  srcStr[len1]=0;
+
+  str_cpy_all(pattern, op_data[2], len2);
+  str_cpy_all(srcStr, op_data[1], len1);
```
DaveBirdsall: I take it the reason we need to copy is because regcomp requires null terminators and the SQL operands are not null-terminated?

core/sql/exp/exp_like.cpp
```diff
@@ -477,6 +501,51 @@ ex_like_clause_base::processNulls(char *op_data[], CollHeap *heap,
...
+
+  z = regcomp(&reg, pattern, cflags);
+
+  if (z != 0){
+    //ERROR
+    ExRaiseSqlError(exHeap, diagsArea, EXE_BAD_ARG_TO_MATH_FUNC);
```
DaveBirdsall: I don't think this will give the user enough information to debug the error. I recommend at the very least including the value of z in the error message (for example, "regcomp call returned %d"), or even better, use the regerror() function to get a human readable error string back.

core/sql/exp/exp_like.cpp
```diff
@@ -477,6 +501,51 @@ ex_like_clause_base::processNulls(char *op_data[], CollHeap *heap,
...
+    **diagsArea << DgString0("REGEXP");
+    return ex_expr::EXPR_ERROR;
+  }
+ 
+  z = regexec(&reg,srcStr , nmatch, pm, 0);
```
DaveBirdsall: From the regexec man page, it looks like regexec returns 0 for a successful match, REG_NOMATCH for a successful mismatch, and an error code for an error. We need to check for the error and report it (for example, using regerror() to obtain an error string).

core/sql/exp/exp_like.cpp
```diff
@@ -477,6 +501,51 @@ ex_like_clause_base::processNulls(char *op_data[], CollHeap *heap,
...
+
+  cflags = REG_EXTENDED|REG_NEWLINE;
+  pattern = new (exHeap) char[len2+1];
+  pattern[len2]=0;
+  srcStr[len1]=0;
+
+  str_cpy_all(pattern, op_data[2], len2);
+  str_cpy_all(srcStr, op_data[1], len1);
+
+  z = regcomp(&reg, pattern, cflags);
```
DaveBirdsall: This is good for now. Later we can think about ways to move the regcomp call to fixup time if the pattern is constant, which will save some execution time path length (but introduces complications in memory management).

core/sql/optimizer/BindItemExpr.cpp
```diff
@@ -6451,7 +6452,13 @@ NABoolean Like::beginEndKeysApplied(CollHeap *heap)
 
   return beginEndKeysApplied_;
 
-} // Like::beginEndKeysApplied()
+} // PatternMatchingFunction::beginEndKeysApplied()
+
+ItemExpr *Regexp::applyBeginEndKeys(BindWA *bindWA, ItemExpr *boundExpr,
+				  CollHeap *heap)
+{
+  return boundExpr;
```
DaveBirdsall: This is the place where we should add a check for non-default collating sequence, and raise error 4392 if found.

DaveBirdsall:
There's another JIRA on this subject, https://issues.apache.org/jira/browse/TRAFODION-114. I will mark that one resolved/fixed when your change is merged in.

core/sql/regress/executor/EXPECTED002.SB
```diff
@@ -6713,6 +6713,90 @@ s3            100-90-40
...
+>>select * from regexp_test where c1 regexp '^(25[0-5]|2[0-4][0-9]|[0-1]{1}[0-9]{2}|[1-9]{1}[0-9]{1}|[1-9])\.(25[0-5]|2[0-4][0-9]|[0-1]{1}[0-9]{2}|[1-9]{1}[0-9]{1}|[1-9]|0)\.(25[0-5]|2[0-4][0-9]|[0-1]{1}[0-9]{2}|[1-9]{1}[0-9]{1}|[1-9]|0)\.(25[0-5]|2[0-4][0-9]|[0-1]{1}[0-9]{2}|[1-9]{1}[0-9]{1}|[0-9])\s*$';
+
+C1
+----------
+
+127.0.0.1
+
+--- 1 row(s) selected.
+>>-- utf-8 code
+>>select * from regexp_test where c1 regexp '[中文测试]';
```
zellerh: I don't quite understand why this matches. Isn't the square bracket notation giving a choice of different characters? So, this would match a string of length 1 with any of the Chinese characters in the list, but not multiple characters?

core/sql/exp/exp_like.cpp
traflm: thanks Dave for pointing this out, yes, this will be a very good optimization, and I will do some research on that, and hope to update it with next PR.

core/sql/common/OperTypeEnum.h
traflm: yes, it is not used, Will remove it,

core/sql/exp/exp_like.cpp
traflm: yes, the current error handling is not quite correct, I misunderstand the regexec return value. Thanks Dave for your carefully review. We should return SQL Error when regexec return error. And enhance the error msg.

PR #843, opened 2016-11-21
[TRAFODION-2358] Implement builtin function SOUNDEX

ryzuo:
Support SQL builtin function SOUNDEX() like Oracle does. It allows you compare words that are spelled differently, but sound alike in English, which is useful for searching people's names that sound alike, e.g. “Ruoyu” and “Roy”

### DIFF
--- a/core/sql/exp/exp_function.cpp
+++ b/core/sql/exp/exp_function.cpp
@@ -213,6 +213,8 @@ ExFunctionSha2::ExFunctionSha2(){};
 ExFunctionIsIP::ExFunctionIsIP(){};
 ExFunctionInetAton::ExFunctionInetAton(){};
 ExFunctionInetNtoa::ExFunctionInetNtoa(){};
+ExFunctionSoundex::ExFunctionSoundex(){};
+
 ExFunctionAscii::ExFunctionAscii(OperatorTypeEnum oper_type,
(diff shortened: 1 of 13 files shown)


core/sql/exp/exp_function.cpp
```diff
@@ -8235,5 +8244,118 @@ ex_expr::exp_return_type ExFunctionIsIP::eval(char * op_data[],
...
+                break;
+            case 'M':
+            case 'N':
+                current = 5;
+                break;
+            case 'R':
+                current = 6;
+                break;
+            default:
+                break;
```
DaveBirdsall: Should we set current to something in this case? If we don't, then it just has the same value as it had before. Perhaps that is intended?

ryzuo: Hi Dave, if it's not a alphabetic letter, it goes to default, I think we consider it does not pronounce. Thus as you see, it will be the same value as it had before, then it will be dumped if you look into the following logic at line 8341.

DaveBirdsall: Looks good, then.

zellerh:
+1

traflm:
please file a JIRA to track that document update for this PR is required. Then I will be good for this PR .

ryzuo:
Filed as 2369

PR #844, opened 2016-11-21
[TRAFODION-2359] Smooth out UEC for last interval when last value is …

DaveBirdsall:
…scarce.

Details in the JIRA.


sureshsubbiah:
Looks good Dave. +1

PR #846, opened 2016-11-21
TRAFODION-2363 Add support to return hive MD info in relational format

anoopsharma00:
See jira for details.

Also added support to correctly return tables and views when
GET TABLES/VIEWS in VIEWS <viewname>
stmt is issued.


core/sql/comexe/ComTdbExeUtil.h
```diff
@@ -2946,37 +2946,65 @@ class ComTdbExeUtilLobShowddl : public ComTdbExeUtil
...
-  { "TABLE_NAME",        2,     COM_USER_COLUMN, REC_BYTE_F_ASCII,    256, FALSE , SQLCHARSETCODE_UNKNOWN, 0, 0, 0, 0, 0, 0, 0, COM_NO_DEFAULT, "",NULL,NULL,COM_UNKNOWN_DIRECTION_LIT, 0}
+  { "CATALOG_NAME",   0,      COM_USER_COLUMN, REC_BYTE_F_ASCII,    256, FALSE , SQLCHARSETCODE_UTF8,     0, 0, 0, 0, 0, 0, 0, COM_NO_DEFAULT, "" ,NULL,NULL,COM_UNKNOWN_DIRECTION_LIT, 0},  
+  { "SCHEMA_NAME",    1,      COM_USER_COLUMN, REC_BYTE_F_ASCII,    256, FALSE , SQLCHARSETCODE_UTF8,     0, 0, 0, 0, 0, 0, 0, COM_NO_DEFAULT, "" ,NULL,NULL,COM_UNKNOWN_DIRECTION_LIT, 0},  
+  { "TABLE_NAME",     2,      COM_USER_COLUMN, REC_BYTE_F_ASCII,    256, FALSE , SQLCHARSETCODE_UTF8,     0, 0, 0, 0, 0, 0, 0, COM_NO_DEFAULT, "" ,NULL,NULL,COM_UNKNOWN_DIRECTION_LIT, 0},
+  { "FILE_FORMAT",    3,      COM_USER_COLUMN, REC_BYTE_F_ASCII,     24, FALSE , SQLCHARSETCODE_ISO88591, 0, 0, 0, 0, 0, 0, 0, COM_NO_DEFAULT, "" ,NULL,NULL,COM_UNKNOWN_DIRECTION_LIT, 0},
+  { "CREATE_TIME",    4,      COM_USER_COLUMN, REC_BIN64_SIGNED,      8, FALSE , SQLCHARSETCODE_UNKNOWN,  0, 0, 0, 0, 0, 0, 0, COM_NO_DEFAULT, "" ,NULL,NULL,COM_UNKNOWN_DIRECTION_LIT, 0},
```
DaveBirdsall: Why do we use LARGEINT for times instead of TIMESTAMP? (Yes, I know we do this too in our metadata.)

anoopsharma00: it is to be consistent with how traf metadata create time field is set.
That way callers (like DBMgr) can use the same/similar queries when looking
at this field for traf and hive tables.

Another related note: create time in hive is set to number of seconds
from epoch (1 jan 1970). When returning it as part of hivemd, we convert 
it to juliantimestamp(in microseconds from 1/1/0001). This is consistent
with the value returned for tables stored in traf metadata.

zellerh:
+1, looks good to me, too.

PR #849, opened 2016-11-22
[TRAFODION-2229]SHA2 function enhancement

xwq:
SHA2 support more funtcions than SHA-256. Now, it support SHA-224,
     SHA-256, SHA-384, SHA-512.

SHA2(str, hash_length)
The first argument is the string to be hashed.
The second is the desired bit length of the result, which must have a
value of 224, 256, 384, 512, or 0(which is same with 256).


core/sql/optimizer/SynthType.cpp
```diff
@@ -1034,6 +1033,37 @@ const NAType *BuiltinFunction::synthesizeType()
...
+        const NAType &typ2 = child(1)->getValueId().getType();
+
+        if (typ2.getTypeQualifier() != NA_NUMERIC_TYPE)
+        {
+          *CmpCommon::diags() << DgSqlCode(-4045) << DgString0("SHA2");
+          return NULL;
+        }
+
+        retType = new HEAP
+          SQLVarChar(1024, typ1.supportsSQLnull());
```
DaveBirdsall: Should we make the output length depend on the function? And make it fixed length? For example, https://en.wikipedia.org/wiki/SHA-2 shows that SHA224 results in 224 bits (which gives a hex string of 56 characters), SHA512 results in 512 bits (128 characters) and so on. Oh, but you made the number of bits an operand to the function and that could be an expression. In that case a VarChar is warranted.

core/sql/parser/sqlparser.y
```diff
@@ -9688,12 +9688,12 @@ misc_function :
                             1, $3);
                 }
 
-     | TOK_SHA2 '(' value_expression ')'
+     | TOK_SHA2 '(' value_expression, ',', value_expression ')'
```
DaveBirdsall: Since a second operand is always required, why do we need 0?

core/sql/optimizer/SynthType.cpp
sureshsubbiah: Yes, a fixed length char would be better for a couple of reasons, other than the varchar length overhead
a) varchars are currently blank padded when they are sent to a UDF
b) formation of a VEG with varchar columns is not enabled by default.

Since the length of the char depends on the second argument, this may mean that 
a) we need different ItemExprs for each of the new lengths that are being added and not use a second argument OR
b) we insist that the second argument be a literal.

Often a customer will choose a SHA function with smaller length since they want to control the amount of data being moved. Since our varchars are expanded in certain cases it seems safer to have a fixed length here. The UDF  issue may be fixed soon, if you feel the code will be cleaner as done now, then we can leave it as is too.

zellerh:
Agree with Dave and Suresh. Maybe you could consider two things:

- In addition to SHA, have a few more functions SHA2_256, SHA2_384, SHA2_512, with a single argument each. Those could be translated in the parser into the original SHA2 function, or you could use the ZZZBinderFunction method in the binder to do the translation. That has the advantage of showing the original function name in EXPLAIN.

- If the second argument is a constant, then synthesize the output type appropriately. Only make it a VARCHAR if the second argument is not a constant.

sureshsubbiah:
I would also have preferred to have the output of these functions be in binary format. If they need to be displayed on screen then converttohex function can be used on them. This will result in space savings on both disk and memory. In some sense, the previous PR on these functions has already started us on the current. Is it too late for change the output of these functions? Thank you.

zellerh:
@sureshsubbiah, wouldn't those numbers with >64 bits be very very inconvenient to use, similarly the need to use converttohex would be inconvenient? I've never seen these SHA numbers being used for anything else than displaying and comparing the hex string.

core/sql/parser/sqlparser.y
xwq: I just want to make it compatible with MySQL.

xwq:
Thank you all for the comments. I think I have a solution to make it return a fixed length result. I will update it.

xwq:
I have added a commit to make the return values be the fixed length char.

For the second operand , I use NUMERIC_LITERAL_EXACT_NO_SCALE instead of value_expression. Here is a problem than we can't use dynamic statement like "prepare s1 from select sha2(?, ?) from dual". 

I think the new change make the code looks not clean. So If some of you think the changes can't be commited, I can roll back it.

core/sql/regress/compGeneral/EXPECTED006.SB
```diff
@@ -1686,9 +1686,10 @@ d3d9446802a44259755d38e6d163e820   852952723  b1d5781111d84f7b3fe45a0852e59758cd
 --- 1 row(s) selected.
 >>select sha2('the original data', 128) from dual;
 
-*** ERROR[8428] The argument to function SHA2 is not valid.
+*** ERROR[4045] The operand of function SHA2 must be numeric.
```
zellerh: The error is a bit misleading, since 128 is numeric. Ideally, the error message should say that we expect 224, 256, 384, or 512.

zellerh:
+1

This is a good solution, in my humble opinion.

core/sql/regress/compGeneral/EXPECTED006.SB
xwq: @zellerh 

Can I use 8428 instead of 4045 or use yyerror('The second operand expect 224, 256, 384 or 512') to throw Internal parse error with message. I think we don't need to add a new error code since it only used once here.

xwq:
Jenkins, retest.

xwq:
jenkins, retest

PR #852, opened 2016-11-23
[TRAFODION-2367] EXPLAIN documentation updates + syntax tweak

DaveBirdsall:
Three changes:

1. Noted in EXPLAIN documentation that some non-reserved keywords can be used for statement names in PREPARE but cause syntax errors when used in EXPLAIN. Noted that as a workaround, one can change the name to a delimited identifier.

2. Tweaked the EXPLAIN syntax in the SQL parser to accept all one-letter non-reserved keywords as statement names. So, for example, "explain c;" will now work instead of raising a syntax error.

3. Fixed the EXPLAIN documentation to note that one can do "EXPLAIN query_text" as well.


zellerh:
+1

Looks good to me. Happy Thanksgiving!

DaveBirdsall:
Wow... fastest code review I've ever received. And Happy Thanksgiving to you, too, @zellerh.

PR #853, opened 2016-11-28
[TRAFODION-1262]ODBC:Varchar col operation fails

rwbzx:
**Issue Causes:**
1. Dealing with Varchar-oriented data input, if length of input string(end with '\0') is larger than 32767, then there will be 4-byte memory assigned to indicate data length, otherwise, 2-byte space will do the job. While odbc driver only allocates varchar-col-length+2 bytes to store input data as well as its length, and when length of input string equals 199999, the insert will return SQL_SUCCESS but input data will be truncated.
2. Say that input data type is Char, and db column type is Varchar, while odbc driver decides how many bytes in memory to indicate length of input data by length of input string, mxosrvr decides how many bytes to retrieve the length of input data by length of column type. Thus, when length of input string is smaller than 32768, odbc assigns 2 bytes for the length but mxosrvr retrieves 4 bytes including first two bytes of input data, and this causes error 8402-string overflow.

**Code Modification:**
1. Calculate memory space according to varchar-col-length: if varchar-col-length is bigger than 32767, then varchar-col-length+4 bytes of space will be allocated when building value list.
2. Odbc driver decides how many bytes in memory to indicate length of input data by length of column type, thus the assign&retrieve principal will be identical.
3. Modify sqltocconvert files to acommendate sql-c conversion with c-sql conversion in case of select failures.


hegdean:
Can you please add test cases

rwbzx:
[ODBCUnitTestForTrafodion.zip](https://github.com/apache/incubator-trafodion/files/617996/ODBCUnitTestForTrafodion.zip)

rwbzx:
for this issue, two schemas(utf8 and iso88591) is created, and in each schema a table named tvarchar is created. 
i. e. 
CREATE TABLE TRAFODION.ISO88591.TVARCHAR
  (
    C1                               VARCHAR(200000) CHARACTER SET ISO88591
      COLLATE DEFAULT DEFAULT NULL SERIALIZED
  )
;
CREATE TABLE TRAFODION.UTF8.TVARCHAR
  (
    C1                               VARCHAR(50000 CHARS) CHARACTER SET UTF8
      COLLATE DEFAULT DEFAULT NULL SERIALIZED
  )
;

rwbzx:
[ODBCUnitTestForTrafodion.zip](https://github.com/apache/incubator-trafodion/files/618039/ODBCUnitTestForTrafodion.zip)
unit test code

ryzuo:
@rwbzx we are not supposed to just put the code files as attachment, the test code also should be under control of git. However, we don't have a proper place to put these code yet, or just say we don't have an agreement yet where to place.
@hegdean we can have a discuss about this later, we can go ahead to review this fix, no need to make it bigger.

svarnau:
Agreed that the test code should be added to version control. License statements need to be added to files. If they are windows platform specific, they can be added as a new test directory under win-odbc64, yes?
Otherwise, maybe under core/conn/trafci?

xwq:
@svarnau @hegdean 
Now the JDBC test code was under dcs/src/test/jdbc_test and jenkins can run them automatically . Under dcs/src/test/ , there are some others directories, like python ODBC to put different test codes.We need discuss how to design the framework to make jenkins running automatically.

I prefer creating a separate PR.

hegdean:
As a first step you should just add test code  to version control under win-odbc64  with instructions on how to run it manually.

ryzuo:
win-odbc64 would not be a proper place, these tests are not only for windows, but linux as well. I prefer put them under dcs/src/test. @rwbzx would you please make these test codes runable on both Windows and Linux? The make file, project file, building script would also be workable.

rwbzx:
all right @ryzuo thanks to @hegdean @svarnau and @xwq for your advise.

rwbzx:
jenkins,retest

ryzuo:
+1. Looks good to me now.
Hi @arvind-narain & @hegdean , do you have any further questions or comments?

rwbzx:
hi @hegdean @svarnau @ryzuo @xwq , as for unit test code, I have opened PR#867 to add them to /dcs/src/test folder.

selvaganesang:
Trafodion SQL engine has the concept of Bulk Move of data enabled by default.  SQL detects if the bulk move can be enabled automatically and moves the data in bulk from CLI layer to its data flow processing queues. This would avoid reformatting the data or copy the data for every column from the CLI layer to the tuple structure that is used its data flow processing queues and thus improves the performance of data ingestion.

To keep this enabled the trusted client like ODBC driver/server needs to gather the descriptor information from SQL after the query is prepared and populate the buffer accordingly. The trusted client driver gets this from the server(mxosrvr) and populates the data according to the expected format to be passed through to the cli layer.

 This change seems to do it other way, deciding some portion of the descriptor information based on the data that is being passed in from the client. 

I am not sure if the code to get the descriptor information from long varchar columns is missing or the information obtained  from Trafodion engine for long varchar column  on the server side is not passed to the driver side correctly. If it is already done, can you please provide me some pointer how this information is handled in this code change

DaveBirdsall:
FYI: I just merged pull request 868, so the unit test is in the repository.

xwq:
@selvaganesang 
As my understand, you concern is about how to determine the the number of bytes for length, 2 bytes or 4 bytes. 
Actually, we have a descriptor item names vc_ind_length to indicate using 2 or 4 bytes for length. However, now the client can't get vc_ind_length from server. The same solution was already in JDBC client. 

If we want to modify the protocol between client and server to support vc_ind_length, we need modify both JDBC, linux ODBC and windows ODBC.  I preferred to create anther JIRA to do it.

win-odbc64/odbcclient/drvr35/cdesc.cpp
```diff
@@ -625,7 +645,6 @@ unsigned long CDescRec::setDescRec(short DescMode, SQLItemDesc_def *SQLItemDesc)
 	case SQL_LONGVARCHAR:
 	case SQL_WCHAR:
 	case SQL_WVARCHAR:
-		m_DescPrecision = m_DescLength;
```
rwbzx: The location of local desciptor value assigning is adjusted. For other column type, param _m_DescPrecision_ is assigned between line 346 and 508, while previously for SQL_CHAR/SQL_VARCHAR/SQL_LONGVARCHAR/SQL_WCHAR/SQL_WVARCHAR columns the param is assigned in another _switch..case_ block. Moreover, param _m_ODBCPrecision_ is assigned in line 509 by param _m_DescPrecision_ and will be passed to function _ConvertCToSQL_ afterwards in this issue. As you see, the assigning of _m_DescPrecision_ is put ahead. This change is based on consideration of correct value passing and unified code style.

xwq: Thanks for the explanation. It's good to me.

ryzuo:
@selvaganesang  @arvind-narain, any further comments on this?

ryzuo:
If no one has any more comments, I'm going to merge this.

PR #854, opened 2016-11-28
[TRAFODION - 2345] Support LOG() function of any base

ryzuo:
Like Oracle does, the function LOG() function supports 2 parameter of call, in this way, the 1st parameter is the radix, and the second one is power, the function returns the logarithm.
This change implement this, and still, the user can call it with only one parameter (form LOG(X)), in which case the parameter X will be the power, and the function just returns the natural logarithm of X. This is also compatible with MySQL.

### DIFF
--- a/core/sql/exp/exp_math_func.cpp
+++ b/core/sql/exp/exp_math_func.cpp
@@ -51,6 +51,7 @@
 #define MathLog(op, err) log(op)
 #define MathLog10(op, err) log10(op)
 #define MathLog2(op, err) log2(op)
+#define MathLogX(op1, op2, err) (log(op2)/log(op1))
 #define MathModf(op, err) modf(op)
 #define MathPow(op1, op2, err) pow(op1, op2)
 //extern double MathPow(double Base, double Power, short &err);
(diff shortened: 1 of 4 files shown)


core/sql/exp/exp_math_func.cpp
```diff
@@ -337,8 +338,22 @@ ex_expr::exp_return_type ExFunctionMath::eval(char *op_data[],
 	  **diagsArea << DgString0("LOG");
 	  return ex_expr::EXPR_ERROR;
 	}
-      
-      *(double *)op_data[0] = MathLog(*(double *)op_data[1], err);
+
+      if(3 == getNumOperands())
+    {
```
DaveBirdsall: The indentation is inconsistent. Perhaps your editor is using tabs? Best to replace those with spaces.

core/sql/exp/exp_math_func.cpp
```diff
@@ -337,8 +338,22 @@ ex_expr::exp_return_type ExFunctionMath::eval(char *op_data[],
...
+        double power = *(double *)op_data[2];
+        double radix = *(double *)op_data[1];
+        if(power <= 0 || 1 == radix)
```
DaveBirdsall: The radix also has to be greater than zero, doesn't it?

core/sql/exp/exp_math_func.cpp
ryzuo: I use space, but somehow the original indentation is not recognized, I see they are the same in my editor, not sure which part of the configuration of vim causes this. Anyway, I tried to re-org it, will push on to see how it looks.

core/sql/exp/exp_math_func.cpp
ryzuo: you are right, I will change this.

zellerh:
+1
Looks good to me (with the fix you mentioned).

ryzuo:
The indention looks good now this time. I anything still goes wrong, please point out.

PR #856, opened 2016-11-28
[TRAFODION-2371] Fix ALTER TABLE RENAME for IDENTITY columns

DaveBirdsall:
When ALTER TABLE RENAME is done on a table having an IDENTITY column, the sequence name of the IDENTITY column was not updated. That caused DML which requires sequence values (e.g., INSERT) to fail. This code change adds logic to take care of renaming the sequence name for IDENTITY columns.


zellerh:
+1
Looks good to me.

PR #860, opened 2016-11-30
[TRAFODION 2381] Enhance awk script to support many variations of distros

prashanth-vasudev:
Make awk script extensible to support any number of distro variations.
Also support #elseif  define.


sureshsubbiah:
+1

PR #861, opened 2016-12-02

selvaganesang:
… in TM

The following problems are fixed:
The loaded classes reference are made global to avoid the class reference from gc-ed
Introduced push/pop frame concepts to avoid memory growth usually observed with
many JNI/java transitions
Logged the errors via log4cxx mechanism instead of stdout
Possibly fixes the tm core dumped in obscure way at times


core/sqf/src/seatrans/tm/hbasetmlib2/javaobjectinterfacetm.cpp
```diff
@@ -352,12 +349,11 @@ JOI_RetCode JavaObjectInterfaceTM::init(char*           className,
...
-      if (javaObj_ == 0 || _tlp_jenv->ExceptionCheck())
-      { 
-        _tlp_jenv->ExceptionDescribe();
-        _tlp_jenv->ExceptionClear();
-        _tlp_jenv->DeleteLocalRef(javaClass);  
-        return JOI_ERROR_NEWOBJ;
+      if (getExceptionDetails(NULL)) {
+         tm_log_write(DTM_TM_JNI_ERROR, SQ_LOG_ERR, (char *)"JavaObjectInterfaceTM::init()", 
+             (char *)_tlp_error_msg->c_str(), -1LL);
+         _tlp_jenv->DeleteLocalRef(javaClass);
```
prashanth-vasudev: Do you need to delete global reference javaclass here?
Also line 361.

selvaganesang: In line 361 the global reference of the JavaClass needs to be retained. This ensures that the class reference is never unloaded via gc and all the method ids associated with the class remains valid. This was the one of the possible reasons for the obscure core seen in TM.

For the error path case, it should be ok to delete the global reference, but the TM process is unusable and it would be aborted anyway

PR #862, opened 2016-12-02
[TRAFODION-2384] Trafodion node needs to be gracefully shutdown when …

selvaganesang:
…the node is brought down

Added a service script for Trafodion at $MY_SQROOT/sysinstall/etc/init.d/trafodion.
As part of cluster install, this file will be copied to /etc/init.d and
configured for service startup.

The service script has the support for stop and status only.

You can do the following:
bin]$ sudo service trafodion status
trafodion on node kenai01.cluster.local is started as logical node 000
bin]$ sudo service trafodion stop
Stopping trafodion node 000 on kenai01.cluster.local
bin]$ sudo service trafodion status
trafodion on node kenai01.cluster.local is not yet started
bin]$ sudo service trafodion start
start: missing job name
Try `start --help' for more information.

This allows init 6, shutdown and reboot to execute the trafodion services to be stopped in
correct sequence.  Hence, it should fix the core dumps seen at the time of node shutdown.

### DIFF
--- a/install/python-installer/traf_setup.py
+++ b/install/python-installer/traf_setup.py
@@ -50,7 +50,7 @@ def run():
     ### kernel settings ###
     run_cmd('sysctl -w kernel.pid_max=65535 2>&1 > /dev/null')
     run_cmd('echo "kernel.pid_max=65535" >> /etc/sysctl.conf')
-    run_cmd('cp $SQ_ROOT/sysinstall/etc/init.d/trafodion /etc/init.d')
+    run_cmd('cp %s/sysinstall/etc/init.d/trafodion /etc/init.d' % SQ_ROOT)
     run_cmd('chkconfig --add trafodion')
     run_cmd('chkconfig --level 06 trafodion on')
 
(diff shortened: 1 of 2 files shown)


install/python-installer/traf_setup.py
```diff
@@ -50,6 +50,9 @@ def run():
     ### kernel settings ###
     run_cmd('sysctl -w kernel.pid_max=65535 2>&1 > /dev/null')
     run_cmd('echo "kernel.pid_max=65535" >> /etc/sysctl.conf')
+    run_cmd('cp $SQ_ROOT/sysinstall/etc/init.d/trafodion /etc/init.d')
```
selvaganesang: Good point. I would leave it as is till /etc/trafodion/trafodion_config is available.  The service script needs to be executed by root as part of shutdown.  The variable SQ_ROOT will not be available to root user unless the service script trafodion is regenerated on the fly after installation or the new installer stores some default info at /etc/default/trafodion

Also, by leaving it as it is possible for the existing installation to configure trafodion as a service by issuing the commands in the traf_setup.py manually

svarnau: I think @mkby is right. The run_cmd will not have SQ_ROOT in the shell environment. The cp command needs to be changed as he suggests.

selvaganesang: @mkby suggested in another chat that  /etc/trafodion/trafodion_config  may not available soon. Hence my earlier response.  I made the change in the script as indicated by him

PR #864, opened 2016-12-04
updated expected files that changed due to a previous err msg checkin

anoopsharma00:


DaveBirdsall:
Thanks, @anoopsharma00, for fixing this. This regression failure is due to my change, https://github.com/apache/incubator-trafodion/pull/858, merged last Wednesday. I failed to run this particular suite and also failed to grep for this particular error message in the regression expected results. The script that I am writing for https://issues.apache.org/jira/browse/TRAFODION-2380 will do the grep, so it should help make folks aware of what regression results need to be changed given an error message text change.

PR #866, opened 2016-12-05
[TRAFODION-2380] New script to help automate Message Guide updating work

DaveBirdsall:
This new script is designed to help automate development work on the Messages Guide. In its initial guise, the script does the following:

1. Compares the text in SqlciErrors.txt with that in the Messages Guide
2. Produces a report of mismatched messages. For each mismatched message it shows:
a. The message text in each of SqlciErrors.txt and the Messages Guide
b. Any enums used in the code to refer to the message number
c. A list of regress tests whose expected files might have examples of the message
d. Optionally, a list of all C++ source files with possible references to the message number (this option is very slow unfortunately)

With this information a developer can quickly identify what messages in the Message Guide are out-of-date or missing. If the developer decides to change the text in SqlciErrors.txt, he/she can also see what regression test expected files need to be updated. If option d is specified (--codeRefs), the developer will also have a list of files in the code where the message is generated. But this option presently takes six or seven hours to run and so is best done overnight.

### DIFF
--- a/core/sqf/sql/scripts/analyzeMessageGuide.py
+++ b/core/sqf/sql/scripts/analyzeMessageGuide.py
@@ -335,8 +335,7 @@ class MessagesTable:
  
     def analyzeCodeReferences(self,directory):
         fileString = directory + "/*/*.cpp " + directory + "/*/*.h " + directory + "/*/*.y" 
-        for key in ["1002","1004","1008", "1013", "1402"]:  # TODO: remove this testing line
-        # for key in self.dict:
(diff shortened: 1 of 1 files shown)


DaveBirdsall:
To run the script, cd $MY_SQROOT/sql/scripts, then do "python analyzeMessageGuide.py [--codeRefs]". If --codeRefs is present, option 2d above is executed; if it is absent, it is not. I ran this overnight last night with --codeRefs and it completed in three hours (not the six or seven feared).

PR #867, opened 2016-12-06
[TRAFODION-2386]ODBC:add odbc unit-test code for large varchar column issue

rwbzx:
Add source code, build files and readme files, and test code runs in following way:
Every time 1 row of data is inserted to a specific table and fetched out into local buffer after insertion is done. And then comparation is made between input buffer and fetch-out buffer, if difference exists, then test fails.


rwbzx:
jenkins,retest

rwbzx:
close this PR for too many commit record, will open another PR soon.

PR #868, opened 2016-12-06
[TRAFODION-2386]ODBC:add odbc unit-test code for large varchar column issue

rwbzx:
Add source code, build files and readme files, and test code runs in following way:
Every time 1 row of data is inserted to a specific table and fetched out into local buffer after insertion is done. And then comparation is made between input buffer and fetch-out buffer, if difference exists, then test fails.

### DIFF
--- a/dcs/src/test/odbc/src/Main.cpp
+++ b/dcs/src/test/odbc/src/Main.cpp
@@ -28,7 +28,7 @@ int main(int argc, char * argv[])
 {
 	if (argc != 7)
 	{
-		std::cout << "Usage: " << argv[0] << " -d [<datasource>][-u <userid>][-p <password>]" << endl;
+		std::cout << "Usage: " << argv[0] << " -d <datasource> -u <userid> -p <password>" << endl;
 		return -1;
 	}
 	char inputArgs[6][100];
@@ -37,7 +37,7 @@ int main(int argc, char * argv[])
 
 	if ((strcmp(inputArgs[0], "-d") != 0) || (strcmp(inputArgs[2], "-u") != 0) || (strcmp(inputArgs[4], "-p") != 0))
 	{
-		std::cout << "Usage: " << argv[0] << " -d [<datasource>][-u <userid>][-p <password>]" << endl;
+		std::cout << "Usage: " << argv[0] << " -d <datasource> -u <userid> -p <password>" << endl;
 		return -1;
 	}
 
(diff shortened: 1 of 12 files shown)


dcs/src/test/odbc/src/Main.cpp
```diff
@@ -0,0 +1,57 @@
...
+**************************************************************************/
+
+#include "common/Global.h"
+#include "test/varchar32k/LargeVarcharCol.h"
+
+int main(int argc, char * argv[])
+{
+	if (argc != 7)
+	{
+		std::cout << "Usage: " << argv[0] << " -d [<datasource>][-u <userid>][-p <password>]" << endl;
```
DaveBirdsall: If argc has to be exactly 7, then the parameters mentioned in the Usage are not optional. (That is, the square brackets should be removed.)

dcs/src/test/odbc/src/Main.cpp
```diff
@@ -0,0 +1,57 @@
...
+		return -1;
+	}
+	char inputArgs[6][100];
+	for (int i = 1; i < argc; i++)
+		strcpy(inputArgs[i - 1], argv[i]);
+
+	if ((strcmp(inputArgs[0], "-d") != 0) || (strcmp(inputArgs[2], "-u") != 0) || (strcmp(inputArgs[4], "-p") != 0))
+	{
+		std::cout << "Usage: " << argv[0] << " -d [<datasource>][-u <userid>][-p <password>]" << endl;
```
DaveBirdsall: Same comment as above.

dcs/src/test/odbc/src/common/Global.cpp
```diff
@@ -0,0 +1,116 @@
...
+	unsigned char localBuf[101];
+	char* tempPtr = (char*)localBuf, *tempPtr1;
+	int i, j, a, b, current, calc, length;
+
+	SQLCHAR tempPrecision;
+	SQLCHAR tempScale;
+	SQLCHAR tempSign;
+	SQLCHAR tmpVal[101];
+
+	if (strlen(rTrim(cTmpBuf)) > sizeof(tmpVal))
```
DaveBirdsall: What happens if strlen(rTrim(cTmpBuf)) == sizeof(tmpVal)? Won't you get a buffer overrun at line 52 because there isn't room for the null terminator?

rwbzx: thanks for the tip, if strlen(rTrim(cTmpBuf)) == sizeof(tmpVal), strlen(localBuf) will be 101, but there will be risk of stack corrupt. fixed it in latest commit.

dcs/src/test/odbc/src/Main.cpp
rwbzx: yeah these params are mandatory.fixed it in latest commit.

PR #869, opened 2016-12-06
[TRAFODION-2388] there should be only one ENV for Trafodion's root directory instead of many.

mkby:
convert $SQ_ROOT, $SQ_HOME, $MY_SQROOT into one variable $MY_SQROOT.

There're too many files using $MY_SQROOT, I'm afraid to bring some unknown issues, so not changing the name to $TRAF_ROOT.


svarnau:
Looks reasonable to me, removing some redundant variables. I understand changing the name throughout would be a very big change.

svarnau:
jenkins, retest -- test automation was passing in SQ_HOME to installer, needs to by MY_SQROOT for this change and forward.

amandamoran:
+1 
I agree that it would be a lot of work to change MY_SQROOT to TRAF_ROOT. But I think we should. Also I suggest maybe TRAF_HOME instead of TRAF_ROOT to be like other Hadoop projects (makes it easier for users). We could do it in stages. (So we would have MY_SQROOT and TRAF_HOME for a while). If you want to make a JIRA on this, you could assign to me, since there is no real rush.

mkby:
Thanks for all suggestions. I converted all MY_SQROOT to TRAF_HOME. let's see if it can pass  jenkins test.

svarnau:
jenkins, retest -- change to TRAF_HOME requires another set of changes in test automation

selvaganesang:
It is a very good change. I didn't review this change, but I think this change can be verified by testing only. So, can you please consider running the full-tests before merging this change in. Also you might need to fix any documents/knowledge base articles that refers to the old variables possibly as a different task from this.

You also need to give some heads up on how to migrate to this change on the existing work space when the  contributors pull in this change into their existing work space.

Also consider obtaining a fresh developer work space with these changes and test the normal contributor activity.

DaveBirdsall:
jenkins, extra tests

mkby:
@selvaganesang 
Trafodion documents are all in doc/ folder, they have also been replaced with the new variable name.
And there's no extra steps needed when contirbutors pull this change, just source sqenv.sh and make.

mkby:
jenkins, retest

DaveBirdsall:
jenkins, extra tests

mkby:
The failed test is TEST042, it's not related to this change. So I think extra test is passed.

traflm:
TEST042 fail doesn't due to this change. So this is ready to merge for me. +1

PR #871, opened 2016-12-07
[TRAFODION-2375]ODBC:SQLGetDiagRec return wrong msg length when data truncate

rwbzx:
Acoording to ODBC reference, in function SQLGetDiagRec,  param _TextLengthPtr_  should point to a buffer in which to return the total number of characters, but it returns number of truncated characters when param _BufferLength_ is smaller than character number of error message text.
Thus, modifications are made relating to win/unix odbc driver code as following:
Unix ODBC: if truncation occurs, return SQL_SUCCESS_WITH_INFO, and set variable strLen as total number of error message text whenever data is truncated.
Windows ODBC: set variable strLen as total number of error message text whenever data is truncated.

### DIFF
--- a/win-odbc64/odbcclient/drvr35/cdiag.cpp
+++ b/win-odbc64/odbcclient/drvr35/cdiag.cpp
@@ -389,8 +389,8 @@ SQLRETURN CDiagRec::GetDiagRec(SQLSMALLINT	RecNumber,
 					translateLengthMax/2, (int *)&translateLength, (char *)errorMsg)) != SQL_SUCCESS )
 				rc = SQL_SUCCESS_WITH_INFO; //ERROR;
 
-			strLen = translateLength; 
-			((wchar_t *)MessageText)[strLen] = L'\0' ; 
+			((wchar_t *)MessageText)[translateLength] = L'\0';
+			strLen = tmpStrLen;
 		}
 		if (TextLengthPtr != NULL)
 		*TextLengthPtr = strLen;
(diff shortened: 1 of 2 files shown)


win-odbc64/odbcclient/drvr35/cdiag.cpp
```diff
@@ -389,8 +389,8 @@ SQLRETURN CDiagRec::GetDiagRec(SQLSMALLINT	RecNumber,
 					translateLengthMax/2, (int *)&translateLength, (char *)errorMsg)) != SQL_SUCCESS )
 				rc = SQL_SUCCESS_WITH_INFO; //ERROR;
 
-			strLen = translateLength; 
-			((wchar_t *)MessageText)[strLen] = L'\0' ; 
+			((wchar_t *)MessageText)[translateLength] = L'\0';
+			strLen = tmpStrLen;
```
selvaganesang: As per the ODBC specification, if the MessageText is NULL, the TextLengthPtr should give the caller the number of characters that will be returned. 

Usually, the ODBC applications are expected to get the number of characters needed for the error message by passing in MessageText as NULL, then allocate that many characters in their space and then request for the error message by passing the pointer to the buffer allocated as the MessageText.  

It looks like the logic of returning the actual number of characters needed when MessageText is NULL is missing

rwbzx: Yes indeed, the logic is added in latest commit. Thanks for the tips.

ryzuo:
+1, Looks good to me.

PR #872, opened 2016-12-07
[TRAFODION-2285] Document TO_DATE, TO_TIME and TO_TIMESTAMP functions

DaveBirdsall:


DaveBirdsall:
@anoopsharma00, could you take a look and make sure I got these right? Thanks.

docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
```diff
@@ -7523,7 +7526,139 @@ TIMESTAMPDIFF (SQL_TSI_WEEK, DATE '2006-01-01', DATE '2006-01-09')
...
++
+is an expression that gives a character value.
+
+[[examples_of_to_timestamp]]
+=== Example of TO_TIMESTAMP
+
+* This function returns the timestamp value '2016-12-07 10:01:00':
++
+```
+TO_TIMESTAMP ('2016-12-07 10:01:00')
```
traflm: At first glance, I am guessing TO_TIMESTAMP() is alias to TO_TIME(), but it seems not. So maybe it should explicitly mentioned that the character-expression must in fixed format of YYYY-MM-DD HH24:MI:SS, and no milli-second part.

docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
```diff
@@ -7523,7 +7526,139 @@ TIMESTAMPDIFF (SQL_TSI_WEEK, DATE '2006-01-01', DATE '2006-01-09')
...
+** 'MM/DD/YYYY'
+** 'YYYY/MM/DD'
+** 'YYYYMMDD'
+** 'YY/MM/DD'
+** 'MM/DD/YY'
+** 'MM-DD-YYYY'
+** 'YYYYMM'
+** 'DD-MM-YYYY'
+** 'DD-MON-YYYY'
+** 'DDMONYYYY'
```
traflm: It will be helpful to have brief description of each of these 'marks'. what is difference between YY and YYYY, and it must be in upper case? It will be clear what can be valid choices for "MON" as well.
But this is not a must to have, just a suggestion.

DaveBirdsall:
Hi @traflm, thanks for the great comments. I think I've addressed them. Please review.

Regarding TO_CHAR, could you write a JIRA? Feel free to assign to me.

anoopsharma00:
A general comment.
Format values that have date fields in them are part of TO_DATE function.
For example:  
 to_date('01.03.2016 10.11.12', 'DD.MM.YYYY HH24.MI.SS') is a valid construct.

Currently we accept date and time format syntax for both to_date and to_time functions.
Some of it is for backward compatibility.

For clarity, we can fix our code to accept:
  format values with date field in them for TO_DATE
  format values without date field for TO_TIME

One other place to see examples of supported date/time functions
is regress/seabase/TEST030.

DaveBirdsall:
Ah, OK. So, TO_TIME should always be just TIMEs, while TO_DATE may be DATE or TIMESTAMP (depending on whether the hours etc. fields are specified). I'll change the documentation accordingly and fix the code as well.

DaveBirdsall:
The latest rework does the following:

1. Restricts TO_DATE to formats that include year, month and/or day fields. Restricts TO_TIME to formats that include only hour, minute or second fields.
2. Fixes four bugs:
a. When an invalid day was given (e.g. March 33), the source data was missing from the 8415 error message.
b. Time fields were not validated in either TO_DATE or TO_TIME. Now they are: An 8415 error is reported for an invalid time field.
c. Date values having the MON format would result in 8810 errors if the month was not recognized. Now an 8415 error will be reported. (Example: values(to_date('04-DEC-2016','DDMONYYYY')); -- note the unwanted hyphens in the data value.)
d. An error path in method ex_function_dateformat::eval tried to report an executor internal error (8001) if a datetime conversion routine returned an error but without reporting a diagnostic. If no diagsArea object had been allocated, the error path would fail to report the 8001 error, instead resulting in an 8810 error.

DaveBirdsall:
@traflm and @anoopsharma00, could you take a look at the latest rework? Thanks.

traflm:
It looks good to me. +1

PR #875, opened 2016-12-11
[TRAFODION-2379] update sql reference manual about group_concat function

traflm:


docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
```diff
@@ -3584,6 +3586,47 @@ Start/Date Time/Shipped               (EXPR)
...
+```
+>> SELECT student_name,
+         GROUP_CONCAT(DISTINCT test_score
+                   ORDER BY test_score DESC SEPARATOR ' ')
+         FROM student
+         GROUP BY student_name;
+STUDENT_NAME     (EXPR)
+--------------  --------------
+scott            80 90 91 56
+tom              77 43 91
```
DaveBirdsall: The test scores aren't in order. When I tried this example, I got "91 90 80 56" for scott's test scores, and "91 77 43" for tom's. So the code seems to work. Just this text seems to be wrong.

traflm: Thanks Dave, it should be ordered, I will update the text.

PR #878, opened 2016-12-13
TRAFODION-1586 Support for ext Traf tables mapped to HBase tables

anoopsharma00:
This is the initial support for HBase mapped tables.
A doc for external manuals will be added later.
regress/seabase/TEST022 contains various testcases.


anoopsharma00:
jenkins retest

traflm:
This is a very cool feature. I just check the TEST022, it is exactly what we are looking forwarding. Need some time to read the changes, this is a big change.
One thing to be clarify: what kind of encode/decode to access the data in native hbase table?

core/sql/sqlcomp/CmpSeabaseDDLindex.cpp
```diff
@@ -518,20 +521,13 @@ void CmpSeabaseDDL::createSeabaseIndex( StmtDDLCreateIndex * createIndexNode,
...
   BindWA bindWA(ActiveSchemaDB(), CmpCommon::context(), FALSE/*inDDL*/);
   CorrName cn(tableName.getObjectNamePart().getInternalName(),
 	      STMTHEAP,
 	      btSchemaNamePart,
 	      btCatalogNamePart);
 
-  NATable *naTable = bindWA.getNATable(cn); 
-
-  // Save parser flags settings so they can be restored later
-  Set_SqlParser_Flags (savedParserFlags);
```
DaveBirdsall: Glad you took this out, as this code is incorrect. (It should have used Assign_SqlParser_Flags at line 534.)

core/sql/generator/GenExpGenerator.h
```diff
@@ -1086,45 +1073,7 @@ class ExpGenerator : public NABasicObject
...
-		     // the next few parameters are here
-		     // as part of a horrible kludge for
-		     // the PartitionAccess::codeGen()
-		     // method, which lacks a SearchKey
-		     // object and therefore exposes
-		     // things like the exclusion
-		     // expressions; with luck, later work
-		     // in the Optimizer will result in a
-		     // much cleaner interface
-		     const NABoolean useTheHorribleKludge = FALSE,
```
DaveBirdsall: Glad to see this finally removed/replaced. I remember putting this in around 1996 to fix some issue with the interaction of MDAM and partition access.

anoopsharma00:
Response to encode/decode question.

Column values in native hbase table can be stored in string format or in native
format (this is when data is stored in native format of datatype. For ex,
an integer is stored as 4-byte value).

Values stored in string format (for ex, integer 12345 is stored as
a string of length 5 : "12345") . These values are read as is and then
interpreted depending on the mapped table definition.

When creating external table mapped to hbase table, one can specify
whether the native format value is serialized or not.  
If data is serialized, then it will be unserialized (or
decoded) during read. This is same behavior to when
a trafodion table is created and column definitions are specified.

When writing to native hbase table, data will be serialized/encoded if needed.

PR #879, opened 2016-12-13
[TRAFODION-2211] trafci can use history cmd

mashengchen:
trafci should support up, down, left, right key in the keyboard

### DIFF
--- a/core/conn/trafci/installer_pom.xml
+++ b/core/conn/trafci/installer_pom.xml
@@ -1,106 +1,112 @@
-<!-- 
- @@@ START COPYRIGHT @@@                                                       
-
- Licensed to the Apache Software Foundation (ASF) under one
- or more contributor license agreements.  See the NOTICE file
- distributed with this work for additional information
- regarding copyright ownership.  The ASF licenses this file
- to you under the Apache License, Version 2.0 (the
- "License"); you may not use this file except in compliance
- with the License.  You may obtain a copy of the License at
-
-   http://www.apache.org/licenses/LICENSE-2.0
-
- Unless required by applicable law or agreed to in writing,
- software distributed under the License is distributed on an
- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
- KIND, either express or implied.  See the License for the
- specific language governing permissions and limitations
- under the License.
-
- @@@ END COPYRIGHT @@@
--->
+<!-- @@@ START COPYRIGHT @@@ Licensed to the Apache Software Foundation (ASF) 
+    under one or more contributor license agreements. See the NOTICE file distributed 
+    with this work for additional information regarding copyright ownership. 
+    The ASF licenses this file to you under the Apache License, Version 2.0 (the 
+    "License"); you may not use this file except in compliance with the License. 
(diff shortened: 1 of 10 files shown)


core/conn/trafci/src/main/java/org/trafodion/ci/ConsoleReader.java
```diff
@@ -172,7 +176,8 @@ public String readLine() throws UnsupportedEncodingException, UserInterruption,
```
traflm: This will remove Ctrl-C signal handler, did you test Ctrl-C in trafci? It should do a 'cancel query' if I remember it correctly. Anyway, you should make sure Ctrl-C has same behavior as before, should not be broken by this change. 
I am not sure, need to check the trafci manual, but wondering if you have tested it or not? Since you comments out the signal handler.

mashengchen: ohh, i really not notice that.. i will research how to add this behavior.

core/conn/trafci/installer_pom.xml
```diff
@@ -1,106 +1,112 @@
...
- Unless required by applicable law or agreed to in writing,
- software distributed under the License is distributed on an
- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
- KIND, either express or implied.  See the License for the
- specific language governing permissions and limitations
- under the License.
-
- @@@ END COPYRIGHT @@@
--->
+<!-- @@@ START COPYRIGHT @@@ Licensed to the Apache Software Foundation (ASF)
```
DaveBirdsall: Not sure what happened to the license text here. Could we put the original back?

core/conn/trafci/src/main/resources/install_jar.xml
```diff
@@ -1,71 +1,64 @@
...
- Unless required by applicable law or agreed to in writing,
- software distributed under the License is distributed on an
- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
- KIND, either express or implied.  See the License for the
- specific language governing permissions and limitations
- under the License.
-
- @@@ END COPYRIGHT @@@
--->
+<!-- @@@ START COPYRIGHT @@@ Licensed to the Apache Software Foundation (ASF)
```
DaveBirdsall: Same comment about the license text.

selvaganesang:
This change is definitely very helpful.  Is the change in java/org/trafodion/ci/ConsoleReader.java alone sufficient to take care of history cmd?  If so, can you please add some description for other changes.

mashengchen:
the most important change is in java/org/trafodion/ci/ConsoleReader.java, and the change in java/org/trafodion/ci/Session.java is help to remove duplicates the prompt.
other changes are help to pass compile and deploy trafci.

mashengchen:
pom.xml & installer_pom.xml : add dependence of jline and format it(replace tab to 4 blanks).
install_jar.xml : add the related jar into the lib folder, and package them together
makefile : add clean part
build.xml : let it work
Installer.java-tmpl : help to do deployment

selvaganesang:
Thanks @mashengchen for your explanation.  Please consider adding all other issues fixed in the JIRA too. +1 Looks good

PR #881, opened 2016-12-20
[TRAFODION-2314] MXOSRVR sometimes exit abnormally with NAMutex assert

selvaganesang:
The test suite fails becuase pthread_mutex_unlock returns EPERM error

pthread_mutex_lock and pthread_mutex_unlock always return success when
a program is not linked with -lpthread option. It is linked this way
when a multi-threaded program is run in single thread mode to avoid the
overhead of locking.

The mxosrvr program is now linked with -lpthread. However, it is not
clear how and why it would fix the EPERM error returned from
pthread_mutex_unlock API. But it is good to ensure that mxosrvr .sos and
the binaries are compiled with -lpthread.

Will revisit this issue if the problem persists.

### DIFF
--- a/core/conn/odbc/src/odbc/nsksrvr/Makefile
+++ b/core/conn/odbc/src/odbc/nsksrvr/Makefile
@@ -121,7 +121,7 @@ DBT_LIBS =
 
 EARLY_LIBS = -ljsig
 
-COMMON_LIBS = -lsbfs -lsbms -levlsq -lwin -ltdm_sqlcli -larkcmp_dll -L../nsksrvrcore -lmxocore -llog4cxx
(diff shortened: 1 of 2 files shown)


selvaganesang:
See the JIRA [TRAFODION-2314] for details about the behavior of a sample program with and without -lpthread

selvaganesang:
jenkins, retest

DaveBirdsall:
I did a little web-surfing and found an example of pthread_mutex_unlock returning EPERM. Perhaps this might give some insight into your issue? That said, your code changes look fine.

http://stackoverflow.com/questions/7422316/how-do-you-handle-pthread-mutex-unlock-failures

selvaganesang:
Thanks @DaveBirIdsall  I did come across this link when I did some research on this issue too. I ran check PR jobs to confirm if I would hit the issue with jdbc_test_cdh again  and I didn't. Though it doesn't conclude this change is indeed the fix for the EPERM issue, I will merge this change in.

PR #882, opened 2016-12-20
TMUDF fixes and related changes

zellerh:
This PR has fixes for 5 JIRAs in a set of 4 commits (sorry, two of the fixes had overlaps, so I bundled them into a single commit). It may be easier to review each commit individually, since they are logically independent.

**Commit 1:**

[TRAFODION-2382] No plan produced when joining two TMUDFs

A simple join query resulted in no plan. One of the reasons is
the way we handle operator types for TMUDFs. Changed this so that
the operator type indicates the arity of the operator.

**Commit 2:**

[TRAFODION-2399] Syntax error when loading from salted table

This was caused by methods to generate an SQL string literal
from a constant value. In some cases it repeated the character
set name introducer.

**Commit 3:**

[TRAFODION-2382] Support tinyint and boolean in tmudfs

Adding support for these new data types in TMUDFs.

- core/sql/generator/LmExpr.cpp
- core/sql/optimizer/UdfDllInteraction.cpp
- core/sql/sqludr/sqludr.cpp
- core/sql/sqludr/sqludr.h
- core/sql/src/main/java/org/trafodion/sql/udr/TupleInfo.java
- core/sql/src/main/java/org/trafodion/sql/udr/TypeInfo.java
- core/sql/src/main/java/org/trafodion/sql/udr/UDRInvocationInfo.java
- core/sql/regress/udr/TEST001
- core/sql/regress/udr/EXPECTED001

TRAFODION-2392 avoid costly sort in reducer TMUDFs

Add a REDUCER_NC function type. The only difference is in the
createContextForAChild function, where we don't generate a required
arrangement for UDFs of this type.

- core/sql/optimizer/OptPhysRelExpr.cpp
- core/sql/sqludr/sqludr.cpp
- core/sql/sqludr/sqludr.h
- core/sql/src/main/java/org/trafodion/sql/udr/UDR.java
- core/sql/src/main/java/org/trafodion/sql/udr/UDRInvocationInfo.java

Misc changes, not related to a JIRA:

- core/sql/regress/compGeneral/EXPECTED071
- core/sql/regress/compGeneral/TEST071

**Commit 4:**

[TRAFODION-2400] wrong results for passthru cols

Wrong results were returned for passthru columns in TMUDFs
when we had equal predicates on the source table. This was
related to VEG rewrite, where the rewritten expression had
the wrong type.

Add a cast node when the type of an expression changes after VEG rewrite


core/sql/optimizer/RelRoutine.cpp
```diff
@@ -1027,7 +1052,7 @@ OperatorTypeEnum PredefinedTableMappingFunction::nameIsAPredefinedTMF(const Corr
   const NAString &funcName = qualName.getObjectName();
```
sureshsubbiah: This comment is for line 1047. It must be the smallest nit recorded. The comment could be changed to REL_ANY_TABLE_MAPPING_UDF from REL_TABLE_MAPPING_UDF.

core/sql/generator/LmExpr.cpp
```diff
@@ -362,7 +362,7 @@ LmExprResult CreateLmOutputExpr(const NAType &formalType,
 // specified type t to be converted to/from C strings. The only 
 // SQL types that do not need to be converted to C strings are:
 // 
-//  INT, SMALLINT, LARGEINT, FLOAT, REAL, DOUBLE PRECISION
+//  INT, SMALLINT, LARGEINT, FLOAT, REAL, DOUBLE PRECISION, BOOLEAN
```
sureshsubbiah: Does Tinyint belong to this set of types? I suppose the answer is No, but thought it safer to check.

core/sql/src/main/java/org/trafodion/sql/udr/TupleInfo.java
```diff
@@ -1079,19 +1230,48 @@ else if (t.getByteLength() == 8)
             row_.putLong(t.getDataOffset(), val);
             break;
 
+        case TINYINT_UNSIGNED:
+            if (val < 0 || val > Byte.MAX_VALUE)
+                if (val > (2 * Byte.MAX_VALUE + 1))
+                    throw new UDRException(
+                        38900,
+                        "Overflow/underflow when assigning to TINYINT UNSIGNED type");
+                else
```
sureshsubbiah: Will this handle underflow correctly? If val < 0 is an exception supposed to be thrown?

core/sql/src/main/java/org/trafodion/sql/udr/TupleInfo.java
```diff
@@ -1079,19 +1230,48 @@ else if (t.getByteLength() == 8)
...
+                    // use the signed value that has the same bit
+                    // pattern as the desired unsigned value, since
+                    // Java doesn't support "unsigned" basic types
+                    val = val + Byte.MIN_VALUE + Byte.MIN_VALUE;
```
sureshsubbiah: I suppose this is basic arithmetic. If the comment could please include an example that will help. Is Byte.MIN_VALUE = -127. If yes, how will adding it twice to val, give the same bit pattern as the unsigned type? I am quite sure there is no defect here, but that I am not understanding something obvious. Thanks.

core/sql/generator/LmExpr.cpp
zellerh: Yes, good point, I'll add tinyint.

core/sql/optimizer/RelRoutine.cpp
zellerh: Maybe small, but still a valid point! I stumbled over this while re-reading the code. Fixed.

core/sql/src/main/java/org/trafodion/sql/udr/TupleInfo.java
zellerh: Thanks for finding this! Fixed in all three cases.

core/sql/src/main/java/org/trafodion/sql/udr/TupleInfo.java
zellerh: Done.

PR #884, opened 2016-12-21
[TRAFODION-2228]fix shift/reduce conflict for AES_ENCRYPT/AES_DECRYPT

xwq:


xwq:
jenkins, extra tests

traflm:
failed at TEST042 which is irrelevant, it is good to merge. This will fix the regression test error due to parser change before. 
+1

PR #885, opened 2016-12-22
[TRAFODION-2414] enhance 8616 commit conflict error message to include tx conflict details

prashanth-vasudev:
commit conflict error detail is available in the region.
This error is raised as a commit conflict exception and carried back to
transaction manager and further back to sql client. 
Careful attention is given to make sure TM behaviour does not change
by making sure to the error codes continue to be returned along with 
new exception strings.   
Current set of change takes are of multi region transaction. 
Additional change will be checked in to support single region transactions.

Error message looks like this:
>>commit work;

*** ERROR[8616] A conflict was detected during commit processing. Transaction has been aborted. Detail :org.trafodion.dtm.TransactionManagerException: org.apache.hadoop.hbase.coprocessor.transactional.CommitConflictException: java.io.IOException: Transaction [[transactionId: 17114 regionTX: false status: PENDING scan Size: 2 write Size: 1 startSQ: 2]] has scan which conflicts with [[transactionId: 1554366944090715 regionTX: true status: COMMITED scan Size: 2 write Size: 1 startSQ: 2 commitedSQ:2]]: region [TRAFODION.SCH.T116T6,,1482359816577.c4a43321f6a96035c3deeb0d85e55ccb.], scanRange[startRow: \x80\x00\x00\x01, endRow: \x80\x00\x00\x01], inserted row[\x80\x00\x00\x01]

### DIFF
--- a/core/sqf/inc/cextdecs/cextdecs.h
+++ b/core/sqf/inc/cextdecs/cextdecs.h
@@ -289,9 +289,12 @@ _declspec(dllimport) short   DNUMOUT
 
 _declspec(dllimport) short   ENDTRANSACTION();
 
-//errStr is allocated if errlen is not zero. Caller must deallocate errStr.
-//Rest of functionality same as ENDTRANSACTION();
+//ENDTRANSACTION_ERR() is same as ENDTRANSACTION(). However,
+//errStr is allocated if errlen is not zero. Caller must deallocate errStr
+//by calling DELLAOCATE_ERR.
+//Rest of functionality same as ENDTRANSACTION.
 _declspec(dllimport) short   ENDTRANSACTION_ERR(char *&errStr, int &errlen);
+_declspec(dllimport) void    DEALLOCATE_ERR(char *&errStr);
 
 _declspec(dllimport) short   FILENAME_COMPARE_
 #ifdef TDM_ROSETTA_COMPATIBILITY_
--- a/core/sqf/inc/cextdecs/cextdecs.h
+++ b/core/sqf/inc/cextdecs/cextdecs.h
@@ -1585,10 +1585,12 @@ _declspec(dllimport) short   DNUMOUT
 
 _declspec(dllimport) short   ENDTRANSACTION();
 
-//errStr is allocated if errlen is not zero. Caller must deallocate errStr.
-//Rest of functionality same as ENDTRANSACTION();
+//ENDTRANSACTION_ERR() is same as ENDTRANSACTION(). However,
+//errStr is allocated if errlen is not zero. Caller must deallocate errStr
+//by calling DELLAOCATE_ERR.
+//Rest of functionality same as ENDTRANSACTION.
 _declspec(dllimport) short   ENDTRANSACTION_ERR(char *&errStr, int &errlen);
-
+_declspec(dllimport) void    DEALLOCATE_ERR(char *&errStr);
 
 _declspec(dllimport) short   FILENAME_COMPARE_
 #ifdef TDM_ROSETTA_COMPATIBILITY_
--- a/core/sqf/inc/cextdecs/cextdecs.h
+++ b/core/sqf/inc/cextdecs/cextdecs.h
@@ -289,6 +289,10 @@ _declspec(dllimport) short   DNUMOUT
 
 _declspec(dllimport) short   ENDTRANSACTION();
 
+//errStr is allocated if errlen is not zero. Caller must deallocate errStr.
+//Rest of functionality same as ENDTRANSACTION();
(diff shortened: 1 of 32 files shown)


core/sqf/inc/cextdecs/cextdecs.h
```diff
@@ -289,6 +289,10 @@ _declspec(dllimport) short   DNUMOUT
 
 _declspec(dllimport) short   ENDTRANSACTION();
 
+//errStr is allocated if errlen is not zero. Caller must deallocate errStr.
+//Rest of functionality same as ENDTRANSACTION();
+_declspec(dllimport) short   ENDTRANSACTION_ERR(char *&errStr, int &errlen);
```
selvaganesang: This would mean that the errStr is allocated by the TM functions using its own heap management scheme or the system memory. The caller wouldn't be aware of how to deallocate if it is anything other than system memory

core/sqf/src/seatrans/tm/hbasetmlib2/javaobjectinterfacetm.cpp
```diff
@@ -428,6 +428,42 @@ bool  JavaObjectInterfaceTM::getExceptionDetails(JNIEnv *jenv)
    return true;
 }
```
selvaganesang: It would be better use the existing getExceptionDetails function to avoid duplication of code. You can either to choose to have default parameter or make the existing getExceptionDetails to call the new functions with retCode passed in

core/sqf/src/seatrans/tm/hbasetmlib2/javaobjectinterfacetm.cpp
```diff
@@ -478,3 +514,23 @@ void JavaObjectInterfaceTM::appendExceptionMessages(JNIEnv *jenv, jthrowable a_e
     jenv->DeleteLocalRef(a_exception);
 }
 
+short JavaObjectInterfaceTM::getExceptionErrorCode(JNIEnv *jenv, jthrowable a_exception)
+{
+  jshort errCode = JOI_OK;
+  if(a_exception != NULL)
+  {
+    //Check to see if this exception has getErrorCode method.
+    //Only custom subclasses have errorcode defined.
```
selvaganesang: Consider caching the methodId or it is an expensive way to get the error code. Is this method called only when error is present

prashanth-vasudev:
Thanks for the review.  Have made changes to address all review comments with the latest commit. Please review. 

Thank You.

prashanth-vasudev:
Jenkins, extra tests

prashanth-vasudev:
CompGeneral TEST042 is the only failure, does not seem to be related with this change.  This branch can be merged.
Thank You.

core/sqf/inc/cextdecs/cextdecs.h
selvaganesang: Thanks @prashanth-vasudev for introducing another api to deallocate the error message. It is not  clear why is it missing  in the second time declaration in this file.

core/sqf/src/seatrans/tm/hbasetmlib2/javaobjectinterfacetm.cpp
```diff
@@ -478,3 +499,17 @@ void JavaObjectInterfaceTM::appendExceptionMessages(JNIEnv *jenv, jthrowable a_e
...
+{
+  jshort errCode = JOI_OK;
+  if(a_exception != NULL)
+  {
+    //Only TransactionManagerException class has errorcode defined.
+    if(jenv->IsInstanceOf(a_exception, gTransactionManagerExceptionClass))
+    {
+      errCode =(jshort) jenv->CallShortMethod(a_exception,
+                                              gGetErrorCodeMethodID);
+    }
```
selvaganesang: Add exception handling code here or at least clear the exception and set the error code to zero

core/sqf/src/tm/tmlib.cpp
```diff
@@ -956,13 +956,39 @@ short BEGINTX(int *pp_tag, int pv_timeout, int64 pv_type_flags)
...
+//--------------------------------------------------------------------
+//DEALLOCATE_ERR
+//
+//Purpose   : Called subsequent to ENDTRANSACTION_ERR
+//Params    : none
+//--------------------------------------------------------------------
+void DEALLOCATE_ERR(char *&errStr)
+{
+  if(errStr)
+    delete errStr;
```
selvaganesang: Should we set the errStr to NULL

prashanth-vasudev:
Thank you for the review comments. I have made changes to address all the feedback comments. Thank You.

PR #888, opened 2016-12-22
[TRAFODION-2417] Installer pre-work for ambari integration

svarnau:
Simplify install operations for dcs and rest by setting default values
in env files. Also make trafci use config files rather than have hard-coded
setting. This reduces code in existing command-line installers.

Add system setting files to RPM. These files are stored in core/sqf/sysinstall.
The existing command-line installers could also take advantage of these files,
but it is not required, and I have not made those changes in this commit.

### DIFF
--- a/core/sqf/sqenvcom.sh
+++ b/core/sqf/sqenvcom.sh
@@ -45,7 +45,7 @@ export PRODUCT_COPYRIGHT_HEADER="2015-2016 Apache Software Foundation"
 # LDAP configuration must also be setup--see
 # $TRAF_HOME/sql/scripts/traf_authentication_config for details.
 ##############################################################
-export TRAFODION_ENABLE_AUTHENTICATION=NO
+export TRAFODION_ENABLE_AUTHENTICATION=${TRAFODION_ENABLE_AUTHENTICATION:-NO}
 
 
 # default SQ_IC to TCP if it is not set in sqenv.sh. Values are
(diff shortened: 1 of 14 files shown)


svarnau:
@mkby & @hegdean please take a look

install/installer/dcs_installer
```diff
@@ -130,34 +130,12 @@ fi
...
-   DCS_DIR=$(tar -tf $DCS_BUILD_FILE | grep --max-count=1 bin | sed -e "s@\(^.*\)/bin/@\1@")
-
-   DCS_DIR=$DCS_INSTALL_PATH/$DCS_DIR
-fi
-
-# set env var in sqenvcom.sh for DCS install directory
-grep --invert-match "DCS_INSTALL_DIR=" $TRAF_HOME/sqenvcom.sh > $TRAF_HOME/sqenvcom.temp
-echo "export DCS_INSTALL_DIR=$DCS_DIR" >> $TRAF_HOME/sqenvcom.temp
-mv $TRAF_HOME/sqenvcom.temp $TRAF_HOME/sqenvcom.sh
+DCS_DIR=$(ls $TRAF_HOME | grep dcs)
```
amandamoran: So happy to see $ONE_TAR_INSTALL being removed. +1

core/sqf/sqenvcom.sh
```diff
@@ -139,6 +139,10 @@ else
 fi
 export TRAF_HOME=$PWD
 
+# normal installed location, can be overridden in .trafodion
+export DCS_INSTALL_DIR=$TRAF_HOME/dcs-$TRAFODION_VER
+export REST_INSTALL_DIR=$TRAF_HOME/rest-$TRAFODION_VER
+
```
mkby: I like this idea, but where does `$TRAFODION_VER` variable come from?

mkby:
Looks good to me, it simplifies the installation procedure, **very nice.**

core/sqf/sqenvcom.sh
svarnau: @mkby TRAFODION_VER is defined very near the beginning of this same file.  This file (sqenvcom.sh) is the location that is modified when we roll to a new release number.

mkby: Thanks, looks good!

PR #893, opened 2016-12-23
[TRAFODION-2418] Allow group by push-down to a fact table

zellerh:
Trafodion has a rule that pushes a groupby down over a join when possible, but it will not push the groupby to the left child of a join - it relies on join commutativity. This might have been ok 20 years ago when this was coded (by me...), but it isn't good for some situations today. Example:

    select d.y, count(f.a), sum(f.b)
    from big_fact f join small_dim d on f.x=d.y
    where d.a = 1
    group by d.y

The plan we would like is a hash join with the group by on big_fact as the left child. To do this, we need to remove the heuristic that prevents this form of push-down.

### DIFF
--- a/core/sql/optimizer/TransRule.cpp
+++ b/core/sql/optimizer/TransRule.cpp
@@ -5178,8 +5178,9 @@ RelExpr * GroupByOnJoinRule::nextSubstitute(RelExpr * before,
   if (reverseT1T2)
     {
       // don't fire the rule with reversed roles if we can apply the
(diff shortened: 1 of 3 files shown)


zellerh:
No, I did not investigate in detail why we don't get the transformation with the current rules. It could be the multi-join rule or some other heuristic. To get the plan I wanted it would have had to apply the commutativity rule twice, once to enable the groupby pushdown and a second time to bring the fact table back to the left side.

PR #897, opened 2016-12-27
TRAFODION-2421

mashengchen:
once down & up a node there will cost a few time , currently the check policy of whether node up is check&wait ( (2^attempts)*5 ), the wait time is 10s-20s-40s-80s-160s-320s, totally will be 630s. 
now after check, normally, it will cost about 11s to up a node, so the first wait should be longger than 10s, and be more smooth, now there use ( log(attempts+15)*5 ), the wait time will be 13.5s-13.8s-14.1s-14.4s-14.7s-14.9s, total time is 75s.


traflm:
Could you describe more? 
What issue this change address, is there any side effect of this change?

Or will any DCS expert help to review this change? It is very hard to understand the impact of this change.

selvaganesang:
I guess it should be ok to sleep for a constant time between retry to check if node has come up and give up if the node doesn't come up after 5 or 10 mins. It is not clear about the need to smooth out the wait time using log.

PR #899, opened 2016-12-28

selvaganesang:
Changes to fix the hive failures due to core at
ExHbaseAccessTcb::buildLoggingFileName when the logging location is not set in TDB.


zellerh:
+1

selvaganesang:
jenkins, retest

PR #900, opened 2017-01-02
[TRAFODION-2422]

eowhadi:
populateSortCols was flagged as major perf offender during profiling
the reason being an unbound iterative string search on huge string.
Added a minor fix: When not setting category in log conf file, extra code is executed, potentially impacting performance

### DIFF
--- a/core/sql/executor/hiveHook.cpp
+++ b/core/sql/executor/hiveHook.cpp
@@ -468,7 +468,8 @@ struct hive_skey_desc* populateSortCols(HiveMetaData *md, Int32 sdID,
                   "populateSortCols::sortCols:],###"))
     return NULL;
   
-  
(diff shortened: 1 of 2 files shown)


core/sql/executor/hiveHook.cpp
```diff
@@ -468,7 +468,8 @@ struct hive_skey_desc* populateSortCols(HiveMetaData *md, Int32 sdID,
                   "populateSortCols::sortCols:],###"))
     return NULL;
   
-  
+  if ((foundE - pos)<=10) //this is important to avoid major performance impact when looking for non existent Order(col over and over, parsing to the end of string. hot spot flagged using gprof
```
zellerh: Would be good to add a comment here that we are testing for the string sortCols:[], which happens to be 10 characters long. Or, better to use strlen("sortCols:[]").

PR #901, opened 2017-01-04
[TRAFODION-2291] Ambari integration

svarnau:
Integrates trafodion with Ambari cluster manager via an mpack (management
pack) plug-in.

As described in install/README.md file, to leverage this feature, two RPMs
are built. One to install the mpack and one used by Amari to install trafodion
itself. The mpack plug-in is built with a URL that points to where the
trafodion RPM is hosted (a yum repo). It can be specified on the make command
line as REPO_URL=...


install/README.md
```diff
@@ -0,0 +1,75 @@
...
+management pack.
+
+#### Packaging
+
+Part of Ambari's job is to set up yum repo files on each node in order to install packages.
+The default URLs are for Hortonworks' public repos. But since your custom-built Trafodion is
+not hosted there, you need to specify a URL for your local yum repo server. To build that into
+the `traf_ambari` package, use make to specify value of `REPO_URL`.
+
+   `make package REPO_URL=http://my.repo.server/repo/...`
```
mkby: What does this `REPO_URL` do? If we set an invalid repo url, does it still work or not?

svarnau: @mkby, The URL is what is we plug in to Ambari, so that it will set it up in a repo file on each node where it will install trafodion. Once the repo file is set up on the node, then ambari will do a yum install of apache-trafodion_server package.  The only way to work around that is to manually install the RPM on each node.  It is much easier to set up a yum repo, if you already have a webserver.  If you have a bad URL (including the default one), it will probably break yum install on that cluster.  If you have suggestion for better default URL, I'd be glad to hear it.

mkby: @svarnau, thanks for the explanation. It seems that traf-ambari rpm package is customized with different REPO_URL, so everytime we install on a different envionment, we have to build a new traf-ambari rpm to apply the new URL, is that true? 
Can we put this URL into ambari webpage so we can configure the URL  from there?

mkby: thanks, sounds good to me

svarnau: @mkby, I was just testing again today and found what I said earlier is incorrect. Ambari does give a chance to edit the URL in the GUI.
image

So that helps a lot.

mkby: Great! So I think `REPO_URL` should only be the default value in the GUI, and user can modify it. It's acceptable to me.

svarnau:
failure was unrelated timeout

PR #906, opened 2017-01-06
[TRAFODION-2430] python installer - need to modify Makefile for building a tar package

mkby:


install/Makefile
```diff
@@ -17,28 +17,42 @@
...
 
-all: pkg-installer 
+all: pkg-installer
 
-pkg-installer: installer/LICENSE installer/NOTICE installer/DISCLAIMER
+pkg-installer: installer/LICENSE installer/NOTICE installer/DISCLAIMER \
+               python-installer/LICENSE python-installer/NOTICE python-installer/DISCLAIMER
 	tar czf ${INSTALLER_TARNAME} installer --exclude=tools
+	tar czf ${PYINSTALLER_TARNAME} python-installer
 	mkdir -p ../distribution
```
svarnau: Rather than build two tarballs in one makefile target, I think it would be cleaner to make another target, and have all target depend on both.

svarnau:
jenkins, retest -- automation tweak to distinguish bash installer package from pyinstaller package

PR #908, opened 2017-01-10
[TRAFODION-2431] python installer - add python lib dependencies

mkby:
@svarnau, please help to review.

### DIFF
--- a/.rat-excludes
+++ b/.rat-excludes
@@ -166,7 +166,7 @@ VariableLengthPKTest.java
 # asciidoc extention file -- MIT license
 google-analytics-postprocessor.rb
 
-# installer JSON config files
+# pyinstaller JSON config files
 mod_cfgs.json
 prompt.json
 script.json
@@ -175,3 +175,7 @@ version.json
 mpack.json
 role_command_order.json
 theme.json
+# pyinstaller dependencies
+# they have BSD-3 license that RAT does not recognize
+prettytable.py
+socks.py
--- a/licenses/lic-pyinstall-src
+++ b/licenses/lic-pyinstall-src
@@ -5,8 +5,8 @@ MIT-like licenses:
 +++++++++++++++++++++++++++++
 
 BSD-3 clause for files in:
-    incubator-trafodion/installer/python-installer/prettytable.py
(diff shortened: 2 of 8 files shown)


licenses/lic-pyinstall-src
```diff
@@ -0,0 +1,63 @@
...
+CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+
++++++++++++++++++++++++++++++
+
+MIT-Expat for files in:
+    incubator-trafodion/installer/python-installer/scripts/httplib2/iri2uri.py
+    incubator-trafodion/installer/python-installer/scripts/httplib2/__init__.py
+
```
svarnau: same as above, start path at installer, you can leave off the incubator-trafodion directory.

licenses/lic-pyinstall-src
```diff
@@ -0,0 +1,63 @@
+===========================================================================
+The python installer dependencies bundles source with BSD-like and
+MIT-like licenses:
+
++++++++++++++++++++++++++++++
+
+BSD-3 clause for files in:
+    incubator-trafodion/installer/python-installer/prettytable.py
+    incubator-trafodion/installer/python-installer/scripts/httplib2/socks.py
+
```
svarnau: leave off incubator-trafodion directory. The workspace can have different name.

.rat-excludes
```diff
@@ -166,8 +166,14 @@ VariableLengthPKTest.java
...
 mod_cfgs.json
 prompt.json
 script.json
 version.json
+
+# pyinstaller dependencies
+prettytable.py
+socks.py
+iri2uri.py
+__init__.py
```
svarnau: These files all have headers. RAT is only confused by two of them (prettytable & socks). So the others should NOT be listed in this file.

Also, comment here should not just state what they are, but why they should be excluded from check -- they have BSD-3 license that RAT does not recognize.

mkby:
@svarnau，please review again, thanks

PR #910, opened 2017-01-10
[TRAFODION-2426] Do not set hbase master log splitting property

svarnau:
This changes all install methods to no longer touch the log splitting
property.


mkby:
looks good

PR #911, opened 2017-01-11
add oauth2 ldap for restful api

rucore:


svarnau:
jenkins, add user

core/rest/src/main/java/org/trafodion/rest/util/TrafLdap.java
```diff
@@ -0,0 +1,94 @@
...
+import javax.naming.ldap.LdapContext;
+
+public class TrafLdap {
+
+    private String ip = "10.10.10.3";
+    private String port = "389";
+    private String basedn = "cn=users,cn=accounts,dc=esgyncn,dc=local";
+    private String username="trafodion";
+    private String password="traf123";
+
```
svarnau: Even for test code, it seems this type of info should be in a config file, not hard-coded.

core/rest/src/main/java/org/trafodion/rest/util/Token.java
```diff
@@ -0,0 +1,194 @@
...
+
+    public static void main(String[] args) {
+        // TODO Auto-generated method stub
+        Token t = new Token();
+        // t.setPasswd("12345678");
+        String content = "test";
+        // String password = "12345678";
+
+        System.out.println("before:" + content);
+        byte[] encryptResult = t.encrypt(content, "fuck-trafodion!");
```
svarnau: That is not appropriate language, even in test code.

venkat1m: Agree. This is totally inappropriate for any code.

DaveBirdsall:
Why is this feature going into the release2.0 branch instead of the master branch?

Also, what JIRA is associated with this work?

core/rest/src/main/java/org/trafodion/rest/TransactionsResource.java
```diff
@@ -57,176 +57,197 @@
...
+
+                MultivaluedMap<String, String> pathParams = uriInfo.getPathParameters();
+                output = " Path Parameters :\n";
+                for (String key : pathParams.keySet()) {
+                    output += key + " : " + pathParams.getFirst(key) + "\n";
+                }
+                LOG.debug(output);
+            }
+
+            Response rs = TokenTool.getResponse(auth);
```
venkat1m: Instead of having to call TokenTool.getResponse(auth) on every REST method, it would be better to implement a servlet filter (javax.servlet.Filter) and have this done in one place. Plus it will guarantee that no REST request can come through without passing the authentication enforced by the filter.

PR #913, opened 2017-01-16
TRAFODION-2439 add support for traf views on hive tables

anoopsharma00:
support for creation of trafodion views on hive tables.
cqd hive_views is set to ON by default to enable it.

During view creation, an implicit external table is created
on underlying hive table if one doesn't already exists.
This is used to track usage info
in views_usage metadata table and to track privileges.
Usage info is used to return views on tables or tables in views
during GET stmts.

trafodion view is created in trafodion catalog/schema and not in hive
catalog/schema.

delete and update operations are not allowed on traf/hive views.

drop of external hive table with CASCADE option will cause all
dependent views to be dropped.
drop of external table without CASCADE option will not drop it
an a dependent view exists on that table.

If the underlying hive table is dropped from hive shell,
then an error is returned at runtime when that view is accessed.

new test regress/hive/TEST007 contains tests for this feature.


core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -13445,6 +13435,14 @@ RelExpr * GenericUpdate::bindNode(BindWA *bindWA)
...
       getOperatorType() == REL_UNARY_DELETE) {
 
+    if (getTableDesc()->getNATable()->isHiveTable())
+      {
+        *CmpCommon::diags() << DgSqlCode(-4223)
+                            << DgString0("Update/Delete on Hive table is");
+        bindWA->setErrStatus();
+        return this;
+      }
+
```
DaveBirdsall: I'm curious why this change was made. Do we support other kinds of delete and update operations on Hive tables?

core/sql/regress/hive/EXPECTED007
```diff
@@ -0,0 +1,370 @@
...
+
+*** ERROR[4028] Table or view TRAFODION.SCH.VHIVE4 is not updatable.
+
+*** ERROR[8822] The statement was not prepared.
+
+>>
+>>-- traf view must be in traf cat/sch
+>>create view vhive5 as select * from thive1;
+
+*** ERROR[3242] This statement is not supported. Reason: DDL operations can only be done on trafodion or external tables.
```
DaveBirdsall: Might be a little clearer to say that views can only be created in a trafodion schema. (The DDL operation is on two objects, thive1 and vhive5, but thive1 is not limited to being a trafodion or external table, and vhive5 doesn't exist yet.)

core/sql/regress/hive/EXPECTED007
```diff
@@ -0,0 +1,370 @@
...
+
+          1            2
+          3            4
+
+--- 2 row(s) selected.
+>>sh echo "drop table thive1;" > TEST007_junk;
+>>sh regrhive.ksh -f TEST007_junk;
+>>execute s;
+
+*** ERROR[8442] Unable to access HDFS interface. Call to ExpLOBInterfaceDataModCheck returned error LOB_DATA_FILE_NOT_FOUND_ERROR(554). Error detail 0.
```
DaveBirdsall: I've never liked this error message. We are able to access the HDFS interface. It's just telling us that the file isn't there. Seems like we should simply say that the object is no longer accessable. A 4082 error maybe?

core/sql/optimizer/BindRelExpr.cpp
anoopsharma00: During binder phase, relational nodes are typed as REL_UNARY_UPDATE or REL_UNARY_DELETE
for update or delete operations. Binder checks for them to determine if it is an update or delete.
They may later get changed to other upd/del like subset/unique/cursor etc.
No update or delete operations are currently supported on hive tables.

anoopsharma00:
jenkins retest

PR #914, opened 2017-01-17

selvaganesang:
LOAD with log error rows failed when ERRORCOUNTER hbase table already exists.
LOAD command now returns the number of error rows skipped.
LOAD with truncate option dumps core when encounters an error.
Similar issue exist with PURGEDATA command too.
POPULATE_INDEXES and UPDATE_STATISTICS now displays the start time stamp when the task is about to begin


core/sql/executor/ExExeUtilLoad.cpp
```diff
@@ -1417,29 +1416,45 @@ short ExExeUtilHBaseBulkLoadTcb::work()
...
+           // Need to clear the diags Area to get the correct rowsAffected for the LOAD COMPLETE command
+           // Hence, we might lose any other warnings created at the time of loading like error during
+           // logging error rows
+           diagsArea->clear();
+        }
         if (rowsAffected_ == 0)
           step_ = LOAD_END_;
 
-        sprintf(statusMsgBuf_,"       Rows Processed: %ld %c",rowsAffected_, '\n' );
+        sprintf(statusMsgBuf_,      "       Rows Processed: %ld %c",rowsAffected_+errorRowCount, '\n' );
```
sureshsubbiah: It will be good to make a documentation change explaining this. RowsProcessed is rows read or some kind of total. To get the actual number of rows loaded user should subtract the two numbers (or maybe it comes out as a summary of the statement?)

core/sql/executor/ExExeUtilLoad.cpp
```diff
@@ -1417,29 +1416,45 @@ short ExExeUtilHBaseBulkLoadTcb::work()
...
+           ComCondition *cond;
+           Lng32 entryNumber;
+           while ((cond = diagsArea->findCondition(EXE_ERROR_ROWS_FOUND, &entryNumber)) != NULL) {
+              errorRowCount = cond->getOptionalInteger(0);
+              diagsArea->deleteWarning(entryNumber);
+           }
+           // Need to clear the diags Area to get the correct rowsAffected for the LOAD COMPLETE command
+           // Hence, we might lose any other warnings created at the time of loading like error during
+           // logging error rows
+           diagsArea->clear();
```
sureshsubbiah: Are we saying here that some diags conditions could be lost? If yes, is there no way to avoid this? Maybe correct rowsAffected could be copied to the appropriate diagsCondition (or another diagsArea/condition be used a messenger to communicate the rowcount to its eventual location?). If conditions could be lost here, I would have preferred to see something to avoid that. Likely I just misunderstood the comment.

core/sql/exp/ExpHbaseInterface.cpp
```diff
@@ -271,7 +271,12 @@ Lng32 ExpHbaseInterface::coProcAggr(
 
 char * getHbaseErrStr(Lng32 errEnum)
 {
-  return (char*)hbaseErrorEnumStr[errEnum - (Lng32)HBASE_MIN_ERROR_NUM];
+  Lng32 lv_errEnum;
+  if (errEnum < HBASE_MIN_ERROR_NUM || errEnum >= HBASE_MAX_ERROR_NUM)
+     lv_errEnum = HBASE_GENERIC_ERROR;
```
sureshsubbiah: It is good that we are adding this to catch unanticipated errors. Did testing reveal any errors that are not on the list above. If yes, do you think we can add to the list of errors in ExpHbaseDefs.h.

core/sql/regress/hive/EXPECTED015
```diff
@@ -1571,6 +1582,7 @@ Task:  CLEANUP         Status: Ended      Time: 2016-12-11 01:58:54.934
 Task:  CLEANUP         Status: Ended      Elapsed Time:    00:00:00.682
 Task:  LOADING DATA    Status: Started    Time: 2016-12-11 01:58:54.934
        Rows Processed: 5000 
+       Error Rows:     0
```
sureshsubbiah: A few tests with error rows will be nice.  Some of the current tests with 0 errors could be altered to have error rows, that way total time taken to run the test will not increase much.

core/sql/executor/ExExeUtilLoad.cpp
selvaganesang: Rows Loaded displayed at the end of the command similar to Rows updated or Rows inserted as shown below

Task:  COMPLETION      Status: Started    Time: 2017-01-17 15:53:06.951
Task:  COMPLETION      Status: Ended      Time: 2017-01-17 15:53:06.262
Task:  COMPLETION      Status: Ended      Elapsed Time:    00:00:00.253

--- 20000 row(s) loaded.
>>--

It should be clear when the output is seen as a whole.  If the document has output shown, the document needs to be updated

core/sql/executor/ExExeUtilLoad.cpp
selvaganesang: Any warning diags could be lost.  The warning were lost before this change too. In fact, the warnings never reach the caller of the executeDirect of any SQL command called with the executor utility commands.  This PR has the fix for this. However, when I attempted to pass this warning to the executor utility command,  I was getting wrong rows loaded  for the load completion command for reasons that needed to debugged further.  Hence, as a interim solution I have cleared the diagnostics after getting the error row count and it behaves the same way as before.

When I have a solution for this anomaly, I will create a new PR

core/sql/exp/ExpHbaseInterface.cpp
selvaganesang: Yes. ERRORCOUNTER hbase table was getting created and this resulted in core dump during my unit testing. Hence the above change. There has been violation of the return codes spread all through the JNI layer. To get around this violation, I have mapped any return codes that didn't fall into known error code range into a generic error code for time being. This needs to be cleaned up later.

core/sql/regress/hive/EXPECTED015
selvaganesang: Yes. I will add the tests when I am done with all the enhancements

PR #915, opened 2017-01-17
[TRAFODION-2440] Replace fixed length buffers with strings in LOB code

DaveBirdsall:
The LOB code has two uses:

1. To read and write LOB column values in Trafodion tables
2. To read Hive files into the Trafodion executor

The LOB code made use of fixed length buffers to store LOB file names. Unfortunately these are too short in general for use with Hive files. Rather than simply make them longer, I have replaced these fixed length buffers with string data types in most places in the code.

Some design notes:

The LOB code is in the middle of a slow transition from using STL classes and the global heap to using the Trafodion NAHeap classes and heaps for memory management. Where a LOB class uses the STL heap, I replaced its fixed length buffer members with an STL string. Where a LOB class uses NAHeap, I used a Trafodion NAString instead.

In one place I left the fixed length buffer alone: I did this where the LOB code calls the CLI function SQL_EXEC_LOBcliInterface. It appears that this code path is traversed only for LOB column values. (If anyone knows otherwise, please comment.)


DaveBirdsall:
@sandhyasun, would you please review this change? Thanks.

DaveBirdsall:
jenkins, retest

core/sql/exp/ExpLOBaccess.cpp
```diff
@@ -198,7 +198,7 @@ Ex_Lob_Error ExLob::fetchCursor(char *handleIn, Int32 handleLenIn, Int64 &outOff
   lobCursors_it it = lobCursors_.find(string(handleIn, handleLenIn));
   char logBuf[4096];
   lobDebugInfo("In ExLob::fetchCursor",0,__LINE__,lobTrace_);
-  char *blackBox = new(getLobGlobalHeap()) char[MAX_LOB_FILE_NAME_LEN+6];
+  char *blackBox = new(getLobGlobalHeap()) char[MAX_LOB_FILE_NAME_LEN+6]; // TODO: do we ever go this way for Hive?
   Int32 blackBoxLen = 0;
```
sandhyasun: This code path is taken for external LOB files. The file name could be provided by the user in the form of file:///<path name>/.../<filename> So perhaps best to change this too.

sandhyasun:
General comment before reviewing any more changes. The LOBGlobals class is instantiated from the ex_globals constructor. The heap_ stored in LOBGlobals is derived from the executor heap. In ExLOB::Initialize, we get the lobGlobalHeap_ from the LOBGlobals->heap_ pointer. So the lobGlobalsHeap_ is not an STL heap.  Hence replacing the "new" from lobGlobalHeap_ with STL calls isn't really correct.

DaveBirdsall:
Not sure I understand. Is your comment in reference to line 1511 (old file), where I replaced

char * saveLobDataFile = new(getLobGlobalHeap()) char[MAX_LOB_FILE_NAME_LEN+6];

with

string saveLobDataFile = lobDataFile_ + "_save";

The reason I did that was that lobDataFile_ is a string, and it made the code cleaner if I used the same model for saveLobDataFile and lobDataFile_. Note that the string is a stack variable, so its destructor is called when it goes out of scope.

I could have coded

NAString saveLobDataFile(getLobGlobalHeap());
saveLobDataFile = lobDataFile_.c_str();
saveLobDataFile += "_save";

but that seems a bit ugly.

Later when lobDataFile_ is changed to an NAString, we could code this more cleanly as

NAString saveLobDataFile(getLobGlobalHeap()) = lobDataFile_ + "_save";

Wouldn't it be better to do this when lobDataFile_ is changed to an NAString?

DaveBirdsall:
Actually, it's not that big a deal. I will change string stack variables to NAStrings and resubmit.

DaveBirdsall:
@sandhyasun, I made an attempt at making these NAStrings instead of strings. It gets ugly pretty fast. I think it better to change these stack variables to NAStrings when the corresponding members change also. So, I'd prefer to leave them as strings, unless I've missed something in your objection.

core/sql/exp/ExpLOBaccess.cpp
DaveBirdsall: OK. Will change.

core/sql/exp/ExpLOBaccess.h
```diff
@@ -515,7 +515,7 @@ class ExLob
 
   public:
 
-    char lobDataFile_[MAX_LOB_FILE_NAME_LEN];
+    string lobDataFile_; // TODO: change to NAString when ExLobCursor is allocated off of an NAHeap
```
sandhyasun: Can this not be allocated from NAString now  ? The ExLob class which contains this is allocated from the executor heap.

sandhyasun:
Some classes like ExLobCursors are derived from the standard heap. My comment above was that for any class like ExLob that is derived from the executor heap,  we shoudl try to keep all the datamembers also allocated on that heap rather than have them on the system heap.

core/sql/exp/ExpLOBaccess.h
DaveBirdsall: Actually, it's not. See line 2338 (new file numbering):

	  //lobPtr = new (lobGlobals->getHeap())ExLob();
	  lobPtr = new ExLob();

It appears that someone experimented with putting it on the executor heap and then retreated.

selvaganesang:
General comment: It is good to have any local variables within the executor portion of the code unless we are sure it is allocated from the stack entirely.  Rationale: If the scope of the variable is at a statement level,  and allocated from the statement heap, it will be automatically deallocated when the statement heap is de-allocated. If any variable is allocated from system heap, it increases the possibility of creating fragmented memory though you have correctly deallocated the variable in question, Some other part of the code might have had some memory leaks from system heap and that caused the block of the memory to be left behind. This bloats the virtual memory of the process.

DaveBirdsall:
Please do not merge. I'm doing reworks to address three issues raised by @sandhyasun and @selvaganesang:

1. Changing ExLob to be allocated off the Lob Heap (and making its lobDataFile_ member an NAString instead of a string)
2. Changing the handling of stack variables to avoid dynamic heap allocations
3. Addressing the potential issue with the black box buffer passed to the CLI

DaveBirdsall:
To @sandhyasun and @selvaganesang: 

I just pushed reworks with the following changes:

1. Changed ExLob to be allocated off the Lob Heap (and making its lobDataFile_ member an NAString instead of a string)
2. Changed the handling of stack variables to avoid global heap allocations
3. Added an error check on the externaltolob function: If the file name is longer than 256 characters (= MAX_LOB_FILE_NAME_LENGTH), then I raise a new error, 8557. By preventing over-long file names from being stored in LOB descriptors, we prevent buffer overruns at the SQL_EXEC_LOBcliInterface level. This is a temporary solution; in the long run the SQL_EXEC_LOBcliInterface call must be enhanced to deal with the case that the proper black box length is not known in advance. (I'll write a separate JIRA on that issue.)

I've pushed these changes so you can review them. Please don't merge just yet, as I'd like to add a check in regression test executor/TEST130 to test for the new error. Also it hasn't been regression tested yet.

core/sql/exp/ExpLOBaccess.cpp
```diff
@@ -198,7 +199,7 @@ Ex_Lob_Error ExLob::fetchCursor(char *handleIn, Int32 handleLenIn, Int64 &outOff
   lobCursors_it it = lobCursors_.find(string(handleIn, handleLenIn));
   char logBuf[4096];
   lobDebugInfo("In ExLob::fetchCursor",0,__LINE__,lobTrace_);
-  char *blackBox = new(getLobGlobalHeap()) char[MAX_LOB_FILE_NAME_LEN+6];
+  char *blackBox = new(getLobGlobalHeap()) char[MAX_LOB_FILE_NAME_LEN+6]; // TODO: do we ever go this way for Hive?
```
sandhyasun: I think this comment can be removed now. Hive does not go through this.

sandhyasun:
These changes look good .

core/sql/exp/ExpLOBaccess.cpp
```diff
@@ -2285,7 +2338,7 @@ Ex_Lob_Error ExLobsOper (
 { 
   Ex_Lob_Error err = LOB_OPER_OK;
   ExLob *lobPtr = NULL;
-  char fn[MAX_LOB_FILE_NAME_LEN];
+  string fn;
```
DaveBirdsall: Missed this one. This needs to be converted to a dynamic char array or an NAString on the Lob globals heap.

DaveBirdsall:
Second rework includes:

1. Replaced a string stack variable with dynamic char arrays to avoid using global heap. (I intended to fix this one on the previous rework but missed it.)
2. Added a test to executor/TEST130 to test the new 8557 message.
3. Added documentation for the new 8557 message to the Messages Guide.

DaveBirdsall:
@sandhyasun and @selvaganesang: I think I've addressed all the issues brought up. I plan to commit this tomorrow evening if no further issues arise.

PR #916, opened 2017-01-17
TRAFODION-2442  timestamp/date comparison bug. And couple other fixes.

anoopsharma00:
-- TRAFODION-2442 timestamp comparison with date returns incorrect results.
  that has been fixed. Date value is expanded to timestamp datatype and
  then compared.
  Fix is for timestamp/date comparison only.
  TIMESTAMP comparison to TIME, and DATE comparison to TIME is not allowed.

-- row length of aligned format tables is now calculated correctly and
  stored in metadata.

-- optimizer cost method for sort had a high limit of MEMORY_UNITS_SIZE
   default set to 20M. For large rows in the range of 10 Mb,
   this caused an overflow.
   Fix is to allocate space for MAXOF default and size of 2 rows.

### DIFF
--- a/core/sql/optimizer/BindItemExpr.cpp
+++ b/core/sql/optimizer/BindItemExpr.cpp
@@ -2424,6 +2424,25 @@ static ItemExpr * ItemExpr_handleIncompatibleComparison(
       tgtOpIndex = 1;
       conversion = 4;
     }
+    else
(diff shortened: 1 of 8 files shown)


anoopsharma00:
TRAFODION-2413 tracks the second fix (row length of aligned format tables)

PR #917, opened 2017-01-19
[TRAFODION-2150] and [TRAFODION-2151] TM long prepare and chore thread Issues

prashanth-vasudev:
Queries such as drop schema cascase involve multiple DDL operations that take some time to process. This cause transaction commits to timeout and cause transaction aborts. Fix is to detect long prepare scenarios and continue to commit the transaction.

Cleanup Chore thread calls were deprecated and required chore service constructs. This should help memory cleanup issues in end point co processor. This should also help some of the regression test failure we see in executor/TEST130

### DIFF
--- a/core/sqf/src/seatrans/tm/hbasetmlib2/src/main/java/org/trafodion/dtm/HBaseTxClient.java
+++ b/core/sqf/src/seatrans/tm/hbasetmlib2/src/main/java/org/trafodion/dtm/HBaseTxClient.java
@@ -1108,9 +1108,28 @@ public class HBaseTxClient {
                             }
                             if (LOG.isDebugEnabled()) LOG.debug("TRAF RCOV THREAD: in-doubt transaction size " + transactionStates.size());
                             for (Map.Entry<Long, TransactionState> tsEntry : transactionStates.entrySet()) {
+                                int isTransactionStillAlive = 0;
                                 TransactionState ts = tsEntry.getValue();
                                 Long txID = ts.getTransactionId();
                                 // TransactionState ts = new TransactionState(txID);
+                                
+                                //It is possible for long prepare situations that involve multiple DDL
+                                //operations, multiple prompts from RS is received. Hence check to see if there
+                                //is a TS object in main TS list and transaction is still active.
+                                //Note that tsEntry is local TS object. 
+                                if (hbtx.mapTransactionStates.get(txID) != null) {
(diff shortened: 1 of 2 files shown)


core/sqf/src/seatrans/tm/hbasetmlib2/src/main/java/org/trafodion/dtm/HBaseTxClient.java
```diff
@@ -1108,9 +1108,28 @@ public void run() {
...
+                                //is a TS object in main TS list and transaction is still active.
+                                //Note that tsEntry is local TS object. 
+                                if (hbtx.mapTransactionStates.get(txID) != null) {
+                                  if (hbtx.mapTransactionStates.get(txID).getStatus().toString().contains("ACTIVE")) {
+                                    isTransactionStillAlive = 1;
+                                  }
+                                  if (LOG.isInfoEnabled()) 
+                                  LOG.info("TRAF RCOV THREAD: TID " + txID
+                                            + " still has ts object in TM memory with state "
+                                            + hbtx.mapTransactionStates.get(txID).getStatus().toString()
```
sbroeder: How do you feel about printing the entire TS object instead of just the state?  Some other useful information could be printed
for example:
LOG.info("TRAF RCOV THREAD: TID " + txID + " still has ts object in TM memory " + hbtx.mapTransactionStates.get(txID).toString()

shangshengtung: Do we plan to move "committing/committed" changes next round?

shangshengtung:
Looks good.

core/sqf/src/seatrans/tm/hbasetmlib2/src/main/java/org/trafodion/dtm/HBaseTxClient.java
prashanth-vasudev: Thanks for the review. I have updated the logging info to include TS object. Regarding committing/committed changes is part of separate Jira.

sandhyasun:
This looks ready to merge ? And it shoudl address the intermittent TEST130 failure due to resource issue on HDP as well ?

prashanth-vasudev:
Yes. It is ready for merge. Should address Test130 failure. Thank You.

PR #918, opened 2017-01-19
[TRAFODION-2420] RMS Enhancements

selvaganesang:
Added yet another offender feature to list query ids that has a total
IO time for any storage engine opertor consuming longer than a given
number of seconds.

./offender -s se_offender

Will list the query ids along with the table name. SEE
$TRAF_HOME/export/limited-support-tools/LSO/README

The "Number of SQL Processes" counter is now made multi-fragment aware
and hence contains the actual number of ESPs used + 1 for master
process.

### DIFF
--- a/core/sql/executor/ExStats.cpp
+++ b/core/sql/executor/ExStats.cpp
@@ -2357,11 +2357,19 @@ Int64 ExHdfsScanStats::getNumVal(Int32 i) const
 
 NABoolean ExHdfsScanStats::filterForSEstats(struct timespec currTimespec, Lng32 filter)
 {
-   blockTime_ = timer_.filterForSEstats(currTimespec);
-   if (blockTime_ >= filter)
-      return TRUE;
+   Int64 sumIOTime;
+
+   if (filter > 0) {
+      blockTime_ = timer_.filterForSEstats(currTimespec);
+      if (blockTime_ >= filter)
+         return TRUE;
+   }
    else
-      return FALSE;
+   if (queryId_ != NULL && (sumIOTime = timer_.getTime()) > 0 && (sumIOTime = sumIOTime /(1000000LL)) >= -filter) {
+      blockTime_ = sumIOTime;
+      return TRUE;
+   }
+   return FALSE;
 }
 
 void ExHdfsScanStats::getVariableStatsInfo(char * dataBuffer,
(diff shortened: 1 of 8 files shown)


core/sql/executor/ExStats.cpp
```diff
@@ -2357,11 +2357,19 @@ Int64 ExHdfsScanStats::getNumVal(Int32 i) const
...
+   Int64 sumIOTime;
+
+   if (filter > 0) {
+      blockTime_ = timer_.filterForSEstats(currTimespec);
+      if (blockTime_ >= filter)
+         return TRUE;
+   }
    else
-      return FALSE;
+   if (queryId_ != NULL && (sumIOTime = timer_.getTime()) > 0 && (sumIOTime = sumIOTime /(1000000LL)) >= -filter) {
```
nonstop-qfchen: Could write as sumIOTime /= 1000000LL

core/sql/executor/ExStats.cpp
```diff
@@ -2741,13 +2749,21 @@ Int64 ExHbaseAccessStats::getNumVal(Int32 i) const
 
 NABoolean ExHbaseAccessStats::filterForSEstats(struct timespec currTimespec, Lng32 filter)
 {
-   blockTime_ = timer_.filterForSEstats(currTimespec);
-   if (blockTime_ >= filter)
-      return TRUE;
+   Int64 sumIOTime;
```
nonstop-qfchen: The same logic seems duplicated between this class and ExHdfsScanStats.  Maybe some refactoring?

core/sql/executor/ExStats.cpp
```diff
@@ -2357,11 +2357,19 @@ Int64 ExHdfsScanStats::getNumVal(Int32 i) const
 
 NABoolean ExHdfsScanStats::filterForSEstats(struct timespec currTimespec, Lng32 filter)
```
nonstop-qfchen: I wonder if the type of filter argument should be Int64 here, since Int64 / 1000000LL could be a value larger than max(Int32).

core/sql/executor/ExStats.h
```diff
@@ -3791,9 +3791,9 @@ NA_EIDPROC
   short  &cmpPriority() {return cmpPriority_;}
   short  &dp2Priority() {return dp2Priority_;}
   short  &fixupPriority() {return fixupPriority_;}
-  inline void setNumSqlProcs(short i) {numSqlProcs_ = i; }
+  void  incNumEspsInUse() { numOfTotalEspsUsed_++; }
   inline void setNumCpus(short i) {numCpus_ = i; }
-  inline short getNumSqlProcs() { return numSqlProcs_; }
+  inline short getNumSqlProcs() { return numOfTotalEspsUsed_+1; }
```
nonstop-qfchen: Should we +1 every time we call getNumSqlProcs()?

core/sql/executor/ExStats.cpp
selvaganesang: Yes. Can be done. I am thinking of getting rid of one of these classes because both have same member variables sometime in future.

selvaganesang: Will take care of your suggestion, when this is done

core/sql/executor/ExStats.h
selvaganesang: Yes.  Because if the query had just master, then the number of SQL procs will be zero. Only the numOfTotalEspsUsed_ is maintained.

core/sql/executor/ExStats.cpp
selvaganesang: The filter is in seconds and comes from passing the user command. The user command is something like
select * from table(statistics(null, 'SE_OFFENDER=<t1_in_secs>)).  The parsing of the second parameter should take care of this.

The blockTime_ is currently in 32bit, because we don't expect the query to be blocked so long. It should timeout much earlier than this from hbase.

sureshsubbiah:
+1. Changes look good. 
I can merge this in by the end of today if there are no concerns. @selvaganesang has promised to address some issues in a subsequent delivery.

PR #919, opened 2017-01-19
[TRAFODION-2453] pyinstaller error on hdp with kerberos enabled.

mkby:


mkby:
@svarnau ,that's a good idea. `trafodion` user should be an application user like `hdfs`, `hbase` and donnot requrie a password.

PR #922, opened 2017-01-20
[TRAFODION-2438] Unnecessary and sometimes wrong delete during upsert…

sureshsubbiah:
…/merge with IM

This change is done by Hans. Design by Selva.
Please see JIRA for an explanation.

### DIFF
--- a/core/sql/generator/GenRelUpdate.cpp
+++ b/core/sql/generator/GenRelUpdate.cpp
@@ -1153,6 +1153,7 @@ short HbaseDelete::codeGen(Generator * generator)
 		      0, // updateTuppIndex
 		      0, // mergeInsertTuppIndex
 		      0, // mergeInsertRowIdTuppIndex
+		      0, // mergeIUDIndicatorTuppIndex
 		      0, // returnedFetchedTuppIndex
 		      0, // returnedUpdatedTuppIndex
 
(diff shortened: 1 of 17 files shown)


core/sql/generator/GenRelUpdate.cpp
```diff
@@ -1319,12 +1320,41 @@ short HbaseUpdate::codeGen(Generator * generator)
...
   ex_cri_desc * work_cri_desc = NULL;
-  work_cri_desc = new(space) ex_cri_desc(10, space);
+  work_cri_desc = new(space) ex_cri_desc(11, space);
+
+  if (getProducedMergeIUDIndicator() != NULL_VALUE_ID) 
+  {
+    mergeIUDIndicatorTuppIndex = 10;
+    iudIndicatorAttr = 
+      (generator->addMapInfo(getProducedMergeIUDIndicator(), 0))->getAttr();
+    iudIndicatorAttr->setAtpIndex(mergeIUDIndicatorTuppIndex);
```
zellerh: Just a comment, no need to change it: These three calls to set ATP, ATPIndex and Offset could be done with a call to ExpGenerator::processAttributes() or ExpGenerator::processValIdList(), but this works fine, too.

PR #924, opened 2017-01-21

selvaganesang:
Warnings are not ignored as part of LOAD command. This should enable
the display of the errors during logging error rows as warnings.

LOAD command now displays "Rows Loaded: " as part of its status output.
This should help the user when the LOAD command is executed from trafci
because trafci just displays "SQL Opertation complete". sqlci displays
"Rows Loaded" already.


sureshsubbiah:
+1. Looks good Selva.

PR #928, opened 2017-01-25
[TRAFODION-2315] Heuristic decision for common subexpressions

zellerh:
This includes various fixes in addition to the heuristic decision,
mainly an improvement on how to navigate between parent and child
common subexpressions and improvement of support for statistics.
Also, see the comment in RelMisc.h (class CommonSubExprRef) for
a definition of some terminology.


sureshsubbiah:
+1

PR #929, opened 2017-01-25
[TRAFODION-2455] Add retry to row count estimation logic

DaveBirdsall:

1. Changes the stack from NATable::estimateHBaseRowCount on down to return detailed error information about any failure.
2. Changes NATable::estimateHBaseRowCount to return a rowcount of 100 million instead of zero when an error occurs. It is safer to overestimate the size of an object than to underestimate it.
3. Changes UPDATE STATISTICS to give an error 9252 "Unable to get row count estimate: Error code $0int0, detail $1int1. Exception info (if any): $0~string0" when an error occurs in or underneath NATable::estimateHBaseRowCount instead of using a rowcount estimate of zero. The information in this message gives details such as what error path was taken, and any Java exception information that may be pertinent.
4. Adds a retry loop to HBaseClient.java method estimateRowCount so that we retry if we encounter a FileNotFoundException. UPDATE STATISTICS will do up to 4 minutes worth of accumulated retries; normal compilation will do up to 5 seconds worth of accumulated retries. Wait times for retries start out at 2 seconds, doubling until topping out at 30 seconds.
5. Adds timestamps to messages in update statistics logging in local time. To get local time, I needed to fix a bug in the monitor (process.cxx) that was incorrectly setting an environment variable TZ to the empty string when TZ was not defined in the monitor itself.

Note: The fix in process.cxx might (or might not!) fix a similar bug in DTM logging where local timestamps are desired but UTC timestamps are produced.


DaveBirdsall:
@narendragoyal and @sbroeder, you might want to review the process.cxx change.

core/sqf/monitor/linux/process.cxx
```diff
@@ -1317,7 +1318,11 @@ bool CProcess::Create (CProcess *parent, int & result)
...
     env = getenv ("TZ");
-    STRCPY (tz, (env?env:""));
+    tz_exists = (env != NULL);
+    if (tz_exists)
+    {
+      STRCPY (tz, env); // see note regarding TZ below
+    }
     env = getenv ("USER");
     STRCPY (user, (env?env:""));
     env = getenv ("HOME");
```
selvaganesang: Though it is in surround code, I thought that I would mention this.  I think getenv is an expensive call that needs to be avoided in this part of code.

core/sql/executor/HBaseClient_JNI.cpp
```diff
@@ -1303,12 +1304,26 @@ HBC_RetCode HBaseClient_JNI::grant(const Text& user, const Text& tblName, const
...
+  if (jenv_ == NULL)
+     if (initJVM() != JOI_OK)
+         return HBC_ERROR_INIT_PARAM;
 
+  breadCrumb = 2;
+  if (jenv_->PushLocalFrame(jniHandleCapacity_) != 0) {
+     getExceptionDetails();
+     return HBC_ERROR_ROWCOUNT_EST_EXCEPTION;
+  }
+  breadCrumb = 3;
```
selvaganesang: I know you wanted to provide breadcrumb for the errors reported. However, I would think you would want to consider leave this initialization code as it is. I have cleaned up the code earlier just to ensure that the initialization to the JNI layer is encapsulated in this method initJNIEnv() so that any changes can be made in one routine.

core/sql/executor/HBaseClient_JNI.cpp
```diff
@@ -1319,17 +1334,21 @@ HBC_RetCode HBaseClient_JNI::estimateRowCount(const char* tblName,
...
                                               js_tblName, jPartialRowSize,
-                                              jNumCols, jRowCount);
+                                              jNumCols, jRetryLimitMilliSeconds, jRowCount);
   jboolean isCopy;
   jlong* arrayElems = jenv_->GetLongArrayElements(jRowCount, &isCopy);
   rowCount = *arrayElems;
   if (isCopy == JNI_TRUE)
     jenv_->ReleaseLongArrayElements(jRowCount, arrayElems, JNI_ABORT);
 
+  jenv_->DeleteLocalRef(js_tblName);
```
selvaganesang: popLocalFrame would do this for you. Again I cleaned this code earlier to remove the unnecessary call to DeleteLocalRef if push/pop local frame is used

selvaganesang:
Changes look good though I have provided some comments for you to consider

core/sqf/monitor/linux/process.cxx
narendragoyal: Thanks Dave for this find! Requesting @zcorrea also to take a look at the process.cxx changes.

narendragoyal: @selvaganesang thanks!  worth a look - perhaps the values of these env variables could be obtained just once in the monitor process' life time.

zcorrea: Change looks good. @selvaganesang has a good point on making the calls to getenv() on each process create. Will have to analyze the impact of caching these environment variables.

DaveBirdsall: @zcorrea, would you like to write a JIRA on this point?

core/sql/executor/HBaseClient_JNI.cpp
DaveBirdsall: My mistake... I seem to have accidentally resurrected some old code. Will fix.

core/sql/executor/HBaseClient_JNI.cpp
DaveBirdsall: Right. I think I accidentally resurrected some old code. Will fix.

PR #931, opened 2017-01-26

selvaganesang:
Load with log error rows won't work if there are multiple ranges
involved in the hdfs scan. The hdfs log file is closed when the
first range of hdfs scan is done.  The error rows in the subsequent
ranges will not be logged. Instead an SQL warning is generated for
each process.  When these warnings are shipped from ESPs to master,
ESP dumps core.

Added tests to test load with log error rows and continue on error

### DIFF
--- a/core/sql/executor/ExExeUtilLoad.cpp
+++ b/core/sql/executor/ExExeUtilLoad.cpp
@@ -1442,7 +1442,8 @@ short ExExeUtilHBaseBulkLoadTcb::work()
            ComCondition *cond;
            Lng32 entryNumber;
            while ((cond = diagsArea->findCondition(EXE_ERROR_ROWS_FOUND, &entryNumber)) != NULL) {
-              errorRowCount = cond->getOptionalInteger(0);
+              if (errorRowCount < cond->getOptionalInteger(0))
+                 errorRowCount = cond->getOptionalInteger(0);
               diagsArea->deleteWarning(entryNumber);
            }
            diagsArea->setRowCount(0);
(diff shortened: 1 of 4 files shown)


selvaganesang:
Changes look good  though I have provided some comments for you to consider.

core/sql/executor/ExExeUtilLoad.cpp
```diff
@@ -1442,7 +1442,8 @@ short ExExeUtilHBaseBulkLoadTcb::work()
            ComCondition *cond;
            Lng32 entryNumber;
            while ((cond = diagsArea->findCondition(EXE_ERROR_ROWS_FOUND, &entryNumber)) != NULL) {
-              errorRowCount = cond->getOptionalInteger(0);
+              if (errorRowCount < cond->getOptionalInteger(0))
```
zellerh: Actually, I do have one question: You mention that each warning is for a different range. Should we add the rowcounts instead of using the max?

selvaganesang: Actually, the changes in ExHdfsScan.cpp is the one needed to take care of the core dump issue.  To honor the command 'stop after n error rows' when the data is loaded into trafodion table, a counter is maintained in hbase to increment the number of error rows from all the ESP process.  Each load prepare command has its own counter id. This counter id is visible to the ExHbaseAccessBulkLoadPrepSQTcb operator. This operator hosted in multiple ESPs will read the counter at the end of its operation and pass it as a warning. So all the ESPs are expected to have the same value but if an esp finishes early it may not have read the final counter value. Hence I am taking the max error counter value passed in the warnings.

selvaganesang: This counter is displayed as "Error Rows: <error_count> " in the load command output now

selvaganesang:
jenkins, retest

PR #935, opened 2017-01-27
[TRAFODION-2464] failure to upsert into a table with an index

sureshsubbiah:
During upsert into table with index there is a temporary scan in optimizer.
If index access path is chosen for this scan, query would raise an error.
This is now fixed by using the correct indexDesc from this temporary scan.


selvaganesang:
The change looks benign. But, is there a need to run full regressions with this change?

sureshsubbiah:
Full regressions were on my workstation. Thanks.

zellerh:
+1
Looks good to me, too. My explanation (more accurately, guess) for why the fix is valid is the following: The SimpleFileScanOptimizer, where this code came from, is using the best index-only index to scan the table. However, here in the HBase update operator, we must always update the clustering index.

sureshsubbiah:
I am sorry that I forgot to mention that fix was done with lots of help from Hans. Thank you.

PR #936, opened 2017-01-28
[TRAFODION-2466]Handle sortReclen limit in sort overflow.

prashanth-vasudev:
Handle sortReclen limit in sort overflow.
Also minor change to commit conflict error text.


core/sql/generator/GenRelMisc.cpp
```diff
@@ -3071,7 +3074,16 @@ short Sort::generateTdb(Generator * generator,
...
   sort_options->sortMaxHeapSize() = (short)getDefault(SORT_MAX_HEAP_SIZE_MB);
   sort_options->mergeBufferUnit() = (short)getDefault(SORT_MERGE_BUFFER_UNIT_56KB);
+  
   sort_options->scratchIOBlockSize() = (Int32)getDefault(SCRATCH_IO_BLOCKSIZE_SORT);
+  if(sortRecLen >= sort_options->scratchIOBlockSize())
+  {
+    Int32 maxScratchIOBlockSize = (Int32)getDefault(SCRATCH_IO_BLOCKSIZE_SORT_MAX);
+    GenAssert(sortRecLen <= maxScratchIOBlockSize, 
+         "sortRecLen is greater than SCRATCH_IO_BLOCKSIZE_SORT_MAX");
+    sort_options->scratchIOBlockSize() = MINOF(sortRecLen * 128, maxScratchIOBlockSize);
```
sandhyasun: What is the reason to pick 128 as the factor for multiplying ?

core/sql/sqlcomp/DefaultConstants.h
```diff
@@ -2585,6 +2585,7 @@ enum DefaultConstants
   SCRATCH_MGMT_OPTION,
   SCRATCH_PREALLOCATE_EXTENTS,
   SCRATCH_IO_BLOCKSIZE_SORT,
+  SCRATCH_IO_BLOCKSIZE_SORT_MAX,
```
sandhyasun: Since we have a new configurable MAX, can the previous CQD be removed and be made a constant in the code ?

core/sql/generator/GenRelMisc.cpp
prashanth-vasudev: The Record is already long at this point, the thought here is to atleast have 128 records in the block. The thought here is to not have 1 record for 1 block.

core/sql/sqlcomp/DefaultConstants.h
prashanth-vasudev: Yes. I will remove the cqd and keep the max one.

core/sql/generator/GenRelMisc.cpp
DaveBirdsall: Not sure I understand. It looks like this code sizes the scratch IO block size to have at most 128 records (rather than at least).

prashanth-vasudev: Sort IO block size default value is 512kb. This block is essentially the buffer used to write into scratch file.  Sort record length received from compiler in most cases is much smaller, so many records fit into default buffer size of 512kb.  In some cases, sort record length is greater than 512kb, this is where SORT_IO_BLOCKSIZE_SORT_MAX kicks in.  Assuming sort record length here is 512kb or more,  512kb times 128 is almost 64MB block size. Sort allocates several multiples of blocks especially during merge phase, hence keeping the block size smaller is preferred, at the same time, having at most one record fit in one block is also undesirable.  User can control the max by SORT_IO_BLOCKSIZE_SORT_MAX cqd at this point. Currently this cqd defaults to 5MB.

sandhyasun: Would be good to add this bit to the document which describes the various knobs for sort parameters .

sandhyasun:
Looks good .  I iwll merge this is in a short while if there are no further comments.

sandhyasun:
Oopes before that  shouldn' there be a testcase to cover this ? I didn't see a new testcase added to the sort regression test ?

core/sql/generator/GenRelMisc.cpp
DaveBirdsall: Thanks for the explanation, @prashanth-vasudev.

DaveBirdsall:
I think merging it is fine. I agree adding a test case would be good; that could be done as a separate pull request.

prashanth-vasudev:
Thanks for the review. I have added a test case to regression test that covers this case.

PR #937, opened 2017-01-30
Trafodion-2175

robertamarton:
Added new regression test privs1/TEST123


zellerh:
+1 looks good to me. Thanks for removing the "get tables" from schema trafodion."_HV_HIVE_", that made the test fail on my workstation pretty much every time.

PR #938, opened 2017-01-30
[TRAFODION-2465] Fix two bugs

DaveBirdsall:
This set of changes fixes two problems (both introduced by me):

1. When UPDATE STATISTICS retries a statement, and the statement continues to fail, the error diagnostics are incorrect. The root cause of the error is not reported, and some irrelevant errors are reported (i.e. 8609, error on transaction rollback, which happens because the Executor sometimes has already rolled back the transaction when UPDATE STATISTICS attempts a ROLLBACK after a statement failure). This bug was introduced when I rewrote the retry logic in JIRA TRAFODION-2341. My unit test was not thorough enough. (ustat/hs_cli.cpp)
2. Performing UPDATE STATISTICS ON EXISTING COLUMNS PERSISTENT would fail on Hive tables with varchar columns over 200000 characters long. I fixed a similar issue in JIRA TRAFODION-2322, but failed to notice that this issue existed separately in the ON EXISTING COLUMNS code path. (ustat/hs_parser.cpp).

### DIFF
--- a/core/sql/ustat/hs_cli.cpp
+++ b/core/sql/ustat/hs_cli.cpp
@@ -401,8 +401,10 @@ Lng32 HSFuncExecTransactionalQueryWithRetry( const char *dml
     return retcode;
   }
 
-  // The HSErrorCatcher captures any diagnostics when it goes out-of-scope.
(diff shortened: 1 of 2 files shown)


zellerh:
+1
Looks good to me.

PR #940, opened 2017-01-31
[TRAFODION-2457] Re-enable compGeneral/TEST042 and try to stabilize

zellerh:
Implemented suggestions made by Suresh and Selva. Suresh found out
that the test gives different results when authorization is
initialized, and that caused problems because it is usually
initialized on developer workstations but not on Jenkins.

Suresh and Selva suggested to suppress the key length, since that can
vary with different CQD settings like SCHEMA.

Added the queries to the log and expected files to see whether that
will make failure diagnosis easier. The queries get filtered out, so
should not cause test failures.

### DIFF
--- a/core/sql/regress/compGeneral/TEST042
+++ b/core/sql/regress/compGeneral/TEST042
@@ -229,6 +229,10 @@ values (600,52,43,4.12345678,6.0932);
 --
 insert into p9tab (c5) values (0E0);
 
+-- caching behavior changes with or without authorization initialized,
+-- make sure authorization is initialized when we run this test
+initialize authorization;
+
 
 ?section test_dml
 set schema TRAFODION.ORDERENTRY;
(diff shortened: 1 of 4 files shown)


zellerh:
jenkins, extra test

core/sql/regress/compGeneral/TEST042
```diff
@@ -229,6 +229,10 @@ values (600,52,43,4.12345678,6.0932);
 --
 insert into p9tab (c5) values (0E0);
 
+-- caching behavior changes with or without authorization initialized,
+-- make sure authorization is initialized when we run this test
+initialize authorization;
```
DaveBirdsall: So, do we get an error if authorization is already initialized? Is the error filtered out of the diffs?

robertamarton: No error is generated if authorization is already initialized.

zellerh: Thanks. From what I saw in my tests, initialize authorization does not return an error, but it may print out some diagnostics when it actually creates or upgrades some tables. That's ok, because this command is not being logged.

zellerh:
The extra test is a success, compGeneral/TEST042 passes, the other tests (and failures) are irrelevant.

PR #941, opened 2017-02-01
[TRAFODION-2467] HBase region start key straddling a datetime key error

zellerh:
Make sure we complete a partial region start key for a datetime column
such that year, month and day are set to at least 1.


zellerh:
Thanks, very good catch! I'll fix this. Somehow had date, timestamp(0) and timestamp in mind instead of realizing that time doesn't have the date portion.

PR #942, opened 2017-02-01
[TRAFODION-2469] TM clients like dtmci don't exit cleanly

selvaganesang:
Created a new API TMCLIENTEXIT in TM library to clean up
opens. dtmci now calls this API before exiting


prashanth-vasudev:
Looks good to me.

PR #943, opened 2017-02-02
[TRAFODION-2470] Fixed one cause of 6004 warnings

DaveBirdsall:
The function literalOfNumericPassingScale (parser/SqlParserAux.cpp) is used to parse literals into ConstValue nodes and encoded values. It was incorrectly handling signed numeric literals of 20 characters in length: it would encode these as an unsigned 64-bit integer, then create a signed SQLLargeInt or SQLNumeric with that unsigned encoded value. One symptom of this is that histograms for NUMERIC datatypes would encounter 6004 errors (out-of-order histogram intervals) if there were some values requiring less than 20 characters and others requiring 20 or more.

The fix here reverts to using SQLBigNum instead for such values.

### DIFF
--- a/core/sql/parser/SqlParserAux.cpp
+++ b/core/sql/parser/SqlParserAux.cpp
@@ -863,8 +863,7 @@ ItemExpr *literalOfNumericPassingScale(NAString *strptr, char sign,
   } else if (strSize <= 19) {
     datatype = (createSignedDatatype ? REC_BIN64_SIGNED : REC_BIN64_UNSIGNED);
     length = sizeof(Int64);
-  } else if (strSize == 20) {
-    createSignedDatatype = FALSE;
+  } else if ((strSize == 20) && (!createSignedDatatype)) {
     datatype = REC_BIN64_UNSIGNED;
     length = sizeof(Int64);
   }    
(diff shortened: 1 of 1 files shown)


core/sql/parser/SqlParserAux.cpp
```diff
@@ -863,8 +863,7 @@ ItemExpr *literalOfNumericPassingScale(NAString *strptr, char sign,
   } else if (strSize <= 19) {
     datatype = (createSignedDatatype ? REC_BIN64_SIGNED : REC_BIN64_UNSIGNED);
     length = sizeof(Int64);
-  } else if (strSize == 20) {
-    createSignedDatatype = FALSE;
+  } else if ((strSize == 20) && (!createSignedDatatype)) {
```
zellerh: This is in the surround, but I wonder whether we shouldn't also look at the TRAF_LARGEINT_UNSIGNED_IO CQD and only generate unsigned largeint literals if that CQD is set to ON.

DaveBirdsall:
Thanks, @zellerh, for the suggestion. I agree, and it was easy to make the suggested change, so I made it.

zellerh:
+1

Thank you for adding this!

sureshsubbiah:
+1

PR #946, opened 2017-02-07
TRAFODION-1562  Enable Vsbb upsert for base table nodes when more than row

sandhyasun:
This is part of a performance enhancment  for the upsert with index feature. The CQD tRAF_UPSERT_TO_EFF_TREE  is also being turned on by default with this PR.  Had to make a few tweaks to the merge transformation so both the old and new ways are intact.  Enhanced the VSBB executor operator to project rows. This is enabled in the generator only when needed - i.e when the operator needs to freturn rows for other operators to consume.


core/sql/executor/ExHbaseIUD.cpp
```diff
@@ -1022,9 +1088,10 @@ ExWorkProcRetcode ExHbaseAccessUpsertVsbbSQTcb::work()
 	    if (NOT hbaseAccessTdb().computeRowsAffected())
 	      matches_ = 0;
 
-	    if ((step_ == DONE) &&
+             if ((step_ == DONE) &&
```
DaveBirdsall: Was this indentation change intentional? It now looks like the "if ((step_ == DONE)..." is executed when "NOT hbaseAccessTdb().computeRowsAffected" is true.

core/sql/executor/ExHbaseIUD.cpp
```diff
@@ -924,6 +938,58 @@ ExWorkProcRetcode ExHbaseAccessUpsertVsbbSQTcb::work()
...
+
+	 	// allocate tupps where returned rows will be created
+                if (allocateUpEntryTupps(
+                         -1,
+                         0,
+                         hbaseAccessTdb().returnedTuppIndex_,
+                         hbaseAccessTdb().returnUpdatedRowLen_,
+                         FALSE,
+                         &rc))
+              return rc;
```
DaveBirdsall: The inconsistent indentation makes this code harder to read.

core/sql/executor/ExHbaseIUD.cpp
```diff
@@ -924,6 +938,58 @@ ExWorkProcRetcode ExHbaseAccessUpsertVsbbSQTcb::work()
...
+		if (exprRetCode == ex_expr::EXPR_ERROR)
+		  {
+		    step_ = HANDLE_ERROR;
+		    break;
+		  }
+		
+		rc = 0;
+		// moveRowToUpQueue also increments matches_
+		if (moveRowToUpQueue(&rc))
+		  return 1;
```
DaveBirdsall: Elsewhere in the code we use enums such as WORK_OK for the return from this procedure. What does 1 mean?

core/sql/generator/GenPreCode.cpp
```diff
@@ -5486,11 +5490,15 @@ RelExpr * HbaseInsert::preCodeGen(Generator * generator,
       ((getInsertType() == Insert::VSBB_INSERT_USER) ||
        (getInsertType() == Insert::UPSERT_LOAD)))
     {
-      if ((inlinedActions || producesOutputs())&& !getIsTrafLoadPrep())
- 	setInsertType(Insert::SIMPLE_INSERT);
+      // Remove this restriction
```
DaveBirdsall: Perhaps the whole block from line 5489 should be deleted?

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -10600,34 +10600,45 @@ Upsert is also converted into merge when TRAF_UPSERT_MODE is set to MERGE and
 there are omitted cols with default values in case of aligned format table or 
 omitted current timestamp cols in case of non-aligned row format
 */
-NABoolean Insert::isUpsertThatNeedsMerge(NABoolean isAlignedRowFormat, NABoolean omittedDefaultCols,
-                                   NABoolean omittedCurrentDefaultClassCols) const
+NABoolean Insert::isUpsertThatNeedsTransformation(NABoolean isAlignedRowFormat, 
+                                                  NABoolean omittedDefaultCols,
+                                                  NABoolean omittedCurrentDefaultClassCols,
+                                                  NABoolean &toMerge) const
```
DaveBirdsall: You've added a parameter to return more information from this function. This is fine. Another way to do it is to change the return type from NABoolean to an enum, then no added parameter would be needed. But this is just a matter of taste; it may or may not be more convenient for the caller.

core/sql/optimizer/Inlining.cpp
```diff
@@ -618,10 +618,14 @@ RETDesc *GenericUpdate::createOldAndNewCorrelationNames(BindWA *bindWA, NABoolea
     rd = new (bindWA->wHeap()) RETDesc(bindWA);
   }
 
+  /*  if ((getOperatorType() != REL_UNARY_INSERT) || 
+      getUpdateCKorUniqueIndexKey() ||
+      ((getOperatorType() == REL_UNARY_INSERT) &&((Insert *)this)->isMerge()) ||
+      ((getOperatorType() == REL_UNARY_INSERT) && ((Insert *)this)->isUpsert() && (CmpCommon::getDefault(TRAF_UPSERT_TO_EFF_TREE) == DF_ON )))  */
```
DaveBirdsall: It's better to not leave this old code here... it just clutters things up and makes reading the code more confusing. We can always recover the old code from the repository if needed.

core/sql/executor/ExHbaseIUD.cpp
```diff
@@ -912,8 +912,22 @@ ExWorkProcRetcode ExHbaseAccessUpsertVsbbSQTcb::work()
 	    copyRowIDToDirectBuffer(rowId_);
 
 	    currRowNum_++;
-	    matches_++;
-
+            // temp code below.WIll remove the envvar after this
+            //has got enough exposure in  testing.
+            if (getenv("TURN_OFF_RETURNROW"))
```
sureshsubbiah: Please consider removing this. From what I can tell from Google, getenv will have a performance impact, Particularly when called so often (once per row?). A cqd or COMP_BOOL can be used to tun feature of temporarily. If a runtime toggle is needed, maybe we call getenv at the top somewhere and store the value in a local variable.

core/sql/optimizer/Inlining.cpp
sandhyasun: Will remove .

core/sql/executor/ExHbaseIUD.cpp
sandhyasun: Sorry about that. I'll definitely fix all the indentation.

core/sql/executor/ExHbaseIUD.cpp
sandhyasun: Thanks - I will remove this. I had used it for quick debugging. I didn't realize it'll have an impact on performance.  Will look into fixing it and submit a new version of this file.

core/sql/executor/ExHbaseIUD.cpp
sandhyasun: This was cloned from other parts of this file. I have changed it in this case to return "rc". That's what is supposed to be returned.

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -10600,34 +10600,45 @@ Upsert is also converted into merge when TRAF_UPSERT_MODE is set to MERGE and
...
+      // table has secondary indexes or
+      (getTableDesc()->hasSecondaryIndexes() ))
+        {
+          toMerge = FALSE;
+          return TRUE;
+        }
+  else if (isUpsert() &&   (NOT getIsTrafLoadPrep()) && 
+           (NOT (getTableDesc()->isIdentityColumnGeneratedAlways() && getTableDesc()->hasIdentityColumnInClusteringKey())) && 
+           (NOT (getTableDesc()->getClusteringIndex()->getNAFileSet()->hasSyskey())) &&         
+           // CQD is set to MERGE
```
sureshsubbiah: The conditions in IF and ELSE IF are duplicated, making it harder to read. Please consider using a local variable to store the result of common conditions. Also, it is not clear what should happen if we have both IM (no syskey etc.) and default values in the upsert. With current code we will do the new efficient transformation. I suppose that means the new efficient transformation can handle the missing default values case too. Is the else if case used as fallback ?

core/sql/regress/executor/EXPECTED015.SB
```diff
@@ -1961,51 +1970,51 @@ TRAFODION_MERGE
...
 
 --- SQL command prepared.
->>execute explainIt;
+>>execute explainItEff;
 
 OPERATOR        
 ----------------
 
-TRAFODION_MERGE 
+SEQUENCE
```
sureshsubbiah: Test cases are nice and robust. I like them.

sureshsubbiah:
Very nice work @sandhyasun. I am excited that we have this feature now.

core/sql/executor/ExHbaseIUD.cpp
```diff
@@ -1042,7 +1104,10 @@ ExWorkProcRetcode ExHbaseAccessUpsertVsbbSQTcb::work()
 
 		numRetries_++;
 		return WORK_CALL_AGAIN;
-	      }
+                
+                step_ = PROCESS_INSERT_AND_CLOSE;
```
DaveBirdsall: I think a bug might have crept in here. This statement can't be reached because of the return at line 1106.

core/sql/executor/ExHbaseIUD.cpp
```diff
@@ -1620,21 +1685,20 @@ ExWorkProcRetcode ExHbaseAccessBulkLoadPrepSQTcb::work()
...
-				   0,
-				   hbaseAccessTdb().returnedTuppIndex_,
-				   hbaseAccessTdb().returnUpdatedRowLen_,
-				   FALSE,
-				   &rc))
+	  ex_queue_entry * up_entry = qparent_.up->getTailEntry();	 
 	    return rc;
-	  
+            // allocate tupps where returned rows will be created
+            if (allocateUpEntryTupps(
```
DaveBirdsall: This statement can't be reached because of the return at 1692.

core/sql/executor/ExHbaseIUD.cpp
```diff
@@ -1620,21 +1685,20 @@ ExWorkProcRetcode ExHbaseAccessBulkLoadPrepSQTcb::work()
...
+                 -1,
+                 0,
+                 hbaseAccessTdb().returnedTuppIndex_,
+                 hbaseAccessTdb().returnUpdatedRowLen_,
+                 FALSE,
+                 &rc))  
+              return 1;
```
DaveBirdsall: Should this be return rc?

core/sql/executor/ExHbaseIUD.cpp
sandhyasun: Removed this inadvertant bug. Thanks !

core/sql/executor/ExHbaseIUD.cpp
sandhyasun: Yes . Not needed. We transition to PROCESS_INSERT_AND_CLOSE only after 3 tries.

core/sql/optimizer/BindRelExpr.cpp
sandhyasun: Good point...Let me rework this to be more readable and may have to add a 3rd check.

core/sql/generator/GenRelUpdate.cpp
```diff
@@ -2844,8 +2848,8 @@ short HbaseInsert::codeGen(Generator *generator)
 	  (noCheck()))
 	hbasescan_tdb->setHbaseSqlIUD(FALSE);
 
-      if (((getInsertType() == Insert::VSBB_INSERT_USER) && 
-                   generator->oltOptInfo()->multipleRowsReturned()) ||
+      if (((((getInsertType() == Insert::VSBB_INSERT_USER) || isUpsert() )&& 
+           generator->oltOptInfo()->multipleRowsReturned())) ||
```
zellerh: A nit, there is an unnecessary pair of parentheses at the beginning of line 2851 and the end of line 2852.

core/sql/generator/GenRelUpdate.cpp
```diff
@@ -2855,8 +2859,8 @@ short HbaseInsert::codeGen(Generator *generator)
            setVsbbInsert(TRUE);
       }
 
-      if ((isUpsert()) &&
-	  (getInsertType() == Insert::UPSERT_LOAD))
+      
+        if (getInsertType() == Insert::UPSERT_LOAD)
```
zellerh: Just to make sure I understand: You removed the isUpsert() here because you know that for upserts that are VSBB inserts, the insert type is UPSERT_LOAD?

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -10600,34 +10600,45 @@ Upsert is also converted into merge when TRAF_UPSERT_MODE is set to MERGE and
...
+                                                  NABoolean omittedDefaultCols,
+                                                  NABoolean omittedCurrentDefaultClassCols,
+                                                  NABoolean &toMerge) const
 {
-  // The necessary conditions to convert upsert to merge and
+  // The necessary conditions to transform upsert 
   if (isUpsert() && 
       (NOT getIsTrafLoadPrep()) && 
-      (NOT (getTableDesc()->isIdentityColumnGeneratedAlways() && getTableDesc()->hasIdentityColumnInClusteringKey())) && 
+      (NOT (getTableDesc()->isIdentityColumnGeneratedAlways() &&
```
zellerh: Maybe add a comment here that tables with an identity or syskey column never encounter any duplicate rows on insert/upsert.

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -10600,34 +10600,45 @@ Upsert is also converted into merge when TRAF_UPSERT_MODE is set to MERGE and
...
+          return TRUE;
+        }
+  else if (isUpsert() &&   (NOT getIsTrafLoadPrep()) && 
+           (NOT (getTableDesc()->isIdentityColumnGeneratedAlways() && getTableDesc()->hasIdentityColumnInClusteringKey())) && 
+           (NOT (getTableDesc()->getClusteringIndex()->getNAFileSet()->hasSyskey())) &&         
+           // CQD is set to MERGE  
+           ((CmpCommon::getDefault(TRAF_UPSERT_MODE) == DF_MERGE) &&
             // omitted current default columns with non-aligned row format tables
             // or omitted default columns with aligned row format tables 
             (((NOT isAlignedRowFormat) && omittedCurrentDefaultClassCols) ||
```
zellerh: Maybe another short comment here that TRAF_UPSERT_MERGE controls the behavior of upserting an existing row with a subset of columns. ON means keep the column values of the existing row for the omitted columns (slower) and OFF means setting the omitted columns to default values (faster). Since there is no standard for UPSERT, either behavior could be considered "correct".

sandhyasun: Yes I am reworking this and making the code clearer with new comments.

sandhyasun:
jenkins extra tests

sandhyasun:
jenkins retest

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -10605,43 +10605,62 @@ NABoolean Insert::isUpsertThatNeedsTransformation(NABoolean isAlignedRowFormat,
...
+  // Transform upsert to efficient tree if none of the above conditions 
+  // are true and the table has secondary indexes 
+  if (isUpsert() &&  
+      mustTryTransform &&
+      (getTableDesc()->hasSecondaryIndexes()))
+    {
+      toMerge = FALSE;
+      return TRUE;
+    }
+
```
sandhyasun: @sureshsubbiah @zellerh  Please take a look at this file's rework and see if it looks better. Also I think we shoudl separately discuss making the default TRAF_UPSERT_OPTIMAL  since it's better to make a faster performing solution the default behaviour and still have support for the merge behaviour if needed.

sureshsubbiah: Looks good @sandhyasun. All my concerns have been addressed in the best possible manner. Thanks.+1

selvaganesang:
Is there a way to avoid white space changes in the review via git? If not, can you please provide the new and old folder so that we can use bcompare

arvind-narain:
Hi @selvaganesang - you could try the following - add w=1 option

https://github.com/apache/incubator-trafodion/pull/946/files?diff=split&w=1

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -10600,37 +10600,67 @@ Upsert is also converted into merge when TRAF_UPSERT_MODE is set to MERGE and
...
+  if (isUpsert()  &&   
+      mustTryTransform &&          
+      ((CmpCommon::getDefault(TRAF_UPSERT_MODE) == DF_MERGE) &&     
+       (((NOT isAlignedRowFormat) && omittedCurrentDefaultClassCols) ||
+        (isAlignedRowFormat && omittedDefaultCols)))
+      ||
+      ((CmpCommon::getDefault(TRAF_UPSERT_MODE) == DF_OPTIMAL) &&
+       ((NOT isAlignedRowFormat) && omittedCurrentDefaultClassCols)))
+    {
+      toMerge = TRUE;
```
selvaganesang: Is there a reason for not choosing the new efficient tree instead?

selvaganesang:
I would suggest test cases displaying query plans are added to ensure vsbb upsert and vsbb_delete is chosen when multiple tuples are involved except for the unique indexes.

core/sql/optimizer/BindRelExpr.cpp
sandhyasun: If we choose the new tree, we won't get merge behaviour. We will get "replace" behaviour. I do think we need to make DF_OPTIMAL the default behaviour so the more efficient one is the default behaviour and if the user wants merge behaviour they can choose the slower method. We can discuss that separately.

sandhyasun:
Test cases to show vsbb are already added and reviewed as being robust.

core/sql/optimizer/BindRelExpr.cpp
selvaganesang: The non-aligned format table needs merge behavior by default. Will the non-aligned format table work as it was before then?

selvaganesang:
Are the test cases part of this PR? Or do we have these test cases existing already?

core/sql/optimizer/BindRelExpr.cpp
sandhyasun: Correct. Yes it will . As specified  in Case 1: and Case 2: comments in the above code.

sandhyasun:
Yes executor/TEST015 where it looks for vsbb_upsert in the plan for the basetable when there is more than one row being upserted during transformation. That is what this PR addresses.

core/sql/optimizer/BindRelExpr.cpp
selvaganesang: The default is to behave like 'MERGE" for both aligned and non-aligned format tables with omitted columns and omitted class columns respectively.  After carefully reading the code, it looks like MERGE is being chosen for this default behavior. Yes. We can a create a new JIRA to track this

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -10600,37 +10600,67 @@ Upsert is also converted into merge when TRAF_UPSERT_MODE is set to MERGE and
...
-            ((NOT isAlignedRowFormat) && omittedCurrentDefaultClassCols))
-        ) 
-     )
-     return TRUE;
-  else
-     return FALSE;
+      ( NOT getIsTrafLoadPrep()) &&
+      (NOT (getTableDesc()->isIdentityColumnGeneratedAlways() && 
+            getTableDesc()->hasIdentityColumnInClusteringKey()))  && 
+      (NOT(getTableDesc()->getClusteringIndex()->getNAFileSet()->hasSyskey())))
```
selvaganesang: Consider making this as a OR condition for better readability because the original condition is now split into two

sandhyasun:
jenkins extra tests

selvaganesang:
Change looks good. I have added two scripts setup.sql and query.sql to [TRAFODION-1546] that I used to test upsert/update command.  I have also attached the output file too from the previous runs. Please consider testing this change with these scripts and make it as part of the trafodion regressions  if you can

PR #947, opened 2017-02-07
[TRAFODION-2472] disable transactional alter command for hbase options.

prashanth-vasudev:
Transaction DDL for alter commands is currently disabled. 
There are few statements such as alter hbase option that is not disabled which results in unpredictable errors. 
Initially fix would be to disable alter statement to not use DDl transaction. 

Following this initial fix,  DDL transaction would be enhanced to support of Alter table statement.


prashanth-vasudev:
Thanks for the review. Have added DDL_ALTER_INDEX_ALTER_HBASE_OPTIONS  as well.

sureshsubbiah:
+1

DaveBirdsall:
I did a no-no... I merged this before tests passed... apologies.

PR #725, opened 2016-09-23
TRAFODION-2231 support transactional ddl exceptions to sql.

prashanth-vasudev:

PR #949, opened 2017-02-08
[TRAFODION-2475] Create table/index or alter does not work with unava…

sureshsubbiah:
…ilable compression options

### DIFF
--- a/core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
+++ b/core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
@@ -96,6 +96,7 @@ import org.apache.hadoop.hbase.regionserver.StoreFileInfo;
 import org.apache.hadoop.hbase.HConstants;
 import org.apache.hadoop.hbase.client.DtmConst;
 import org.apache.commons.codec.binary.Hex;
+import org.apache.hadoop.hbase.util.CompressionTest;
 
 import com.google.protobuf.ServiceException;
 
@@ -245,7 +246,8 @@ public class HBaseClient {
    private ChangeFlags setDescriptors(Object[] tableOptions,
                                       HTableDescriptor desc,
                                       HColumnDescriptor colDesc,
-                                      int defaultVersionsValue) {
+                                      int defaultVersionsValue) 
+       throws IOException {
        ChangeFlags returnStatus = new ChangeFlags();
        String trueStr = "TRUE";
        for (int i = 0; i < tableOptions.length; i++) {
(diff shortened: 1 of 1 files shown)


core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
```diff
@@ -294,15 +296,27 @@ private ChangeFlags setDescriptors(Object[] tableOptions,
...
+		   CompressionTest.testCompression(Algorithm.LZO);
                    colDesc.setCompressionType(Algorithm.LZO);
+	       }
                else if (tableOption.equalsIgnoreCase("NONE"))
                    colDesc.setCompressionType(Algorithm.NONE);
                else if (tableOption.equalsIgnoreCase("SNAPPY"))
+	       {   // throws IOException
+		   CompressionTest.testCompression(Algorithm.SNAPPY);
                    colDesc.setCompressionType(Algorithm.SNAPPY); 
+	       }
```
DaveBirdsall: Should there be an exception if the tableOption isn't recognized? (Right now the code just ignores it.)

core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
```diff
@@ -354,16 +368,28 @@ else if (tableOption.equalsIgnoreCase("PREFIX_TREE"))
...
                    colDesc.setCompactionCompressionType(Algorithm.LZO);
+	       }
                else if (tableOption.equalsIgnoreCase("NONE"))
                    colDesc.setCompactionCompressionType(Algorithm.NONE);
-               else if (tableOption.equalsIgnoreCase("SNAPPY"))
+               else if (tableOption.equalsIgnoreCase("SNAPPY")) {
+		   // throws IOException
+		   CompressionTest.testCompression(Algorithm.SNAPPY);
                    colDesc.setCompactionCompressionType(Algorithm.SNAPPY); 
+	       }
```
DaveBirdsall: Same remark as above.

core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
sureshsubbiah: Good catch Dave. I was confident that invalid compression options like "SNAPY" would be caught earlier in the parse/bind stage. Now I see that validation of input arguments is not thorough. I will add a commit to this PR by tomorrow with that change. Much validation can be done on the C++ side before we call the create()/alter() Java function. For an allowed compression setting usable by HBase, the only verification I know is the Java one provided by HBase itself. Therefore this validation could be done once in Java side and results stored in C++ side or we can do as this PR does, validate and raise error in Java side when necessary. The method provided by HBase uses static variables to store results of previous validations, so we will not be fully validating the same compression algorithm twice in a given session.

This is how the error message looks like now.
create table test5 (a int not null primary key, b int) hbase_options (compression = 'SNAPPY') salt using 2 partitions ;

*** ERROR[8448] Unable to access Hbase interface. Call to ExpHbaseInterface::create() returned error HBASE_CREATE_ERROR(701). Cause: org.apache.hadoop.hbase.DoNotRetryIOException: java.lang.RuntimeException: native snappy library not available: this version of libhadoop was built without snappy support.
org.apache.hadoop.hbase.util.CompressionTest.testCompression(CompressionTest.java:102)
org.trafodion.sql.HBaseClient.setDescriptors(HBaseClient.java:317)
org.trafodion.sql.HBaseClient.createk(HBaseClient.java:490) Caused by 
java.lang.RuntimeException: native snappy library not available: this version of libhadoop was built without snappy support.
org.apache.hadoop.io.compress.SnappyCodec.checkNativeCodeLoaded(SnappyCodec.java:64)
org.apache.hadoop.io.compress.SnappyCodec.getCompressorType(SnappyCodec.java:133)

selvaganesang:
Isn't that we are getting proper exception to sql due to #725 https://github.com/apache/incubator-trafodion/pull/725

sureshsubbiah:
@selvaganesang, I am interpreting your comment to mean that with PR 725 we may get a proper exception even without this change.

It seems safer to check for appropriate compression libs are installed and available to Hbase before creating/altering a table. When we rely on HBase master/RS code to raise the exeption we find that (a) for LZO no error is raised, but created table is not useful (this is in my dev env where LZO is not installed) (b) for SNAPPY an exception is raised but it will several minutes due to retry and some other aspect of HBase code. I doubt PR 725 will address both these issues.

DaveBirdsall:
@selvaganesang, I was experiencing hangs when using uninstalled compression on a workstation as recently as last week. PR 725 went in in November, so evidently that didn't fix the issue Suresh is trying to address.

selvaganesang:
I concur with both the comments above.  I missed reading the line CompressionTest.testCompression. Hence I was wondering how an exception would be raised here.

PR #950, opened 2017-02-09
[TRAFODION-2471] pyinstaller: error command when grant privilege in HBase.

mkby:
The old command doesn't work.
also change the owner of HDFS directory /user/trafodion


robertamarton:
looks good to me.

PR #951, opened 2017-02-10
[TRAFODION-2314] MXOSRVR sometimes exits abnormally with NAMutex assert

arvind-narain:
Fixes as suggested by @selvaganesang and @sandhyasun 

We were hitting the code to make heap thread safe repeatedly while running
jdbc and phoenix tests.

PROCESS_LOB flag was being set incorrectly in some cases.

Unlocking a mutex owned by a different thread lead to the assertion.

This could also result in heap corruption.

### DIFF
--- a/core/sql/generator/GenPreCode.cpp
+++ b/core/sql/generator/GenPreCode.cpp
@@ -4177,9 +4177,11 @@ RelExpr * FileScan::preCodeGen(Generator * generator,
 	 TRUE);
 
       if (isHiveTable())
-	// assign individual files and blocks to each ESPs
-	((NodeMap *) getPartFunc()->getNodeMap())->assignScanInfos(hiveSearchKey_);
-       generator->setProcessLOB(TRUE);
+        {
+          // assign individual files and blocks to each ESPs
+          ((NodeMap *) getPartFunc()->getNodeMap())->assignScanInfos(hiveSearchKey_);
+          generator->setProcessLOB(TRUE);
+        }
     }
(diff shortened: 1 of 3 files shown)


selvaganesang:
Thanks @arvind-narain for these changes. This should also fix the memory corruption with hive access if seen already.

core/sql/generator/GenPreCode.cpp
```diff
@@ -4177,9 +4177,11 @@ RelExpr * FileScan::preCodeGen(Generator * generator,
 	 TRUE);
 
       if (isHiveTable())
-	// assign individual files and blocks to each ESPs
-	((NodeMap *) getPartFunc()->getNodeMap())->assignScanInfos(hiveSearchKey_);
-       generator->setProcessLOB(TRUE);
+        {
+          // assign individual files and blocks to each ESPs
+          ((NodeMap *) getPartFunc()->getNodeMap())->assignScanInfos(hiveSearchKey_);
+          generator->setProcessLOB(TRUE);
```
sandhyasun: Need to ensure this is set to TRU for LOB datatype  access too not just for hive .

sandhyasun:
jenkins extra tests

core/sql/generator/GenPreCode.cpp
selvaganesang: Isn't setLOBAccess set to TRUE in LOBoper::preCodeGen for LOB datatype.

PR #953, opened 2017-02-11
[TRAFODION-2478] Reduce the number of memory monitoring threads in Tr…

selvaganesang:
…afodion SQL processes

The memory monitor thread is now moved to mxsscp process and the collected data is
stored in RMS shared segment. All SQL proceses have access to it.

### DIFF
--- a/core/sql/bin/ex_esp_main.cpp
+++ b/core/sql/bin/ex_esp_main.cpp
@@ -357,15 +357,18 @@ Int32 runESP(Int32 argc, char** argv, GuaReceiveFastStart *guaReceiveFastStart)
   cliGlobals->initiateDefaultContext();
   NAHeap *espIpcHeap = cliGlobals->getIpcHeap();
   IpcEnvironment *ipcEnvPtr = cliGlobals->getEnvironment();
-  //
-  // Start the  memory monitor for dynamic memory management
-  Lng32 memMonitorWindowSize = 10;
-  Lng32 memMonitorSampleInterval = 10;
-  MemoryMonitor memMonitor(memMonitorWindowSize,
+  if (statsGlobals != NULL)
+     cliGlobals->setMemoryMonitor(statsGlobals->getMemoryMonitor());
+  else 
+  {
+     // Start the  memory monitor for dynamic memory management
+     Lng32 memMonitorWindowSize = 10;
+     Lng32 memMonitorSampleInterval = 10;
+     MemoryMonitor memMonitor(memMonitorWindowSize,
                            memMonitorSampleInterval,
                            espExecutorHeap);
-  cliGlobals->setMemoryMonitor(&memMonitor);
-
+     cliGlobals->setMemoryMonitor(&memMonitor);
+  }
   // After CLI globals are initialized but before we begin ESP message
   // processing, have the CLI context set its user identity based on
   // the OS user identity.
--- a/core/sql/bin/ex_esp_main.cpp
+++ b/core/sql/bin/ex_esp_main.cpp
@@ -364,10 +364,11 @@ Int32 runESP(Int32 argc, char** argv, GuaReceiveFastStart *guaReceiveFastStart)
      // Start the  memory monitor for dynamic memory management
      Lng32 memMonitorWindowSize = 10;
      Lng32 memMonitorSampleInterval = 10;
-     MemoryMonitor memMonitor(memMonitorWindowSize,
+     MemoryMonitor *memMonitor = new (espExecutorHeap) 
+                           MemoryMonitor(memMonitorWindowSize,
                            memMonitorSampleInterval,
                            espExecutorHeap);
-     cliGlobals->setMemoryMonitor(&memMonitor);
+     cliGlobals->setMemoryMonitor(memMonitor);
   }
   // After CLI globals are initialized but before we begin ESP message
   // processing, have the CLI context set its user identity based on
--- a/core/sql/cli/Globals.cpp
+++ b/core/sql/cli/Globals.cpp
@@ -235,19 +235,7 @@ void CliGlobals::init( NABoolean espProcess,
   if (! espProcess)
   {
     // Create the process global ARKCMP server.
-    // In R1.8, Each context has its own mxcmp.
     sharedArkcmp_ = NULL;
-
-    //sharedArkcmp_->setShared(FALSE);
-    // create the process global memory monitor. For now with
-    // defaults of 10 window entries and sampling every 1 second
-    Lng32 memMonitorWindowSize = 10;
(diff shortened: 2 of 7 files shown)


core/sql/cli/memorymonitor.cpp
```diff
@@ -102,12 +102,18 @@ MemoryMonitor::MemoryMonitor(Lng32 windowSize,
...
-  char buffer[1024];
+  char buffer[2048];
   char *currPtr;
   size_t bytesRead;
   fd_meminfo_ = fopen("/proc/meminfo", "r");
   if (fd_meminfo_) {
-    bytesRead = fread(buffer, 1, 1024, fd_meminfo_);
+    bytesRead = fread(buffer, 1, 2048, fd_meminfo_);
+    if (ferror(fd_meminfo_))
+       assert(false);
```
zellerh: Could you remind me what this assert would do? Would it just write a log entry or would it kill the process and cause a core dump? I would say we shouldn't cause a failure of the process.

core/sql/cli/memorymonitor.cpp
```diff
@@ -102,12 +102,18 @@ MemoryMonitor::MemoryMonitor(Lng32 windowSize,
...
   // if the windowSize is 0, we do not need memory monitor.
   assert(windowSize);
-  char buffer[1024];
+  char buffer[2048];
   char *currPtr;
   size_t bytesRead;
   fd_meminfo_ = fopen("/proc/meminfo", "r");
   if (fd_meminfo_) {
-    bytesRead = fread(buffer, 1, 1024, fd_meminfo_);
+    bytesRead = fread(buffer, 1, 2048, fd_meminfo_);
```
zellerh: Can I suggest to use sizeof(buffer) here instead of 2048? Same for lines 230 and 271 below.

core/sql/cli/memorymonitor.cpp
```diff
@@ -223,12 +229,12 @@ void MemoryMonitor::update(float &scale) {
...
-	        scale = 6;
-		pressure_ = 0;
-		return;
-        }
+        if (ferror(fd_meminfo_))
+           assert(false); 
+        if (feof(fd_meminfo_))
+           clearerr(fd_meminfo_); 
+        else
+           buffer[bytesRead] = '\0';
```
zellerh: If we get an EOF, which should be the normal case, then shouldn't we also set buffer[bytesRead] to \0? In either case, to do that, we would need to allocate one extra byte to avoid a buffer overrun.

core/sql/cli/Globals.cpp
```diff
@@ -322,6 +310,18 @@ void CliGlobals::init( NABoolean espProcess,
     capacities_.setHeap(defaultContext_->exCollHeap());
     freespaces_.setHeap(defaultContext_->exCollHeap());
     largestFragments_.setHeap(defaultContext_->exCollHeap());
+    if (statsGlobals_ != NULL) 
+       memMonitor_ = statsGlobals_->getMemoryMonitor();
```
zellerh: Now that the memory monitor resides in another process, I think we need to make any of its methods that shouldn't be called by the executor thread private, like update() and updatePageFaultRate(). It would also be good to update the comments in memorymonitor.h to indicate which calls are meant for consumers and which ones are potentially in a separate thread.

core/sql/cli/memorymonitor.cpp
selvaganesang: It will cause a failure of the mxsscp

core/sql/cli/memorymonitor.cpp
selvaganesang: I can. Usually, I try not to use sizeof of arrays because it can cause confusion if the array is passed around. I will change it to a define.

core/sql/cli/memorymonitor.cpp
selvaganesang: fread cpp reference says the following:
Number of objects read successfully, which may be less than count if an error or end-of-file condition occurs. 

It looks like feof is treated like an error condition. 
This code assumed that buffer is always big enough. I can change it to get rid of this assumption

core/sql/cli/Globals.cpp
selvaganesang: I will make both the methods private.

core/sql/bin/ex_esp_main.cpp
```diff
@@ -357,15 +357,18 @@ Int32 runESP(Int32 argc, char** argv, GuaReceiveFastStart *guaReceiveFastStart)
...
-  Lng32 memMonitorSampleInterval = 10;
-  MemoryMonitor memMonitor(memMonitorWindowSize,
+  if (statsGlobals != NULL)
+     cliGlobals->setMemoryMonitor(statsGlobals->getMemoryMonitor());
+  else 
+  {
+     // Start the  memory monitor for dynamic memory management
+     Lng32 memMonitorWindowSize = 10;
+     Lng32 memMonitorSampleInterval = 10;
+     MemoryMonitor memMonitor(memMonitorWindowSize,
```
selvaganesang: Yes. I will change the old code to use this espExecutorHeap and make it as a pointer

core/sql/cli/memorymonitor.cpp
selvaganesang: And possibly 10 core dumps of mxsscp process because mxsscp process is retried 10 times before giving up

core/sql/cli/memorymonitor.cpp
zellerh: @selvaganesang, I like that you added the terminating \0 to the buffer read, since we are using strstr below and since fread does not add this terminator. My comment was that we should be doing this unconditionally, otherwise strstr could potentially go beyond the buffer - although this is unlikely to happen in real life.

core/sql/cli/memorymonitor.cpp
zellerh: What do you think, maybe just log an error instead and continue (returning pressure=0), like the old code did?

core/sql/cli/memorymonitor.cpp
zellerh: I see. This is my favorite way of doing it for char arrays that are used in methods like snprintf and fread here. Just a style comment, please feel free to ignore, no need for a change.

core/sql/cli/memorymonitor.cpp
selvaganesang: I am not sure. This should never happen and hence I am thinking it should be ok to leave it as assert

zellerh:
+1
Looks good!

prashanth-vasudev:
+1 
Looks Good!

PR #957, opened 2017-02-13
TRAFODION-2441 user has only select privilege on a table can do ...

robertamarton:
TRAFODION-2409 support privilege control(column privileges) for hive tables
TRAFODION-2423 any user can perform 'initialize trafodion, drop'
TRAFODION-2435 Any user can perform TRUNCATE on native Hive tables.
TRAFODION-2463 Hive: Any user can do update statistics for hive tables

Fixed issues found while testing privileges with native Hive.

TRAFODION-2441:
  changed code that initializes owner privileges for views.

TRAFODION-2409:
  returning error message 1328 during attempt to grant unsupported column level
  privilege on hive table.

TRAFODION 2423:
  added privilege checks for all initialize commands, error 1017 is returned if
  not DB__ROOT

TRAFODION-2435:
  Returning error 1051 if TRUNCATE is attempted on a hive table where the
  current user has no privilege

TRAFODION-2463:
  Privilege checks added for Hive table during update statistics


core/sql/common/ComUser.cpp
```diff
@@ -348,6 +348,33 @@ Int16 ComUser::getAuthNameFromAuthID(Int32   authID,
...
+// is found
+//
+//  Returns:  true - role found
+//            false - role not found
+// ----------------------------------------------------------------------------
+bool ComUser::currentUserHasRole(Int32 roleID)
+{
+  Int32 numRoles = 0;
+  Int32 *roleIDs = 0;
+  if (SQL_EXEC_GetRoleList(numRoles, roleIDs) < 0)
```
robertamarton: It is the latter.  The cli allocates space from its heap.  It is implemented similar to other CLI requests,  if memory is required, it is allocated from the heap.  The destructor removes it.

PR #958, opened 2017-02-14
[TRAFODION-2235] Added monitoring of local monitor's znode to detect …

zcorrea:
…errors with Zookeeper quorum.


zcorrea:
Reviewers recommended:
Narendra Goyal
Trina Krug

core/sqf/monitor/linux/zclient.cxx
```diff
@@ -472,6 +469,65 @@ void CZClient::CheckCluster( void )
...
+        }
+        if ( ! IsZNodeExpired( Node_name, zerr ) )
+        {
+            if ( zerr == ZCONNECTIONLOSS || zerr == ZOPERATIONTIMEOUT )
+            {
+                // Ignore transient errors with the quorum.
+                // However, if longer than the session
+                // timeout, handle it as a hard error.
+                clock_gettime(CLOCK_REALTIME, &currentTime);
+                if (currentTime.tv_sec > myZNodeFailedTime_.tv_sec)
```
zcorrea: The desired behavior is to continually reset the local nodes ZNode expiration time when the ZNode has not expired, which is the normal state. The only errors that can return from IsZNodeExpired and that are handled at this point are communication errors with the Zookeeper quorum, i.e., the connection loss and operation timeout. These errors can be transient, but if they persist beyond the myZNodeFailedTime they indicate that the local ZNode has gone beyond the local monitor's session expiration time window and the local monitor must bring itself down.

So yes, setting resetMyZNodeFailedTime_ to true is to reset myZNodeFailedTime on each iteration, every zcMonitoringRateValue in seconds, to push the expiration time out until there is a communication failure with the Zookeeper quorum.

trinakrug: Sounds good Zalo!  Thumbs up from me then!

PR #961, opened 2017-02-16
[TRAFODION-2854] Improve SET STATISTICS command to display the inform…

selvaganesang:
…ation in new format

Currently SET STATISTICS ON | OFF alone is supported. It is improved to take some more options

SET STATISTICS PERTABLE ;
SET STATISTICS PROGRESS ;
SET STATISTICS DEFAULT ;
This will display pertable, progress, operator level statistics for the these options respectively.

SET STATISTICS ALL will display both progress and operator level statistics.

These commands are valid both in sqlci and trafci.


prashanth-vasudev:
+1 looks good to me.

PR #965, opened 2017-02-16
[TRAFODION-2475] Improved error handling for HBASE_OPTIONS clause

sureshsubbiah:
Two new attributes have also been added.


selvaganesang:
My opinion is that Trafodion shouldn't do semantics checking of hbase options because if there is a change in hbase for hbase options then Trafodion needs to be updated accordingly. Hbase options are just pass through for Trafodion.  But, Trafodion can do limited syntax checking

sureshsubbiah:
Thank you @selvaganesang for the review. This change does not do semantics checking on attribute value. What we do is syntax checking (Check if an enum value is to specified, then it is one of the allowed values). This HBase JIRA talks about semantics checking. https://issues.apache.org/jira/browse/HBASE-10591. This PR does not do anything to interfere with that.

The original comment by Dave that inspired this change was that the following create table statement succeeded. HBase itself silently disregarded the incorrect compress option. Our metadata had the incorrect value.

create table t (a int) hbase_options (compression = 'snapy') 
With this change we will raise a nice error early on.

;

selvaganesang:
Thanks @sureshsubbiah for the clarification.  So, when a new compression is supported in hbase, then Trafodion needs to be updated to allow this compression type.  I understand that the metadata would contain the incorrect information in hbase options because hbase allowed it. However it should cause no harm in Trafodion processing of the table. Isn't it?

If so and if hbase options are just pass through to hbase from Trafodion, then personally I think we should let hbase do the checking.  If these options are used to influence Trafodion processing then it makes sense to do the validation in Trafodion.

sureshsubbiah:
Thank you for the comments @selvaganesang. I don't quite understand the statement "However it should cause no harm in Trafodion processing of the table. Isn't it". Can you please explain? Maybe an example to illustrate your concern will help.

If you are saying that it is acceptable to not raise an error for the following statement, I don't agree.
create table t (a int) hbase_options (compression = 'snapy').
It is preferable in my opinion to fail fast. If this statement were to succeed, the user will imagine that the table is compressed while is its not. It could be a while before the problem is detected.

The way current code is constructed, if new compression options are supported by HBase, new code will have to added in Trafodion. This PR does not really make that any worse. 

In createTable API that is used in HBaseClient.java, we have to pass in an enum for the compression setting. Thought I would not want to, I don't  think there is an API to just pass the string that we got from the user.

zellerh:
Hi, just wanted to add my two cents. You are both bringing up very good points. My preference would also be that we just pass HBase table options through to HBase and let it do the checking. However, from what I heard and saw, HBase does not do a good job with such checking. We have seen HBase options that are accepted by HBase but then cause the region server to hang, e.g. when specifying an invalid split policy. So, I guess that Suresh's approach is the lesser of the two evils. Note, however, that it may still not prevent a problem in cases where - for example - the specified compression library is not installed.

DaveBirdsall:
I agree with @zellerh. Another thing we can do is make a contribution to HBase to more properly evaluate its parameters. As Trafodion moves onto the newer HBase, we can then remove our checking.

sureshsubbiah:
Thank you for the review. A few specific responses

1) " that it may still not prevent a problem in cases where - for example - the specified compression library is not installed". This is now addressed in Trafodion. We call an HBase API to check if library is installed and accessible by hbase before creating the table. This was done in https://github.com/apache/incubator-trafodion/pull/949

2) HBase does do some checking of create option. It is more detailed than what we do. I am not sure if it does all the small things we do though. However it has to be enabled by setting the attribute  hbase.table.sanity.checks in the global configuration. From what I can tell currently HDP does this by default while CDH does not. Please see https://issues.apache.org/jira/browse/HBASE-10591

3) There is a basic problem that is currently best addressed in our layer. We get a String from the user in the CREATE statement, specifying compression type. We have to pass in an Enum value to HBase when we use the HBaseAdmin create API. When mapping the string to an enum, if we find that there is a no mapping possible, then it seems best that Trafodion raise the error. I suppose in the future we could extend the HBase API to take a string and push this error handling to HBase too.

zellerh:
As always, Suresh is much better informed than I am, thanks for the clarification! Good to know how it really works. I probably need to revisit the code I added for the split policy, to see whether it needs a change. If any of you thinks that is the case, please let me know and I'll fix it. This code has a hard-coded list of split policies in Trafodion from which the user must choose.

sureshsubbiah:
This PR is modelled after your split-policy error handling code. I replaced the previous 8448 error in bad arguments with 8449 error you added for split policy. The only minor change for split policy is that we will now raise an error if the split policy is specified twice in the same CREATE statement. I think no change is needed for splitPolicy (unless we were to implement the nice dynamic model Selva outlined where we can be transparent to additions in HBase)

zellerh:
Thanks for letting me know.

PR #967, opened 2017-02-17
i added one profile used for publish jdbct4 to maven central, i need to submit first, otherwise could not publish.

AlexPeng19:
i added one profile used for publish jdbct4 to maven central, i need to submit first, otherwise could not publish.

### DIFF
--- a/core/conn/jdbcT4/pom.xml
+++ b/core/conn/jdbcT4/pom.xml
@@ -27,19 +27,18 @@
 -->
 
   <modelVersion>4.0.0</modelVersion>
-  <!--<parent>
+  <parent>
     <groupId>org.apache.trafodion</groupId>
     <artifactId>trafodion</artifactId>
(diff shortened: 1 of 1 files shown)


core/conn/jdbcT4/pom.xml
```diff
@@ -27,11 +27,54 @@
 -->
 
   <modelVersion>4.0.0</modelVersion>
-  <groupId>org.trafodion.jdbc.t4</groupId>
+  <!--<parent>
+    <groupId>org.apache.trafodion</groupId>
```
hegdean: Package name for jdbcT4  is org.trafodion.jdbc.t4. Just changing the group id to contain apache in the package name without the package structure change in jdbcT4 code will not make it work.

AlexPeng19:
too many commits, let me new pull request.

PR #970, opened 2017-02-20
[TRAFODION-2489] [TRAFODION-2490] fix for pyinstaller

mkby:
[TRAFODION-2489] pyinstaller - sqconfig on single node
[TRAFODION-2490] pyinstaller - admin user setup gives syntax error


sureshsubbiah:
+1

PR #971, opened 2017-02-20
i made some change used to publish jdbcT4 to maven central, not sure should i commit it or not.

AlexPeng19:
i made some change used to publish jdbcT4 to maven central, not sure should i commit it or not.


AlexPeng19:
i need to commit to release 2.1 for this change.

PR #975, opened 2017-02-22
TRAFODION-2441 user has only select privilege on a table can do ...

robertamarton:
This is the second delivery that updates "get" commands so users can only view
information where they have been granted privileges.  This delivery handles:

get components;
get privileges on component;
get privileges on component for <role>;
get privileges on component for <user> [cascade];
  cascade returns privileges for user plus any roles the user has been granted

### DIFF
--- a/core/sql/comexe/ComTdbExeUtil.h
+++ b/core/sql/comexe/ComTdbExeUtil.h
@@ -2604,6 +2604,10 @@ public:
   {(v ? flags_ |= IS_HBASE : flags_ &= ~IS_HBASE); };
   NABoolean isHbase() { return (flags_ & IS_HBASE) != 0; };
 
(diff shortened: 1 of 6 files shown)


sureshsubbiah:
Change looks great.
Following sandhya's comments yesterday,
1) Does this change need a documentation JIRA? If yes, could you please file one? I think the model we are adopting is that the new JIRA have some notes on what the documentation content should be.
2) A query in some regression test would be nice for a later delivery.

robertamarton:
Actually, it shouldn’t require documentation changes.  However, the documentation for GET commands needs some tweaks.
There are more privilege changes planned to support all statements with priv checks.
It might be good to wait until they are done but not sure when this will happen.
I will file a JIRA with a description on what needs to be done.

     Roberta

PR #976, opened 2017-02-22
TRAFODION-2492 Support for native hive views, plus few more changes

anoopsharma00:
-- added support for native hive views created externally. These views
   can be accessed in a traf sql query.

   There are some restrictions as listed below:
   -- privileges support has not yet been added. That will be done, as
      needed, in the next checkin.
   -- usage info for tables contained in a native hive view is not maintained.
   -- native hive views cannot be inserted/updated/deleted
   -- cannot update stats on a native hive view

-- support for GET statements to retrieve hive data
   get tables/views/objects in schema hive.sch;
    -- this will return all tables/views in specified schema(hive database)
   get tables/views/objects in catalog hive;
    -- this will return all tables/views in hive
   get schemas in catalog hive;
    -- this will return all schemas(databases) that exist in hive

-- GET statements to retrieve hive data now supports optional MATCH clause.
   This can be used to filter results.

-- showddl will show if an external table on hive was created implicitly
   for internal usage (upd stats, privileges, views) or explicitly by user.
   Internal creation of external table now use 'create implicit external table'
   syntax and that info is stored in metadata.

-- drop external table will drop that table even if underlying hive table
   is missing. Currently it returns an error.

-- at runtime, accessing a missing hive table now returns a better error
   message instead of an internal "data mod check failed" message.

-- regress/tools/runallsb has 2 new options:
    -basetests   (this will run core,compGeneral,executor,seabase,hive)
    -othertests  (this will run all other tests suites)

-- optimizer/NATable.cpp was incorrectly indented by some checkin from
   a few days back. TABs were used instead of spaces which caused code
   to become all crooked. That has been fixed.

-- regress/hive/TEST007 has been updated with additional hive view tests


core/sql/optimizer/NATable.cpp
```diff
@@ -5008,541 +5009,546 @@ NABoolean createNAFileSets(TrafDesc * table_desc       /*IN*/,
...
 
-  // change partFunc for base table if PARTITION clause has been used
-  // to limit the number of partitions that will be accessed.
-  if ((qualifiedName_.isPartitionNameSpecified()) ||
-		  (qualifiedName_.isPartitionRangeSpecified())) {
-	  if (filterUnusedPartitions(corrName.getPartnClause())) {
-		  return ;
-	  }
-  }
+       NABoolean  errorOccurred =
```
zellerh: Just a question: Do we support constraints on Hive exernal tables? For example a unique or check constraint?

anoopsharma00: No, we dont support constraints on hive external tables. An error will be returned if unique, foreign, or check constraint is specified while creating a hive external table. 

On the other question of IMPLICIT keyword, so that can only be used when the create stmt
is issued internally and the internal flag is set. If used externally, a syntax error will be returned.
From functionality pov, there is no difference between an explicit or implicit external table.

In showddl output, the implicit keyword shows up as an informational comment:
  create /*implicit*/ external table...
This is similar to how we also show traf internally created system indexes.

zellerh: Thanks for the info. If I have plenty of spare time someday I'll look into non-enforced constraints on Hive tables.

PR #977, opened 2017-02-23
Fix for a potential stack memory corruption.{TRAFODION-2495]

sandhyasun:
The lobHdfsServer was being overwritten by some code right below the line that read the default value and caused problems trying to make an HDFS connection.


core/sql/optimizer/NATable.cpp
```diff
@@ -5481,7 +5481,9 @@ NABoolean createNAFileSets(TrafDesc * table_desc       /*IN*/,
 
   if(postCreateNATableWarnings != preCreateNATableWarnings)
 	  tableConstructionHadWarnings_=TRUE;
-  const char *lobHdfsServer = CmpCommon::getDefaultString(LOB_HDFS_SERVER);
+  char  lobHdfsServer[256] ; // max length determined by dfs.namenode.fs-limits.max-component-length(255)
+  memset(lobHdfsServer,0,256);
+  strncpy(lobHdfsServer,CmpCommon::getDefaultString(LOB_HDFS_SERVER), strlen(CmpCommon::getDefaultString(LOB_HDFS_SERVER)));
```
zellerh: Doesn't that make the problem worse? If the string is larger than 256 bytes, it will overrun the lobHdfsServer buffer? Should the length passed in be MINOF(strlen(...), 256)?

Also, I use sizeof(lobHdfsServer) instead of 256 in these cases, but others have disagreed with me on that.

Also, shouldn't there be a global constant for the max. length of a server?

sureshsubbiah: The comment in the next file tells us that there is expectation the max valid string length will be 255 bytes for this attribute. However with user error there could be a longer invalid string. Maybe some error handling code that checks if the specifies string is less that 255 bytes will address the concern above.

I do have another question.
We are passing a pointer to a local variable (lobHdfsSever) to the CLI call SQL_EXEC_LobddlInterface(). Is there a guarantee that this call will NOT save the pointer somewhere that could cause the pointer to be accessed after the call has returned. I suppose I am not clear on changing the storage of this actual data from system heap used by NAString (old code) to stack used by newly created char array will avoid the corruption problem. I am often confused by such question, so consider this a question, rather than a issue to be resolved.

sandhyasun: Yes the length param , I will change to 256.  That shoudl be sufficient wouldn't it. It would copy the strlen of CmpCommon::getDeafulatString(...) length upto a max of 256.  To answer Suresh,The corruption is caused not by passing in, the  lobHdfsServer.  But to be safe, I will do a strcpy into a hap allocated string within the CLI function of the passed in param. The corruption was  caused by the code a few lines below : 
ComString newSchName = "\"";
      newSchName += catalogNamePart;
      newSchName.append("\".\"");
      newSchName.append(schemaNamePart);
      newSchName += "\"";
      NABoolean lobTrace = FALSE;
The ComString's small_  data member overwrote what loHdfsServer was pointing to. So what got passed into the LOB ddl function instead of "default"  was "TRAFODION.SEABASE"! . But I can make the cli call opy over the passed in param right away.

core/sql/optimizer/NATable.cpp
```diff
@@ -5485,9 +5485,13 @@ NABoolean createNAFileSets(TrafDesc * table_desc       /*IN*/,
 
    Lng32 postCreateNATableWarnings = CmpCommon::diags()->getNumber(DgSqlCode::WARNING_);
 
+
    if(postCreateNATableWarnings != preCreateNATableWarnings)
      tableConstructionHadWarnings_=TRUE;
-   const char *lobHdfsServer = CmpCommon::getDefaultString(LOB_HDFS_SERVER);
+   char  lobHdfsServer[256] ; // max length determined by dfs.namenode.fs-limits.max-component-length(255)
+   memset(lobHdfsServer,0,256);
+   strncpy(lobHdfsServer,CmpCommon::getDefaultString(LOB_HDFS_SERVER), sizeof(lobHdfsServer));
```
DaveBirdsall: If strlen(CmpCommon::getDefaultString(LOB_HDFS_SERVER)) is greater than or equal to sizeof(lobHdfsServer), then lobHdfsServer will not be null-terminated. So you might get some garbage characters on the end.

Consider nulling out the last byte. (You might want to make your buffer 256+1 bytes if the maximum length supported is 256.)

If strlen(CmpCommon::getDefaultString(LOB_HDFS_SERVER)) is greater than 256 bytes, this code will truncate it. Is that OK?

If not, consider using a dynamically sized array instead.

core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
```diff
@@ -4275,7 +4275,10 @@ short CmpSeabaseDDL::dropSeabaseTable2(
   short *lobNumList = new (STMTHEAP) short[numCols];
   short *lobTypList = new (STMTHEAP) short[numCols];
   char  **lobLocList = new (STMTHEAP) char*[numCols];
-  const char *lobHdfsServer = CmpCommon::getDefaultString(LOB_HDFS_SERVER);
+
+  char  lobHdfsServer[256] ; // max length determined by dfs.namenode.fs-limits.max-component-length(255)
+  memset(lobHdfsServer,0,256);
+  strncpy(lobHdfsServer,CmpCommon::getDefaultString(LOB_HDFS_SERVER),sizeof(lobHdfsServer));
```
DaveBirdsall: Same comment as above.

zellerh: Just a small, optional, addition: We could also get rid of the (char *) lobHdfsServer cast on line 4334 below, now that lobHdfsServer is a non-const. Same for line 5538 in file NATable.cpp.

core/sql/optimizer/NATable.cpp
sandhyasun: OK will add a null terminator or do a strncpy for 255 since we re allready memsetting the array. The length is supposed to be 255 max. We do want to limit it and not allow the CQD setting to be something aritrarily large. If it is we will get a hdfs connection error so it's ok. We do need better error reporting though. WIll address that separately.

core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
sandhyasun: Will do.

sandhyasun:
Will merge this change  if no objections .

sureshsubbiah:
+1

PR #978, opened 2017-02-23
[TRAFODION-2496] pyinstaller - better handle constants

mkby:
Put constants together into a new constants.py file. It will be easy to modify them to fit new Hadoop environment.

along with some minor fixes:
1. kill krb5service and delete status file in uninstall script
2. fix hbase_lib_path issue while install trafodion on apache hadoop
3. typo in db_config_default


sureshsubbiah:
+1

PR #979, opened 2017-02-24
TRAFODION-2498 Add support to run hive stmts from traf interface

anoopsharma00:
Syntax:
  process hive statement '<string>';

  <string>: hive statement starting with create/drop/alter/truncate.
      These are the only stmts currently supported.
  Ex:
   >>process hive statement 'create database trafsch';
   will create hive database 'trafsch'

   >>process hive statement 'create table trafsch.t (a int)';
   will create hive table 't' in hive schema 'trafsch'.

  'process hive statment ..' could be issued from any traf interface
   (sqlci/trafci/jdbc...)

### DIFF
--- a/core/sql/executor/ExExeUtilMisc.cpp
+++ b/core/sql/executor/ExExeUtilMisc.cpp
@@ -2509,3 +2509,128 @@ ExExeUtilHiveTruncatePrivateState::ExExeUtilHiveTruncatePrivateState()
 ExExeUtilHiveTruncatePrivateState::~ExExeUtilHiveTruncatePrivateState()
 {
 };
+
+////////////////////////////////////////////////////////////////
+// Constructor for class ExExeUtilHiveQueryTdb
+///////////////////////////////////////////////////////////////
+ex_tcb * ExExeUtilHiveQueryTdb::build(ex_globals * glob)
+{
+  ExExeUtilTcb * exe_util_tcb;
+
+  exe_util_tcb = new(glob->getSpace()) ExExeUtilHiveQueryTcb(*this, glob);
+  exe_util_tcb->registerSubtasks();
+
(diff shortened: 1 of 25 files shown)


anoopsharma00:
Somehow the syntax got chopped off.
Syntax is:
  process hive statement 'hive-string';
  hive-string:  hive statement starting with create/drop/alter/truncate

anoopsharma00:
For hive ddl operations, we need a hive_admin kind of traf privilege. 
If a user has that privilege, then they should be allowed to do hive DDL operations.
This would be relevant for explicit hive ddl, or for temp table ddl (as Hans mentioned)
or future support for volatile tables in hive.
Hive ddl privilege support need to be added.

For the first phase, only ddl operations are supported. We would extend for
some dml operations in a subsequent checkin.

core/sql/executor/ExExeUtilMisc.cpp
```diff
@@ -2509,3 +2509,128 @@ ExExeUtilHiveTruncatePrivateState::ExExeUtilHiveTruncatePrivateState()
...
+  while (1)
+    {
+      switch (step_)
+        {
+        case INITIAL_:
+          {
+            step_ = PROCESS_QUERY_;
+          }
+          break;
+
```
robertamarton: If authorization is enabled, need a check to make sure that the current user has the authority to manage hive tables.   Need to determine if the currently connected user has admin privileges in Hive.  If not, should not allow the operation.  If this is too difficult, then just prevent create, alter, and drops of Hive tables when authorization is enabled.
When Sentry is enabled, then return an error - Sentry enabled hive tables are only managed by Hive.  We could, in the future, call Sentry to see if the currently connected user has admin privileges and allow these operations.

PR #981, opened 2017-02-27
Reverting minor version back to 1

sandhyasun:


zellerh:
Actually, I have a question for @svarnau, who told me the answer a while ago: Somehow I thought there were more places where the version number is stored. Is this the only place where we need a change? Maybe no longer, since it looks like DCS also changed their pom.xml file to get the version from an environment variable?

svarnau:
I believe all of the other places are fixed to use TRAFODION_VER.  
I may need to re-visit the ambari plugin so that it is configured to handle 2.1 and 2.2, but that is a downstream task.

PR #982, opened 2017-02-28
[TRAFODION-2462] TRAFCI gui installer does not work

AlexPeng19:
fix jira 2462


DaveBirdsall:
If you put [TRAFODION-2462] into the title for your pull request, then the JIRA will automatically be updated. (As it is we would have to hand-update it.)

DaveBirdsall:
I took a look at the JIRA. It lists a number of issues. I'm guessing this change addresses only some of them. Could you comment on which issues you think are fixed with this change?

AlexPeng19:
@DaveBirdsall ok, i added comment under the jira, please check.

AlexPeng19:
sorry, i did not give much infomation in the git log. the pull are expected to fix two issues, first one, the trafci gui installer, second is : trafci multi line extra SQL> prompt issue. while input multi line, it will print as +>SQL>, it should be +>.

PR #984, opened 2017-03-01
[TRAFODION-2499] TMUDF does not pass error to client

zellerh:
Added check to avoid clearing out the error when source and
destination diags area are the same.


sureshsubbiah:
+1. Looks good.

zellerh:
jenkins, retest

PR #987, opened 2017-03-02
[TRAFODION-2517] Allow scalar UDFs with delimited identifiers

zellerh:
Added a parser rule for this case. Fortunately the new rule did not
increase the number of parser conflicts. Note that TMUDFs already
allowed delimited identifiers.


sureshsubbiah:
+1

PR #985, opened 2017-03-01
[TRAFODION-2514] Obscure cores seen in Trafodion while running jenkin…

selvaganesang:
…s tests with RH7

The global object gv_sb_thread_table was not constructed in a
thread safe manner.

There was buffer overrun while creating explain fragment involving hive tables.

PR #988, opened 2017-03-03
[TRAFODION-2514] Obscure cores seen in Trafodion while running jenkin…

selvaganesang:
…s tests with RH7

The global object gv_sb_thread_table was not constructed in a
thread safe manner.

There was buffer overrun while creating explain fragment involving hive tables.


selvaganesang:
This change is same as #985 . This PR is created to push into release2.1

sureshsubbiah:
+1

PR #994, opened 2017-03-07
[TRAFODION-2524]As debugged by Arvind: Increase size of the buffer that reads cmdline…

kevinxu021:
https://github.com/esgyn/traf-plus/commit/2724a86859f3ae5efac114d61f4be436ea876b47

### DIFF
--- a/core/sqf/src/seabed/src/ms.cpp
+++ b/core/sqf/src/seabed/src/ms.cpp
@@ -1391,6 +1391,7 @@ SB_THROWS_FATAL {
                         true, // args
                         &lv_cmdline,
                         lv_cmdline.size());
+#if 0
 #ifdef SQ_PHANDLE_VERIFIER
     sprintf(&lv_line, "msg_init - p-id=%d/%d" PFVY ", tid=%ld, pname=%s, cmdline=%s, %s\n",
 #else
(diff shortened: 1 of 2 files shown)


arvind-narain:
Looks good Kevin. Thanks for the change. 
This is needed for Centos 6.8+ systems where the cmdline in /proc/<process ID> can be > 4K.

Could you please also include the following change or equivalent? This will be needed once we get over the sqInit() failure: 

`--- a/core/sqf/src/seabed/src/ms.cpp
+++ b/core/sqf/src/seabed/src/ms.cpp
@@ -1391,6 +1391,7 @@ SB_THROWS_FATAL {
                         true, // args
                         &lv_cmdline,
                         lv_cmdline.size());
+#if 0
 #ifdef SQ_PHANDLE_VERIFIER
     sprintf(&lv_line, "msg_init - p-id=%d/%d" PFVY ", tid=%ld, pname=%s, cmdline=%s, %s\n",
 #else
@@ -1405,6 +1406,7 @@ SB_THROWS_FATAL {
             ga_ms_su_pname,
             &lv_cmdline,
             ms_seabed_vers());
+#endif`

DaveBirdsall:
The change looks good. There needs to be a JIRA for this though.

kevinxu021:
@arvind-narain Thanks! The changes has been in.
@DaveBirdsall Thanks! JIRA ID has been added.

DaveBirdsall:
+1

PR #996, opened 2017-03-07
Fix regression issue with privs1 and privs2

robertamarton:
Always call reg_users.sql before running privilege tests.
It does not hurt to run this scripts even if authorization already enabled


DaveBirdsall:
Looks like you picked up a lot of changes that you did not intend?

svarnau:
perhaps PR was intended for release2.1 instead of master?

robertamarton:
went to the wrong branch

PR #1005, opened 2017-03-13
[TRAFODION-2533] pyinstaller - check ssh pam settings during installation

mkby:


mkby:
let's check this in for now,  unless we can find a way to start trafodion without PAM.

svarnau:
works for me. On balance it is helpful, since I'm sure PAM is the prevalent configuration.

PR #1006, opened 2017-03-13
[TRAFODION-2523] Fixes drop schema cascade memory corruption.

prashanth-vasudev:


DaveBirdsall:
jenkins, retest

core/sql/sqlcomp/CmpSeabaseDDLschema.cpp
```diff
@@ -1006,7 +1006,7 @@ void CmpSeabaseDDL::alterSeabaseSchema(StmtDDLAlterSchema * alterSchemaNode)
 
    bool isVolatile = (memcmp(schName.data(),"VOLATILE_SCHEMA",strlen("VOLATILE_SCHEMA")) == 0);
    int32_t length = 0;
-   int32_t rowCount = 0;
+   Int64 rowCount = 0;
```
DaveBirdsall: It doesn't look like this particular "rowCount" variable is used anywhere. I suppose we could delete it. The code change is harmless though.

prashanth-vasudev: This rowCount is used at line 889. It is a lengthy method , hence not visible unless file is expanded. Thanks.

DaveBirdsall: Indeed! I missed that. Thanks.

core/sql/cli/Cli.cpp
```diff
@@ -9717,7 +9717,7 @@ Lng32 SQLCLI_LOBcliInterface
 	// set parserflags to allow ghost table
 	currContext.setSqlParserFlags(0x1);
 	
-	Int64 numChunks = 0;
+	Lng32 numChunks = 0;
```
sandhyasun: This can cause corruption. The len coming back will be 8. So there is a potential for corruption. The rowcount which is the container needed to hold the return value needs to be large enough to hold the return value always.

core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
```diff
@@ -7791,7 +7791,7 @@ void CmpSeabaseDDL::alterSeabaseTableAddPKeyConstraint(
   // this new primary key.
   // Do this optimization in mode_special_4 only.
   Lng32 len = 0;
-  Int64 rowCount = 0;
+  Lng32 rowCount = 0;
```
sandhyasun: Same comment as above. The container is not large enough to hold the return value from executeImmediate() . And since it's on stack it can corrupt another value on the stack.

DaveBirdsall:
Apologies, @sandhyasun, I committed this before seeing your comments.

It sounds like @prashanth-vasudev fixes are correct, but there are more places where similar fixes are needed?

sandhyasun:
Actually his first commit in this PR was correct . The second commit  undoes a couple of changes that were initially correct. Even though they don't exhibit corruption now, in the debugger it is seen that the 4 byte container in these places is not sufficient and have the potential to cause corruption as was seen in the "drop schema" case.

core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
sandhyasun: Checked again and even though the variable is named rowcount, we are retrieving the value "[ANY 1]" into that. So this does return a 4 byte interger and we're ok.  Only the (count*) return values are 8 bytes long.

PR #1009, opened 2017-03-14
Various fixes, details below

anoopsharma00:
-- max length limited to 16777216 bytes (16M)
   for char cols and functions (repeat, concat).
  (optimizer/SynthType.cpp, common/ComSmallDefs.h, sqlcomp/nadefaults.cpp)

-- previous max length change requires adding of a new hbase property
   called hbase.client.keyvalue.maxsize so large key/values could be
   handled in hbase cell.

   Following scripts have been updated to handle that.
   (sqf/sql/scripts/install_local_hadoop,
    install/ambari-installer/traf-mpack/common-services/TRAFODION/2.1/
         service_advisor.py
    install/installer/traf_hortonworks_mods
    install/python-installer/configs/mod_cfgs.json)

   Developers can also update hbase-site.xml with this property if
   they dont want to reinstall local hadoop.
      <property>
        <name>hbase.client.keyvalue.maxsize</name>
        <value>0</value>
      </property>

-- while accessing a hive table as an external table, the hive table and
   corresponding external table definitions are validated to be the
   same. This validates that corresponding columns have the same
   data attributes (type, length, scale, etc).
   This check causes failures if hive column is of 'string' datatype.
   That is because hive 'string' column length can be changed by
   a cqd but the corresponding external table has predefined length
   set when the table is created.
   The validation check now ignores the length attribute if the hive column is
   of 'string' datatype.
   (optimizer/BindRelExpr.cpp, common/CharType.*, NAType.*)

-- data moved into direct buffer would sometimes cause overflow and crash.
   The max direct buffer length used to send/retrieve hbase data
   is now limited to 1G  (executor/ExHbaseAccess.cpp)

-- errors during vsbb upsert are now handled correctly
  (executor/ExHbaseIUD.cpp)

-- support for GET CATALOGS command
   (generator/GenRelExeUtil.cpp, executor/ExExeUtilGet.cpp)

-- An incorrect computation would sometimes cause group by rollup to
   crash compiler in NAHeap::unlinkLargeFragment()
  (generator/GenRelGrby.cpp)

-- sort of a large row that exceeded sort pre-set buffer size would crash
   if the row size exceeded the max buffer size.
   That has been fixed by allocating space for atleast one row.
  (generator/GenRelMisc.cpp)

-- purgedata on a delimited name was failing. That has been fixed.
   (optimizer/RelExeUtil.h)

-- regress/tools/runregr_privs1/privs2 fixed to handle running of
   subset of tests

-- regress/seabase/TEST031 updated with new tests

### DIFF
--- a/core/sql/bin/SqlciErrors.txt
+++ b/core/sql/bin/SqlciErrors.txt
@@ -1198,7 +1198,7 @@ $1~String1 --------------------------------
 4126 42000 99999 BEGINNER MAJOR DBADMIN The row value constructors in a VALUES clause must be of equal degree.
 4127 42000 99999 BEGINNER MAJOR DBADMIN Type $0~String0 cannot be assigned to type $1~String1.
 4128 42000 99999 BEGINNER MAJOR DBADMIN Default volume and subvolume information could not be retrieved from =_DEFAULTS define - DEFINEINFO error $0~int0.
-4129 ZZZZZ 99999 BEGINNER MAJOR DBADMIN The character-typed result of the function $0~String0 is longer than 32768 octets.
+4129 ZZZZZ 99999 BEGINNER MAJOR DBADMIN The character-typed result of the function $0~String0 is longer than max supported size.
 4130 ZZZZZ 99999 BEGINNER MAJOR DBADMIN $0~string0 is a read-only DEFAULTS attribute and cannot be updated.
 4131 42000 99999 BEGINNER MAJOR DBADMIN Current_time, current_date, or current_timestamp is not allowed in a check constraint.
 4132 ZZZZZ 99999 BEGINNER MAJOR DBADMIN Current_user, session_user, or system_user is not allowed in a check constraint.
--- a/core/sql/common/CharType.cpp
+++ b/core/sql/common/CharType.cpp
@@ -752,16 +752,21 @@ NABoolean CharType::isCharSetAndCollationComboOK() const
 // -----------------------------------------------------------------------
 // Equality comparison
 // -----------------------------------------------------------------------
+NABoolean CharType::compareCharAttrs(const NAType& other) const
+{
+  return caseinsensitive_ == ((CharType&)other).caseinsensitive_ &&
+    nullTerminated_ == ((CharType&)other).nullTerminated_ &&
+    upshifted_	 == ((CharType&)other).upshifted_ &&
+    charSet_	 == ((CharType&)other).charSet_ &&
+    collation_	 == ((CharType&)other).collation_ &&
+    coercibility_	 == ((CharType&)other).coercibility_ &&
+    encodingCharSet_ == ((CharType&)other).encodingCharSet_ ;
+}
+
 NABoolean CharType::operator==(const NAType& other) const
 {
   return NAType::operator==(other) &&
-	 caseinsensitive_ == ((CharType&)other).caseinsensitive_ &&
-	 nullTerminated_ == ((CharType&)other).nullTerminated_ &&
-	 upshifted_	 == ((CharType&)other).upshifted_ &&
-	 charSet_	 == ((CharType&)other).charSet_ &&
-	 collation_	 == ((CharType&)other).collation_ &&
-	 coercibility_	 == ((CharType&)other).coercibility_ &&
-	 encodingCharSet_ == ((CharType&)other).encodingCharSet_ ;
+    compareCharAttrs(other);
 }
 
 // -----------------------------------------------------------------------
@@ -965,6 +970,28 @@ SQLVarChar::SQLVarChar(const CharLenInfo & maxLenInfo,
       clientDataType_(collHeap()),  // Get heap from NABasicObject. Can't allocate on stack.
       wasHiveString_(FALSE)
 {}
+
+// -----------------------------------------------------------------------
+// Equality comparison
+// -----------------------------------------------------------------------
+NABoolean SQLVarChar::operator==(const NAType& other) const
+{
+  NABoolean compResult = FALSE; // false, failed. true, passed.
+
+  // if this or other was originally a hive string type, then
+  // do not compare lengths.
+  if ((wasHiveString()) ||
+      ((other.getTypeName() == LiteralVARCHAR) &&
+       (((SQLVarChar&)other).wasHiveString())))
+    compResult = equalIgnoreLength(other);
+  else
+    compResult = NAType::operator==(other);
+
+  if (NOT compResult)
+    return FALSE;
+
+  return compareCharAttrs(other);
+}
 
 // -----------------------------------------------------------------------
 // The SQL/C Preprocessor rewrites a VARCHAR declaration (Ansi 19.4)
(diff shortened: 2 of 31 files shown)


DaveBirdsall:
@anoopsharma00, Is there a JIRA for these changes?

core/sql/generator/GenRelGrby.cpp
```diff
@@ -105,9 +105,13 @@ short GroupByAgg::genAggrGrbyExpr(Generator * generator,
...
   // find the number of aggregate and groupby entries
-  ULng32 numAttrs
-    = ((NOT aggregateExpr.isEmpty() ) ? aggregateExpr.entries() : 0)
-    + ((NOT groupExpr.isEmpty() ) ? groupExpr.entries() : 0);
+  ULng32 numAttrs = 0;
+  if (NOT aggregateExpr.isEmpty())
+    numAttrs += aggregateExpr.entries();
+  if (isRollup() && (NOT rollupGroupExprList.isEmpty()))
+    numAttrs += rollupGroupExprList.entries();
+  else if (NOT groupExpr.isEmpty())
```
DaveBirdsall: I take it that we cannot stick a GROUP BY clause after a ROLLUP?

core/sql/generator/GenRelMisc.cpp
```diff
@@ -3045,8 +3045,11 @@ short Sort::generateTdb(Generator * generator,
   
   CostScalar bufferSize = getDefault(GEN_SORT_MAX_BUFFER_SIZE);
 
-  ULng32 bufferSize_as_ulong = 
-    (ULng32)(MINOF(CostScalar(UINT_MAX), bufferSize)).getValue(); 
+  UInt32 bufferSize_as_ulong =
```
DaveBirdsall: This is fine, but maybe we should change the variable name to "bufferSize_as_uint"? Or "...uint32"?

core/sql/sqlcomp/nadefaults.cpp
```diff
@@ -5936,8 +5937,39 @@ enum DefaultConstants NADefaults::validateAndInsert(const char *attrName,
      }
      break;
 
-      default:  break;
-      }
+     // max char col length is 10M (10485760).
```
DaveBirdsall: Below we say 16M. Why do we say 10M here?

core/sql/regress/seabase/EXPECTED031
```diff
@@ -888,6 +960,143 @@ CREATE INDEX T031T1I1 ON TRAFODION.SCH.T031T1
...
+A1          B2                    4                    15
+A1          B2                    ?                    26
+A1          B3                    5                    19
+A1          B3                    6                    23
+A1          B3                    ?                    42
+A1          B4                    7                    27
+A1          B4                    8                    31
+A1          B4                    ?                    58
+A1          ?                     ?                   136
+A1          ?                     ?                   136
```
zellerh: Shouldn't this row have 3 question marks? Maybe a VEGie problem?

sureshsubbiah:
The purgedata issue is tracked by this JIRA. 
https://issues.apache.org/jira/browse/TRAFODION-1702
It is now assigned to Anoop.

core/sql/bin/SqlciErrors.txt
```diff
@@ -1198,7 +1198,7 @@ $1~String1 --------------------------------
 4126 42000 99999 BEGINNER MAJOR DBADMIN The row value constructors in a VALUES clause must be of equal degree.
 4127 42000 99999 BEGINNER MAJOR DBADMIN Type $0~String0 cannot be assigned to type $1~String1.
 4128 42000 99999 BEGINNER MAJOR DBADMIN Default volume and subvolume information could not be retrieved from =_DEFAULTS define - DEFINEINFO error $0~int0.
-4129 ZZZZZ 99999 BEGINNER MAJOR DBADMIN The character-typed result of the function $0~String0 is longer than 32768 octets.
+4129 ZZZZZ 99999 BEGINNER MAJOR DBADMIN The character-typed result of the function $0~String0 is longer than max supported size.
```
anoopsharma00: will fix this for a later checkin.

core/sql/common/CharType.cpp
```diff
@@ -752,16 +752,21 @@ NABoolean CharType::isCharSetAndCollationComboOK() const
 // -----------------------------------------------------------------------
 // Equality comparison
 // -----------------------------------------------------------------------
+NABoolean CharType::compareCharAttrs(const NAType& other) const
+{
```
anoopsharma00: By the time this method is called, it has already gone through the test of
validating that 'this' and 'other' are part of same class.

core/sql/common/CharType.cpp
```diff
@@ -967,6 +972,28 @@ SQLVarChar::SQLVarChar(const CharLenInfo & maxLenInfo,
...
+// -----------------------------------------------------------------------
+NABoolean SQLVarChar::operator==(const NAType& other) const
+{
+  NABoolean compResult = FALSE; // false, failed. true, passed.
+
+  // if this or other was originally a hive string type, then
+  // do not compare lengths.
+  if ((wasHiveString()) ||
+      ((other.getTypeName() == LiteralVARCHAR) &&
+       (((SQLVarChar&)other).wasHiveString())))
```
anoopsharma00: literalVARCHAR is used for all sql varchar datatypes.
The bytesPerChar and character set parameters to the char
classes defined in common/CharType.h determine what
kind of char datatype it is.

core/sql/executor/ExExeUtilGet.cpp
```diff
@@ -266,6 +266,12 @@ static const QueryString getComponentPrivileges[] =
   {" ; "}
 };
 
+static const QueryString getCatalogsQuery[] =
+{
+  {" select * from (values ('TRAFODION'), ('HIVE')) "},
```
anoopsharma00: Currently, trafodion and hive are true catalogs that can be traversed through DBMgr
or GET statements. Details about their contents (schemas, tables, etc) can
be returned.
'hbase' is not really a catalog, it is the first part of a 3-part name used
to access native hbase tables. We can think about it if it makes sense to treat it
as a catalog.

core/sql/generator/GenRelGrby.cpp
anoopsharma00: that is correct. Only one of 'group by' or 'group by rollup' can be used.

core/sql/generator/GenRelMisc.cpp
anoopsharma00: Will change the variable name for a later checkin.

core/sql/sqlcomp/nadefaults.cpp
anoopsharma00: this is an obsolete comment. We used to support upto 10M but now increased it to 16M.
will fix it for a later checkin.

core/sql/regress/seabase/EXPECTED031
anoopsharma00: You are right, this is an existing bug. 
For a statement of the form:
  select a, sum(a) from t where a = 1 group by rollup (a)
the VEG transformation replaces the 'a' in select list with '1' and
that causes that column to not return a null('?') as part of the
final rollup group. We probably need to have VEG transformation
use the column 'a' instead of '1'.
Will file a separate jira for this case.

PR #1010, opened 2017-03-14
TRAFODION-2538 Revoking privileges from role not invoking query inval…

robertamarton:
…idation

Fixed a issue where query invalidation keys were not being sent correctly when
a privilege was revoked from a role.

When a table is cached, a list of all the query invalidation keys for the user
are stored.  Later, when a query is run, the compiler picks the relevant keys
and places them in the plan.  When a revoke occurs, a key is sent to RMS and
the executor processes check for keys at the next execution. If the key affects
any caches, the cache entries are refreshed and plans recompiled.

Incorrect keys were being created when privileges were revoked from roles, so
queries continued to work even though the user had no more privileges.

### DIFF
--- a/core/sql/regress/privs1/EXPECTED120
+++ b/core/sql/regress/privs1/EXPECTED120
@@ -15,7 +15,7 @@
 >>--
 >>--    games     - multiple roles giving same privileges
 >>--    teams     - multiple privileges through different roles
->>--    players   - control, not roles involved in privileges
+>>--    players   - control, no roles involved in privileges
 >>--    standings - used to test sequence privileges and revoke role
 >>--    stats     - tests revoke PUBLIC authorization ID
 >>-- =================================================================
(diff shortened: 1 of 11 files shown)


sureshsubbiah:
+1

core/sql/regress/privs1/EXPECTED120
```diff
@@ -746,6 +748,109 @@ TEAM_NUMBER  TEAM_NAME
...
+          4  Indians             
+          5  Tigers              
+
+--- 5 row(s) selected.
+>>-- now works and query recompiled (8597)
+>>execute insert_teams;
+
+*** WARNING[8597] Statement was automatically retried 1 time(s). Delay before each retry was 0 seconds. See next entry for the error that caused this retry.
+
+*** WARNING[8583] This statement contains no generated plan to execute at runtime. An error during query compilation caused this condition.
```
DaveBirdsall: A strange warning. Seems something like "plan was recompiled due to authorization setting changes" might be more appropriate.

robertamarton:
Error 8583 is a generic error returned when aqr is invoked and a previous compilation failed.  Need to investigate further to see if a different (or better) warning can be returned.

PR #1015, opened 2017-03-16
[TRAFODION-2526] pyinstaller - don't use sudo command in root user during installation

mkby:


sandhyasun:
Does this affect the installation procedure for 2.1 and is this fix needed for 2.1 release  ? Or can it go into the master branch for the next release ?  we are doing an RC2 but don't want to destablize what we have.

mkby:
will create a new PR for this to master branch.

PR #1017, opened 2017-03-17
[TRAFODION-2541]"UPDATE LOB" default max lob size error

xwq:
the default lob max size for "UPDATE LOB" operator should be 10000MB
rather than 10000B


xwq:
Hi, I am implementing the LOB support for JDBC. I found the limit of lob size for "update lob" support is not correct computed. So I created this Pull request.

- Weiqing

xwq:
jenkins, retest.

PR #1019, opened 2017-03-20
[TRAFODION-2522] Removing carriage returns from the ends of lines.

zellerh:
This file was checked in in MS-DOS format with CRLF as line separators. Removing the extra CRs.


zellerh:
I compared the files before and after the dos2unix command and they compared the same in BeyondCompare. To me, that indicates that BeyondCompare ignores the CRs and that nothing else in the file changed.

DaveBirdsall:
+1 Looks good to me.

PR #1022, opened 2017-03-23
[TRAFODION-2516]Support for insert-select from varchar/char colummns into lob columns…

sandhyasun:
…. Support  for insert with parameter which gets typed by default to a string type so caller can pass in a string as the value for the param. Added tests in the regression tests for insert from hive varchar column, traafodion fixed char and trafodion varchar .


core/sql/optimizer/BindItemExpr.cpp
```diff
@@ -6173,7 +6173,35 @@ ItemExpr *Assign::bindNode(BindWA *bindWA)
...
+    {
+      ValueId vid1 = child(1)->castToItemExpr()->getValueId();  
+      // Add a stringToLob node
+      ItemExpr *newChild =  new (bindWA->wHeap()) LOBinsert( vid1.getItemExpr(), NULL, LOBoper::STRING_, FALSE);    
+      newChild->bindNode(bindWA);
+      if (bindWA->errStatus())
+	return boundExpr; 
+      setChild(1, newChild);
+    }
+  /* if ((sourceType == NA_UNKNOWN_TYPE) && (targetType = NA_LOB_TYPE))
```
DaveBirdsall: I think there was a CQD added awhile back that caused params to be treated as strings. Perhaps this commented code was intended to handle the case where the CQD is set to 'OFF'?

sandhyasun: No I will remove the commented out code. I had initilaly tried to do this cast in the binder but it was not the right place. The type coercion needed to be done in the ::doSynthType() method.

PR #1023, opened 2017-03-23
[TRAFODION-2552] Use column stats when SUBSTR takes a column prefix

DaveBirdsall:
Scenario: A join on two tables, on a column that is badly skewed in one of the tables. The join predicate is of the form SUBSTR(skewedcolumn,1,n) = SUBSTR(othercolumn,1,n). We use the default distribution for this predicate, so we miss an opportunity to use skew buster.

The fix simply allows the base table column statistics to be used in this case.

### DIFF
--- a/core/sql/optimizer/ItemExpr.cpp
+++ b/core/sql/optimizer/ItemExpr.cpp
@@ -1062,7 +1062,7 @@ NABoolean ItemExpr::useStatsForPred()
         {
           NABoolean negate = FALSE;
           ConstValue * c = startPosition->castToConstValue(negate);
-          if (c->canGetApproximateNumericValue() &&
+          if (c->canGetExactNumericValue() &&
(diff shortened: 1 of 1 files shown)


DaveBirdsall:
Please do not merge yet (but feel free to review of course). I will be running full regressions against this change over the weekend.

DaveBirdsall:
I ran full regressions last night on this change on a workstation, and they pass. So, barring negative review comments, this change is ready for merging.

DaveBirdsall:
I noticed today that I had mistakenly coded c->canGetApproximateNumericValue() when I should have coded c->canGetExactNumericValue(). This was a cut-and-paste error. The rework corrects it. The problem turns out to be an unobservable bug though; the binder guarantees that the second argument of SUBSTR is in fact an exact numeric.

PR #1024, opened 2017-03-24
[TRAFODION-2551]Exception is thrown, when using TrafT4DataSource in jdbcT4

kevinxu021:
decrease counter while connection error

### DIFF
--- a/core/conn/jdbcT4/src/main/java/org/trafodion/jdbc/t4/TrafT4PooledConnectionManager.java
+++ b/core/conn/jdbcT4/src/main/java/org/trafodion/jdbc/t4/TrafT4PooledConnectionManager.java
@@ -131,18 +131,27 @@ public class TrafT4PooledConnectionManager implements javax.sql.ConnectionEventL
 							pc.close();
 						} catch (Exception e) {
 						} // cleanup, ignore any errors
+						if (removeInUseConnection(pc, false)) {
+							--count_;
+						}
 					}
 				} else {
 					throw TrafT4Messages.createSQLException(null, null, "max_pool_size_reached", null);
 				}
 			} else {
 				pc = (PooledConnection) free_.get(0);
-				if (removeFreeConnection(pc, true)) {
-					TrafT4Connection c = (TrafT4Connection) pc.getConnection();
-					try {
-						c.ic_.enforceT4ConnectionTimeout(c);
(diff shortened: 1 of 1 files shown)


core/conn/jdbcT4/src/main/java/org/trafodion/jdbc/t4/TrafT4PooledConnectionManager.java
```diff
@@ -137,12 +137,18 @@ public Connection getConnection() throws SQLException {
...
+					c.ic_.enforceT4ConnectionTimeout(c);
+					if (removeFreeConnection(pc, true)) {
 						validConnection = true;
-					} catch (SQLException sqlEx) {
+					}
+				} catch (Exception e) {
+					//ignore errors
+				} finally {
+					if (!validConnection) {
+						if (free_.remove(0) != null)
```
selvaganesang: removeFreeConnection returns false when the pooledConnection was not present in the free_. If so, it is not clear why are removing the 0 entry, If line 142 returns an exception, you might want to remove the pooled connection from free_. But,  still you can't remove the 0th entry.  It looks like this code needs to be reworked

kevinxu021: @selvaganesang Thanks for you comments. i don't quit understand. It gets 0th entry first, then remove it in removeFreeConnection( here it still in 0th). If it cannot be removed, that means the connection is not in free_ which is not reasonable( unless memory issue or something else, rarely happens but doesn't matter to remove again and throw exception, then the program exit). It should be fine to remove 0th here. 
In standard doc for free_.remove: true if this list contained the specified element. it might not be represent PRESENT.

selvaganesang: What gets 0th entry first?

kevinxu021: Thanks Selva. it was refected in https://github.com/apache/incubator-trafodion/pull/1035. Please review again.

kevinxu021: Thanks Selva. it was refected in #1035. Please review again. current branch has something wrong.

PR #1028, opened 2017-03-28
[TRAFODION-2449] dtmci - Adding 'status regions' command

bucaojit:
Prints the full region name, hostname:port, and startkey/endkeys in hex format.

### DIFF
--- a/core/sqf/src/seatrans/tm/hbasetmlib2/hbasetm.cpp
+++ b/core/sqf/src/seatrans/tm/hbasetmlib2/hbasetm.cpp
@@ -1196,6 +1196,8 @@ HMN_RetCode HashMapArray::init()
       JavaMethods_[JM_GET_HOSTNAME   ].jm_signature  = "(I)Ljava/lang/String;";
       JavaMethods_[JM_GET_PORT       ].jm_name       = "getPort";
       JavaMethods_[JM_GET_PORT       ].jm_signature  = "(I)Ljava/lang/String;";
+      JavaMethods_[JM_GET_REGINFO    ].jm_name       = "getRegionInfo";
+      JavaMethods_[JM_GET_REGINFO    ].jm_signature  = "(J)Ljava/lang/String;";
 
       return (HMN_RetCode)JavaObjectInterfaceTM::init(className, javaClass_, JavaMethods_, (int32)JM_LAST, false);
     }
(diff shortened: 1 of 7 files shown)


bucaojit:
jenkins, retest

core/sqf/src/seatrans/tm/hbasetmlib2/hbasetm.cpp
```diff
@@ -1218,6 +1220,22 @@ char* HashMapArray::get(int tid)
...
+   jstring js_val = (jstring)(_tlp_jenv->CallObjectMethod(javaObj_, JavaMethods_[JM_GET_REGINFO].methodID, tid));
+   if (getExceptionDetails(NULL)) {
+      tm_log_write(DTM_TM_JNI_ERROR, SQ_LOG_ERR, (char *)"HashMapArray::getRegionInfo()", (char *)_tlp_error_msg->c_str(), -1LL);
+      return NULL;
+   }
+   if(js_val == NULL){
+       return NULL;
+   }
+
+   const jbyte* jb_val = (jbyte*)_tlp_jenv->GetStringUTFChars(js_val, NULL);
```
DaveBirdsall: Do we need to worry about memory leaks from C++ references to Java variables? I don't see any code in this module that does worry about this, but the executor code (see for example $TRAF_HOME/../sql/executor/HBaseClient_JNI.cpp) is very careful to push and pop frames to avoid memory leaks. Perhaps the TM is handling this issue in a different way?

core/sqf/src/seatrans/tm/hbasetmlib2/src/main/java/org/trafodion/dtm/HBaseTxClient.java
```diff
@@ -1264,64 +1265,40 @@ public HashMapArray callRequestRegionInfo() throws IOException {
...
-          port = String.valueOf(it7.next().getPort());
-          while(it7.hasNext()){
-              port = port + ";" + String.valueOf(it7.next().getPort());
-          }
-          hm.addElement(tnum, "Port", port);
+          
+          while(it.hasNext()) {
+              TransactionRegionLocation trl = it.next();
+              tablename = trl.getRegionInfo().getTable().getNameAsString();
+              if(tablename.contains("_MD_"))
```
DaveBirdsall: This test might be "good enough" but it is not sharp if your intent is to exclude just metadata tables. I might, for example, create a table X_MD_YZ. Even "._MD." isn't good enough because I can use a delimited identifier and create a table "X._MD_.YZ". I'm curious why we want to exclude metadata tables at all?

core/sqf/src/seatrans/tm/hbasetmlib2/hbasetm.cpp
bucaojit: Hi Dave, thanks for the review.  You make excellent points.  I did find that I was not deleting a reference of the map on the C++ side in tm.cpp.  I will add that change.  As for the frames, I don't see that on the JNI for the TM side.  From what I understand once the reference is gone it is up to the Java gc to manage the memory cleanup.

core/sqf/src/seatrans/tm/hbasetmlib2/src/main/java/org/trafodion/dtm/HBaseTxClient.java
bucaojit: Good point here.  I will change the check to "TRAFODION._DTM_."  

I decided to not include the metadata and include the explicit tables.  When testing with the metadata tables included it would print about 9 metadata regions and I was more interested in seeing the table I was inserting or deleting from.  I started working on a 'status regions detail' that will add the metadata as well but I was hitting a character limit with so many regions when testing.  Message limit for the TM message CTmTxMessage looks to be around 4k.

bucaojit:
Hi Selva, Appreciate the review.  I made the change to the parameter name to 'tid'.  
Sure, I can look into converting it to a byte array in the next version.  You're right about it not being called often, the HashMapArray code path is only run when the 'status regions' command is used through dtmci.  

I've committed another patch with the suggestions added by Dave and Selva.

PR #1033, opened 2017-03-30
TRAFODION-2534 CLASSPATH for Java UDRs

zellerh:
Two changes in the custom Java class loader used for UDRs.

First, the class loader now adds any jar files stored in
$TRAF_HOME/udr/public/external_libs to its CLASSPATH.
The choice of this directory is anticipating work on
isolated UDFs, see TRAFODION-2561.

Second, the custom class loader LmClassLoader now uses
java.net.URLClassLoader to do most of the work. This allows
us to delete a lot of code in this file. When this class
was first written, many years ago, the URLClassLoader was
probably not yet available.

Finally, with the new class loader, the JDBC predefined
UDF can also be simplified.

### DIFF
--- a/core/sql/langman/LmLangManagerJava.cpp
+++ b/core/sql/langman/LmLangManagerJava.cpp
@@ -114,8 +114,6 @@ if (result == LM_PARAM_OVERFLOW)			\
 // Class LmLanguageManagerJava
 //
 //////////////////////////////////////////////////////////////////////
-char* LmLanguageManagerJava::classPath_ = NULL;
-char* LmLanguageManagerJava::sysClassPath_ = NULL;
 ComUInt32 LmLanguageManagerJava::maxLMJava_ = 0;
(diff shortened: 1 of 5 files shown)


zellerh:
Another comment that might help understanding this code: The old LmClassLoader used the "delegation model" of class loaders, meaning that it was first calling its parent class loader to load a class. That meant that it would always look at the entire Trafodion CLASSPATH first and that the methods to add or remove the CLASSPATH from its own list did not have any effect. The new LmClassLoader follows the delegation model as well, but it no longer adds the CLASSPATH to its own logic, since this is unnecessary. The new logic validates once, on UDR creation, that the referenced class is actually in the container. However, if the same class also exists in the Trafodion CLASSPATH, we will pick up the class from there when we execute the UDR (as we did with the old code). This should not pose a security risk, as the UDR is trusted and can call anything in the Trafodion CLASSPATH from within the UDR.

PR #1034, opened 2017-03-30
[TRAFODION-2544] JDBCT4 group id needs to be changed to org.apache.trafodion in JAR 

AlexPeng19:
[TRAFODION-2544] JDBCT4 group id needs to be changed to org.apache.trafodion in JAR

### DIFF
--- a/core/conn/jdbcT4/.gitignore
+++ b/core/conn/jdbcT4/.gitignore
@@ -3,3 +3,5 @@
 /LICENSE
 /NOTICE
 src/main/java/org/trafodion/jdbc/t4/Vproc.java
+src/main/resources/META-INF/NOTICE
+src/main/resources/META-INF/LICENSE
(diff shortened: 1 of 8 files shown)


AlexPeng19:
jenkins, retest

selvaganesang:
Is this a requirement from Apache? Changing the package name would mean the existing application needs to be updated to load the new driver isn't?

selvaganesang:
I think there are 2 other issues with this change.  I think the directory of the java programs needs to be in sync with package name for use in eclipse and other IDEs.  @zellerh can confirm this.  In addition, I didn't see the corresponding change in trafci. I wonder if trafci is broken with this change

selvaganesang:
I understood from @arvind-narain that this change is not changing the package name of T4 driver and it just changes the group id in maven.  I misunderstood the change earlier. If this is what is expected in the PR, then the change looks good.

DaveBirdsall:
@selvaganesang, so is this ready to be merged?

DaveBirdsall:
Pinging again, @selvaganesang, is this ready to be merged?

selvaganesang:
Yes. I think so.

DaveBirdsall:
Great! I'll merge it.

PR #1036, opened 2017-03-30
[TRAFODION-2554]Convert into Maven project

kevinxu021:
Change releated dependency


core/conn/jdbc_type2/Makefile
```diff
@@ -57,7 +57,15 @@ T2_OBJS  = $(OUTDIR)/CommonDiags.o \
...
-
+CSES := org.trafodion.jdbc.t2.JdbcDebug \
+		org.trafodion.jdbc.t2.DataWrapper \
+		org.trafodion.jdbc.t2.SQLMXCallableStatement \
+		org.trafodion.jdbc.t2.SQLMXConnection \
+		org.trafodion.jdbc.t2.SQLMXDatabaseMetaData \
+		org.trafodion.jdbc.t2.T2Driver \
+		org.trafodion.jdbc.t2.SQLMXPreparedStatement \
+		org.trafodion.jdbc.t2.SQLMXResultSet \
+		org.trafodion.jdbc.t2.SQLMXStatement
```
arvind-narain: Would this change be needed if we touch the package name in T2 source to include apache ? Should we change the package name (org.apache.trafodion...) in the source and the file layout  to make it consistent ?

kevinxu021:
@arvind-narain @selvaganesang Please help to review it. Thanks!

selvaganesang:
I am not clear on the need to change the package name. If so, I think the directory of the java programs needs to be in sync with package name for use in eclipse and other IDEs. @zellerh can confirm this

zellerh:
Sorry, can't confirm for sure, as I have not tried that, but here is what I believe (not know).
My hope was that we could just change the group id in Maven to org.apache.trafodion.x and leave the package name unchanged, org.trafodion.x. My assumption is that Maven does not insist that the group id matches all the classes inside the artifact.

It looks like you already went ahead and changed the package names. In that case I agree with Selva, we should then also change the directories. I'm surprised that Maven would build with a mismatch between directory names and package names (or is this built with a Makefile?). If you do this move, make sure to do it with git mv, so that git knows the files have been moved and aren't new. Otherwise you may cause a lot of merge conflicts or lose changes.

kevinxu021:
@zellerh the directory has been changed. Yes, if the directory does not change, maven can also work fine. To keep consistent, it has been changed. For git mv, yes, as you see, already applied.

selvaganesang:
+1 if regress/udr test passes

selvaganesang:
The impact of changing the package name in Type2 driver is confined to the server applications running on the Trafodion cluster.  However, please consider sending a notice to dev and user list about this upcoming change

kevinxu021:
The notice has been sent out a day ago. Please let me know if any more suggestion.

core/conn/jdbc_type2/src/main/java/org/apache/trafodion/jdbc/t2/SQLMXPooledConnectionManager.java
```diff
@@ -23,14 +23,14 @@
...
-package org.trafodion.jdbc.t2;
+package org.apache.trafodion.jdbc.t2;
 
 import java.sql.*;
 import javax.sql.*;
 import java.util.Properties;
 import java.util.LinkedList;
 import java.io.PrintWriter;
-import org.trafodion.jdbc.t2.*;
+import org.apache.trafodion.jdbc.t2.*;
```
zellerh: Just a minor question: Do we really need to import the content of our own package?

DaveBirdsall:
@zellerh, @selvaganesang: Is this change ready to merge?

zellerh:
It looks good to me, but I have only limited knowledge about JDBC.

PR #1037, opened 2017-03-30
[TRAFODION-2565] JDBC T2 Vproc needs updating . Causes build or initi…

arvind-narain:
…alize issues for 2.1 release

Problem seen when a build is attempted in a non-git environment.

Not seen in 2.0 since revision was being set to empty in non-git environment (svn). 
https://github.com/apache/incubator-trafodion/commit/e93d518


sandhyasun:
+1

PR #1041, opened 2017-04-02
Rework the fix for [TRAFODION-2516] to returntype for param as LOB in…

sandhyasun:
…stead of coercing the type by default to char type since it confuses JDBC and Catalog APIs. The expected input param still is expected as a varchar format. Also added  testcases both for the regression as well as the clitestdriver program. Added support for update with param as well. Fixed a problem with GC limit where GC was happening too often when the "update lob" interface was called becuase the limit was not taken as MB and was taken as bytes instead.


sandhyasun:
Do not merge yet. Needs JDBC testing first.

core/sql/bin/clitest.cpp
```diff
@@ -316,7 +317,9 @@ case 7:
 	  return retcode;
           }
         cout << "LOB handle for "<< columnname << ": " << lobHandle << endl;
-	retcode = updateAppendBufferToLobHandle(cliGlob,lobHandle);
+        char *source= new char[300];
```
DaveBirdsall: This is a test program so I suppose we don't care about the memory leak here?

sandhyasun: I can fix this.

sandhyasun:
Any other comments ? If not I think  it's ready to merge

PR #1043, opened 2017-04-03
[TRAFODION-25] Temporary code for UPDATE costing

DaveBirdsall:
This adds some temporary code for UPDATE costing.

The existing code returns a constant value for cost on all UPDATE operations; it is a stub that was put in when the predecessor product was ported to Hadoop and HBase. The temporary code is a copy of the costing code for DELETE. It is being put in because we have an immediate need for parallel UPDATE plans in a particular application, which you don't get with the current stub.

This temporary code, of course, uses the DELETE theory of operation, which understates the costs for UPDATE in general. So it is a stop-gap until better code can be written.

To turn on this code, one must issue CQD HBASE_UPDATE_COSTING 'ON'. Without this CQD, the old stub, a constant cost of 1e32, will be used.

### DIFF
--- a/core/sql/optimizer/ScmCostMethod.cpp
+++ b/core/sql/optimizer/ScmCostMethod.cpp
@@ -4162,9 +4162,9 @@ CostMethodHbaseUpdate::scmComputeOperatorCostInternal(RelExpr* op,
 if ( CmpCommon::getDefault( OPTIMIZER_PRINT_COST ) == DF_ON )
     {
       pfp = stdout;
-      fprintf(pfp, "HbaseDelete::scmComputeOperatorCostInternal()\n");
+      fprintf(pfp, "HbaseUpdate::scmComputeOperatorCostInternal()\n");
       updateCost->getScmCplr().print(pfp);
-      fprintf(pfp, "HBase Delete elapsed time: ");
+      fprintf(pfp, "HBase Update elapsed time: ");
       fprintf(pfp,"%f", updateCost->
               convertToElapsedTime(
                    myContext->getReqdPhysicalProperty()).
--- a/core/sql/optimizer/ScmCostMethod.cpp
+++ b/core/sql/optimizer/ScmCostMethod.cpp
@@ -3868,7 +3868,7 @@ CostMethodHbaseUpdate::scmComputeOperatorCostInternal(RelExpr* op,
   // -----------------------------------------
   countOfStreams = countOfStreams_;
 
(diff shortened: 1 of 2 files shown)


DaveBirdsall:
Note: I'm running full regressions overnight (with the CQD 'OFF'), and will report results tomorrow. Please do not merge until then.

core/sql/optimizer/ScmCostMethod.cpp
```diff
@@ -3845,8 +3845,336 @@ CostMethodHbaseUpdate::scmComputeOperatorCostInternal(RelExpr* op,
...
+  CostScalar probeRowSize = delOp->getIndexDesc()->getKeyLength();
+  Cost * updateCost = 
+    scmCost(tuplesProcessed, tuplesProduced, tuplesSent, randomIOs, sequentialIOs, noOfProbes_,
+	    rowSize, csZero, outputRowSize, probeRowSize);
+
+#ifndef NDEBUG
+if ( CmpCommon::getDefault( OPTIMIZER_PRINT_COST ) == DF_ON )
+    {
+      pfp = stdout;
+      fprintf(pfp, "HbaseDelete::scmComputeOperatorCostInternal()\n");
```
DaveBirdsall: Just noticed that this says "HbaseDelete". It should be "HbaseUpdate". I will change it.

core/sql/optimizer/ScmCostMethod.cpp
```diff
@@ -3845,8 +3845,336 @@ CostMethodHbaseUpdate::scmComputeOperatorCostInternal(RelExpr* op,
...
+      updateCost->getScmCplr().print(pfp);
+      fprintf(pfp, "HBase Delete elapsed time: ");
```
DaveBirdsall: Here too.

DaveBirdsall:
In the rework, I added a script, analyzeOptimizerTrace.py, which is useful for seeing the big picture of plan choice in the Optimizer when looking at an optimizer debugging trace.

DaveBirdsall:
@sureshsubbiah noticed in the previous version of these changes that I had incorrectly cast a RelExpr node to an HBaseDelete node. It should have been HBaseUpdate. This is now fixed.

DaveBirdsall:
A full regression run of these changes on a workstation passed overnight. I will commit this now.

DaveBirdsall:
I merged this up this morning, but the notification part of the Apache infrastructure seems to not be working this morning.

PR #1045, opened 2017-04-04
[TRAFODION-2566] Reduce the virtual memory allocated in Trafodion pro…

selvaganesang:
…cesses with JDK1.8

Added two new configurable JVM options -XX:CompressedClassSpaceSize=128m and
-XX:MaxMetaSpaceSize=128m.

You can modify the default value of these options using environment variables
JVM_COMPRESSED_CLASS_SPACE_SIZE and JVM_METASPACE_SIZE respectively.

With this change the virtual memory of any SQL processes and tm is reduced by 1GB.

In JDK1.8 this space is used in lieu of permanent generation. Hence, I reduced the
default Java Objects Max heap size for SQL master, ESP and arkcmp processes to 256 MB from 512MB.
The udrserv process continues to have 512MB of java objects space.

### DIFF
--- a/core/sql/udrserv/udrserv.cpp
+++ b/core/sql/udrserv/udrserv.cpp
@@ -179,6 +179,8 @@ static Int32 processSingleCommandFromFile(FILE *f, UdrGlobals &glob);
 // Changed the default to 512 to limit java heap size used by SQL processes.
 // Keep this define in sync with executor/JavaObjectInterface.cpp
 #define DEFAULT_JVM_MAX_HEAP_SIZE 512
+#define DEFAULT_COMPRESSED_CLASSSPACE_SIZE 128
+#define DEFAULT_MAX_METASPACE_SIZE 128
 
 static NAString initErrText("");
 /*************************************************************************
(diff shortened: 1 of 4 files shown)


prashanth-vasudev:
+1

selvaganesang:
Yes,  I was trying to combine mxudr and SQL executor JVM startup to one. In case of SQL executor JVM startup parameter is purely environment variable driven. But in case of udr it is both CQD (control query default) and environment variable driven. Hence it become difficult to combine.  TM code is merely a clone of SQL executor code with subtle differences to sync it with their respective components.

core/sql/udrserv/udrserv.cpp
```diff
@@ -236,7 +238,27 @@ void InitializeJavaOptionsFromEnvironment(LmJavaOptions &javaOptions
    int maxHeapEnvvarMB = DEFAULT_JVM_MAX_HEAP_SIZE;
    sprintf(maxHeapOption, "-Xmx%dm", maxHeapEnvvarMB);
    javaOptions.addOption((const char *)maxHeapOption, TRUE);
-   
+
```
zellerh: You talked in the conversation thread about avoiding duplication of code. Selva and I also talked about that. The Language manager has the nice feature of being able to pass JVM options from the outside. If we allow this for JavaObjectInterface::createJVM() as well, couldn't we then combine the code? If I ever have time to do this, I'll attempt it, but right now it doesn't look like that will happen anytime soon.

For now, this looks good to me. If you like we can file a JIRA to combine JVM startup and to call JavaObjectInterface::createJVM() in the language manager instead of using its own code. Note that we may call this LanguageManager code from the executor as well, although it's probably highly unlikely right now.

DaveBirdsall: Filing a JIRA sounds like a good idea.

zellerh: Created [TRAFODION-2581](https://issues.apache.org/jira/browse/TRAFODION-2581).

PR #1046, opened 2017-04-05
[TRAFODION-2576] Fix long varchar bug in incremental UPDATE STATISTICS

DaveBirdsall:
The fix consists of two parts:

1. Moving two methods, addTruncatedSelectList and addTruncatedColumnReference from the HSSample class to the HSTableDef class in the first case and the HSColumnStruct struct in the second case. The methods more naturally belong to those objects, and it makes it easier to reference them in new contexts, such as the next item below.
2. In the logic that generates UPSERT statements to sample table increments, truncate column references where needed by calling the two methods above.


core/sql/ustat/hs_globals.cpp
```diff
@@ -1882,6 +1882,55 @@ NABoolean HSColumnStruct::operator==(const HSColumnStruct& other) const
...
+        qry += " FOR ";
+        
+        char temp[20];  // big enough for "nnnnnn)"
+        sprintf(temp,"%d)", maxLengthInBytes / CharInfo::maxBytesPerChar(charset));
+        qry += temp;
+        if (isFixedChar)
+          {
+            qry += " AS CHAR(";
+            qry += temp;
+            qry += ")";
```
zellerh: This is a question to this pre-existing code: We added character and byte semantics, like Oracle has it, to UTF-8 character columns. Do we need something similar here? Maybe as simple as adding the keyword "BYTES" when we deal with UTF-8 character columns with a length specified in bytes?

zellerh: ... or, actually, make it universally BYTES when the charset is UTF-8. That probably makes more sense, as the information content is then roughly the same.

zellerh: ... thinking about this again, we don't have a SUBSTRING method that will give the first n _bytes_ of a UTF-8 string, truncated to the next UTF-8 character. So, without such a function we can't really do what I suggested above.

If we have a long UTF-8 column c and maxLengthInBytes is 256, then we will take SUBSTRING(c for 64) here. In many situations and languages, this will be about 64 bytes of information, not 256.

DaveBirdsall: Thanks for these comments. Your analysis is correct. The code is limiting strings to 256 bytes, which for UTF-8 means 64 characters since the longest UTF-8 character is 4 bytes long.

By the way, the 256 byte limit is controllable by CQD so a customer or user can increase this if they wish. The CQD is USTAT_MAX_CHAR_COL_LENGTH_IN_BYTES.

PR #1051, opened 2017-04-07
[TRAAFODION-2568] [TRAFODION-2574]

sureshsubbiah:
2568 - Predicates that are always false at compile time will cause their scan to have begin and end keys reversed. No rows will be scanned. Idea due to Hans.
Files : ValueDesc.h/.cpp, SearchKey.h/.cpp and GenPreCode.cpp
2574 - Update on table with an index in which SET and WHERE clause contain an index key column can now get a index driven scan. Thank you Qifan and Hans for your help.
Files : NormRelExpr.cpp, RelExpr.cpp and RelScan.h
Two other minor changes that will help get index plans. I am not able to get reproducible test cases to open JIRAs on them. Thank you Eric and Dave for your help.
Files : ScmCostMethod.cpp and Analyzer.cpp
Thank you Ming for finding all these issues.

### DIFF
--- a/core/sql/optimizer/RelExpr.cpp
+++ b/core/sql/optimizer/RelExpr.cpp
@@ -8580,9 +8580,10 @@ void Scan::addIndexInfo()
     {
       IndexDesc *idesc = ixlist[indexNo];
 
-
+      NABoolean dummy; // halloweenProtection is decided using updateableIndex
+      // in GU::normalizeNode. Here this parameter is not used.
       // Determine if this index can be used for a scan during an update.
-      if (updatingCol AND NOT updateableIndex(idesc))
+      if (updatingCol AND NOT updateableIndex(idesc, preds, dummy))
         continue;
 
       ValueIdSet indexColumns(idesc->getIndexColumns());
@@ -9055,19 +9056,33 @@ void Scan::setTableAttributes(CANodeId nodeId)
   setBaseCardinality(MIN_ONE (tableDesc->getNATable()->getEstRowCount())) ;
 }
 
-NABoolean Scan::updateableIndex(IndexDesc *idx)
+NABoolean Scan::updateableIndex(IndexDesc *idx, ValueIdSet& preds, 
+                                NABoolean & needsHalloweenProtection)
 {
   //
-  // Returns TRUE if the index (idx) can be used for a scan during 
-  // an UPDATE. Halloween problem is protected with a sort using
-  // Scan::requiresHalloweenForUpdateUsingIndexScan(). 
-  // Returns FALSE only for certain embedded updates now.
+  // Returns TRUE if the index (idx) can be used for a scan during an UPDATE.
+  // Returns TRUE with needsHalloweenProtection also set to TRUE, if the index
+  // needs a blocking sort for Halloween protection
+  // Otherwise, returns FALSE to prevent use of this index. 
+  // Using the index in this case requires Halloween protection, but is likely 
+  // inefficient since there are no useful preds on the index key columns, so
+  // we don't add this index to list of candidates.
   //
 
-  if (!getGroupAttr()->isEmbeddedUpdate())
-    return TRUE ;
+  // The conditions of when this index returns TRUE can also be expressed as
+  // returns true for an index if it is 
+  // a) a unique/clustering index or 
+  // b) has a unique predicate on its key or 
+  // c) has an equals or range predicate on all the index columns that 
+  // get updated. If one of the key columns being updated has a range predicate
+  // then needsHalloweenProtection is set to TRUE.
+  // Note that if a key column is being updated and has no predicate on it then
+  // we return FALSE.
 
+  // preds has predicate in non-RangeSpec form, 
+  // while pred will be in RangeSpec form if feature is enabled.
   ValueIdSet pred = getSelectionPredicates(), dummySet;
+
   SearchKey searchKey(idx->getIndexKey(),
 	              idx->getOrderOfKeyValues(),
 	              getGroupAttr()->getCharacteristicInputs(),
@@ -9076,10 +9091,11 @@ NABoolean Scan::updateableIndex(IndexDesc *idx)
                       dummySet, // needed by the interface but not used here
                       idx
                       );
+
   // Unique index is OK to use.
   if (searchKey.isUnique())
     return TRUE;
-
+  
   const ValueIdList colUpdated = getTableDesc()->getColUpdated();
   const ValueIdList indexKey = idx->getIndexKey();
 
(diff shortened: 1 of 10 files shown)


selvaganesang:
jenkins, extra-tests

selvaganesang:
Running extra-tests to ensure that this change didn't cause any other issue

sureshsubbiah:
Both privs suites failed completely during extra-tests. These suites passed on my workstation though.

core/sql/optimizer/SearchKey.h
```diff
@@ -1042,6 +1045,10 @@ class HbaseSearchKey : public SearchKey
    ValueIdSet requiredOutputColumns_;
 
    ValueIdSet nonKeyPreds_;
+
+   // is a constant folded pred that returns FALSE always. SK has 
+   // min/max for begin/end reversed.
+   NABoolean isFalsePred_;
```
zellerh: In the longer term it would be nice to make this part of the generic SearchKey, not just HBaseSearchKey.

core/sql/optimizer/RelExpr.cpp
```diff
@@ -9121,38 +9090,85 @@ NABoolean Scan::updateableIndex(IndexDesc *idx)
...
+  for (CollIndex indexNo = 0; indexNo < ixlist.entries(); indexNo++)
+  {
+    IndexDesc *idx = ixlist[indexNo];
+    if (idx->isClusteringIndex() || 
+        (idx->isUniqueIndex() && 
+        (CmpCommon::getDefault(UPDATE_CLUSTERING_OR_UNIQUE_INDEX_KEY) 
+         == DF_ON)))
+      continue ; // skip this idesc
+
+    const ValueIdList indexKey = idx->getIndexKey();
```
zellerh: Could save a few nanoseconds here by not making a copy of the key and using a reference.

core/sql/optimizer/RelExpr.cpp
```diff
@@ -9121,38 +9090,85 @@ NABoolean Scan::updateableIndex(IndexDesc *idx)
...
 
-      if ((NOT(idx->isUniqueIndex() || idx->isClusteringIndex()))
-	     ||
-	  (CmpCommon::getDefault(UPDATE_CLUSTERING_OR_UNIQUE_INDEX_KEY) == DF_OFF))
+  // preds are in RangeSpec form
+  ValueIdSet preds = getSelectionPredicates();
+  const ValueIdList colUpdated = getTableDesc()->getColUpdated();
+  const LIST(IndexDesc *) & ixlist = getTableDesc()->getIndexes();
+
+  if ((colUpdated.entries() == 0) || (preds.entries() == 0) ||
```
zellerh: Can you explain why we return FALSE here when there are no predicates? I am thinking of the classical case that all of us employees like to use: update emp set salary = salary*1.1; There are no predicates, yet an index on salary has the Halloween problem. Is this handled in some different way?

core/sql/optimizer/RelExpr.cpp
```diff
@@ -9121,38 +9090,85 @@ NABoolean Scan::updateableIndex(IndexDesc *idx)
...
+  return TRUE;
+} // Scan::updateableIndex
+
+NABoolean Scan::requiresHalloweenForUpdateUsingIndexScan()
+{
+
+  // Returns TRUE if any non clustering index can be used for a scan 
+  // during an UPDATE and a key column of that index is in the SET
+  // clause of the update. If this method returns TRUE we will use a 
+  // sort node to prevent the "Halloween Update Problem".
```
zellerh: I wonder why you are writing a new method to determine this. There already was a method Scan::updateableIndex() that tests an index, which you changed above. We could call this updateableIndex() method here in a loop, if we need the info for all indexes. I also have comments on a few places in this code that are different from the Scan::updateableIndex() method, see below. Maybe you need this different behavior and that's why this new method is here?

core/sql/optimizer/RelExpr.cpp
```diff
@@ -9121,38 +9090,85 @@ NABoolean Scan::updateableIndex(IndexDesc *idx)
...
+        (idx->isUniqueIndex() && 
+        (CmpCommon::getDefault(UPDATE_CLUSTERING_OR_UNIQUE_INDEX_KEY) 
+         == DF_ON)))
+      continue ; // skip this idesc
+
+    const ValueIdList indexKey = idx->getIndexKey();
+    // Determine if the columns being updated are key columns. Each key
+    // column being updated must have an associated equality clause in
+    // the WHERE clause of the UPDATE for it to be used.
+    for (CollIndex i = 0; i < colUpdated.entries(); i++)
```
zellerh: I don't quite understand this comment and the logic below. The existing method, Scan::updateableIndex(), allows an index if it is a) a unique index or b) has a unique predicate on its key or c) has an equals predicate on all the index columns of a non-unique index. Do you want to achieve something different here?

core/sql/optimizer/RelExpr.cpp
```diff
@@ -9062,42 +9055,19 @@ void Scan::setTableAttributes(CANodeId nodeId)
...
-
 NABoolean Scan::updateableIndex(IndexDesc *idx)
 {
   //
-  // Returns TRUE if the index (idx) can be used for a scan during an UPDATE.
-  // Otherwise, returns FALSE to prevent the "Halloween Update Problem".
+  // Returns TRUE if the index (idx) can be used for a scan during 
+  // an UPDATE. Halloween problem is protected with a sort using
+  // Scan::requiresHalloweenForUpdateUsingIndexScan(). 
+  // Returns FALSE only for certain embedded updates now.
```
zellerh: See also comments below. I wonder why you changed this method, which seemed good to me, to something that no longer corresponds to its name?

core/sql/optimizer/ValueDesc.h
```diff
@@ -1481,6 +1481,16 @@ class ValueIdSet : public ClusteredBitmap
   // ---------------------------------------------------------------------
   NABoolean containsAnyTrue(ValueId &refAnyTrue ) const;
 
+  // ---------------------------------------------------------------------
+  // Returns a boolean to indicate if this vid set has a vid that 
+  // corresponds to a FlaseConstant. Such a constant is generated
```
zellerh: typo: Flase

core/sql/optimizer/SearchKey.h
sureshsubbiah: In the SearchKey begin and end keys are already reversed as we want for the isAFalseConstant predicate Please see SearchKey.cpp around Line 147.

core/sql/optimizer/RelExpr.cpp
sureshsubbiah: Yes, we can now use a an index to drive an update of an indexed column, regardless of the type of WHERE predicate. A blocking SORT will be added for anything other than an equality predicate without a constant. As such there is no concept of an updateableIndex anymore (not considering embeddedUpdate which is still retained in the old method). Every index is updateable, some predicate/index combinations require a blocing sort while others do not.  This method is to help decide whether a blocking sort is needed for an index driven scan of an update.

core/sql/optimizer/RelExpr.cpp
sureshsubbiah: The old updateableIndex would have returned TRUE if there was no predicate on the SET clause column. That is why I wrote the current method this way. I agree that we are open to the halloween problem as the case you mentioned. Since it is an existing bug I would like to address it in my next checkin which will be in a day or two.

core/sql/optimizer/RelExpr.cpp
sureshsubbiah: True. Will fix this next time.

core/sql/optimizer/RelExpr.cpp
sureshsubbiah: Yes. Please see response above. We  should now accept any index.  Certain index/predicate combinations will require a blocking sort. We will make the choice to create that sort, even if that particular index is not chosen.

core/sql/optimizer/ValueDesc.h
sureshsubbiah: Will fix next time.

core/sql/optimizer/RelExpr.cpp
sureshsubbiah: I did not quite understand the comment. This may be a repetition of the responses below. I will state my reasoning and then we can determine if it is reasonable.

1) With a blocking sort any index can be used to drive the scan of an update. Columns of unique index and keys of clustering index will not get here since they would have been changed to a delete + insert already.
2) IsUpdateable Index method is not needed anymore, but it has some logic about embeddedUpdate that I did not want to move. SoI kept the isUpdateableIndex method but it now has only the embeddedUpdate check.
3) IsUpdateableIndex used to return FALSE for non-equalityPreds, that logic has been moved to the new method. Here FALSE means we need a blocking sort. We go through all indexes, if any needs a blocking sort, then we flag the update as needing the sort.

core/sql/optimizer/ValueDesc.cpp
```diff
@@ -2764,6 +2764,54 @@ NABoolean ValueIdSet::containsAsEquiLocalPred(ValueId x) const
...
+        {
+          ItemExpr* childExpr = expr->child(i);
+          found = ((VEGReference*)(childExpr))->getVEG()
+            ->getAllValues().contains(x);
+        } else
+          found = (expr->child(i) == x);
+        
+        if (found) {
+          CollIndex j = (i==0) ? 1 : 0;
+          ValueIdSet vset(expr->child(j));
```
zellerh: Would it make sense to bypass creation of a ValueIdSet with one member and call method ItemExpr::doesExprEvaluateToConstant(FALSE, TRUE) directly? Maybe we don't even need this entire method if we go with the proposal above?

core/sql/optimizer/RelExpr.cpp
```diff
@@ -9093,79 +9109,83 @@ NABoolean Scan::updateableIndex(IndexDesc *idx)
...
-  // during an UPDATE and a key column of that index is in the SET
-  // clause of the update. If this method returns TRUE we will use a 
-  // sort node to prevent the "Halloween Update Problem".
-  //
+{ 
+  // Returns TRUE if any index that is in the list of indexes that will be 
+  // added later in addIndexInfo() to drive the scan for an UPDATE requires 
+  // Halloween protection. This is decided by using Scan::updateableIndex(). 
+  // If this method returns TRUE we will use a sort node to prevent the 
+  // "Halloween Update Problem".
```
zellerh: Thanks, that comment really helps.

core/sql/optimizer/RelExpr.cpp
```diff
@@ -9055,19 +9056,33 @@ void Scan::setTableAttributes(CANodeId nodeId)
...
   //
 
-  if (!getGroupAttr()->isEmbeddedUpdate())
-    return TRUE ;
-
+  // The conditions of when this index returns TRUE can also be expressed as
+  // returns true for an index if it is 
+  // a) a unique/clustering index or 
+  // b) has a unique predicate on its key or 
+  // c) has an equals or range predicate on all the index columns that
```
zellerh: I would have thought that we look for any key predicates for this  index, not just on the updated columns.  We are forming a SearchKey below and check for uniqueness. Couldn't we just examine that SearchKey instead, for example by using methods areAllBeginKeysMissing() and areAllEndKeysMissing() or by calling getKeyPredicates() and checking whether the returned set is empty?

Two examples, let's assume we have a base table with two indexes, one on column A and one on column B:

```
-- example 1:
update t set b=b+1 where a > 10; 
-- We would like to use index on a (no Halloween protection needed)
-- and we would like to exclude the index on b because it is not useful
-- and it requires Halloween protection.

-- example 2:
update t set b=b+1 where b > 10;
-- Here the index on a is not useful, but there is also no need to exclude it
-- since it does not require Halloween protection. Index on b is useful but
-- requires Halloween protection.
```

Would this work? Sorry for providing so much detail, this is just an attempt to provide detailed feedback, and not meant to say that this is the only solution:

* Rename the method to "indexIsUsefulForUpdateOrDelete"
* Let _ixcols_ be the set of columns the index is defined on, excluding the clustering key columns added to non-unique indexes, if necessary.
* Return TRUE if the predicates can be used to build a SearchKey for _ixcols_. Note that this excludes MDAM - in a future change we could check for MDAM predicates as well. Note also that it includes poor search key predicates - cost-based optimization should take care of that. Return FALSE otherwise (we would have to do a full index scan).
* Return FALSE for the needsHalloweenProtection parameter only if a) _ixcols_ do not contain any columns that are updated or b) the index is unique (that includes the clustering index) or c) it has a unique predicate on the columns it is defined on (excluding the clustering key columns added to the end of the key) or d) we have an equals predicate on all the columns that get updated. Return TRUE otherwise - the index needs Halloween protection.
* CQD UPDATE_USE_HALLOWEEN_INDEXES (Note: Shouldn't we use a separate CQD here, since this logic is very different from that for updating clustering keys?) be used to override the condition for Halloween protection:
  * OFF: Always return FALSE for indexes that require Halloween protection, so we never add a sort node
  * ON (the default): Follow the logic above
  * AGGRESSIVE: Always return FALSE for needsHalloweenProtection, meaning that we never force a sort and that we could encounter the Halloween problem - this would mean data corruption!

core/sql/optimizer/ValueDesc.cpp
sureshsubbiah: Yes, this would be better. I can make this change in a subsequent checkin.

core/sql/optimizer/RelExpr.cpp
sureshsubbiah: I think what is proposed here would work, it is an elegant alternative to what we have here.

zellerh: Just updated the comment above to simplify the condition for when we need Halloween protection. Checking for a unique key on _ixcols_ is not needed, all we have to check are equality predicates on the updated columns.

zellerh: In the process of mentally going through these conditions (see the previous messages above) and by talking to Suresh, I finally came to realize that what I was thinking matches what Suresh says in the comment here. Sorry that it took me a long time to arrive at this conclusion.

core/sql/optimizer/RelExpr.cpp
```diff
@@ -9121,39 +9106,90 @@ NABoolean Scan::updateableIndex(IndexDesc *idx)
...
+          else {
+            needsHalloweenProtection = FALSE;
+            return FALSE;
+          }
+        } // index key col is being updated
+      } // not a clustering or unique index
+    } // loop over index key cols
+  } // loop over cols being updated
 
   return TRUE;
```
zellerh: If we never hit line 9130 then isn't needsHalloweenProtection left at whatever value was passed in?

core/sql/optimizer/RelExpr.cpp
sureshsubbiah: I am sorry @zellerh. I misunderstood the comment. I did not realize you were asking to check if these two examples work as explained in this comment. I tested them now and get these results. They do show the desired behaviour with the current change. I am certain the alternate way discussed above will also lead to the desired result. Thank you.

create table tbl (k int not null primary key, a int, b int) ;
create index ixa on tbl(a) ;
create index ixb on tbl(b) ;

control query shape nested_join(nested_join(nested_join(cut,cut),cut),cut) ;
prepare s1 from update tbl set b=b+1 where a > 10;
explain options 'f' s1 ;
explain options 'f' s1 ;

10   .    11   root                            x                     6.59E+001
6    9    10   nested_join                                           6.59E+001
7    8    9    blocked_union                                         2.00E+000
.    .    8    trafodion_vsbb_upser            IXB                   1.00E+000
.    .    7    trafodion_vsbb_delet            IXB                   1.00E+000
4    5    6    nested_join                                           3.29E+001
.    .    5    trafodion_update                TBL                   1.00E+000
1    3    4    nested_join                                           3.29E+001
2    .    3    probe_cache                                           1.00E+000
.    .    2    trafodion_vsbb_scan             TBL                   1.00E+000
.    .    1    trafodion_index_scan            IXA                   3.29E+001

--- SQL operation complete.

control query shape nested_join(nested_join(cut,cut),cut) ;
prepare s2 from update tbl set b=b+1 where b > 10;
>>explain options 'f' s2 ;

8    .    9    root                            x                     6.69E+001
4    7    8    nested_join                                           6.59E+001
5    6    7    blocked_union                                         2.00E+000
.    .    6    trafodion_vsbb_upser            IXB                   1.00E+000
.    .    5    trafodion_vsbb_delet            IXB                   1.00E+000
2    3    4    nested_join                                           3.29E+001
.    .    3    trafodion_update                TBL                   1.00E+000
1    .    2    sort                                                  3.29E+001
.    .    1    trafodion_index_scan            IXB                   3.29E+001

--- SQL operation complete.

core/sql/optimizer/RelExpr.cpp
sureshsubbiah: True. I guarded against this by making sure needsHalloweenProtection is reinitialized in the normalizer, each time we call this function in the normalizer. Clearly it is better to reinitialize at the top of this method too. I will do that at that in my next checkin along with the vidset issue mentioned above. Thanks much for your ideas. Without them, this change would not have been possible,

core/sql/optimizer/RelExpr.cpp
zellerh: Well, I am sorry, @sureshsubbiah. Given that the current code does what I wanted, there really was no reason to suggest something else. So, this was just an exercise for myself to convince myself that the code works as is.

PR #1055, opened 2017-04-10
[TRAFODION-2483] Trafodion treat '\' as NULL in hive table

andyyangcn:
[TRAFODION-2483] fix issue: '\'  is treated as NULL in hive table

### DIFF
--- a/core/sql/executor/ExHdfsScan.cpp
+++ b/core/sql/executor/ExHdfsScan.cpp
@@ -1653,7 +1653,7 @@ char * ExHdfsScanTcb::extractAndTransformAsciiSourceToSqlRow(int &err,
                   // for non-varchar, length of zero indicates a null value.
                   // For all datatypes, HIVE_DEFAULT_NULL_STRING('\N') indicates a null value.
                   if (((len == 0) && (tgtAttr && (NOT DFS2REC::isSQLVarChar(tgtAttr->getDatatype())))) ||
-                      ((len > 0) && (memcmp(sourceData, HIVE_DEFAULT_NULL_STRING, len) == 0)))
+                      ((len > 1) && (memcmp(sourceData, HIVE_DEFAULT_NULL_STRING, len) == 0)))
                     {
                       *(short *)&hdfsAsciiSourceData_[attr->getNullIndOffset()] = -1;
                     }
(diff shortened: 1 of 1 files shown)


core/sql/executor/ExHdfsScan.cpp
```diff
@@ -1653,7 +1653,7 @@ char * ExHdfsScanTcb::extractAndTransformAsciiSourceToSqlRow(int &err,
                   // for non-varchar, length of zero indicates a null value.
                   // For all datatypes, HIVE_DEFAULT_NULL_STRING('\N') indicates a null value.
                   if (((len == 0) && (tgtAttr && (NOT DFS2REC::isSQLVarChar(tgtAttr->getDatatype())))) ||
-                      ((len > 0) && (memcmp(sourceData, HIVE_DEFAULT_NULL_STRING, len) == 0)))
+                      ((len > 1) && (memcmp(sourceData, HIVE_DEFAULT_NULL_STRING, len) == 0)))
```
traflm: This can handle the '\' problem, but if (not possible) the user define '\' as null, this logic will break.
So here, it should be to compare the whole HIVE_DEFAULT_NULL_STRING, and make sure length is equal, as I understand.
\N is NULL, but \NA is not NULL, what will happen if there is \NAA ?

sureshsubbiah:
jenkins run tests

sureshsubbiah:
jenkins, add user

core/sql/executor/ExHdfsScan.cpp
andyyangcn: For the current fix \NAA will not be treated as NULL. and I agree that making sure the length is equal is better.

core/sql/executor/ExHdfsScan.cpp
```diff
@@ -1653,7 +1653,7 @@ char * ExHdfsScanTcb::extractAndTransformAsciiSourceToSqlRow(int &err,
                   // for non-varchar, length of zero indicates a null value.
                   // For all datatypes, HIVE_DEFAULT_NULL_STRING('\N') indicates a null value.
                   if (((len == 0) && (tgtAttr && (NOT DFS2REC::isSQLVarChar(tgtAttr->getDatatype())))) ||
-                      ((len > 0) && (memcmp(sourceData, HIVE_DEFAULT_NULL_STRING, len) == 0)))
+                      ((len == strlen(HIVE_DEFAULT_NULL_STRING)) && (memcmp(sourceData, HIVE_DEFAULT_NULL_STRING, len) == 0)))
```
selvaganesang: I am not sure if the c++ compiler would optimize this. You could improve this further by getting the strlen only once in the TCB

selvaganesang:
Another option to optimize this code is to replace strlen with sizeof(HIVE_DEFAULT_NULL_STRING)

selvaganesang:
In the absence of further comments from @andyyangcn I will merge this PR.

PR #1059, opened 2017-04-11
[TRAFODION-2586] UPSERT USING LOAD running slower than UPSERT with tr…

sureshsubbiah:
…ansactions disabled

Showplan indicates problem is likely in initial queue length. Particularly for rowset upserts that are used for trickle loading, the number of rows in each statement si not sufficient for the queue resize to kick in.


sureshsubbiah:
jenkins, retest

PR #1061, opened 2017-04-11
[TRAFODION-2588] monitor failed to start if part of zookeeper server is down

mkby:
retry 3 times when connecting to zookeeper.

@narendragoyal @zcorrea please review.

### DIFF
--- a/core/sqf/monitor/linux/zclient.cxx
+++ b/core/sqf/monitor/linux/zclient.cxx
@@ -53,6 +53,9 @@
 // The monitors register their znodes under the cluster znode
 #define ZCLIENT_CLUSTER_ZNODE               "/cluster"
 
+// zookeeper connection retry count
+#define ZOOKEEPER_RETRY_COUNT                3
+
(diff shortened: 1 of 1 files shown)


core/sqf/monitor/linux/zclient.cxx
```diff
@@ -53,6 +53,9 @@
 // The monitors register their znodes under the cluster znode
 #define ZCLIENT_CLUSTER_ZNODE               "/cluster"
 
+// zookeeper connection retry count
+#define ZOOKEEPER_RETRY_COUNT                3
```
venkat1m: Wouldn't it be better to make this an environment variable or monitor config setting that can be changed instead of a hardcoded retry count of 3.

mkby: Based on the test result, actually retry for once is enough for this case. So it's not quite necessary to modify this value.

DaveBirdsall:
Is this ready to merge?

PR #1063, opened 2017-04-13
[TRAFODION-2592] Better error message when HBase is down at startup.

zellerh:
This code was written by Narendra Goyal.

### DIFF
--- a/core/sqf/sql/scripts/hbcheck
+++ b/core/sqf/sql/scripts/hbcheck
@@ -1,0 +1,35 @@
+#!/bin/bash
+#
+# @@@ START COPYRIGHT @@@
+#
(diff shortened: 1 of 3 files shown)


selvaganesang:
Can you please consider retrying hbcheck at least few times after sleeping for some time before exiting from sqstart

zellerh:
@selvaganesang, interesting, I thought the arguments "4 30" provided caused 4 retries, but looking at the code it seems like hbcheck is ignoring these arguments. The java program does set hbase.client.retries.number to 3, however, and it does some retries. I tested this by running "sqstart" and then starting HBase a few seconds later, and hbcheck did wait long enough for the "sqstart" command to succeed.

PR #1066, opened 2017-04-14
[TRAFODION-2584] Add support to register hive objects in traf metadata

anoopsharma00:
Syntax:
  register hive table/view [if not exists] <object-name> [cascade]
  unregister hive table/view  [if exists] <object-name> [cascade]
  cascade option: register/unregister all underlying objects that are
                  part of the specified view

update statistics, grant/revoke, traf views or external table creation
on hive objects will automatically and internally register those objects
in trafodion metadata.

invoke/showddl will show if this object is registered and whether
that registration was internal or explicit.

Get command extensions:
  get hive registered tables/view/objects in catalog trafodion;
  get hive external tables in catalog trafodion;

Cleanup command extensions:
  cleanup metadata command will clean up  inconsistent hive objects
   (underlying hive object is missing but object is registered or
    an external table exists)

   cleanup [hive table | hive view] on <object-name>;

Existing hive objects that had implicit or explicit external tables
created prior to this checkin will have no change in behavior.
ObjectUID of those external tables will continue to be used for
relevant operations.
One can drop those external tables and explicitly register them,
or a subsequent operation(upd stats, grant, etc) that needs objectUID will
automatically register them.

minor changes based on review comments of previous checkin

get all objects   command on hive metadata no longer fails.
get views on objects   return 3-part name that could be used to
  differentiate between a traf and hive view.

regress/hive/TEST007 has been extended.

TBD: Add register/unregister privileges


anoopsharma00:
jenkins retest

PR #1068, opened 2017-04-17
[TRAFODION-2594] Trafodion logs the same message twice in most of its…

selvaganesang:
… log4j and log4cxx logs

[TRAFODION-2478] Memory monitor Thread enhancements

a) Changed the configuration to have log4j.additive property set to false
b) ScratchIOSize is initialized so that this counter doesn't show invalid
   value
c) Cleaned up memory monitor concept and added code to log the memory monitoring
   counters to sccp log file every 5 minutes or when there is a pressure difference

### DIFF
--- a/core/sqf/conf/log4cxx.monitor.mon.config
+++ b/core/sqf/conf/log4cxx.monitor.mon.config
@@ -29,7 +29,7 @@ trafodion.log.filename.suffix=${TRAFODION_LOG_FILENAME_SUFFIX}
 log4j.threshhold=ALL
 
 # Define the root logger to the system property "trafodion.root.logger".
-trafodion.root.logger=FATAL
+trafodion.root.logger=INFO, monAppender
 log4j.rootLogger=${trafodion.root.logger}
 
 # Rolling File Appender
@@ -46,4 +46,4 @@ log4j.appender.monAppender.Append=true
 log4j.additivity.monAppender=false 
 
 # Foundation Monitor
-log4j.logger.MON=INFO, monAppender
+log4j.logger.MON=INFO
--- a/core/sqf/conf/log4cxx.monitor.mon.config
+++ b/core/sqf/conf/log4cxx.monitor.mon.config
@@ -22,17 +22,16 @@
 #
 
 # Define some default values that can be overridden by system properties
-trafodion.root.logger=INFO, monAppender
 trafodion.log.dir=${TRAF_HOME}/logs
 trafodion.log.filename.suffix=${TRAFODION_LOG_FILENAME_SUFFIX}
-
-# Define the root logger to the system property "trafodion.root.logger".
-log4j.rootLogger=${trafodion.root.logger}
 
 # Logging Threshold
 log4j.threshhold=ALL
 
-#
+# Define the root logger to the system property "trafodion.root.logger".
+trafodion.root.logger=FATAL
+log4j.rootLogger=${trafodion.root.logger}
+
 # Rolling File Appender
 #
 log4j.appender.monAppender=org.apache.log4j.RollingFileAppender
@@ -43,6 +42,8 @@ log4j.appender.monAppender.addPid=false
 log4j.appender.monAppender.layout=org.apache.log4j.PatternLayout
 log4j.appender.monAppender.layout.ConversionPattern=%d, %p, %c, %m
 log4j.appender.monAppender.Append=true
+
+log4j.additivity.monAppender=false 
 
 # Foundation Monitor
 log4j.logger.MON=INFO, monAppender
--- a/core/sqf/conf/log4cxx.trafodion.masterexe.config
+++ b/core/sqf/conf/log4cxx.trafodion.masterexe.config
@@ -22,15 +22,16 @@
 #
 
 # Define some default values that can be overridden by system properties
-trafodion.root.logger=INFO,  mxoAppender
+trafodion.root.logger=INFO, appender0 
 trafodion.log.dir=${TRAF_HOME}/logs
 trafodion.log.filename.suffix=${TRAFODION_LOG_FILENAME_SUFFIX}
 
-# Define the root logger to the system property "trafodion.root.logger".
-log4j.rootLogger=${trafodion.root.logger}
-
 # Logging Threshold
 log4j.threshhold=ALL
+
+# Define the root logger to the system property "trafodion.root.logger".
+log4j.rootLogger=${trafodion.root.logger}
+log4j.appender.appender0.file=${trafodion.log.dir}/master_exec${trafodion.log.filename.suffix}
 
 #
 # Rolling File Appender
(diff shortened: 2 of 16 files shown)


selvaganesang:
@zellerh Can you please review this PR at least the memory monitor counters related changes

core/sqf/conf/log4cxx.trafodion.masterexe.config
```diff
@@ -44,15 +45,15 @@ log4j.appender.mxoAppender.layout=org.apache.log4j.PatternLayout
...
 log4j.logger.MXOSRVR=ERROR, mxoAppender
 
 # SQL
 # uncomment the following line if you want to enable debug logging
-#log4j.logger.SQL.EXE=DEBUG
 log4j.logger.SQL=INFO
 log4j.logger.SQL.COMP=ERROR
-log4j.logger.SQL.EXE.Memory=INFO
-log4j.logger.SQL.ESP=ERROR
+log4j.logger.SQL.EXE=INFO
```
sandhyasun: This will by default log all the informational events . In most cases we don't want those and normally turn them on by default only when needed.

core/sqf/conf/log4cxx.trafodion.masterexe.config
```diff
@@ -44,15 +45,15 @@ log4j.appender.mxoAppender.layout=org.apache.log4j.PatternLayout
...
+log4j.logger.SQL.ESP=INFO
```
sandhyasun: Same comment as above.

core/sqf/conf/log4cxx.trafodion.masterexe.config
```diff
@@ -44,15 +45,15 @@ log4j.appender.mxoAppender.layout=org.apache.log4j.PatternLayout
 log4j.appender.mxoAppender.layout.ConversionPattern=%d, %p, %c, %m%n
 
 #MXOSRVR
+log4j.additive.mxoAppender=false 
 log4j.logger.MXOSRVR=ERROR, mxoAppender
 
 # SQL
 # uncomment the following line if you want to enable debug logging
```
DaveBirdsall: Should this comment be removed? (Since the line it refers to was removed.)

core/sqf/conf/log4cxx.trafodion.masterexe.config
selvaganesang: The hiearchy of the logging level is TRACE, DEBUG, INFO,WARN,ERROR and FATAL,  It is good to have the logging level set to INFO to ensure that the users can decipher the problem without the need to prompte the logging level to the higher hierarchy. When a message is logged at the incorrect level, it can be corrected later

sandhyasun: Ok . But we are definitely changing behaviour already. There is a generic method SQLMXLoggingArea::logExecRTInfo that uses INFO as the level. Am guessing it could affect performance since it's used in BMO code paths.    Not sure we wnt the user to see all these messages by default. If you do a  quick grep on our code you can see what is being logged using this generic method.

sandhyasun: Also in the same block compiler messages are logged under ERROR instead of INFO ? log4j.logger.SQL.COMP=ERROR

venkat1m: I'm not familiar with this code path. But generally we want to avoid flood of INFO messages. It makes debugging difficult when we have to scan through the logs for real problems.

selvaganesang: SQL doesn't log much in INFO level in C++ part of the code.  Also, INFO level is preferred level. I

selvaganesang: Removed log4j.logger.SQL.COMP=ERROR in the 2nd commit. But kept ERROR level for MXOSRVR for now because MXOSRVR seems to log INFO level messages quite enthusiastically.

selvaganesang:
The second commit consists of changes in config files because there were errors reported in monitor logs about the process inability to open the rootLogger appender, but the process would come up. When I properly configured the rootLogger appender like another RollingLogAppender or ConsoleAppender, initLog4cxx fails while opening the appender. So, I added just the rootLogger without any appender. This seems to work without having duplicate messages in the log file. The additivity flag in the appender didn't work as expected because the same appender is used for both rootLogger and the  child logger.

core/sqf/conf/log4cxx.monitor.mon.config
```diff
@@ -22,17 +22,16 @@
...
 
-# Define the root logger to the system property "trafodion.root.logger".
-log4j.rootLogger=${trafodion.root.logger}
-
 # Logging Threshold
 log4j.threshhold=ALL
 
-#
+# Define the root logger to the system property "trafodion.root.logger".
+trafodion.root.logger=FATAL
```
sandhyasun: Can you make the output atleats go to stdout ? The hierarchy of logging is supposed to be setup this way.

core/sqf/conf/log4cxx.trafodion.masterexe.config
```diff
@@ -43,19 +38,10 @@ log4j.appender.mxoAppender.addPid=false
...
-log4j.logger.SQL.COMP=ERROR
-log4j.logger.SQL.EXE.Memory=INFO
-log4j.logger.SQL.ESP=ERROR
-log4j.logger.SQL.PRIVMGR=INFO
-#
-log4j.logger.SQL.HDFS.JniTop=FATAL
-log4j.logger.SQL.HDFS.SeqFileReader=FATAL
-log4j.logger.SQL.HDFS.SeqFileWriter=FATAL
-log4j.logger.SQL.HBase=FATAL
+log4j.logger.SQL=INFO, mxoAppender
```
sandhyasun: How come the granularity has changed ? Sometimes we want to log only Hbase  or Hdfs messages. Having the different levels helps. Plus if we have to make the default INFO  our code needs to change  to not use INFO in sqlmxevents APIs to log debugging type of messages which it does today. Those are not meant for the end user.

core/sqf/conf/log4cxx.monitor.mon.config
selvaganesang: I tried and the process didn't come up as I explained in my comments.   Web searching revealed that it is ok not to configure rootLogger without appender though there is no default rootLogger appender

core/sqf/conf/log4cxx.trafodion.masterexe.config
selvaganesang: Yes. We will change these sqmxevents to DEBUG or TRACE level  as and when we come across messages that are becoming dominating in the logs. INFO will be the production level setting.

core/sqf/conf/log4cxx.monitor.mon.config
selvaganesang: I found another way to configure rootLogger, but at the same time not causing duplicate messages to be written. Changed the configuration files accordingly in commit 3

venkat1m:
+1

PR #1069, opened 2017-04-19
JIRA 2485,2540,2582,2591, plus other bug fixes

anoopsharma00:
[TRAFODION-2485]   Add support for REVERSE function
     (OperTypeEnum.h, exp/*, sqlparser.y,
      BindItemExpr.cpp, ItemExpr.cpp, SynthType.cpp,
      GenItemFunc.cpp)

[TRAFODION-2540]   Alter Drop column on table with identity
                 "generated always" fails
                 (file CmpSeabaseDDLtable.cpp)

[TRAFODION-2582]   "timestamp - timestamp" returns "days" instead of "seconds"
                 (file: BindItemExpr.cpp)

[TRAFODION-2591]   Create index on added column returns incorrect results.
                 (file generator/GenRelScan.cpp)

Bug: Invalid "set transaction" stmt syntax no longer causes assertion failure
              or abend. An error is issued instead.
                 (file BindRelExpr.cpp)

Bug: Hive inserts with mismatched tgt/src datatypes would sometimes do not
     insert or insert incorrect values.
                 (file GenFastTransport.cpp)

Bug: an error indication is now returned if buildEncodeTree method is
    called to encode a "NULL" value for a non-null column.
                  (file EncodedKeyValue.cpp)

hive/TEST007 now does initialize auth and register user


core/sql/exp/exp_function.cpp
```diff
@@ -2484,6 +2491,79 @@ ex_expr::exp_return_type ExFunctionTokenStr::eval(char *op_data[],
...
+
+          tgtPos -= charLen;
+          str_cpy_all(&tgt[tgtPos], &src[srcPos], charLen);
+          srcPos += charLen;
+        }
+    }
+  else
+    {
+      const char *csname = CharInfo::getCharSetName(cs);
+      ExRaiseSqlError(heap, diagsArea, EXE_INVALID_CHARACTER);
```
DaveBirdsall: Is this the error you want? Looks like it is error 8433 with text, "Invalid $0~string0 character encountered in $1~string1." But the error here is an invalid character set.

PR #1072, opened 2017-04-25
[TRAFODION-2598] Lob insert-select work from/to lob and external lob columns

sandhyasun:
Support for moving LOB data from source to target table containing LOB columns.


core/sql/generator/GenPreCode.cpp
```diff
@@ -5570,18 +5570,36 @@ RelExpr * HbaseInsert::preCodeGen(Generator * generator,
...
+                                            << DgInt0((Int32)srcCol->lobStorageType())
+                                            << DgInt1((Int32)tgtCol->lobStorageType())
+                                            << DgString0(tgtCol->getColName());
                         GenExit();
                       }
-		  li->lobStorageLocation() = col->lobStorageLocation();
+                    }
+                    else
+                      if ((child1Expr->getOperatorType() == ITM_BASECOLUMN)||
+                          (child1Expr->getOperatorType() == ITM_INDEXCOLUMN))
```
DaveBirdsall: Can a LOB column be an index column?

core/sql/exp/ExpLOBaccess.cpp
```diff
@@ -2454,7 +2560,55 @@ Ex_Lob_Error ExLobsOper (
...
+    case Lob_InsSel:
+      {
+        
+        ExLob *srcLobPtr;
+        Int16 flags;
+        Lng32  lobNum;
+        Int32 lobType;
+        Int64 uid, inDescSyskey, descPartnKey;
+        short schNameLen;
+        char schName[1024];
```
DaveBirdsall: I see several fixed-length char arrays here. Do we have guarantees that these won't overrun?

core/sql/generator/GenPreCode.cpp
sandhyasun: This is restricted elsewhere in the binder.

sandhyasun:
The rework includes replacing the fixed buffers with defines. Replaced several more occurences in various other places in the file.

PR #1073, opened 2017-04-26
[TRAFODION-2550] update sql reference manual about metadata clean up command

liuyu000:
Add "CLEANUP Statement".

### DIFF
--- a/docs/sql_reference/src/asciidoc/_chapters/sql_utilities.adoc
+++ b/docs/sql_reference/src/asciidoc/_chapters/sql_utilities.adoc
@@ -31,6 +31,7 @@ This section describes the {project-name} SQL utilities:
 
 [cols=","]
 |===
+| <<cleanup_statement,CLEANUP Statement>>                     | Cleans up objects whose information is inconsistent in the metadata.
 | <<load_statement,LOAD Statement>>                           | Uses the {project-name} Bulk Loader to load data from a source table, either
 a {project-name} table or a Hive table, into a target {project-name} table.
 | <<populate_index_utility,POPULATE INDEX Utility>>           | Loads indexes.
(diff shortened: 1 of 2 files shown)


docs/sql_reference/src/asciidoc/_chapters/sql_utilities.adoc
```diff
@@ -48,6 +49,326 @@ to connect to and issue SQL utilities, see the
...
 <<<
+[[cleanup_statement]]
+== CLEANUP Statement
+
+The CLEANUP Statement cleans up objects whose information is inconsistent in the metadata. This command can be used to clean up individual objects or all objects in a schema. This command can also be used to locate all inconsistent objects in the Trafodion metadata and then clean them up.
+
+Inconsistent objects can be cleaned up and removed from the metadata and HBase by using one of the following options.
+
+```
+CLEANUP {obj-type} {obj-name} [ ,UID {object-uid} ]
```
DaveBirdsall: Reading parser/sqlparser.y, it looks like [ , RETURN DETAILS ] is supported for CLEANUP obj-type obj-name  as well but oddly not when [ , UID object-uid ] is present. So perhaps modify this to CLEANUP obj-type obj-name [ , UID object-uid | , RETURN DETAILS ] ?

docs/sql_reference/src/asciidoc/_chapters/sql_utilities.adoc
```diff
@@ -48,6 +49,326 @@ to connect to and issue SQL utilities, see the
...
+==== Security and Privileges
+
+To clean up an object, one needs to have DROP privileges. If object privileges cannot be retrieved from the metadata or if metadata cleanup is being done, then one needs to be DB_ROOT to perform the cleanup.
+
+[[consistency_checks_during_access]]
+==== Consistency Checks During Access
+
+If an object exists in the metadata but does not exist in HBase, then error 4254 will be returned if that object is accessed during any DML or DDL operation.
+
+An object is also marked as inconsistent during any complex DDL operation. For example, if a CREATE TABLE also creates implicit indexes or constraints, then the object will be marked as inconsistent until the complex DDL operation is completed. If an abnormal error happens (like a process crash or a node failure), then that object remains inconsistent. Any attempt to access that object will return error 4254.
```
DaveBirdsall: "(like a process crash...)"  Wordsmith nit. "(such as a process crash...)" is a bit better because here you are citing process crashes and node failures as examples of errors, rather than characterizing all errors to be similar to process crashes or node failures, which is what "like" implies.

docs/sql_reference/src/asciidoc/_chapters/sql_utilities.adoc
liuyu000: Thanks Dave.
Yes, the {project-name} appears many times as a variable for Trafodion in .adoc files.
But here I'm afraid, the usage of curly brace, is also appropriate, as the Notation Conventions defined in the _Trafodion SQL Manaul_:
> { } Braces
Braces enclose **required syntax items**.
...
A group of items enclosed in braces is a list from which you are required to choose one item.

Here is an example of the syntax of GRANT Statement in the .adoc file and on the website:
``
GRANT {privilege [,privilege]... |ALL [PRIVILEGES]}
``
``
ON [object-type] [schema.]object
``
`` 
TO {grantee}
``
``
[WITH GRANT OPTION]
``
``
[[GRANTED] BY grantor]
``

As for this statement, `obj-type` and `obj-name` are required syntax items. :)

liuyu000: Thanks Dave. :)
But I'm afraid `RETURN DETAILS` is not supported here.
Ming helped me perform a test using the sqlci, it returned a syntax error, then we checked the parser/sqlparser.y and found that `RETURN DETAILS is not yet supported` , details are as below (circled in red):
return details is not yet suported

docs/sql_reference/src/asciidoc/_chapters/sql_utilities.adoc
liuyu000: Thanks Dave.
Your comments has been incorporated. :)

docs/sql_reference/src/asciidoc/_chapters/sql_utilities.adoc
DaveBirdsall: I should have been more clear.

The Trafodion SQL Manual describes the case where there are multiple syntax items, where you must choose one. Notice in the GRANT example you mention that one has to supply either a list of privileges or the keywords "ALL [ PRIVILEGES ]". But in the CLEANUP example, there is just one choice, not multiple choices. I suppose one could argue that { a } means I have to choose a, but then so does specifying "a" by itself.

DaveBirdsall: OK. Thanks for checking this out.

liuyu000: Thanks Dave.

Could I change to:
`CLEANUP object-type object-name [ , UID {object-uid} ]` 
?

If so, the syntax of GRANT Statement should be changed to:
`....`
`TO grantee`
`...`

(delete the curly braces around `grantee` since just one choice?

DaveBirdsall: Yes, I agree.

liuyu000: Got, thanks Dave. :)

PR #1074, opened 2017-04-27
[TRAFODION-2294] Fix permissions hole in Explain

DaveBirdsall:

1. Changes EXPLAIN so that it does not cache the query plan of whatever query it is explaining. This closes a permissions hole that otherwise allows one to execute the query afterward without privilege checking.

2. Extract the SqlParser_Flags_Enum from parser/SqlParserGlobalsCmn.h into a new file parser/SqlParserGlobalsEnum.h, to make it easier to use these literals in executor code. I changed a few hard-coded constants to their associated literals in executor/ExExeUtilExplain.cpp as well.

3. Update privs1/TEST120 regression test to include a test of the hole fixed by item 1 above.


DaveBirdsall:
I should mention the net effect of these changes is to allow anyone to do EXPLAINs of any query (as is the case before these changes), but subject all access to the data in the tables being queried to privilege checks. Thus a performance analyst can look at query plans without having direct access to the underlying data.

DaveBirdsall:
seabase/TEST022 failed with a transient failure that we've seen before. I will retest.

DaveBirdsall:
jenkins, retest

core/sql/arkcmp/CmpStatement.cpp
```diff
@@ -1072,7 +1082,8 @@ CmpStatement::process(const CmpMessageDDLwithStatus &statement)
 				 odbcProcess,
 				 noTextCache,
 				 aqrPrepare,
-				 standaloneQuery))
+				 standaloneQuery,
+				 doNotCachePlan))
```
sandhyasun: Have you fixed the code paths for both embedded compiling and separate compiler process compilation ?

DaveBirdsall: Yes. In the separate compiler case, the CmpCompileInfo object is sent to the tdm_arkcmp process. In the embedded case, it is passed to CmpContext::compileDirect. Either way, the bit travels to CmpStatement::process and ultimately down to CmpMain::sqlcomp. I've unit tested both paths.

PR #1078, opened 2017-05-02
[TRAFODION-2602] memory out of bound when fetch LONGVARCHAR UCS2

Weixin-Xu:


svarnau:
jenkins, add user

core/conn/unixodbc/odbc/odbcclient/unixcli/cli/sqltocconv.cpp
```diff
@@ -1555,6 +1555,7 @@ unsigned long ODBC::ConvertSQLToC(SQLINTEGER	ODBCAppVersion,
 								return IDS_22_003;
 							}
 							srcLength = TransStringLength + 1;
+                            DataLen = sizeof(DATE_STRUCT);
```
DaveBirdsall: I don't understand why this code change is correct. It looks like we are converting between a variety of C numeric types and ODBC varchar types. Why would DATE enter into it?

xwq: Hi Dave, thanks for review. The case here is application bind a DATE_STRUCT to LONGVARCHAR, and the charset is UCS2.  In the code, we reuse the variable "DataLen" to represent the length of the data from server side, since the charset is UCS2, so the length is 2 times of the DATE_STRUCT. After converting the data. DataLen will be used to represent the length of DATE_STRUCT which user binded.

BYW, this code was removed from the early commit which for VARHCAR (> 32K) support. but this line should not be removed. That's why we need this change.

xwq:
@arvind-narain Hi Arvind, Please help to review this pull request.

PR #1079, opened 2017-05-02
Fixing a broken link (already fixed on the actual site).

zellerh:


sandhyasun:
+1  Checked the new url and it works

PR #1080, opened 2017-05-02
[TRAFODION-2603] Remove obsolete utilities and component privileges

DaveBirdsall:

1. Removes the PUBLISH and UNPUBLISH utility statements. They are from a predecessor product, and do not work in Trafodion. And there is no specific requirement for them in Trafodion.

2. Removes the INITIALIZE TRAFODION, CREATE SEQUENCE utility statement. Now that the SEQ_GEN table is automatically created as part of the metadata schema, this utility is no longer needed.

3. Changes GRANT COMPONENT PRIVILEGE to give an unsupported statement error (SQL code -1010) if one of the DML_* privileges is given. They don't work, and there is consideration being given to removing them.

Note: If you upgrade your development instance to include this change, you may experience a make failure complaining about the lack of file parser/StmtDDLPublish.h.

The problem is that parser make dependency files still include a reference to this file. "make clean" does not delete the dependency files. (Maybe it should!)

To recover from this make failure, do the following:

cd $TRAF_HOME/../sql/nskgmake/parser/linux/64bit/debug/depend

rm *

cd $TRAF_HOME/../sql/parser

touch sqlparser.y

These commands remove the make dependency files for the parser, and tell make to build the parser file. When you next do a "make all", the dependency files will be regenerated.


core/sql/parser/ParKeyWords.cpp
```diff
@@ -832,7 +832,6 @@ ParKeyWord ParKeyWords::keyWords_[] = {
   ParKeyWord("PROTECTION",         TOK_PROTECTION,  NONRESTOKEN_),  
   ParKeyWord("PROTOTYPE",          TOK_PROTOTYPE,   COMPAQ_|RESWORD_),
   ParKeyWord("PUBLIC",             TOK_PUBLIC,      ANS_|RESWORD_),
-  ParKeyWord("PUBLISH",            TOK_PUBLISH,     NONRESTOKEN_),
```
robertamarton: sqlci/sqlci_yacc.y defines PUBLISH and UNPUBLISH as token.  May want to remove this also.

robertamarton:
Looks good.

core/sql/parser/ParKeyWords.cpp
DaveBirdsall: Nice catch. Fixed it. Thanks.

PR #1081, opened 2017-05-02
[TRAFODION-2604] sort merge phase memory pool improvements

prashanth-vasudev:


core/sql/executor/ex_sort.cpp
```diff
@@ -218,7 +304,8 @@ ExSortTcb::ExSortTcb(const ExSortTdb & sort_tdb,
   sortStats_ = NULL;
   childTcb_ = &child_tcb;
 
-  CollHeap * space = glob->getSpace();
+  //Create heap to be used by sort.
+  sortHeap_ = new(getHeap()) NAHeap("Sort Heap", (NAHeap *)getHeap(), 204800);
```
DaveBirdsall: What's the reasoning behind the hard-coded size, 204800?

sandhyasun: Isn't there a SORT_MAX_HEAP_SIZE_MB CQD ?

sandhyasun:
+1 Looks good to me too.

selvaganesang:
Please attempt to use NADELETE instead of delete.

core/sql/executor/ex_sort.cpp
prashanth-vasudev: SortHeap is used to allocate internal data structures to accommodate initial default number of 1 million records. The hard coded value of 200kb is the block size hint given to NAheap. This value will help allocations that are not too big or too small for sort.  SortHeap is initialized only at one place in sort and also this value is not a tune-able parameter , hence kept it as is and did not make it as a define or cqd. 

SORT_MAX_HEAP_SIZE_MB cqd defaults to 800MB , this is the max limit that will be enforced if quota system is not enforced for sort.

PR #1082, opened 2017-05-03
[TRAFODION-2294] Add known diffs file for executor/TEST140

DaveBirdsall:
It turns out the fix for this JIRA has a flaw. Not caching a query plan actually changes the plan! This shows up in executor/TEST140. This pull request provides a DIFFs file for that test. A permanent fix will be done later via JIRA TRAFODION-2605.


zellerh:
+1

PR #1083, opened 2017-05-04
[TRAFODION-2098] Move all hdfs directories trafodion needs/uses uder /user/trafodion

sandhyasun:


core/sql/sqlcomp/nadefaults.cpp
```diff
@@ -1956,7 +1956,7 @@ SDDkwd__(EXE_DIAGNOSTIC_EVENTS,		"OFF"),
   DDkwd__(HIVE_DEFAULT_CHARSET,            (char *)SQLCHARSETSTRING_UTF8),
   DD_____(HIVE_DEFAULT_SCHEMA,                  "HIVE"),
   DD_____(HIVE_FILE_CHARSET,                    ""),
-  DD_____(HIVE_FILE_NAME,     "/hive/tpcds/customer/customer.dat" ),
```
selvaganesang: Do you now why is it a specific table name is used as a default?

core/sql/sqlcomp/nadefaults.cpp
```diff
@@ -3378,7 +3378,7 @@ XDDkwd__(SUBQUERY_UNNESTING,			"ON"),
   DDkwd__(TRAF_READ_OBJECT_DESC,                       "OFF"),   
 
   DDkwd__(TRAF_RELOAD_NATABLE_CACHE,                   "OFF"),
-  DD_____(TRAF_SAMPLE_TABLE_LOCATION,                  "/sample/"),
```
selvaganesang: Would be good to come up with a different location rather than sample.  The name 'sample' connotes to kind of example table which might be deleted unknowingly.

core/sql/sqlcomp/nadefaults.cpp
sandhyasun: The CQD is not used anymore so removed it.  Must have been some prototype code.

core/sql/sqlcomp/nadefaults.cpp
sandhyasun: Barry and Dave worked on this sample feature for ustat. WIll let them decide.

PR #1087, opened 2017-05-09

selvaganesang:
New counters introduced as part of BMO

InterimRowCount - Counter to keep track intermediate rows produced by hash join and for intermediate runs in sort
phase           - BMO Phase for hash-join and sort

get statistics for statement s1, options 'cs' now works.

Removed attaching BMO stats to Non-root Paritial and leaf partition hash group by operators.
These operators are non-blocking BMOs.

Introduced 3 phases for hash grby operator to be visible in RMS Stats

options 'sl' now works with get statistics for qid.

The relevant new counters are now part of accumulated statistics type.

Extended the support up to 512 operators from 256 operators in RMS

Fixed both Type 2 and mxosrvr code to obtain all the externalized counters in accumulated stats correctly.


selvaganesang:
@zellerh Can you please review the code related to the new BMO counters

core/sql/executor/ex_hash_grby.h
```diff
@@ -302,6 +312,7 @@ NA_EIDPROC
   void resetClusterAndReadFromChild(); // Tmobile.
 
 public:
+  static const char *HashGrbyPhaseStr[];
```
zellerh: Minor comment: One way to avoid the static, global variable would be to use a method that returns a constant rather than this array.

core/sql/executor/ex_hash_grby.cpp
```diff
@@ -452,6 +463,8 @@ short ex_hash_grby_tcb::work() {
             hasFreeTupp_ = TRUE;
     	};
       };
+      if (bmoStats_)
+         bmoStats_->setBmoPhase(PHASE_END-HGB_READ_PHASE);
```
zellerh: This is executed for every row. Would it be cheaper to do this in the two places where we set state_ to HASH_GRBY_READ_CHILD?
Another question not about correctness but more related to personal taste: Why not simply assign values 1, 2, 3 etc. in the enum instead of subtracting two enums here? Feel free to ignore if you like this solution better.

selvaganesang: For your latter question, I didn't want to disturb the existing enum values and hence kept the existing value. But, for BMO stats, I wanted show the phase of the instance that hasn't moved to the next higher phase.  When the BMO stats is merged from different esp instances,  higher phase value is used. Hence, the phase enum value is reversed for BMO stats

selvaganesang: With the lack of test case to test out the phase changes, I took a safer approach of setting it many times. I will move this code to the two places mentioned after I did some testing with TPCDS queries

core/sql/executor/ex_hash_grby.h
selvaganesang: That's correct. Static global variables are also thread safe and I am hoping these phases are not changed forever and hence it shouldn't be error prone.

core/sql/executor/ex_hash_grby.cpp
zellerh: Thanks for the explanation!

prashanth-vasudev:
Looks good. +1

PR #1090, opened 2017-05-10
TRAFODION-2610 Register native HBase tables in traf MD plus HB Mapped table fixes

anoopsharma00:
-- support for registering hbase native cell/row format tables in traf MD
   Syntax:
       register [internal] hbase table [if not exists] <table-name>
       unregister [internal] hbase table [if exists] <table-name>

-- support to get info related to registered hbase tables
   Syntax: get hbase registered table in catalog trafodion

-- enhancements to HBase mapped tables and bug fixes

-- disabling of creation of external tables on native hbase
   tables unless they are mapped tables

-- support for case-insensitive format string in TO_DATE function
   (exp/exp_datetime.h)

-- regress/seabase/TEST022 enhanced with additional tests


core/sql/regress/seabase/EXPECTED030
```diff
@@ -7,7 +7,7 @@
 2016-03-01
 
 --- 1 row(s) selected.
->>select to_date('03/01/2016', 'MM/DD/YYYY') from (values(1)) x(a);
+>>select to_date('03/01/2016', 'mm/dd/yyyy') from (values(1)) x(a);
```
DaveBirdsall: Does "ddmonyy" also work in lower case?

anoopsharma00: yes, all TO_DATE formats are now case-insensitive.

core/sql/regress/hive/EXPECTED007
```diff
@@ -353,9 +355,7 @@ A            B
 
 >>unregister hive table if exists hive.hive.thive1;
 
-*** ERROR[4263] Object HIVE.HIVE.THIVE1 has invalid state and cannot be accessed. It is registered in trafodion metadata but the correponding hive object does not exist. Use 'cleanup hive table/view' command to cleanup this object.
-
-*** ERROR[3251] UNREGISTER operation could not be completed.
+*** ERROR[3251] UNREGISTER operation could not be completed. Reason: Specified object THIVE1 does not exist.
```
anoopsharma00: Good eye.
The error message is incorrect. It should return error that table has invalid state and need to be cleaned up.
Check for 'if exists' only apply to tables with valid state.
Will fix this in a subsequent release.

PR #1094, opened 2017-05-12
Fix for [Trafodion-2614]

sbroeder:
HBaseTxClient throws TableNotFoundException for TRAFODION._DTM_.TDDL


prashanth-vasudev:
Looks good to me. +1

PR #1095, opened 2017-05-14
TRAFODION-2462 jline not work well

mashengchen:
use -j yes or no to decide whether using jline
yes means use jline
no means not use jline 
default is not use jline


mashengchen:
jenkins, retest

mashengchen:
@kevinxu021  can you help to review it

kevinxu021:
+1 Good to go.

PR #1097, opened 2017-05-17
[TRAFODION-647]Make threshold of compression configurable

CoderSong2015:
In Windows,you need to modify the registry to set the compression threshold.
In Linux,you need to set the threshold by DSN file like:
[Default_DataSource]
...
Compression = balance
Compressionlimits = 3000
...


svarnau:
jenkins, add user

PR #1099, opened 2017-05-18
[TRAFODION-2605] Rework of fix for TRAFODION-2294

DaveBirdsall:
This fix reworks part of the fix for JIRA TRAFODION-2294. It contains the following changes:

1. The "useQueryCache" logic at the sqlcomp/CmpMain.cpp level has been changed. Formerly, there were just two options, useQueryCache == TRUE and useQueryCache == FALSE. For EXPLAIN, we want to pretend like we are using the query cache, but at the last minute choose not to cache the plan. (In this way, we get the same query plan as would be in cache, without actually caching it and thereby opening the security hole that TRAFODION-2294 fixed.)

2. The known differences file for regression test executor/TEST140 has been removed.

3. The expected results file for test executor/TEST140 has been updated. In one place, we do an EXPLAIN after executing the same statement. The new logic picks up the cached query plan from the statement for purposes of EXPLAIN, but that plan was compiled without the GENERATE_EXPLAIN CQD. So, the line "GENERATE_EXPLAIN .... ON" no longer appears on one of the EXPLAIN outputs. (I looked at trying to inhibit EXPLAIN from using cached plans, but this appears to be difficult: the logic that parameterizes plans for caching is tightly bound with the logic that does the cache lookup. In any case, using a cached plan does not open any security hole as privilege checking has already been done on the cached plan.)

4. In two places I replaced an obscure use of double Set_SqlParser_Flags calls with a single equivalent call to Assign_SqlParser_Flags. When the original lines were written, the Assign_SqlParser_Flags function did not exist.

### DIFF
--- a/core/sql/arkcmp/CmpStatement.cpp
+++ b/core/sql/arkcmp/CmpStatement.cpp
@@ -504,7 +504,7 @@ CmpStatement::process (const CmpMessageSQLText& sqltext)
     cmpmain.sqlcomp(qText, 0, 
 		    &(reply_->data()), &(reply_->size()),
 		    reply_->outHeap(), CmpMain::END, 
-		    &fragmentDir, typ, !doNotCachePlan);
+		    &fragmentDir, typ, doNotCachePlan ? CmpMain::EXPLAIN : CmpMain::NORMAL);
   }
   catch (...)
   {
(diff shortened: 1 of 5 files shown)


DaveBirdsall:
Note: It is possible that there will be other expected results files that need updating for lack of the line "GENERATE_EXPLAIN ... ON". I am running the full regression suite overnight on a workstation, and will make any needed updates later.

zellerh:
+1

Looks good to me, nice solution. I think preventing EXPLAIN from using cached plans may generate other differences. Changing the expected file is the best solution IMHO.

DaveBirdsall:
Full regressions passed on my workstation, so there are no additional expected files needing update.

PR #1101, opened 2017-05-23

selvaganesang:
Fixed a possible corruption issue in Type 2 JDBC driver while
getting the accumulated statistics.

The SE_IO_Bytes is now changed to SE_IO_KBytes. It used to return
in MB and now it is changed to KB.

Some of the IO counters were not displayed in hdfs scan operators


zellerh:
+1

PR #1103, opened 2017-05-25
[TRAFODION-2516] Further work on this original JIRA

sandhyasun:
JDBC testing revealed that some obsolete code in dinder caused issues . Removed that code. 
Added option to turn of GC completely if needed.
Added code to handle CLOB datatype in some places where it wasn't handled like BLOB datatype.


zellerh:
+1 looks good to me

PR #1109, opened 2017-06-02
TRAFODION - 2632 Performing update statistics on metadata tables caus…

robertamarton:
…es issues

Added code to return error and cleanup code a bit:
*** ERROR[9205] UPDATE STATISTICS is not supported for object <metadata table>

Also fixed a problem where privileges on Hive tables that included external
table gave false "no priv" error for the SHOWDDL <hive table> command.

### DIFF
--- a/core/sql/ustat/hs_globals.cpp
+++ b/core/sql/ustat/hs_globals.cpp
@@ -3776,7 +3776,7 @@ NABoolean HSGlobalsClass::isAuthorized(NABoolean isShowStats)
        if (privs == NULL)
          {
            *CmpCommon::diags() << DgSqlCode(-1034);
-            authorized = FALSE;
+            return FALSE;
          }
 
        // Requester must have at least select privilege
@@ -3788,7 +3788,7 @@ NABoolean HSGlobalsClass::isAuthorized(NABoolean isShowStats)
             << DgSqlCode( -4481 )
             << DgString0( "SELECT or MANAGE_STATISTICS" )
             << DgString1( objDef->getNATable()->getTableName().getQualifiedNameAsAnsiString() );
-            authorized = FALSE;
+            return FALSE;
           }
        }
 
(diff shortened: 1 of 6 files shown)


core/sql/ustat/hs_globals.cpp
```diff
@@ -3776,7 +3776,7 @@ NABoolean HSGlobalsClass::isAuthorized(NABoolean isShowStats)
        if (privs == NULL)
          {
            *CmpCommon::diags() << DgSqlCode(-1034);
-            authorized = FALSE;
+            return FALSE;
```
DaveBirdsall: Changing this to an immediate return breaks the timer at line 3795. Please either add a call to LM->LogTimeDiff or revert this change.

core/sql/ustat/hs_globals.cpp
```diff
@@ -3788,7 +3788,7 @@ NABoolean HSGlobalsClass::isAuthorized(NABoolean isShowStats)
             << DgSqlCode( -4481 )
             << DgString0( "SELECT or MANAGE_STATISTICS" )
             << DgString1( objDef->getNATable()->getTableName().getQualifiedNameAsAnsiString() );
-            authorized = FALSE;
+            return FALSE;
```
DaveBirdsall: Ditto

core/sql/ustat/hs_parser.cpp
```diff
@@ -178,89 +178,24 @@ Lng32 AddTableName( const hs_table_type type
...
-            if (SqlParser_NAMETYPE == DF_NSK)
-              {
-                catName = SqlParser_MPLOC.getSysDotVol();
-                hs_globals->tableType = GUARDIAN_TABLE;
-                hs_globals->tableFormat = SQLMP;
-              }
-            else
-            // LCOV_EXCL_STOP
-              catName = ActiveSchemaDB()->getDefaultSchema().getCatalogName();
+            catName = ActiveSchemaDB()->getDefaultSchema().getCatalogName();
```
DaveBirdsall: Thank you for getting rid of the NSK anachronisms.

core/sql/ustat/hs_globals.cpp
robertamarton: The return was added because the next statement "if ( privs->hasSelectPriv() )" cores if privs is NULL.     But will change the code to not break the time.

PR #1112, opened 2017-06-06
[TRAFODION-2626] make location of tmp directory configurable

svarnau:
This is one small part of this jira.  Default path of tmp is
the same old $TRAF_HOME/tmp location, but now it is configurable
via the TRAF_VAR variable. Enables pointing it to somewhere such
as /var/run/trafodion.


mkby:
do we need to make $TRAF_HOME/logs configurable either?

svarnau:
Yes, that will be another step for this JIRA. I just did not want to try to do everything at once.

PR #1113, opened 2017-06-07
[TRAFODION-2636] Memory leak in metadata context

zellerh:
Fixing a memory leak found by Selva. See [[TRAFODION-2636]](https://issues.apache.org/jira/browse/TRAFODION-2636) for details. Also adding a method that helped me in debugging this issue.

### DIFF
--- a/core/sql/common/NAMemory.h
+++ b/core/sql/common/NAMemory.h
@@ -468,6 +468,11 @@ NA_EIDPROC
   NAMemoryType getType() {  return type_; }
 NA_EIDPROC
   NABoolean getUsage(size_t* lastSegSize, size_t* freeSize, size_t* totalSize);
+  // for debugging
+NA_EIDPROC
+  NABoolean containsAddress(void *addr)
+        { return NABlock::blockHolding(firstBlk_, addr) != NULL; }
+
   // #ifdef NA_WINNT
 #ifndef __EID
   NABoolean  getSharedMemory() { return sharedMemory_; }
--- a/core/sql/optimizer/ControlDB.cpp
+++ b/core/sql/optimizer/ControlDB.cpp
@@ -108,7 +108,12 @@ ControlDB::~ControlDB()
 
 void ControlDB::setRequiredShape(ControlQueryShape *shape)
 {
-  delete requiredShape_;
+  if (requiredShape_)
+    {
+      delete requiredShape_->getShape();
+      delete requiredShape_;
+    }
+
   if (!shape
 	  ||
 	  !shape->getShape()


core/sql/optimizer/ControlDB.cpp
```diff
@@ -108,7 +108,12 @@ ControlDB::~ControlDB()
 
 void ControlDB::setRequiredShape(ControlQueryShape *shape)
 {
-  delete requiredShape_;
+  if (requiredShape_)
+    {
+      delete requiredShape_->getShape();
+      delete requiredShape_;
```
selvaganesang: Though delete would also work fine because of the macro, I would prefer NADELETE here with the correct heap.

core/sql/common/NAMemory.h
```diff
@@ -468,6 +468,11 @@ NA_EIDPROC
   NAMemoryType getType() {  return type_; }
 NA_EIDPROC
   NABoolean getUsage(size_t* lastSegSize, size_t* freeSize, size_t* totalSize);
+  // for debugging
+NA_EIDPROC
+  NABoolean containsAddress(void *addr)
+        { return NABlock::blockHolding(firstBlk_, addr) != NULL; }
+
```
selvaganesang: Is this some kind of debugging function?

selvaganesang: Please ignore this comment. I should have read the commit message

zellerh: Yes, I added the following code to the GroupAttributes constructor to find this leak, out of thousands of allocations it did:

>  const char *loc;
> 
>  if ((long) this < 0x8000000)
>    loc = "c++h";
>  else if ((long) this > 0x7ff000000000)
>    loc = "stck";
>  else if (CmpCommon::statementHeap() == NULL)
>    loc = "nost";
>  else if (CmpCommon::statementHeap()->containsAddress(this))
>    loc = "stmt";
>  else if (CmpCommon::contextHeap()->containsAddress(this))
>    loc = "ctxt";
>  else
>    loc = "unkn";
>
>  printf("GroupAttributes1, %#018lx %s\n", (long) this, loc);

core/sql/optimizer/ControlDB.cpp
zellerh: @selvaganesang can you explain this more? When I look at the definition of NADELETE, it explicitly says that NADELETE is **not** meant for NABasicObjects, and we are dealing with NABasicObjects here. Is it that you don't trust the method in which we put the heap pointer into NABasicObjects?

selvaganesang: If the NADELETE macro can't be used, then it is fine to stick with delete. However, NABasicObject also has the default constructor and hence it is possible to create objects without heap. I guess overloaded delete operator would take care of this too.

I preferred NADELETE macro because it is easy to read that it is being deleted from the heap.

zellerh: Yes, one thing I don't like about NABasicObject is that the heap pointer is junk for NABasicObjects that are allocated on the stack. Fortunately that's not a problem here. The allocation/deallocation of objects is pretty clear in this file. Thanks again for finding this leak!

PR #1114, opened 2017-06-07
[TRAFODION-2635] Fix two bugs re: HBASE CELL access on Traf objects

DaveBirdsall:
This set of changes fixes two bugs.

1.	In ValueIdList::computeEncodedKey (optimizer/ValueDesc.cpp), if we used HBASE.\”_CELL_\” access for a salted Trafodion table, and we had no begin/end key (e.g. select count(*) from hbase.\”_CELL_\”.\”TRAFODION.SCH.XYZ\”), the code would dereference a null pointer.
2.	In NATableDB::get, if we were creating an NATable object on the NATable cache for an HBASE.\”_CELL_\” access to a Trafodion table, the HBase-related descriptors were allocated on the statement heap instead of the context heap, resulting in a dangling pointer (with unpredictable failure modes) when the object was accessed that way in a second statement.

The change to optimizer/ValueDesc.cpp fixes the first bug.

The remaining changes fix the second bug.

Why so many changes for the second bug?

Partly, it is because routines in generator/Generator.cpp that wrote these descriptors into compiled plans were reused to obtain these descriptors for compile time logic. Unfortunately, the memory models are different: For compiled plans we use ComSpace for heap allocation (as we need the allocations to be contiguous in memory). For compiler access, we use an NAHeap. Before this set of changes, this was hidden in the GENHEAP macro in sqlcat/TrafDDLdesc.h; the code would use a ComSpace object if the space parameter to several methods was non-null, and use the statement heap if the space parameter was null.

The need to (sometimes) use the context heap in NATableDB::get meant that this GENHEAP scheme could no longer work. Since ComSpace and NAHeap inherit from NAMemory, one possibility was to change all relevant descriptor methods to use an NAMemory * parameter instead of ComSpace *. This set of changes attempts this possibility. There is a nuance: The NAMemory class hierarchy does not use virtual methods. So, when there is a need for polymorphism, the calling routines must explicitly cast the NAMemory pointer to the proper class. Fortunately, there was only one place where this was necessary. I’m not completely satisfied with this solution as it seems a dangerous coding practice. No simple alternative seemed to be available however.

After making this change, I encountered numerous test failures due to uninitialized variables in the various descriptors. Perhaps objects on the Statement heap are guaranteed (or vastly more likely) to be zeroed out while those on the Context heap are not. As a result, I added initializers for many of the descriptor members in sqlcat/TrafDDLdesc.h. I am not confident that I got all of the ones needed (I did initialize all pointer members so those are safe).

Please do not merge this until a full regression suite run has passed, which I’m doing on a workstation.

### DIFF
--- a/core/sql/common/NAMemory.h
+++ b/core/sql/common/NAMemory.h
@@ -473,6 +473,9 @@ NA_EIDPROC
   NABoolean  getSharedMemory() { return sharedMemory_; }
   void setSharedMemory() { sharedMemory_ = TRUE; }
 #endif
+
+  NABoolean isComSpace(void) { return derivedClass_ == COMSPACE_CLASS; } ;
+
 protected:
 
   // set the name of this memory
--- a/core/sql/generator/GenRelExeUtil.cpp
+++ b/core/sql/generator/GenRelExeUtil.cpp
@@ -284,6 +284,7 @@ TrafDesc *ExeUtilExpr::createVirtualTableDesc()
 {
   TrafDesc * table_desc =
     Generator::createVirtualTableDesc(getVirtualTableName(),
+				      NULL, // let it decide what heap to use
 				      ComTdbExeUtil::getVirtTableNumCols(),
 				      ComTdbExeUtil::getVirtTableColumnInfo(),
 				      ComTdbExeUtil::getVirtTableNumKeys(),
@@ -316,6 +317,7 @@ TrafDesc *ExeUtilDisplayExplain::createVirtualTableDesc()
   TrafDesc * table_desc = 
     Generator::createVirtualTableDesc
     (getVirtualTableName(),
+     NULL, // let it decide what heap to use
      ComTdbExeUtilDisplayExplain::getVirtTableNumCols(),
      vtci,
      ComTdbExeUtil::getVirtTableNumKeys(),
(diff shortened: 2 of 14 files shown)


zellerh:
+1 Looks good to me. Wow, pretty complicated.

DaveBirdsall:
I ran full regressions on a workstation overnight. I saw the same failure in seabase/TEST022 that Jenkins did.  Oddly, seabase/TEST028 failed in Jenkins, but in my overnight run, it passed, which suggests some non-determinism. All other suites passed. In any case, I need to debug the seabase test failures.

DaveBirdsall:
Learned something just now. TrafDesc objects are allocated using a factory method, TrafAllocateDDLdesc (sqlcat/TrafDDLdesc.cpp). This method has logic to zero out the object if the "space" parameter is non-null. Otherwise it assumes that the memory is already zeroed out. Evidently, this is true for ComSpace objects. My change above made the "space" parameter non-null for non-ComSpace objects. Hence my need for initializers above. And the regression test failures turned out to be evidence of more initializers needed. So, what I should have done was to change the logic in TrafAllocateDDLdesc rather than adding initializers in TrafDDLdesc.h.

DaveBirdsall:
The rework properly initializes the TrafDesc objects, removing my reticence concerning that aspect.

Once again, please don't merge yet; I will rerun full regressions on a workstation overnight.

zellerh:
+1, looks good.

DaveBirdsall:
Full regressions passed on my workstation. I will merge this now.

PR #1116, opened 2017-06-12
[TRAFODION-2643] Release date for 2.1

svarnau:
Used release date documented in board report.

This PR is for release2.1 branch!

### DIFF
--- a/docs/shared/revisions.txt
+++ b/docs/shared/revisions.txt
@@ -26,7 +26,7 @@
 [cols="2",options="header"]
 |===
 | Version    | Date
-| 2.1.0      | TBD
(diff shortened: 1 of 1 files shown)


zellerh:
If I understand this correctly, we will not create any new R2.1 artifacts, but you probably do want to update the web site? If so, would it make sense to just make this change in the master branch and update the site?

svarnau:
@zellerh - I agree for most content, but I think we need to retain ability to update docs for release branches, in the case of patch releases, or small changes like this of adding release date.

zellerh:
+1

Yes, makes sense, thanks for reminding me of patch releases that will also go into the 2.1 branch.

PR #1118, opened 2017-06-14
[TRAFODION-2035] Not able to copy table where there is column with type timestamp(0), error out with ERROR[8102]

Weixin-Xu:
Whatever the precision of oracle timestamp's second is 0-9, the char buffer is like YYYY-MM-DD HH24:MI:SS.FFFFFFFFF.
So data size is always equal the display size of timestamp(9) ( SQL_TIMESTAMP_LEN + '.' + 9 for precision )


xwq:
+1 , It looks good to me.

sureshsubbiah:
+1

PR #1119, opened 2017-06-14
[TRAFODION-2639] parallelize commmands with randomizes execution order core dumped

SuJinpei:
root cause: generated random index out of range.
fix by: narrow down random range, and the last element no need to be exchanged.


zellerh:
jenkins, add user

zellerh:
+1 Looks good to me.

xwq:
+1. 

Weiqing

PR #1120, opened 2017-06-15
[TRAFODION-2584] Add support to register hive schemas in traf MD

anoopsharma00:
This checkin extends jira TRAFODION-2584 to add support to register
hive schemas(referred to as databases in hive) in trafodion metadata.
Until now, only hive tables and views were registered.
Registering hive schemas would allow certain operations, like
granting/revoking privileges, to be done on a hive schema/database.

Syntax to register/unregister hive schema
  -- register/unregister hive schema hive.sch;

Syntax to show registred hive schemas
  -- get hive registered schemas in catalog trafodion

"showddl schema hive.sch" will now show the ddl for hive database
creation and whether this schema is registered.

showddl and 'get tables in schema' will now return error if that
schema doesn't exist.

regress/hive/TEST007 has been extended with additional tests.


sureshsubbiah:
+1

PR #1121, opened 2017-06-16
[TRAFODION-2346] ODB copy incorrect rows when using rowset and parallel

Weixin-Xu:
fix jira TRAFODION-2346 and TRAFODION-2373 bug as follows:
When using parallel, max rows/ parallel will cause rows cutting because of divisibility.
When using rowset, per thread rows/rowset will cause one more round contained the same rowset.
For example, rows = 5000 ,max = 500000 ,parallel = 6 :
per thread rows(size_t) = max rows / parallel = 500000 / 6 = 83333
round = per thread rows / rowset = 83333 / 5000 = 16 (actually do one more round)
Finally , 5000 * 17 * 6 = 510000 rows will be load and copy


Weixin-Xu:
jenkins, retest

sureshsubbiah:
+1

xwq:
+1

PR #1122, opened 2017-06-16
[TRAFODION-2646]mxosrvr change to AVAILABLE after timeout if no clien…

xwq:
…t connect

### DIFF
--- a/core/conn/odbc/src/odbc/nsksrvr/SrvrConnect.cpp
+++ b/core/conn/odbc/src/odbc/nsksrvr/SrvrConnect.cpp
@@ -1661,7 +1661,10 @@ void __cdecl SRVR::ASTimerExpired(CEE_tag_def timer_tag)
 
 		if( !shutdownThisThing )
 		{
-			static long long timeout = JULIANTIMESTAMP();
+            timeval tv;
+            int retcode = gettimeofday(&tv, NULL);
+            long long current_time = tv.tv_sec * 1000LL + tv.tv_usec / 1000LL;
+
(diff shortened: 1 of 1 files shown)


core/conn/odbc/src/odbc/nsksrvr/SrvrConnect.cpp
```diff
@@ -1680,47 +1683,49 @@ void __cdecl SRVR::ASTimerExpired(CEE_tag_def timer_tag)
...
-						timeout = JULIANTIMESTAMP();
-					}
+                    // Third token is dialogue ID
+                    tkn = strtok(NULL, ":");
+                    if (tkn == NULL)
+                        goto HandleNoTokens;
+
+                    currDialogueId = atoi(tkn);
+                    prevDialogueId = currDialogueId;
+
```
arvind-narain: Sorry for the delay Weiqing.

We don't seem to be using prevDialogueId now ? Earlier this was introduced for problems seen in doing multiple connects/disconnects via endurance test. If similar tests have been run to test this change then please go ahead and merge.
https://issues.apache.org/jira/browse/TRAFODION-156

xwq: I have run the all phx test. It haven't reproduced. And I also keep the logic as the old. So I think it's OK.

arvind-narain: Thanks for confirming. +1

PR #1123, opened 2017-06-19
[TRAFODION-2337] odb on windows can not use parameter with absolute path

Weixin-Xu:
When using absolute path on windows such as src=d:/src.txt , ':' will be used as separator.
So I add judging condition to avoid dividing when ':' and '/' is apart.


sureshsubbiah:
+1

PR #1098, opened 2017-05-18
TRAFODION-2385 jdbcT4 support boolean, tinyint, largeint

mashengchen:

PR #1124, opened 2017-06-19
TRAFODION-2385 bug fix

mashengchen:
there exists bug in #1098 , this PR is to fix it.


arvind-narain:
+1 thanks for the change.

PR #1125, opened 2017-06-19
[TRAFODION-1975] odb load error with different system formatted LF

Weixin-Xu:
Linux use '\n' as LF, Windows use "\r\n" and Mac use '\r'.
It's not yet distinguished in code. So that data inserted will be wrong, or odb will crash because the buffer is too long but not divided.


xwq:
+1

sureshsubbiah:
+1

PR #1126, opened 2017-06-19
Multiple fixes: REPLACE func maxlen, hbase unregister, DDLExpr cleanup

anoopsharma00:
-- REPLACE function max length is no longer limited to 32K
-- HBase object unregister was not working correctly. That is fixed.
-- class DDLExpr has been cleaned up to have one constructor for all cases.
   All options are now set using flags instead of constructor params.
-- "showddl <tab>, explain " command now returns an error instead of crash.


sureshsubbiah:
+1

PR #1128, opened 2017-06-20
[TRAFODION-2652] run cmd “./odb64luo -l src=nofile:tgt=ldata:map=map:max=300000:rows=K1:parallel=16” core dumped

SuJinpei:
root cause: "etab[eid].r = etab[eid].rbs / etab[eid].s" may be zero, this value will use for memory allocation and loop condition. so etab[eid].r = 0 is unreasonable, at least one.
fixed by: if etab[eid].r < 0 let it be 1.


SuJinpei:
jenkins, retest 
maven download failed

xwq:
+1

PR #1129, opened 2017-06-20
[TRAFODION-2626] Make conf directory location configurable

svarnau:
For user configuration files, use TRAF_CONF variable instead of
TRAF_HOME/conf.

In cases where we want to install config files to an alternate location
(say TRAF_CONF=/etc/trafodion), we would copy all the default files from
TRAF_HOME/conf to TRAF_CONF and make modifications in the TRAF_CONF location.

In development workspace, both locations can be one in the same.

One case where I left the original path is for the install_features file.
This file identifies version features and is not a user-configurable file.
Hence it should be read from the TRAF_HOME/conf location.


sureshsubbiah:
+1

selvaganesang:
I merged in the PR 1143. This changed the TrafConfiguration.java to look for trafodion-site.xml in the config directory. Please consider using TRAF_CONF variable in this file too.

PR #1130, opened 2017-06-20
[TRAFODION-2642] dcs install with enable HA

svarnau:
Also tweak some documentation on ambari installation.
Earlier change to bashrc caused install issue due to extraneous output. Re-route to /dev/null.
Enable HDP-2.6 stack installation.
Re-instate the scratch disk setting in python-installer per TRAFODION-2647.


svarnau:
This was merged with squash commit.

PR #1131, opened 2017-06-20
[TRAFODION-2644] Obsolete bash installer

svarnau:


amandamoran:
+1 but also -1000! Great work all!

svarnau:
@amandamoran  All the great work you did on bash installer has been inherited into the new generation.

PR #1136, opened 2017-06-21
[TRAFODION-2655] Fix 2 MDAM optimizer bugs. Update optimizer simulator.

DaveBirdsall:
This set of changes does three things:

1. Increases the weight of cumulative MDAM probe costs by a factor of 3 (configurable via the new CQD MDAM_PROBE_TAX), to encourage consideration of MDAM plans on fewer key columns. This addresses the issue with the first query in this JIRA.
2. Changes the default of CQD MDAM_APPLY_RESTRICTION_CHECK from '2' to '0'. This heuristic was put in place before cumulative probes were taken into account for MDAM probes. Now that this is the case, and now that their weight has been adjusted above, this heuristic should no longer be needed. The heuristic prevents MDAM plans in some cases where they are beneficial, such as the second query in this JIRA.
3. The predecessor product contained a feature called the "Optimizer Simulator". This feature allows one to capture DDL, histogram statistics, various cluster configuration information, and queries on one system, then load that information on a workstation for optimizer debugging. It is useful for precisely the sorts of plan and costing issues dealt with in this JIRA. The changes here update the Optimizer Simulator so that it now works on Trafodion. This should be considered a "technology preview" feature. This particular set of changes is a group effort: Credit for this work lays mostly with @HowardQin, with additional changes from @nonstop-qfchen, @zellerh and @eowhadi. 

Changes for the first two items are in optimizer/ScanOptimizer.cpp, sqlcomp/DefaultConstants.h and sqlcomp/nadefaults.cpp. The remaining files are from the third item.

### DIFF
--- a/core/sql/arkcmp/CmpContext.cpp
+++ b/core/sql/arkcmp/CmpContext.cpp
@@ -131,6 +131,8 @@ CmpContext::CmpContext(UInt32 f, CollHeap * h)
   sqlTextBuf_(NULL),
   uninitializedSeabaseErrNum_(0),
   hbaseErrNum_(0),
+  numSQNodes_(0),
+  hasVirtualSQNodes_(FALSE),
   trafMDDescsInfo_(NULL),
   transMode_(TransMode::IL_NOT_SPECIFIED_,    // init'd below
              TransMode::READ_WRITE_,
--- a/core/sql/arkcmp/CmpContext.h
+++ b/core/sql/arkcmp/CmpContext.h
@@ -395,7 +395,12 @@ public :
   void setArkcmpEnvDirect(const char *name, const char *value,
                           NABoolean unset);
 
+  // variables managed by the HHDFSMasterHostList class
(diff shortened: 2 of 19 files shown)


DaveBirdsall:
Instructions for using the "Optimizer Simulator" (aka OSIM):

To capture OSIM information from a source system in a sqlci session:

control osim capture location '\<directory name\>';
prepare xx from \<problematic query\>;
control osim capture stop;

Then tar the contents of the directory, and move the tar to a workstation. Untar it there.

On the workstation, to load the OSIM information:

control osim load from '\<directory\>';
control osim simulate start;
prepare xx from \<problematic query \>;

When done debugging the query,

control osim unload '\<directory\>';

DaveBirdsall:
Please do not merge this change yet. I will run a full set of regression tests overnight on a workstation.

zellerh:
+1
Good to have OSIM on Trafodion!

DaveBirdsall:
A full regression run on a workstation for this change passed overnight. So this is good to merge.

PR #1138, opened 2017-06-22
[TRAFODION-2649] Fixed 'rmscheck' method for obtaining status

zcorrea:
It now generates the 'rmscheck.sql' logic in a temporary file which obtains the
node-ids from the 'trafconf' utility. The tempory file is created at 'rmscheck'
invocation.

### DIFF
--- a/core/sqf/sql/scripts/rmscheck
+++ b/core/sqf/sql/scripts/rmscheck
@@ -22,7 +22,7 @@
 # @@@ END COPYRIGHT @@@
 #
 
-if [ -z $TRAF_HOME ]; then
+if [[ -z $TRAF_HOME ]]; then
     echo
     echo "The TRAF_HOME environment variable does not exist."
     echo "Please ensure sqenv.sh has been sourced."
@@ -38,11 +38,5 @@ if [[ $sq_stat == 255 ]]; then
    exit 1
 fi
 
-# Generate the 'rmscheck.sql' file
-temp_rmscheck_sql=`mktemp -t`
-sqgenrmscheck $temp_rmscheck_sql
-echo temp_rmscheck_sql=$temp_rmscheck_sql
-
 echo "Timestamp                           Id    Status "
-sqlci -i $temp_rmscheck_sql | grep 'Node\|ERROR' | grep -v varchar
-rm -f $temp_rmscheck_sql
+sqgenrmscheck | sqlci | grep 'Node\|ERROR' | grep -v varchar
--- a/core/sqf/sql/scripts/rmscheck
+++ b/core/sqf/sql/scripts/rmscheck
@@ -21,7 +21,28 @@
 #
 # @@@ END COPYRIGHT @@@
 #
-# Most of this was stolen from WMScheck...  which may have started as NDCScheck
+
+if [ -z $TRAF_HOME ]; then
+    echo
+    echo "The TRAF_HOME environment variable does not exist."
+    echo "Please ensure sqenv.sh has been sourced."
+    echo
+    exit 1;
+fi
+
+# Check whether the Trafodion environment up.
+sqcheck -i 1 -d 1 > /dev/null 2>&1
+sq_stat=$?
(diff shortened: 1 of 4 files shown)


core/sqf/sql/scripts/rmscheck
```diff
@@ -21,7 +21,28 @@
 #
 # @@@ END COPYRIGHT @@@
 #
-# Most of this was stolen from WMScheck...  which may have started as NDCScheck
+
+if [ -z $TRAF_HOME ]; then
+    echo
```
svarnau: Need double brackets here, or an empty variable will give shell syntax error.

core/sqf/sql/scripts/sqconfig
```diff
@@ -20,7 +20,7 @@
 # @@@ END COPYRIGHT @@@
 
 begin node
-_virtualnodes 2
+_virtualnodes 4
 end node
```
svarnau: Intentional to change to 4 virtual nodes for single-nodes? Or was this just for testing?

core/sqf/sql/scripts/sqgenrmscheck
```diff
@@ -0,0 +1,123 @@
...
+}
+
+###########################################################
+# MAIN portion of sqgen begins
+###########################################################
+
+SQLCI_IN_FILE=rmscheck.sql
+
+if [ -z $TRAF_HOME ]; then
+    echo
```
svarnau: double brackets again.

core/sqf/sql/scripts/sqgenrmscheck
```diff
@@ -0,0 +1,123 @@
...
+    echo "The TRAF_HOME environment variable does not exist."
+    echo "Please ensure sqenv.sh has been sourced."
+    echo
+    exit 1;
+fi
+
+rm -f $SQLCI_IN_FILE
+
+cd $TRAF_HOME/sql/scripts
+
```
svarnau: We are writing temp files into sql/scripts?  Ugh.
I'm busy trying to move such things to $TRAF_VAR.

core/sqf/sql/scripts/rmscheck
```diff
@@ -21,7 +21,28 @@
...
+
+# Generate the 'rmscheck.sql' file
+temp_rmscheck_sql=`mktemp -t`
+sqgenrmscheck $temp_rmscheck_sql
+echo temp_rmscheck_sql=$temp_rmscheck_sql
 
 echo "Timestamp                           Id    Status "
-sqlci -i $TRAF_HOME/sql/scripts/rmscheck.sql | grep 'Node\|ERROR' | grep -v varchar
+sqlci -i $temp_rmscheck_sql | grep 'Node\|ERROR' | grep -v varchar
+rm -f $temp_rmscheck_sql
```
svarnau: Would it be easier to generate commands to stdin of sqlci instead of using a temp file?

core/sqf/sql/scripts/sqconfig
zcorrea: No. I need to fix this, I was testing and forgot to change it back.

core/sqf/sql/scripts/rmscheck
zcorrea: Got it!

core/sqf/sql/scripts/sqgenrmscheck
zcorrea: Good point! Perhaps your next suggestion on using stding will remove this issue altogether. Right?

core/sqf/sql/scripts/rmscheck
zcorrea: Yes, this would be better. Scripting is not one of my strengths.

core/sqf/sql/scripts/sqgenrmscheck
svarnau: yep.

PR #1139, opened 2017-06-22
TRAFODION [2641] User who has MANAGE_STATISTICS privilege can't do up…

robertamarton:
…date

   statistics on HIVE tables

TRAFODION [2175] a user should only see specific schemas/tables that he has
   privileges to
Updated the following get commands:
  get schemas (in catalog)
  get tables, indexes (in schema)
  get sequences, views (in schema, in catalog)
  get libraries, procedures, functions, table mapping functions

TRAFODION [1573] Additional GET commands for privileges
  get privileges on table
  get privileges on view
New regression test privs1/TEST125

Fixed bug:  user granted MANAGE privilege does not have MANAGE sub-privs
Changed REGISTER_HIVE_OBJECT to be treated as a sub-priv under CREATE
Changed UNREGISTER_HIVE_OBJECT to be treaed as a sub-priv under DROP

### DIFF
--- a/core/sql/executor/ExExeUtil.h
+++ b/core/sql/executor/ExExeUtil.h
@@ -2497,6 +2497,11 @@ private:
     const char *catName,
     const char *schName,
     const char *objName);
+
+  NAString getGrantedPrivCmd(
+    const NAString &roleList,
+    const char * cat);
+
   char * getRoleList(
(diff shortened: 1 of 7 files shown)


sureshsubbiah:
+1

PR #1140, opened 2017-06-22
[TRAFODION-2663] Simplify HBase config settings in installers

svarnau:
Remove settings that are now in per-table trafodion-site.xml config.

For Ambari, 2.2 service_adviser over-rides the 2.1 version. This allows
possibility of single plug-in to manage 2.1 & 2.2 installs, though this is not
tested.

Difference in 2.2 service adviser is just removal of these same hbase properties.


svarnau:
@selvaganesang  notice that hbcheck failed on all the HDP nodes.

install/python-installer/scripts/hadoop_mods.py
```diff
@@ -122,10 +122,6 @@ def mod(self):
         }
         self.p.post('%s/config_groups' % cluster_url, hbase_config_group)
 
-        if dbcfgs['secure_hadoop'].upper() == 'Y':
```
selvaganesang: Should these lines stay around for secure hadoop?

selvaganesang:
Can I have access to the Jenkins VM?

install/python-installer/scripts/hadoop_mods.py
svarnau: These are not values we are adding, just values we are trying to avoid deleting.  Since we are no longer writing value to hbase.coprocessor.region.classes, there is no danger of over-writing these security values.

svarnau:
jenkins, retest -- The test volumes are not retained that long, so need to re-run to generate failed VM to debug.

sureshsubbiah:
+1

PR #1141, opened 2017-06-22
[TRAFODION-2637] Library management with Cloudera parcels

zellerh:
Also in this commit:

[TRAFODION-2581] Combine JVM startup code of SQL executor and UDR server

We have library management SPJs that allow a user to deploy library
files from a client machine. These SPJs make use of the pdcp command
that is installed on many clusters, but not always. The new method
uses temporary HDFS files and a TMUDF to distribute the library
file to all nodes of the cluster.

For now I left the old PUT method in place and added a new PUTFILE
method. We can remove the old method later.

A few other things fixed in the process:

- Make sure that when a UDR server calls the executor and uses a JVM,
  we share the same JVM. To do this, we need to use the same startup
  method, and I needed to change the executor startup method to accept
  the extra options used for the UDR server.

- Got rid of the "HBase compatibility mode" in JavaObjectInterface.

- A common method to execute SQL in the library management code.

- removed the obsolete init_libmgmt.sh file.

- We have special code that makes sure a UDR can request to run in
  parallel, including a request to run on all nodes. Expanded this
  special code from predefined UDFs to any UDF.


zellerh:
@selvaganesang we talked about the fix to the JVM startup code before. If you could take a look at the changes in files JavaObjectInterface.* that would be great.

robertamarton:
What changes are needed to the installation procedure to pick up the new library.  For tenants, the installation procedure makes a call to initialize trafodion, add tenant usage.

core/sql/udrserv/udrserv.cpp
```diff
@@ -230,35 +230,6 @@ void InitializeJavaOptionsFromEnvironment(LmJavaOptions &javaOptions
 {
    char *val;
 
-   if ((val = getenv("CLASSPATH")))
-   {
-      javaOptions.addSystemProperty("java.class.path", val);
-   }
-   char maxHeapOption[64];
-   int maxHeapEnvvarMB = DEFAULT_JVM_MAX_HEAP_SIZE;
```
selvaganesang: You can remove all the defines related to this code at the top of this file. For some reason, I was not able to comment at that line.

zellerh: Thanks for catching that, will do.

zellerh: I pushed [another commit](https://github.com/zellerh/incubator-trafodion/commit/8b38be76203dabcb47a03c09a72b14bd305d1564) to remove the 3 lines, but for some reason it does not show up here. If the rework doesn't get committed I'll fix that in a later PR.

core/sql/langman/LmJavaOptions.cpp
```diff
@@ -134,6 +134,17 @@ void LmJavaOptions::addOptions(const char *options,
   NADELETEBASIC(copy, collHeap());
 }
 
+CollIndex LmJavaOptions::findByPrefix(const char *prefix) const
+{
+  size_t prefixLen = strlen(prefix);
+
+  for (CollIndex i=0; i<options_.entries(); i++)
+    if (strncmp(options_[i], prefix, prefixLen) == 0)
```
DaveBirdsall: Is it possible for one prefix to be a subset of another? Just making up a possible example: -Xms vs. -Xms2? Would this code behave correctly then?

core/sql/lib_mgmt/src/main/java/org/trafodion/libmgmt/SyncLibUDF.java
```diff
@@ -0,0 +1,170 @@
...
+    public SyncLibUDF()
+    {}
+
+    @Override
+    public void describeParamsAndColumns(UDRInvocationInfo info)
+        throws UDRException
+    {
+        // this TMUDF takes no table-valued inputs, two string
+        // parameters, and generates a table with a single integer
+        // value. It assumes that it was created with the following
```
DaveBirdsall: "...with a single integer value". From the example, it looks like you meant, "...with a single integer column." I'm wondering what the column values mean.

core/sql/lib_mgmt/src/main/java/org/trafodion/libmgmt/SyncLibUDF.java
```diff
@@ -0,0 +1,170 @@
...
+
+    @Override
+    public void describeDesiredDegreeOfParallelism(UDRInvocationInfo info,
+                                                   UDRPlanInfo plan)
+        throws UDRException
+    {
+        boolean usesVirtualNodes = false;
+
+        // check for configurations with virtual nodes. Run the UDF serially
+        // in those cases, since all the virtual nodes share the same node.
```
DaveBirdsall: This is an interesting bit of smarts. Is it universal among our system UDFs? I wonder too about our optimizer: does it reduce DoP when we run on virtual nodes?

core/sql/langman/LmJavaOptions.cpp
zellerh: Yes, if that happens we may need to do some additional parsing.

core/sql/lib_mgmt/src/main/java/org/trafodion/libmgmt/SyncLibUDF.java
zellerh: Yes, I'll fix that next time. The column value is the instance number. So, for n instances, we would expect values 0 ... n-1 back.

core/sql/lib_mgmt/src/main/java/org/trafodion/libmgmt/SyncLibUDF.java
zellerh: This particular code was added for the event log reader. For these types of maintenance UDFs, we want to run exactly one instance per physical node. The event log reader used to return all the events twice before I added this code. I don't think we need to adjust the DoP for regular queries. As far as I know we don't support virtual nodes for production.

zellerh:
Closing this PR. It somehow got messed up. The first commit actually got committed, but the second commit put this PR into a bad state. I will commit that part in a new PR.

PR #1142, opened 2017-06-22
[TRAFODION-2662] Fix incr stats bug when enormous sample tables are used

DaveBirdsall:
Refactored some of the incremental update statistics (IUS) logic. The creation and population of an incremental sample table depended on one column's data fitting in memory. That dependency has been removed; the incremental sample is created and populated even if no columns fit in memory.

This is a classic example of an "empty set" bug, where logic doesn't work correctly on empty sets.


zellerh:
+1 looks good to me.

sureshsubbiah:
+1

PR #1143, opened 2017-06-23
[TRAFODION-2654] Change the location of trafodion-site.xml from $TRAF_HOME/etc  to config 

selvaganesang:
[TRAFODION-2653] Sort operator loops at times 

All HBase client connections from Trafodion uses Trafodion configuration concept. Any client side property can now be set in trafodion-site.xml. 

 The scratch files are now open in blocking mode. Hence any scratch file operation shouldn't return EAGAIN.  

The internal table TDDL of table is now created without any table co-processors attached to it

The WAL sync can now be synchronized if needed to avoid DWE


core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/TrxRegionEndpoint.java.tmpl
```diff
@@ -4015,6 +4015,25 @@ CoprocessorService, Coprocessor {
...
+     try {
+       if (txid == 0)
+          wal.sync();
+       else
+          wal.sync(txid);
+     } catch (IOException wale) {
+        wale.fillInStackTrace();
+        LOG.error("commitRequest txId: " + transactionId + " HLog seq " + txid + " Caught IOException in HLOG sync ", wale );
+        try {
+           Thread.sleep(1000);          ///1000 milliseconds is one second.
```
prashanth-vasudev: Why thread.sleep if IO exception?

selvaganesang: This is to stagger the sync operation assuming that it was causing the concurrentModificationException. But @shangshengtung came up with different change that took care of concurrentModificationException. He will be delivering his change in a separate PR soon

PR #1144, opened 2017-06-23
TRAFODION-2668 one mxosrvr cant startup, dcsServer down

mashengchen:


dcs/src/main/java/org/trafodion/dcs/server/ServerManager.java
```diff
@@ -414,32 +431,40 @@ public Boolean call() throws Exception {
...
-                       if (!retryCounter.shouldRetry()) {
-                          throw new IOException("Node " + nid + " is not Up");
-                       } else {
-                           retryCounter.sleepUntilNextRetry();
-                           retryCounter.useRetry();
-                       }
-                   }
-                   serverHandlers[childInstance-1] = new ServerHandler(childInstance);
-                   completionService.submit(serverHandlers[childInstance-1]);
+                        if (!retryCounter.shouldRetry()) {
```
selvaganesang: I think you need to fix retryCounterFactory to do the looping for retry count or get rid of retryCounter completely and use the restartAttempt variable.  For the latter case, the line no 445 shouldn't be using retryCounter.shouldRetry() method.

How was this change unit tested?

dcs/src/main/java/org/trafodion/dcs/server/ServerManager.java
```diff
@@ -400,7 +415,9 @@ public Boolean call() throws Exception {
             // /bin/dcs-daemon script
             // which DOES NOT set childServers count.
             for (int childInstance = 1; childInstance <= childServers; childInstance++) {
-                serverHandlers[childInstance-1] = new ServerHandler(childInstance);
+                ServerHandler serverHandler = new ServerHandler(childInstance);
+                serverHandler.setRestartAttempts(this.restartAttempts);
```
kevinxu021: suggest to add a new constructor new ServerHandler(this.conf, childInstance), then if you add more KVs in the future, the only changes may from ServerHandler

dcs/src/main/java/org/trafodion/dcs/server/ServerManager.java
kevinxu021: Agree with Selva, please keep retryCounterFactory  and get rid of new retry logic.

dcs/src/main/java/org/trafodion/dcs/server/ServerManager.java
```diff
@@ -414,32 +431,40 @@ public Boolean call() throws Exception {
...
-                   serverHandlers[childInstance-1] = new ServerHandler(childInstance);
-                   completionService.submit(serverHandlers[childInstance-1]);
+                        if (!retryCounter.shouldRetry()) {
+                            throw new IOException("Node " + nid + " is not Up");
+                        } else {
+                            retryCounter.sleepUntilNextRetry();
+                            retryCounter.useRetry();
+                        }
+                    }
+                    if (isRunning) {
```
kevinxu021: Merge the logic with line 444

dcs/src/main/java/org/trafodion/dcs/server/ServerManager.java
```diff
@@ -414,32 +431,40 @@ public Boolean call() throws Exception {
...
+                        restartAttempts = this.restartAttempts;
+                        LOG.info("mxosrvr " + nid + "," + pid + " still running");
+                    }
+                    else {
+                        restartAttempts--;
+                        LOG.info("mxosrvr " + nid + "," + pid + " exited, restarting, restart time : " + restartAttempts);
+                    }
+                    if (restartAttempts > 0) {
```
kevinxu021: Merge the logic with line 444

PR #1146, opened 2017-06-26
[TRAFODION-2661] Improve MDAM costing for RangeSpecs

DaveBirdsall:
JIRA TRAFODION-1641 fixed an issue with MDAM costing, where MDAM costing was not accounting for the cost of recursion. That is, it was not accounting for the accumulated costs of probes as we recurse through a set of key columns. When analyzing that JIRA, I uncovered an issue with RangeSpec costing which I partially fixed.

JIRA TRAFODION-2655 tuned the above fix further, increasing the weight of the cost of accumulated probes. So for plans lacking RangeSpecs, or plans whose RangeSpecs lack ORs, MDAM costing should now be doing a much better job of picking prefixes when it should.

This fix addresses the last bit of the problem by addressing RangeSpecs. There was some code that was forcing MDAM plans to traverse all key columns when a RangeSpec containing an OR was present. I have simply deleted that code.

I do not know why that code was originally there. It does predate the fixes noted above. I will speculate that before that change, we were picking MDAM in some cases where we should not, and adding that bit of code increased the total cost of MDAM to such a point that we no longer did.

### DIFF
--- a/core/sql/common/str.cpp
+++ b/core/sql/common/str.cpp
@@ -1620,9 +1620,11 @@ Int32 str_convertToHexAscii(const char * src,               // in
 //    char * dataPointer = getDataPointer();
 //    Lng32 len = tupp_.getAllocatedSize();
 //
-//    printBrief(dataPointer, len) 
+//    printBrief(dataPointer, len) if you want an end of line
+//
+//    printBrief(dataPointer, len, FALSE) if you don't 
 //    
-void printBrief(char* dataPointer, Lng32 len) 
+void printBrief(char* dataPointer, Lng32 len, NABoolean endLine) 
 {
    // We don't know what the data type is, but we do know how
    // long the field is. So we will guess the data type.
@@ -1771,7 +1773,7 @@ void printBrief(char* dataPointer, Lng32 len)
        }
      }    
    cout << local;
-   // cout << *(Lng32 *)dataPointer;
-   // End test change
+   if (endLine)
+     cout << endl;
  }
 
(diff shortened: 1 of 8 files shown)


DaveBirdsall:
Please do not merge this change yet. I am running full regressions on a workstation overnight with this change.

DaveBirdsall:
@zellerh, you raise a good point. I should do a bit more due diligence on this. Please do not merge yet.

DaveBirdsall:
Full regressions passed overnight. But please do not merge yet; I want to do some due diligence on @zellerh's remarks.

DaveBirdsall:
I will be submitting additional changes for this pull request, so please do not merge.

DaveBirdsall:
Notes about the rework for [TRAFODION-2661]:
    
    This rework fixes a bug in the pre-code-gen time code. The bug was
    that in the presence of partitioning key predicates, logic that
    communicates stop column information to the generator was being
    bypassed. This would result in traversal to the last key predicate
    at run time, with poor performance if the stop column was prior to
    the last key predicate.
    
    Also included in this rework are some tweaks to the run-time debug
    code for MDAM (the code had become non-functional from not being used
    for a long time), and a fix to a run-time memory leak bug that shows
    up as an assert when that debug code is present.

DaveBirdsall:
I will be running full regressions on this change overnight on a workstation, so please do not merge until I give the "all clear" on that.

zellerh:
+1
Looks good to me.

DaveBirdsall:
The test failure was in seabase/TEST010, which I believe has been separately addressed by https://github.com/apache/incubator-trafodion/pull/1161.

DaveBirdsall:
I have run this through the full regression suite on a workstation instance and it passes (apart from a couple of false negatives). This can be merged now.

PR #1152, opened 2017-06-28
[TRAFODION-2658] pyinstaller - support elastic adding nodes

mkby:
init version of pyinstaller elastic feature.


core/sqf/sql/scripts/sqregen
```diff
@@ -41,8 +41,15 @@ function Usage {
 
 function RegenConfig {
     echo "Regenerating Trafodion configuration: $SQCONFIG_FILE"
+    cd $TRAF_HOME/sql/scripts
     ./regensq.pl $SQCONFIG_FILE
-    exit $?;
+    if [ $? != 0 ]; then
+        echo "Failed to run regensq.pl."
```
DaveBirdsall: The error message is incorrect. We did run regensq.pl. The problem is that regensq.pl failed.

core/sqf/sql/scripts/sqregen
```diff
@@ -41,8 +41,15 @@ function Usage {
...
+        exit $?
+    fi
+    echo "Copying Trafodion configuration file $SQCONFIG_FILE to all nodes"
+    pdcp `trafconf -wname` $SQCONFIG_FILE $TRAF_HOME/sql/scripts/
```
DaveBirdsall: Should there be an error message if pdcp fails?

core/sqf/sql/scripts/sqregen
mkby: sounds good, I will change the message, thanks.

core/sqf/sql/scripts/sqregen
mkby: I think pdcp itself will show the error message if fails

mkby:
jenkins, retest

PR #1154, opened 2017-06-28
[TRAFODION-2673] Improve incremental update stats performance

DaveBirdsall:
This change improves the performance of incremental UPDATE STATISTICS by using "WITH NO ROLLBACK" when deleting rows from the persistent sample table.

A CQD, USTAT_DELETE_NO_ROLLBACK, has been added. When set to 'ON' (the default), we use DELETE WITH NO ROLLBACK on the persistent sample table. When set to 'OFF', we do the old behavior, that is, a vanilla transactional DELETE. This CQD is here mostly in order to test and verify the performance improvement. It can be removed later.

While I was at it, I noticed another CQD that is no longer used anywhere and removed it.


zellerh:
+1
Looks good.

PR #1155, opened 2017-06-29
[TRAFODION-2664] Instance will be down when the zookeeper on name node has been down

mkby:
do retry on every 'zoo_exists' call when we get a ZCONNECTIONLOSS failure.
@zcorrea , could you have to review the changes?


mkby:
I think this quick fix will only handle zookeeper HA failure(e.g. 1 of 3 ZK server is down), so it will retry immediately to get another available ZK server.   I'm not sure if 60 seconds retry wait will cause some other issues, may need more time to investigate.

zcorrea:
Got it! Thanks.

PR #1157, opened 2017-06-30
[TRAFODION-2675] Fix sqlci core dump issue.

andyyangcn:


sureshsubbiah:
jenkins, retest

zellerh:
Oops, sorry, failed to look at the checkin test. I hope it will continue to run and display results.

zellerh:
There is actually a problem in this code. The information we are adding contains NUL bytes, which act as a string terminator. The old code would stop at those bytes, while the new code includes them. Note that it's not good that we have binary data and try to put that into the EXPLAIN, but the code seems to handle the characters other than NULs in an ok way, at least those that are showing up in the explain. I'll be submitting a fix soon - given that I merged a change that broke the test.

PR #1159, opened 2017-07-02
[TRAFODION-2674] Create index sees error 2006 internal assertion

sureshsubbiah:
RangePartitionFunction was missing region boundaries for indexes and using table's
boundaries incorrectly
Turned OFF OPT_PCODE_CACHE to avoid mem leak.
Unrelated testware change for compGeneral/TEST023


core/sql/sqlcomp/nadefaults.cpp
```diff
@@ -2839,7 +2839,7 @@ SDDkwd__(ISO_MAPPING,           (char *)SQLCHARSETSTRING_ISO88591),
   DD_____(PCODE_DEBUG_LOGDIR,        ""  ), // Pathname of log directory for PCode work
   DDint__(PCODE_EXPR_CACHE_CMP_ONLY, "0" ), // PCode Expr Cache compare-only mode
   DDint__(PCODE_EXPR_CACHE_DEBUG,    "0" ), // PCode Expr Cache debug (set to 1 to enable dbg logging)
-  DDint__(PCODE_EXPR_CACHE_ENABLED,  "1" ), // PCode Expr Cache Enabled (set to 0 to disable the cache)
+  DDint__(PCODE_EXPR_CACHE_ENABLED,  "0" ), // PCode Expr Cache Enabled (set to 0 to disable the cache)
```
DaveBirdsall: Is the idea here that there is a memory leak in the PCODE cache that we intend to track down later?

sureshsubbiah: Yes @DaveBirdsall . JIRA 2179 tracks this problem.  I am sorry for not explaining this in the checkin message.

sureshsubbiah: On EXPECTED023, I was trying to change the text of an error message which incorrectly had the default schema as SEABASE instead of SCH. But I think that change had already been made (between the time I checked out and created this PR). Instead of my actual change, these DIFFs now show all the timestamp differences that are likely handled by the FILTER? I was not aware of the issue  in HSTable::makeSampleName. Thank you for explaining.

DaveBirdsall: Right. That was my screw-up. I had run compGeneral/TEST023 without running core/TEST000 first, which sets the default schema. I had fixed that issue, but the non-determinism in sample table name still causes random failures due to the line wrap issue.

PR #1160, opened 2017-07-03
[TRAFODION-2676]remove log redirection lines

kevinxu021:
When the user is using T2, the lines of the log before getConnection will be printed on console, while after getConnection the lines redirected into trafodion.sql.java.log. In another word, the caller should take responsibility of the direction of logs in stead of redirecting by default. Usually, it's on console by default, or the caller will specify the log file themselves( here the logs in HBaseClient.java even could be printing out) .

### DIFF
--- a/core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
+++ b/core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
@@ -150,18 +150,6 @@ public class HBaseClient {
     }
 
     static {
-    	//Some clients of this class e.g., DcsServer/JdbcT2 
-    	//want to use use their own log4j.properties file instead
-    	//of the lo4j.hdf.config so they can see their
-    	//log events in their own log files or console.
-    	//So, check for alternate log4j.properties otherwise
-    	//use the default HBaseClient config.
-    	String confFile = System.getProperty("trafodion.log4j.configFile");
-    	if (confFile == null) {
-    		System.setProperty("trafodion.sql.log", System.getenv("TRAF_HOME") + "/logs/trafodion.sql.java.log");
-    		confFile = System.getenv("TRAF_CONF") + "/log4j.sql.config";
-    	}
-    	PropertyConfigurator.configure(confFile);
         config = TrafConfiguration.create(TrafConfiguration.HBASE_CONF);
     }
 


ey4295:
@selvaganesang if we move the deleted lines to org.trafodion.sql( the caller), it should work as well. Then we don't have to remember the specific settings. There is a default property name 'log4fFile' for log4j, why not using existing one which is well-known. No matter what the solution is should be decided by the callers.

kevinxu021:
@selvaganesang if we move the deleted lines to org.trafodion.sql( the caller), it should work as well. Then we don't have to remember the specific settings. There is a default property name 'log4fFile' for log4j, why not using existing one which is well-known, that we do nothing because it covered by log4j lib internally.

selvaganesang:
That's correct.  But we want to control the location of the log files to be $TRAF_HOME/logs directory and the setup the configuration for the logs to enable us to use the standard log browser and other tools to look at the log files. If the T2 driver application needs to have its own log files, it can be configured in the application layer  but I wouldn't recommend changing the configuration of org.trafodion.sql and other C++ log entries via log4j to be modified. If the application needs to see these logs in their log, it can do so by enabling the appender in our configuration. In that case, I would propose a separate configuration files file for T2 driver and turn the appender on for this configuration so that the entries are logged both in the log file pointed by Trafodion configuration and the application configuration.

-1 for the change in the current form

kevinxu021:
@selvaganesang, do please take a look https://logging.apache.org/log4j/2.x/manual/configuration.html. if we want to specify the log file, at least 3 ways in stead of hardcoded. It's harmful for the current code because if you run your t2 process, you should know the logs before getConnection is not in the log file(trafodion.sql.java.log), and it's error level by default. if we change it to debug, all the other logs useless details will be printed because it's a shared config file. I was ask time and time again. Where is my logs after getConnection. DCS is using T2, it has its own log, but it was redirected trafodion.sql.java.log. true, we can add the VAR as you said, but if we want to change the settings like level, every time we have to remember to change 2 config files, one is in dcs while another is in $TRAF_HOME/conf. here the problem comes, $TRAF_HOME/conf is a shared config folder which will make all related modules taking effective. there will be a lot of useless logs while we only need the logs triggered by DCS. SPJ/Phoniex test/other potential project are all the other examples.

selvaganesang:
Trafodion SQL engine is a library and hence it can't assume that the caller application could have enabled the log4j or log4cxx. It needs to configure and decide its level of logging and the location of the files.  DCS being Trafodion component, is it possible for the DCS components to use the config file and the log directory like the rest of the other components instead of having its own config directory and the log directory.

My earlier comment applies to generic T2 application.

kevinxu021:
it should be more reasonable to use existing variable, why not? I don't understand here. For log4j, again, it has other solutions instead of hardcode. Yes, it's true, we can move DCS logs into the directory as the others. Do you mean all the logs should go into one file( trafodion.sql.java.log)? Possible, but not very good maybe.  For SPJ we do the same thing? More and more programs are using T2. Every time the issues occur, we have to tell them that there is an variable with all other logs useless for your program. And another problem i was experienced, if i don't use log4j in my program, expected printing on console, the logs after getConnection still being in trafodion.sql.java.log. OpenTSD collector is also using T2, do you mean we have to add log4j config file? Collectors expected console output. I would suggest to use existing variable instead of a new one. whatever you create jvm by c++ or not, the settings could be configured in CLASSPATH, not hard coded. I dont agree to hardcode only.

venkat1m:
I'm missing something here. When you run a T2 application, the T2 jdbc, executor and compiler run within SQL processes.  The T2 client application is a java client and everything else is treated as server code. The java client application can log or trace any exception returned by the T2 calls. But the T2 internal commands are logged in the sql log file. If you use a HBase java client analogy, when you write a Java client that uses Regionserver APIs. the regionserver logs go into the hbase region server log but any exceptions thrown can be traced and logged on the Java Client side. The logging or tracing on the client side is entirely optional. This is true for T4 connections too. All errors on mxosrvr side are logged to the common trafodion sql log.

Again, it seems there is a disconnect here or I'm missing something in the interpretation.

selvaganesang:
It looks like you want the application writers to control where the log goes even when the log entry comes from Trafodion engine.  By redirecting the log entry to application logs, would make the trouble shooting of Trafodion unmanageable. Hence, I would always recommend that the trafodion log entries to be logged in its own log file.  

If you want to fall back of creating a separate log file  for each master, please refer to PR  1070 
 to use an environment variable TRAF_MULTIPLE_SQL_LOG_FILE

kevinxu021:
Fine for Selva's explanation, if there is no redirection things.

PR #1162, opened 2017-07-03
[TRAFODION-2662] Remove some non-determinism from compGeneral/TEST023

DaveBirdsall:
This is yet another attempt to fix compGeneral/TEST023. The sample table names generated for persistent sample tables could vary in length, as they were functions of object UIDs and timestamps. The changes here pad these numbers with leading zeroes out to a fixed length, which should eliminate variability in line wrap in the test results. While I was at it, I found a bug: there was a buffer overrun in this code (my bad); the buffer is now the proper length.


DaveBirdsall:
The test failure was in seabase/TEST010, which I believe has been addressed separately by https://github.com/apache/incubator-trafodion/pull/1161.

PR #1163, opened 2017-07-04
[TRAFODION-2622] Left join with other_join_predicate on table alone i…

sureshsubbiah:
…s slow

This change is due to Hans Zeller.
The other join predicate in a hash join is evaluated before the hash join,
so that failing rows can be nulll instantiated directly. This avoids
a temporary join explosion when the join predicate matches most rows but
the other_join_predicate fails most rows. This change will be extended
to merge joins later.

### DIFF
--- a/core/sql/generator/GenPreCode.cpp
+++ b/core/sql/generator/GenPreCode.cpp
@@ -2459,6 +2459,24 @@ RelExpr * Join::preCodeGen(Generator * generator,
 
         }
 
+      // if we have both equi join preds and a beforejoin pred
+      // Set a flag that will cause beforeJoinPred to be evaluated prior
+      // join equi pred during execution. This helps with join explosion
+      // if there are frequent matching values and the beforeJoinPred is
+      // highly selective. There is no downside to evaluating beforeJoinPred
+      // early, if it contains vids from outer only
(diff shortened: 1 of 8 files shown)


core/sql/generator/GenPreCode.cpp
```diff
@@ -2459,6 +2459,24 @@ RelExpr * Join::preCodeGen(Generator * generator,
...
+      // if we have both equi join preds and a beforejoin pred
+      // Set a flag that will cause beforeJoinPred to be evaluated prior
+      // join equi pred during execution. This helps with join explosion
+      // if there are frequent matching values and the beforeJoinPred is
+      // highly selective. There is no downside to evaluating beforeJoinPred
+      // early, if it contains vids from outer only
+      if (!(getEquiJoinPredicates().isEmpty() || getJoinPred().isEmpty() || 
+	    isAntiSemiJoin()))
+      {
+	ValueIdSet dummy1, dummy2, dummy3, uncoveredPreds ;
```
zellerh: Just for a future checkin, I think dummy3 could be removed.

sureshsubbiah: Thank you @zellerh. Will do.

PR #1165, opened 2017-07-05
[TRAFODION-2677] odb load source with src=-file is failure

SuJinpei:
root cause: each thread try to call splitcso to get Catalog/Schema/Table from the same tgt buffer, but splitcso will modify tgt buffer, so only one thread can get the correct result.
fix way: 1. Catalog/Schema/Table can be determined when creating etab,  so move it forward.
             2. etab[*].tgt is used elsewhere and is assumed not split, so make a copy of tgt.
             3. add globalPointers to save pointers to buffers which may be shared by many thread and length determined at run time.


sureshsubbiah:
+1

xwq:
+1

PR #1166, opened 2017-07-06
[TRAFODION-2679] odb crash when extracting data to xml file

SuJinpei:
root cause: write tag to file which had been closed.
fix by: don't close file until tag writing finished.


xwq:
It's good to me.

PR #1167, opened 2017-07-06
[TRAFODION-2308]JDBC support insert LOB

xwq:


core/sql/exp/ExpLOBinterface.cpp
```diff
@@ -728,9 +728,9 @@ Lng32 ExpLOBInterfaceUpdate(void * exLobGlob,
   Ex_Lob_Error status;
   Int64 cliError = -1;
   Int64 sourceLen = 0;
-  if((so == Lob_Memory) || (so == Lob_External_File))
+  if(so == Lob_External_File)
     sourceLen = strlen(srcLobData);
-  else if (so == Lob_Buffer)
+  else if ((so == Lob_Buffer) || (so == Lob_Memory))
```
sandhyasun: For Lob_Memory,  for eg stringtolob() , won't this break since we don't have the target length ?

core/sql/optimizer/BindItemExpr.cpp
```diff
@@ -6105,12 +6105,21 @@ ItemExpr *Assign::bindNode(BindWA *bindWA)
...
+                         CmpCommon::getDefaultNumeric(LOB_BATCH_SIZE));
+              }
+              else {
+              newType = new SQLBlob((CmpCommon::getDefaultNumeric(LOB_MAX_SIZE)*1024*1024),
                                              lobType.getLobStorage(), 
                                              TRUE, FALSE, TRUE, 
-                                             CmpCommon::getDefaultNumeric(LOB_MAX_CHUNK_MEM_SIZE)*1024*1024); 
-              //              vid1.coerceType(desiredType, NA_LOB_TYPE); 
+                                             CmpCommon::getDefaultNumeric(LOB_BATCH_SIZE));
+              }
```
sandhyasun: For if a param is passed in for LOB, then are we limiting it to 4096 bytes now  for the input ?

core/sql/parser/sqlparser.y
```diff
@@ -11826,11 +11827,40 @@ blob_optional_left_len_right: '(' NUMERIC_LITERAL_EXACT_NO_SCALE optional_blob_u
...
+	    {
+	      $$ = (Int64)100000;
+	    }
+	  else
+	    {
+	      $$ = (Int64)CmpCommon::getDefaultNumeric(LOB_MAX_SIZE)*1024*1024;
+
+	    }
+        }
+
```
sandhyasun: Did you run TESTTOK2 ? Perhaps better to run extra tests. I'll trigger those.

sandhyasun:
jenkins extra test

core/sql/exp/ExpLOBinterface.cpp
xwq: I have run the regression test in sql/regress, it passed.  Here the problem is the CLOB will be truncate by '\0'.  Some charset, it uses 2 or more bytes for one character, one of the byte may be '\0'.   It seems the tgtLobLen is the real length. Could you help me to check it and give me some suggestions.

core/sql/parser/sqlparser.y
xwq: I am not run TESTTOK2, thank you for running the extra tests.

core/sql/exp/ExpLOBinterface.cpp
sandhyasun: Let me try a few cases and check.

core/sql/optimizer/BindItemExpr.cpp
xwq: I think it's better to compare the CQD and this 'param'. If the param is larger, I will use CQD, else use the param.  I will change it.

sandhyasun: But will this batch size affect non-batch inputs too ? For eg, if a statement is like " insert into tlob values (1,?) "    Then the data passed in for ? should not be limited to 4096 for the non batch cases. Can you test and make sure of that ?

core/sql/exp/ExpLOBinterface.cpp
sandhyasun: Yes this looks correct. I tried it out.  But you do need to mak ea similar change in the interface ExpLOBInterfaceUpdateAppend

sandhyasun:
Looks like the tests passed. Just need to  run that requested test above : non rowset insert using parameter and provide input > 4K - i.e something like  2 or 3 G.

xwq:
jenkins, retest.

xwq:
jenkins, extra test

xwq:
jenkins, extra test

sandhyasun:
The Etra test failure in executor/TEST106 is unrelated to this change . It needs to be investigated separately - seems like a timing related failure on the VM on which it's running. This PR seemes ready to be merged now since the functionality it touches is TEST130 and that is passing fine.

PR #1169, opened 2017-07-06
[TRAFODION-2680] Add CSV_FORMAT function to Trafodion SQL

DaveBirdsall:
This change adds a new function, CSV_FORMAT, to Trafodion SQL.

Syntax: CSV_FORMAT(argument1, ... ,argumentn)

Semantics: The result of CSV_FORMAT is a VARCHAR string consisting of a comma-separated list of the argument values. If the argument itself contains a comma, it is surrounded with double quotes.

This is a prototype for evaluation. For example, it does not escape double quotes within a double quoted string; the need for that is to be determined. Also, a regression test for this needs to be added.

### DIFF
--- a/core/sql/optimizer/BindItemExpr.cpp
+++ b/core/sql/optimizer/BindItemExpr.cpp
@@ -10851,6 +10851,96 @@ ItemExpr *ZZZBinderFunction::bindNode(BindWA *bindWA)
       }
     break;
 
+    case ITM_CSV_FORMAT:
+      {
+	bindChildren(bindWA);
+	if (bindWA->errStatus()) 
+	  return this;
+
+        // The way the arguments of CSV_FORMAT are represented in
+        // the parse tree is as a tree of ItemList nodes; so
+        // CSV_FORMAT(a,b,c,d) is represented as
+        //
+        //    this
+        //    /  \
+        //   a   ItemList
+        //         /  \
+        //        b   ItemList
+        //              /  \
+        //             c    d
+        //
+        // The code below traverses accordingly.
+       
+        ItemExpr * next = child(0)->castToItemExpr();     
+        ItemExpr * resultBoundTree = NULL;
+        while (next)  // while arguments remain to process
+          {
+            ItemExpr * childk = NULL;
+            if (next->getOperatorType() == ITM_ITEM_LIST)
+              {
+                childk = next->child(0);
+                next = next->child(1);
+              }
+            else
+              {
+                childk = next;
+                next = NULL;
+              }
+           
+            const NAType &type = childk->getValueId().getType();
+            switch (type.getTypeQualifier())
+              {
+                case NA_BOOLEAN_TYPE:
+                case NA_DATETIME_TYPE:
+                case NA_INTERVAL_TYPE:
+                case NA_NUMERIC_TYPE:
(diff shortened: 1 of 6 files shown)


core/sql/optimizer/BindItemExpr.cpp
```diff
@@ -10851,6 +10851,96 @@ ItemExpr *ZZZBinderFunction::bindNode(BindWA *bindWA)
...
+                    // operand has an unsupported data type
+                    *CmpCommon::diags() << DgSqlCode(-4018)
+                                        << DgString0("CSV_FORMAT")
+                                        << DgString1(type.getTypeName().data());
+                    bindWA->setErrStatus();
+                    return this;
+                  }
+              }
+
+            strcpy(buf,"CASE WHEN POSITION(',' IN @A1) > 0 THEN '\"' || @A1 || '\"' ELSE @A1 END");
```
zellerh: You mention the current lack of handling of embedded double quotes. Those could be handled by adding a REPLACE(@A1, '"', '""'). Also, since neither commas nor double quotes occur in booleans, datetimes, intervals, or numerics, would it be ok to apply this transformation on character types only?

core/sql/optimizer/BindItemExpr.cpp
```diff
@@ -10851,6 +10851,96 @@ ItemExpr *ZZZBinderFunction::bindNode(BindWA *bindWA)
...
+        //
+        //    this
+        //    /  \
+        //   a   ItemList
+        //         /  \
+        //        b   ItemList
+        //              /  \
+        //             c    d
+        //
+        // The code below traverses accordingly.
```
zellerh: There is a class ItemExprList that can convert such trees into a collection class. Other binder methods use that to traverse lists.

core/sql/optimizer/BindItemExpr.cpp
zellerh: One more thought: I think that in a CSV file, white space between fields is ignored. So, it may be best if this code adds a TRIM, so that the padding around fixed chars is not included.

core/sql/optimizer/BindItemExpr.cpp
DaveBirdsall: Thank you, Hans. Will use that.

core/sql/optimizer/BindItemExpr.cpp
DaveBirdsall: Re: booleans and so on. I think you're right. I need to check if we have a European form of CAST that uses a comma for the decimal point for numbers. Will look into this.

Re: double quotes: Thanks very much for the tip. I need to do a little experimentation with Excel spreadsheets to see how they behave with double quotes. Then I'll make changes accordingly.

Re: white space: I wondered about that. I suspect programs such as Excel do ignore trailing white space (not sure about leading). Seems like adding TRIM would be redundant in that case? There is an aesthetic case to be made though in getting rid of the extra white space in the output.

DaveBirdsall:
Please do not merge. I intend to take another approach (improving the CSV format feature in trafci). I will close this request when that approach is ready.

DaveBirdsall:
Closing this pull request. Will take another approach, as discussed in the JIRA.

PR #1170, opened 2017-07-06
[TRAFODION-2628] SQL engine work for Elasticity

zellerh:
It turns out that the compiler already kept a list of node ids,
not assuming that node ids are consecutively numbered 0...n-1.
However, this list was contained in lots of other data structures
that are dead code, coming from the time before Trafodion was
open-sourced. The change does two things: a) getting rid of obsolete
info and b) maintining a list of node ids in the NAClusterInfo
class.


core/sql/cli/Globals.cpp
```diff
@@ -187,18 +187,9 @@ void CliGlobals::init( NABoolean espProcess,
...
-  myNumCpus_ = 0;
-  SEGMENT_INFO * segs = new(&executorMemory_) SEGMENT_INFO[MAX_NO_OF_SEGMENTS];
-  ComRtGetSegsInfo(segs, MAX_NO_OF_SEGMENTS, myNumSegs_,
-           (NAHeap *)&executorMemory_);
-  for (Lng32 i = 0; i < myNumSegs_; i++)
-    {
-      myNumCpus_ += segs[i].noOfCpus_;
-    }
-  NADELETEARRAY(segs, MAX_NO_OF_SEGMENTS, SEGMENT_INFO, &executorMemory_);
+  myNumCpus_ = ComRtGetCPUArray(cpuArray_, (NAHeap *)&executorMemory_);
```
selvaganesang: Do you think it is good to assert when myNumCpus_ is zero

core/sql/optimizer/NAClusterInfo.cpp
```diff
@@ -350,14 +244,8 @@ NAClusterInfo::NAClusterInfo(CollHeap * heap)
         if ((nodeInfo[i].type & MS_Mon_ZoneType_Aggregation) != 0 ||
```
selvaganesang: Node types are not considered in ComRtUtils.cpp, but it is done here. Is that ok?

core/sql/cli/Globals.cpp
zellerh: Good question, what do you think? I'm a bit scared to assert, because that could make the whole system inaccessible. Could we log an error event instead? But I'm not sure whether that is possible at the time when we create the CLI globals, maybe the event log code is dependent on something that has not been created yet. I'll check and add either an assert or a log event.

core/sql/optimizer/NAClusterInfo.cpp
zellerh: Yes, I would say we should look at the node type. Here is what I found in file core/sqf/export/include/seabed/ms.h:

    MS_Mon_ZoneType_Undefined   = 0x0000,   // No zone type defined
    MS_Mon_ZoneType_Edge        = 0x0001,   // Zone of service only nodes
    MS_Mon_ZoneType_Aggregation = 0x0002,   // Zone of compute only nodes
    MS_Mon_ZoneType_Storage     = 0x0004,   // Zone of storage only nodes
    MS_Mon_ZoneType_Excluded    = 0x0010,   // Excluded cores

Maybe we should only look at nodes of type "MS_Mon_ZoneType_Aggregation", just like NAClusterInfo does it? If you agree I can make that change.

zellerh:
About the naming of cpuArray vs. nodeArray: The term "cpu" is used in many places in the code and it is fairly consistent and unambiguous. The term "node" is used too, but it is very ambiguous and in most cases it means something very different than what we think of today (NSK nodes). So, I try to stay away from the term "node" (except in comments, where I usually say "Linux node").

core/sql/optimizer/NAClusterInfo.cpp
selvaganesang: I would think Edge and Aggregation should be considered as the valid nodes. Please check with @zcorrea too

zellerh: From what I remember from the old times, edge nodes were not meant for ESPs, those should all go to aggregation nodes.

PR #1176, opened 2017-07-13
[TRAFODION-646] ODBC driver AppUnicodeType setting is not in DSN level

Weixin-Xu:
Originally, AppUnicodeType is set in [ODBC] section of odbc.ini, by environment variable or default utf8.
This change result in modifing AppUnicodeType on the base of dsn.


selvaganesang:
@venkat1m , @arvind-narain Can you please review this PR

PR #1178, opened 2017-07-13
[TRAFODION-2596] Improve the log4j and log4cxx infrastructure in Traf…

selvaganesang:
…odion

Removed setting the root logger to be the appender logger in most of
the configuration files. This would allow T2 driver applications
to have control over the logging in their layer to its preferred way.
But, the logging in SQL engine layer will be controlled by the Trafodion
engine itself.

Changed the code so that the logging location and the file name is
picked up from the config file.


selvaganesang:
jenkins, retest

selvaganesang:
jenkins, retest

zellerh:
+1
Looks good to me - with my limited understanding of log4cxx.

PR #1180, opened 2017-07-14
[TRAFODION-2685] int type doesnot check boundary

mashengchen:


mashengchen:
jenkins, retest

mashengchen:
jenkins, retest

core/conn/jdbcT4/src/main/java/org/trafodion/jdbc/t4/TrafT4ResultSet.java
```diff
@@ -1372,7 +1372,7 @@ public long getLong(int columnIndex) throws SQLException {
 						}
 					} else {
 						throw TrafT4Messages.createSQLException(connection_.props_, connection_.getLocale(),
-								"numeric_out_of_range", null);
+								"numeric_out_of_range", bd.longValue());
```
selvaganesang: Will the longValue convert the bd into  a proper long value? Should we display the original bd value instead?

mashengchen: agree ,i will change this

mashengchen:
they are the same result after the process, even not use toString

DaveBirdsall:
Is this change ready to merge?

selvaganesang:
On second look, there seems to be lot of unneeded code/confusion in the surround code. The error message that is supposed to contain the original data is not consistently done in all the error conditions. This change seems to take care a certain error message paths alone. 

So, I will let the author to decide if the code change needs to go in or out.  I am ok with either of the conclusions.

mashengchen:
this change is try to fix int type not check boundary which given by client, so the emphasis is in TrafT4PreparedStatement.java, check Integer Boundary when type is INTEGER (sign), and check Unsigned Integer Boundary when type is INTEGER UNSIGNED.
and The error message is try to let the user know what's the value they given in the request, this will help user quickly locate the issues

PR #1182, opened 2017-07-15
[TRAFODION-2688] Need a way to find queries that are stuck in compila…

selvaganesang:
…tion via RMS

When a query is stuck in compilation, the mxosrvr remains in connected state forever.
Connection idle timer doesn't kick in. The query timeout can't cancel this query
because the query hasn't start executing. There is a need to find the process id of the
mxosrvr(master) process that is stuck in compilation via RMS. Then the mxosrvr process can be killed.

To find the queries stuck in compile for more than 30 seconds

cd $TRAF_HOME/export/limited-support-tools/LSO
./offender -s queries_in_compile

This script lists queries that are being compiled for more than the given number of seconds

CURRENT_TIMESTAMP           TIME_IN_SECONDS       NODE         PID          QUERY_ID   `                                                                  SOURCE_TEXT
--------------------------  --------------------  -----------  -----------  ----------------------------------------------------------------------------  ----------------------

2017-07-15 01:54:14.909780                  1860            0         5505  MXID11000005505212366841736965000000000000206U3333300_278___SQLCI_DML_LAST__ "select * from partsupp ;


zellerh:
+1
Looks good to me. Thanks for adding that.

PR #1184, opened 2017-07-17

selvaganesang:
Added four new counters to provide information about multiple executes
of a prepared statement.

No. of times executed
Min. Execute Time
Max. Execute Time
Avg. Execute Time


zellerh:
+1
Looks good to me

PR #1185, opened 2017-07-18
[TRAFODION-2689] Fixed log file creation of WDG and PSD processes by

zcorrea:
fixing process creation logic to set TRAF_CONF and TRAF_VAR environment
variables in child processes in addition to TRAF_HOME.
Updated log4cxx configuration settings and removed obsolete
configuration files.
Changed logfile names by removing timestamp and pid from name. The only
exception is the use monitor pid in the file names when in virtual cluster
environment.


selvaganesang:
Looks good. You can consider to remove setting the rootLogger in both wdt and psd to keep it in sync with the rest of the config files of Trafodion. Also, set the immediateFlush=true to ensure that the logs are written.  I think Wdt and Psd logs are kept to a minimum and hence it shouldn't affect the performance

zcorrea:
jenkins, retest

zcorrea:
Not sure what you mean by removing the rootLogger in wdt and psd as all other config files are set the same way?

Zalo

selvaganesang:
Removed these lines in the config file
# Define the root logger to the system property "trafodion.root.logger".
 trafodion.root.logger=INFO, psdAppender		 trafodion.root.logger=INFO, psdAppender
 log4j.rootLogger=${trafodion.root.logger}		 log4j.rootLogger=${trafodion.root.logger}

and modify the line as
log4j.logger.WDG=INFO,psdAppender

And add
log4j.appender.psdAppender.immediateFlush=true

Need to do the similar changes with wdt config file too

zcorrea:
jenkins, retest

zcorrea:
jenkins, retest

PR #1189, opened 2017-07-28
[TRAFODION-2696] control query cancel qid fails with error 8031 somet…

selvaganesang:
…imes

With this change, ssmp will not be stopped when the query
is stuck at the time of close for more than 5 minutes. The ssmp is forced
to dump core only.

Fix for executor/TEST106 failure caused by PR 1188


zellerh:
+1
Do you need to do both, add a fix and also remove some of the tests?

selvaganesang:
@svarnau confirmed in the email message to dev dlist that core files tags are spurious.  The code change in executor/ex_root.cpp is for [TRAFODION-2696]. The code change in runtimestats/ssmpipc.cpp is to take care of potential boundary condition issue with nowait depth checking. executor/TEST106 and executor/EXPECTED106 changes are the regression test failure of TEST106

selvaganesang:
I would think there were no core files created by this change. @svarnau, can you please confirm it

svarnau:
Correct. The code that tags corefiles in the test summary is new and had bug.

PR #1192, opened 2017-07-31
[TRAFODION-2692] Fixed monitor startup when hostname is in various forms

zcorrea:
[TRAFODION-2001] Updated 'sqgen' to not generate script now supported by 'persist' commands.


core/sqf/monitor/linux/clio.cxx
```diff
@@ -377,6 +379,21 @@ Local_IO_To_Monitor::Local_IO_To_Monitor(int pv_pid)
         tmpptr++;
     }
 
+    // Remove the domain portion of the name if any
+    char str1[MPI_MAX_PROCESSOR_NAME];
```
DaveBirdsall: Is MPI_MAX_PROCESSOR_NAME the maximum name length? Or one more than the maximum name length? If the former, then la_node_name, la_short_node_name and str1 should be declared as char[MPI_MAX_PROCESSOR_NAME+1] to leave room for a trailing null. (The use of strcpy here made me think of that.)

core/sqf/monitor/linux/commaccept.cxx
```diff
@@ -33,6 +33,7 @@ using namespace std;
 #include <signal.h>
 #include <unistd.h>
 
+extern CCommAccept CommAccept;
```
DaveBirdsall: Global objects cause race conditions at exit time. Their destructors are called in non-deterministic order. Does this object depend on any other objects? Does its destructor traverse to any other objects? If so, then this will probably break at some time in the future.

core/sqf/monitor/linux/monitor.cxx
```diff
@@ -1089,6 +1090,22 @@ int main (int argc, char *argv[])
         tmpptr++;
     }
 
+    // Remove the domain portion of the name if any
```
DaveBirdsall: This is the third copy I've seen of code to remove the domain portion. Maybe they should be refactored into a separate function?

core/sqf/monitor/linux/cluster.cxx
```diff
@@ -749,7 +732,22 @@ void CCluster::HardNodeDown (int pnid, bool communicate_state)
...
+            // Remove the domain portion of the name if any
+            char my_node_name[MPI_MAX_PROCESSOR_NAME];
+            char str1[MPI_MAX_PROCESSOR_NAME];
+            memset( str1, 0, MPI_MAX_PROCESSOR_NAME );
+            strcpy (str1, node->GetName() );
+        
+            char *str1_dot = strchr( (char *) str1, '.' );
+            if ( str1_dot )
+            {
+                memcpy( my_node_name, str1, str1_dot - str1 );
```
zcorrea: Good catch! I missed one.

core/sqf/monitor/linux/clio.cxx
zcorrea: It's a larger buffer than needed, but I should look at all the uses of MPI_MAX_PROCESSOR_NAME. This is targeted for cleanup is MPI usage which is quite minimal in the current code base.

core/sqf/monitor/linux/commaccept.cxx
zcorrea: There are no dependencies or danger or danger of race conditions, but it is good to remember this.

core/sqf/monitor/linux/monitor.cxx
zcorrea: There are actually four places and unfortunately in the scope of different processes. I will keep this in mind as there is some utility code that could perhaps be refactored to contain functions like this.

PR #1196, opened 2017-08-03
[TRAFODION-2687] the value of Numeric Struct need not to be changed to BidEndian

Weixin-Xu:
For our server , we need not to swap value to big endian


xwq:
+1

PR #1201, opened 2017-08-11
[TRAFODION-2709] getpwuid is not a Thread-safe function

Weixin-Xu:
Using multi-threads app with linux-odbc to connect trafodion will make dcs down


xwq:
+1. 

Without this fix, sometimes, ODBC will send wrong message to DCS in multi threads.

selvaganesang:
I will merge this change in. Check-PR failed in jdbc_test-cdh which is a known failure

PR #1202, opened 2017-08-13
[TRAFODION-2703] Add *Load XML Files* for ODB User Guide

liuyu000:

1. Add **Load XML Files**

### DIFF
--- a/docs/odb_user/src/asciidoc/_chapters/load.adoc
+++ b/docs/odb_user/src/asciidoc/_chapters/load.adoc
@@ -455,7 +455,7 @@ JOB:FIXED:23:10               <- insert into JOB characters starting at position
 Load as follows:
 
 ```
-$ odb64luo –u user –p xx –d dsn \ 
+$ odb64luo –u user –p xx –d dsn \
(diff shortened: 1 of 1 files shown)


docs/odb_user/src/asciidoc/_chapters/load.adoc
```diff
@@ -637,6 +637,64 @@ odb considers the string following the “em” character as the path of the fil
 
 NOTE: odb does not load rows where the size of the input file is greater than the target database column.
 
+[[load_xml_files]]
+== Load XML Files
+Trafodion odb supports loading xml files into tables, the key construct for xml files can be an element or an attribute.
```
DaveBirdsall: XML is an acronym (eXtensible Markup Language), and so should be capitalized when it appears in prose. If it appears as a file extension (for example, xyz.xml) or as part of a code example then lower case is fine. Also, this appears to be a run-on sentence. Better perhaps would be "Trafodion odb supports loading XML files into tables. The key construct for XML files may be an element or an attribute."

liuyu000:
@DaveBirdsall 
Thanks for your detailed explanations!
Yes I should have used uppercase for acronyms. God is in the details. :)
Your comment has been incorporated.

PR #1203, opened 2017-08-14
[TRAFODION-2710] Add *Alter Column (Alter Table Statement)* for Trafodion SQL Reference Manual

liuyu000:

1. Add **Alter Column** for Alter Table Statement

2. Fix two typos:
- ISO8891→ISO8859-1
- DATEPART(text, datetime-expr)→DATE_PART(text, datetime-expr)


liuyu000:
@DaveBirdsall 
Thanks for your detailed explanations! Your comment has been incorporated. :)

- ISO88591
I've checked and corrected ISO8859-1 to ISO88591, since the occurrences of the latter are more than that of the former.

- UTF-8
All cases are UTF8 except the following one: 
> UTF8 (UTF-8) is a variable-length (1 to 4 bytes) encoding of Unicode characters including those in supplementary planes. It is compatible with the US-ASCII character set.

This is an appropriate usage, so no changes have been made on this.

PR #1206, opened 2017-08-14
[TRAFODION-2617] Optionally push estimate row count logic into HBase coprocessor

DaveBirdsall:
When HBase encryption is used, client-side logic to estimate row count in encrypted Trafodion tables fails because the Trafodion ID lacks permission to read the HBase key store file. This set of changes pushes the row count logic down into a coprocessor, so the logic can run under the HBase ID.

The feature is turned on via CQD HBASE_ESTIMATE_ROW_COUNT_VIA_COPROCESSOR 'ON'. The default is set to 'OFF'. There is a small performance overhead when using the coprocessor, so it is best to leave this set to 'OFF' unless HBase encryption is employed.

Note that this change is not sufficient to support encrypted Trafodion tables using HBase encryption. In order to specify at create time that a Trafodion table should be encrypted, one needs an extension to the DDL HBASE_OPTIONS (e.g., "ENCRYPTION = 'AES'"). This has not yet been implemented.

Nevertheless, one can still test these changes on a Trafodion instance by using the HBase shell. After creating a table within Trafodion, one can use the 'alter' command in the HBase shell to turn encryption on. For example: alter "TRAFODION.SEABASE.T1",NAME=>'#1',ENCRYPTION=>'AES'. 

Other pitfalls to watch out for: Configure your instance for HBase encryption as described in the HBase documentation. In addition, set the hfile.format.version value to three. (In Trafodion, by default it is two.) Oddly, HBase allows one to turn on encryption for format two tables, but it does not work; you get file corruption errors.


DaveBirdsall:
Release build failed due to a glitch in obtaining Java artifacts. Will retest.

DaveBirdsall:
jenkins, retest

DaveBirdsall:
@sbroeder, could you take a look at these changes (especially the stuff I put into the DTM files)?

core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
```diff
@@ -1395,6 +1422,169 @@ private boolean estimateRowCountBody(String tblName, int partialRowSize,
...
+    
+      Map<byte[], TrafEstimateRowCountResponse> result = null;
+      try {
+        result = table.coprocessorService(TrxRegionService.class, null, null, callable);
+      } catch (Throwable e) {
+        throw new IOException("Exception from coprocessorService caught in estimateRowCountViaCoprocessor",e);
+      }      
+
+      for (TrafEstimateRowCountResponse response : result.values()) {
+        boolean hasException = response.getHasException();
```
sbroeder: Question:  would it be better to send the coprocessor request only to the first region rather than broadcast to all regions?

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/TrxRegionEndpoint.java.tmpl
```diff
@@ -7335,6 +7338,245 @@ TrxTransactionState isTransactionOnCommittedList(final long transactionId)
...
+      // we'll just use an empty list instead
+      storeFileList = new ArrayList<String>();
+    }
+
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("Trafodion estimate row count sees " + storeFileList.size() + " files.");
+      for (String sfn : storeFileList) {
+        LOG.debug("*** " + sfn);
+      }
+      if (startKey == null)
```
DaveBirdsall: That's what I found as well during unit test. I saw a non-null startkey with a length of zero. Since this is debugging code, though, I wrote it to take into account both possibilities. (In fact, this debug code is how I found out that you are right!)

core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
DaveBirdsall: The code as written adds up the sizes of all the HFiles. I suppose we could assume that all regions are the same size and just sample the first one. That might not be safe, though; one region might have split while another hasn't yet.

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/TrxRegionEndpoint.java.tmpl
```diff
@@ -7335,6 +7338,245 @@ TrxTransactionState isTransactionOnCommittedList(final long transactionId)
...
+              // adjust counts for that
+              putKVsSampled--;
+              sampleRowCount--;
+            }
+          }  // code for first file
+        } catch (IOException exp1) {
+          if (LOG.isErrorEnabled()) LOG.error("Trafodion estimate row count encountered exception ", exp1);
+          t = exp1;
+        } finally {
+          reader.close(false);
```
DaveBirdsall: Good catch... I'll check on it.

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/TrxRegionEndpoint.java.tmpl
```diff
@@ -7335,6 +7338,245 @@ TrxTransactionState isTransactionOnCommittedList(final long transactionId)
...
+
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("Trafodion estimate row count sees " + storeFileList.size() + " files.");
+      for (String sfn : storeFileList) {
+        LOG.debug("*** " + sfn);
+      }
+      if (startKey == null)
+        LOG.debug("startKey is null.");
+      else
+        LOG.debug("startKey.length is " + startKey.length + ", startKey is hex " + bytesToHex(startKey));
```
DaveBirdsall: No, because we are still getting entry and byte counts from each of the HFiles no matter what region we are reading.

core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
```diff
@@ -1395,6 +1422,169 @@ private boolean estimateRowCountBody(String tblName, int partialRowSize,
...
+            }
+            return response;        
+          }
+      };
+    
+      Map<byte[], TrafEstimateRowCountResponse> result = null;
+      try {
+        result = table.coprocessorService(TrxRegionService.class, null, null, callable);
+      } catch (Throwable e) {
+        throw new IOException("Exception from coprocessorService caught in estimateRowCountViaCoprocessor",e);
```
selvaganesang: It is good to throw exception, but I think that the estimate earlier returned some default values if there is any issue with getting estimates.  It would be better to retain the same semantics to avoid surprises.

core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
```diff
@@ -1395,6 +1422,169 @@ private boolean estimateRowCountBody(String tblName, int partialRowSize,
...
+        logger.debug("HBaseClient.estimateRowCountViaCoprocessor(" + tblName + ") called.");
+        logger.debug("numCols = " + numCols + ", partialRowSize = " + partialRowSize);
+      }
+
+      boolean retcode = true; 
+      rc[0] = 0;
+
+      HConnection connection = null;
+      HTableInterface table = null; 
+      connection = HConnectionManager.createConnection(config);
```
selvaganesang: Can you use getConnection instead. It ensures that there is only one connection per process

core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
DaveBirdsall: I've kept the same error-handling semantics as in the existing code. Higher layers do convert the error into a default value.

core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
DaveBirdsall: Will do. Thanks.

core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
sbroeder: Yes, good point.

core/sql/sqlcomp/nadefaults.cpp
```diff
@@ -1759,6 +1763,7 @@ SDDkwd__(EXE_DIAGNOSTIC_EVENTS,		"OFF"),
  // exposure.
  DDkwd__(HBASE_DELETE_COSTING,		             "ON"),
  DDflt0_(HBASE_DOP_PARALLEL_SCANNER,             "0."),
+ DDkwd__(HBASE_ESTIMATE_ROW_COUNT_VIA_COPROCESSOR,   "OFF"),
```
selvaganesang: Don't you think it should be on by default?

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/TrxRegionEndpoint.java.tmpl
DaveBirdsall: HFileScanner doesn't have a close method until HBase 2.0 unfortunately. In the rework, though, I'll add a comment.

core/sql/sqlcomp/nadefaults.cpp
DaveBirdsall: No. In my tests, there was a small performance penalty for using the coprocessor. We must have the coprocessor if HBase encryption is being used, but otherwise we can just use the client code.

Your next question might be: why is there a performance penalty? I looked into that a little bit. I think it is because we do dynamic coprocessor loading. It may be that the penalty is not real: this coprocessor is packaged with the DTM coprocessors. I think they all get loaded together. Assuming this is true, we are simply paying this penalty earlier (at estimate row count time) rather than at transaction start time.

selvaganesang: I like the co-processor approach because it encapsulates the HFile to region server code and enables SQL to use well defined and externalized  APIs of HBase. From this perspective, it should be good to remove the old estimate code entirely even with small performance penalty. This would help to maintain the code better with future releases of hbase

DaveBirdsall: I can buy that. May I suggest that we leave the old code there for the moment until this can be thoroughly exercised? I can remove the old code in a later check-in.

DaveBirdsall:
@sbroeder and @selvaganesang, I've pushed a rework. In the coprocessor, I tried adding a scanner.close(), but it turns out the close() API is not available for HFileInterface until HBase 2.0. So I put a comment there for now. In HBaseClient.java, I replaced the use of the deprecated HConnection with the more current Connection, as you suggested Selva.

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/TrxRegionEndpoint.java.tmpl
```diff
@@ -7335,6 +7338,245 @@ TrxTransactionState isTransactionOnCommittedList(final long transactionId)
...
+          if (LOG.isDebugEnabled()) LOG.debug("Trafodion estimate row count file name " + path + " is not an HFile.");
+          continue;
+        }
+        HFile.Reader reader = HFile.createReader(fileSystem, path, cacheConf, config);
+        try {
+          totalEntries += reader.getEntries();
+          totalSizeBytes += reader.length();        
+          if (ROWS_TO_SAMPLE > 0 &&
+              numCols > 1 &&  // only need to do this for non-aligned format
+              (startKey == null || startKey.length == 0) &&  // first region only
```
sbroeder: Minor nit perhaps, but startKey should never be null here.  If it is, that would be some sort of error, but you're allowing it.  I think it would be better to test for the HConstants.EMPTY_START_ROW, which is different than a null, or remove the test for null and just your test of length of 0 would also be sufficient.

DaveBirdsall:
jenkins, retest

DaveBirdsall:
jenkins, retest

DaveBirdsall:
The test failure is in the JDBC suite; it is a problem with nowait statement depth. It is not related to these changes.

PR #1207, opened 2017-08-15
[TRAFODION-2704]close unused statment in JDBC regression

xwq:


dcs/src/test/jdbc_test/src/test/java/org/trafodion/jdbc_test/BatchTest.java
```diff
@@ -158,38 +182,51 @@ public void testBatchInsertPkDuplicate() throws Exception {
...
-        ResultSet rs = _conn.createStatement().executeQuery(strSelect);
-        while(rs.next()) {
-            System.out.println("ID = " + rs.getString(1) + ", Name = " + rs.getString(2));
-            assertEquals(expectedIdArray[rs.getRow()-1], rs.getInt(1));
-            assertEquals(expectedNameArray[rs.getRow()-1], rs.getString(2));
-            rowCount++;
-        }
-        rs.close();
-        insertStmt.close();
+            if (insertStmt != null) {
```
DaveBirdsall: Could the selectStmt.close() throw an exception, preventing us from trying to close insertStmt?

dcs/src/test/jdbc_test/src/test/java/org/trafodion/jdbc_test/TestBoolean.java
```diff
@@ -75,9 +75,17 @@ public void JDBCBoolean() throws SQLException {
...
             assertNull(e.getMessage());
         } finally {
-            stmt.close();
-            prepStmt.close();
-            conn.close();
+            if (stmt != null) {
+                stmt.close();
+            }
+
+            if (prepStmt != null) {
```
DaveBirdsall: Could the stmt.close() throw an exception, preventing us from trying to close prepStmt and conn?

dcs/src/test/jdbc_test/src/test/java/org/trafodion/jdbc_test/TestTinyInt.java
```diff
@@ -77,9 +77,16 @@ public void JDBCTinyIntSigned() throws SQLException {
             e.printStackTrace();
             assertNull(e.getMessage());
         } finally {
-            stmt.close();
-            prepStmt.close();
-            conn.close();
+            if (stmt != null) {
+                stmt.close();
+            }
+            if (prepStmt != null) {
```
DaveBirdsall: Same comment as above

dcs/src/test/jdbc_test/src/test/java/org/trafodion/jdbc_test/TestTinyInt.java
```diff
@@ -123,9 +130,17 @@ public void JDBCTinyIntUnsigned() throws SQLException {
             e.printStackTrace();
             assertNull(e.getMessage());
         } finally {
+            if (stmt != null) {
             stmt.close();
-            prepStmt.close();
-            conn.close();
+            }
+
+            if (prepStmt != null) {
```
DaveBirdsall: And here too

dcs/src/test/jdbc_test/src/test/java/org/trafodion/jdbc_test/BatchTest.java
xwq: For now, if the selectStmt.close() throw an exception. it seems we don't have a good solution to close it. 
In the real case, when connection is closed, all the statement resource will be freed.

DaveBirdsall: Apologies... I don't think I was being clear. I'm wondering if we should do something like:

try {
  selectStmt.close();
}
catch (Exception e) {
  // ignore it
}

insertStmt.close();

That is, make sure we attempt insertStmt.close() even if there was a problem doing selectStmt.close().

xwq: you are right. 

For now, many of the test cases have the same issues. And the mechanism for handling the exceptions also has big problems. I planed to improve them, but it need some time to finish it. 
I think we can merge this to improve the jenkins test first.

DaveBirdsall: I agree. This is fine.

DaveBirdsall:
@selvaganesang, even with all the statement and connection closes that @xwq has added, the JDBC tests still fail hitting statement limits. Any ideas on why that might be?

selvaganesang:
You can run with ABORT_ON_ERROR=2026 and get the dump of the process that returns this error. This would help to debug this issue further

selvaganesang:
I am running in my environment to reproduce this problem

selvaganesang:
There were around 13 statements in OPEN or FETCH statement with the source string
"select dataOffset, chunkLen, stringParam from table(ghost table \"TRAFODION\".\"T4QA\".\"LOBDescChunks__08574991311373262837_0002\") where descPartnKey = <val1> and descSyskey = <val2>"

These internal statements should be closed for the jdbc_test_cdh to pass or comment out the tests that triggers these internal statements for time being.

Changing the application wouldn't help

selvaganesang:
It is strange how these tests were enabled without going through the check-PR jobs. Check-PR jobs should have failed in the PR that enabled these tests

selvaganesang:
Interestingly, Trafodion T4 driver uses WeakReference to close the statements even when the application fails to close  the statement but the statement is gc-ed by JVM. See TrafT4Connection.gcStmts and WeakReference objects/Queues in TrafT4Statement.java.  For the statements to be gc-ed, it is important to use a small new object heap in JVM. I undestand from @arvind-narain that the jdbc tests scripts are run within the maven. I would suggest that maven is started with appropriate max heap size rather than the default . The default would be 1/4 of the physical memory of the jenkins VM.

sandhyasun:
The tests also have to be modified so the DDL gets created consistently. Here re the findings ..
------------------------------
From: Arvind Narain 
Sent: Tuesday, August 22, 2017 10:00 PM
To: Sandhya Sundaresan <sandhya.sundaresan@esgyn.com>; Weiqing <weiqing.xu@esgyn.cn>
Cc: Selva Govindarajan <selva.govindarajan@esgyn.com>; Kevin <kaihua.xu@esgyn.cn>
Subject: RE: JDBC test failures on official run having to do with DDL

I looked up a little bit more. With mvn since there is no order in which the tests get picked up, the success/failure depends on which test is executed before TestBlob.

Jdbc tests seem to utilize three different ways of setting up each test suite. 

One way is to use Utils to setup the connection. Here if the schema does not exist, then it gets created.

Another way is to have BeforeClass methods that setup connection and the schema, tables needed. AfterClass goes aheads and cleans the tables and sometimes the schema – this is what happens when BatchTest is run.

Third is what TestBlob is doing – using DriverManager.GetConnection to setup the connection. Here schema doesn’t get created. Hence if the expected schema has been dropped by the previous test suite (BatchTest) then the test fails since it cannot create the tables. 

So yes, tests need to be looked at again so that DriverManager.GetConnection call is replaced in TestBlob*,TestClob*,TestGetIndexInfo*.

core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
```diff
@@ -6591,6 +6612,10 @@ odbc_SQLSrvr_UpdateLob_sme_(
         exception_->u.ParamError.ParamDesc = SQLSVC_EXCEPTION_EXECUTE_FAILED;
     }
 
+    if (QryLobUpdateSrvrStmt != NULL) {
+        QryLobUpdateSrvrStmt->Close(SQL_CLOSE);
+    }
+
```
sandhyasun: This doesn't really close the internal cursor statements that SQL maintains to retrieve the LOB data. The JDBC driver has to issue a special syntax to close it. Once PR  https://github.com/apache/incubator-trafodion/pull/1216  is merged in the special syntax needs to be issued to close the stream output ( SQL extract )

selvaganesang: It looks like only one Extract statement can be active from a connection. Is that ok? What happens when there are two LOB columns in a table and that are being fetched simultaneously by the application

xwq:
I will create anther PR for 2.3 branch

PR #1208, opened 2017-08-15
[TRAFODION-2716] Query compilation gets stuck at listSnapshots() at t…

selvaganesang:
…imes

To reduce the impact on the compilation when there are many snapshots,
call Admin.listSnapshots  only when cqd TRAF_TABLE_SNAPSHOT_SCAN is set to
LATEST or SUFFIX.


selvaganesang:
I am merging this PR in though check-PR failed. The failure in jdbc_test-cdh is being worked on

PR #1209, opened 2017-08-16
TRAFODION-2717 odbc give wrong data and dcsmaster hang

mashengchen:


dcs/src/main/java/org/trafodion/dcs/master/listener/ListenerWorker.java
```diff
@@ -86,39 +86,43 @@ public void run() {
...
+                    default:
+                        clientData = requestUnknown.processRequest(clientData, s);
+                        break;
+                }
+                // Return to sender
+                int requestReply = clientData.requestReply;
+                key.attach(clientData);
+                server.send(new PendingRequest(key, requestReply));
+            } catch (Exception e){
+                LOG.error("Unexpected Exception", e);
```
selvaganesang: Handling the exception in a generic way is not a good option. Also, this handler just sinks the exception other than just logging. This can lead to unexpected behavior later

dcs/src/main/java/org/trafodion/dcs/master/listener/RequestUnknown.java
```diff
@@ -76,7 +76,11 @@ ClientData processRequest(ClientData clientData, Socket s) {
             clientData.hdr = hdr;
         } catch (UnsupportedEncodingException ue){
             if(LOG.isErrorEnabled())
-                LOG.error("Exception in RequestUnknown: " + s.getRemoteSocketAddress() + ": " + ue.getMessage() );
+                LOG.error("UnsupportedEncodingException in RequestUnknown: " + s.getRemoteSocketAddress() + ": " + ue.getMessage(), ue);
+            cancelConnection = true;
+        } catch (Exception e){
```
selvaganesang: ditto the earlier comment

dcs/src/main/java/org/trafodion/dcs/master/listener/ListenerWorker.java
mashengchen: in this case if not catch Exception, there will throw the exception and lead to dcsmaster thread died, then the client side will never get connection, and once there catch Exception the cancelConnection will set to true, and then requestReply will be REQUST_CLOSE, and client side will get response for connect failure, this will not influence other client side

kevinxu021: @selvaganesang it keeps the process here. otherwise, the unexpected behavior will be the process being there without ListenerWorker thread which lead to hang. what's your suggestion?

selvaganesang: You need to handle only the exceptions that can be ignored and possibly have a mechanism to allow certain number of retries only. ,  By handling the generic exception, you would ignore the runtime exceptions like null pointer exceptions etc. This can mask the real issues that could remain unnoticed for long time.

kevinxu021: The problem is if we dont catch exception, you could not realize the problem. It's a thread that if there is any exception un-catched happens, the thread will exit while the DCSMaster process being there. The hung issue is because the ListenerWorker thread exiting. So the difference is hanging or not, both of them are covered all exceptions.

selvaganesang: So, you want to do something like the program below that prints 100 times null pointer exception instead of exiting the process. 

public class TestS
{
   public static void main(String args[])
   {
      for (int i = 0; i < 100; i++)
      {
         String s = null;
         try
         {
            System.out.println ("Length of the string " + s.length());
         } catch (Exception e) {
            e.printStackTrace();
         }
      }
   }
}

But serious errors like OutOfMemoryError will anyway won't be caught by this handler. So, do you think it is ok for DCSMaster to hang then.  

My earlier comment is to ensure that all unchecked exceptions (Runtime Exceptions or any exception that need not be caught or thrown) should also be treated like critical error as above.  I would say if you want DCS Master not to hang at all conditions, then it can't be done by handling exceptions. 

If you attempt to avoid this hang by handling Throwable like catch (Throwable t), then it very bad because the only way to recover from this situation is to kill the process manually, which would have happened automatically if you didn't handle Throwable.

kevinxu021: I would suggest to print 100 times instead of exiting. Because 100 NullPointExceptions belongs to 100 requests, that will not hang the rest of the requests. it'll not impact each other. By the way, as you see, there are some changes in a internal request catching Exception and return back to the client, then the outer catch is for the exception out of the request. It doesn't have so much code which rarely has chance for the exception out of control. if you still worry about it, we can add a limitation for the number of the exceptions in a reasonable time range.

OutOfMemoryError will lead to process exiting and client will get socket timeout exception in 60s, at this time, DCSMaster will not hang.

Again, in the past, the request from client lead to the issue, as you see, in the loop, there should be a different request for each round of loop. Unless the client send the same request again and again, it should be a big problem.

kevinxu021: I would say unless the client send the same request again and again, it should **_NOT_** be a big problem.

selvaganesang: I am confused here. If process exit due to out of memory error can cause the socket to timeout exception, it should behave the same way when the process exits due to runtime exception too. Looking at the JIRA, it is a question of if we should consider java.nio.BufferUnderFlowException as a serious error that needs to exit the process or as an recoverable error. In general, I would think it is a serious error because it is derived from runtimeexception. If it is not an serious error, it could have been derived from IOException like any other read/write error.  Did the process exit when you got BufferUnderFlowException before this change?

kevinxu021: This is an issue for memory allocation failed, it doesn't mean the memory has been ran out. I dont think it is reasonable to exit the process instead of returning back the issue to the client. By the way, even we have socket timeout, the real issue will be hidden. whatever the issue is on server-side, you will always get socket timeout for every client request. Can customer really accept the behavior? Again, the outside CATCH is to avoid process exiting because of a very particular bad request. If it is a real OUT OF MEMORY， i mean run out the memory, absolutely the process will be exiting.  More, OutOfMemoryError is an ERROR which not belongs to exception and will not be covered by EXCEPTION block. In another word, the process will exit.

selvaganesang: I am unable to follow the explanation. I will let the community to decide on this PR going forward

PR #1211, opened 2017-08-18
TRAFODION-2722 Change purgedata to use HBase truncate api

anoopsharma00:
Traf now uses HBase truncate api to delete all rows from
a table and indexes, instead of dropping/recreating underlying HBase
objects.

A second change has also been added to fix a bug where compiler cache
was being cleared whenever a DDL statement was issued.
This change was added a while back before Query Invalidation(QI) feature
was available and was not removed when QI was added.
It has now been removed.

### DIFF
--- a/core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
+++ b/core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
@@ -648,6 +648,8 @@ public class HBaseClient {
            }
            else {
               TableName tableName = TableName.valueOf(tblName);
+              if (admin.isTableEnabled(tableName))
+                  admin.disableTable(tableName);
               admin.truncateTable(tableName, preserveSplits);
            }
         } finally {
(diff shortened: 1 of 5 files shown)


core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
```diff
@@ -648,6 +648,8 @@ public boolean truncate(String tblName, boolean preserveSplits, long transID)
            }
            else {
               TableName tableName = TableName.valueOf(tblName);
+              if (admin.isTableEnabled(tableName))
+                  admin.disableTable(tableName);
```
selvaganesang: Why do we disable the table before truncate? And when will it be enabled.

anoopsharma00: HBase truncate api requires table to be disabled.
truncate api enables the table after truncation.

selvaganesang: Thanks @anoopsharma00.  I had used  truncate to delete the data in "_DTM_".TDDL hbase table.  It looks like the errors were ignored in that case and hence it didn't detect this requirement

PR #1213, opened 2017-08-20
implement isValid function

luoyunpeng:


luoyunpeng:
@kevinxu021  please help review

luoyunpeng:
jenkins, test

svarnau:
jenkins, okay to test

luoyunpeng:
@selvaganesang @kevinxu021  please help review

core/conn/jdbcT4/src/main/java/org/trafodion/jdbc/t4/TrafT4Connection.java
```diff
@@ -1876,9 +1876,57 @@ public SQLXML createSQLXML() throws SQLException {
...
 
-	public boolean isValid(int timeout) throws SQLException {
-		// TODO Auto-generated method stub
-		return false;
+    public boolean isValid(int timeout) throws SQLException {
+        if (props_.t4Logger_.isLoggable(Level.FINE) == true) {
+            Object p[] = T4LoggingUtilities.makeParams(props_);
+            props_.t4Logger_.logp(Level.FINE, "TrafT4Connection", "isValid", "iValid", p);
+        }
+        if (props_.getLogWriter() != null) {
```
selvaganesang: Does this mean the log will be written always?

luoyunpeng: no,  the log will only be written when user of client set the log level to true and log writer is not null,

selvaganesang: I meant to ask if the log will be written only when it is FINE. If so, you need to avoid create all these objects in java and then throw it out while logging because the log level is not set to FINE.

PR #1214, opened 2017-08-21
[TRAFODION-2702] fix wrong offset while converting sql to c with ucs2

Weixin-Xu:
For last PR , the bug on windows has not been mentioned


Weixin-Xu:
jenkis , retest

Weixin-Xu:
Jenkins , retest

PR #1216, opened 2017-08-23
TRAFODION-2724 

sandhyasun:
Fix to ensure internal statements used to retrieve LOB descriptor chunks information during an extract are closed properly. The issue was if the the caller of extract passed in the exact length, the ""readCursorData method would return the correct data  but would leave the "open" to the internal table in a FETCH state. This change ensures that a second "fetch" is done so EOD would be seen and as a result the internal statement would get closed within the  method ExLob::fethCursor which would call SQL_EXEC_LOBcliInterface(handleIn, handleLenIn, ....
                                     LOB_CLI_SELECT_CLOSE, .....)


core/sql/exp/ExpLOBaccess.cpp
```diff
@@ -1970,7 +1970,14 @@ Ex_Lob_Error ExLob::readCursorData(char *tgt, Int64 tgtSize, cursor_t &cursor, I
               continue;
          }
       }
-      
+      if (operLen == tgtSize)
+        {
+          //The cursor still has data available but wait until redrive since
+          //we have fetched all the data for this extract operation
```
selvaganesang: The comment seems to be confusing. What is being waited to do until redriver?

sandhyasun: The cursor on the full lob data indicates that there is data left. But the extract operation can retrieve justt a portion of the total. The state is maintained in the cursor and will get updated with each "execute" of the extract statement issued by caller of the extract statement.  Agree this comment is confusing. I can just say "return the  data to caller".

sandhyasun:
Let me rework this a bit so it is explicitly clear as to where we are closing the internal cursor on the descriptor table. Will make it more readable and easier to maintain.  Please don't merge yet.

anoopsharma00:
+1

PR #1218, opened 2017-08-25
[TRAFODION-2725] SQL types are REAL, FLOAT, and DOUBLE. Some values are inserted, a stack overflow occurs when SQLGetData is executed

SuJinpei:
root cause: write stack variable out of range.
unsigned long ODBC::ConvertSQLToC(...., targetLength) {
    ...
    CHAR		cTmpBuf[132];
    ...
    double_to_char (dTmp, FLT_DIG + 1, cTmpBuf, targetLength)  **// when targetLength > 132, the stack will be broken.** 
    ...
}
fixed by:limit the max size to sizeof(cTmpBuf)


SuJinpei:
Jenkins retest

SuJinpei:
test failed seems caused by hive issue, not by the driver.  I close it to fix the similar issue on WCHAR.

PR #1219, opened 2017-08-30
 [TRAFODION-2727] Memory leak in the compiler part of the code in Trafodion

selvaganesang:
Compiler objects were refactored to use the heap infrastructure to ensure that it is accounted and allocated within the trafodion memory management.
    
The commonly used NAType and derived classes are mostly created via heap, but the heap was not passed in the constructor. The embedded objects used the system heap though the container object is allocated in the compiler heap. This could result in memory leak when the container object is destroyed.


sureshsubbiah:
+1. Looks great Selva.

PR #1220, opened 2017-08-30
[TRAFODION-2725] SQL types are REAL, FLOAT, and DOUBLE. Some values are inserted, a stack overflow occurs when SQLGetData is executed.

SuJinpei:
**_root cause:_** write stack variable out of range.
unsigned long ODBC::ConvertSQLToC(...., targetLength) {
...
CHAR	cTmpBuf[132];
...
double_to_char (dTmp, FLT_DIG + 1, cTmpBuf, targetLength) **// when targetLength > 132, the stack will be broken.**
...
}
**_fixed by:_** limit the max size to sizeof(cTmpBuf)
**_test result:_** run coast to check, no new issue introduced.

### DIFF
--- a/win-odbc64/odbcclient/drvr35/drvrglobal.cpp
+++ b/win-odbc64/odbcclient/drvr35/drvrglobal.cpp
@@ -876,7 +876,7 @@ bool double_to_char (double number, int precision, char* string, short size)
     precision = precision < size ? precision : size - 1;
 
     // precission should be limit to a reasonable range.
-    if ((precision < 0) && (precision >(DBL_MANT_DIG - DBL_MIN_EXP))) {
+    if ((precision < 0) || (precision >(DBL_MANT_DIG - DBL_MIN_EXP))) {
         goto fun_exit;
     }
 
--- a/win-odbc64/odbcclient/drvr35/drvrglobal.cpp
+++ b/win-odbc64/odbcclient/drvr35/drvrglobal.cpp
@@ -867,7 +867,7 @@ bool double_to_char (double number, int precision, char* string, short size)
 {
     bool rc = false;
     char format[16];
-    char buf[512];
+    char buf[MAX_DOUBLE_TO_CHAR_LEN];
 
     sprintf(format, "%%.%dl%c", precision, number < 1e-6 ? 'g':'f');
     sprintf(buf, format, number);
--- a/win-odbc64/odbcclient/drvr35/drvrglobal.cpp
+++ b/win-odbc64/odbcclient/drvr35/drvrglobal.cpp
@@ -866,8 +866,8 @@ bool use_gcvt(double number, char* string, short size)
 bool double_to_char (double number, int precision, char* string, short size)
 {
     bool rc = false;
-    char format[16];
-    char buf[MAX_DOUBLE_TO_CHAR_LEN];
+    char format[16] = { '\0' };
+    char buf[MAX_DOUBLE_TO_CHAR_LEN] = { '\0' };
 
     sprintf(format, "%%.%dlg", precision);
     sprintf(buf, format, number);
--- a/win-odbc64/odbcclient/drvr35/drvrglobal.cpp
+++ b/win-odbc64/odbcclient/drvr35/drvrglobal.cpp
@@ -867,16 +867,42 @@ bool double_to_char (double number, int precision, char* string, short size)
 {
     bool rc = false;
     char format[16] = { '\0' };
+    size_t actualLen = 0;
+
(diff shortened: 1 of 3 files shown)


selvaganesang:
While reviewing the double_to_char function to determine if the size includes null termination, I found that ecvt function is being used to convert the double to character. ecvt function is not thread safe. Can you please fix the surround code too.  

Also, it looks like size shouldn't include null termination.  The refactored code using sprintf can decide how the size should be used

SuJinpei:
Hi Selva, thank you for your comments.  seems ecvt function is obsolete, sprintf is recommended.  Can I use sprintf to refactor the function double_to_char?

selvaganesang:
Yes. you can using sprintf

win-odbc64/odbcclient/drvr35/drvrglobal.cpp
```diff
@@ -865,77 +865,18 @@ bool use_gcvt(double number, char* string, short size)
...
-	{
-		buffer [current_location] = '.' ;
-		for(count = current_location;
-			count< abs(decimal_spot)+current_location; count++)
-			buffer [count + 1] = '0' ;
-		strcpy (&buffer [count + 1], temp) ;
-	}
+    bool rc = false;
+    char format[16];
+    char buf[MAX_DOUBLE_TO_CHAR_LEN];
```
xwq: need initialize. **_char buf[MAX_DOUBLE_TO_CHAR_LEN] = {0};_**

win-odbc64/odbcclient/drvr35/drvrglobal.cpp
```diff
@@ -865,77 +865,18 @@ bool use_gcvt(double number, char* string, short size)
...
 
-	rSup(buffer);
-	length = strlen(buffer);
-	if (buffer[0] == '.' || (buffer[0] == '-' && buffer[1] == '.')) length++;
+    sprintf(format, "%%.%dlg", precision);
+    sprintf(buf, format, number);
```
xwq: snprint is better.

SuJinpei: I don't think there is a risk of writing out of range.  So I think sprintf is better.

selvaganesang: There is a risk of writing out of range because precision is not checked to be within the buf size. Also, you need to allow the numbers to be truncated after a decimal even when there is no sufficient length passed by the caller.  So  size can be less than strlen(buf) as long as the fraction part alone is getting cut.

SuJinpei: @selvaganesang from original code logic, size should larger than strlen(buf).  if we still want to return reasonable value though size less than strlen(buf), I think we should use other algorithm instead of snprintf.

selvaganesang:
+1. The concepts for this change look good. I am assuming that the boundary conditions are unit tested.

SuJinpei:
@selvaganesang Yes, I tested some boundary conditions, the result is expected.  test program like below:
int main()
{
    char buf[5120];
    bool rc = false;
    int p = DBL_MANT_DIG-DBL_MIN_EXP;
    unsigned long long dsb[] = {
        0x3ff0000000000000,
        0x3ff0000000000001,
        0x0000000000000001,
        0x000fffffffffffff,
        0x001fffffffffffff,
        0x0010000000000000,
        0x7fefffffffffffff,
        0x800fffffffffffff,
        0x8000000000000001
    };
    int precisions[] = {
        0,1,2,4,8,15,DBL_MANT_DIG - DBL_MIN_EXP
    };

    int sizes[] = {
        -1, 0, 1
    };

    double *ds = (double *)&dsb;

    printf("sizeof(dsb) / sizeof(double) = %d\n", sizeof(dsb) / sizeof(double));

    for (size_t i = 0; i < sizeof(dsb) / sizeof(double); ++i) {
        for (size_t j = 0; j < sizeof(precisions) / sizeof(int); ++j) {
            for (size_t k = 0; k < sizeof(sizes) / sizeof(int); ++k) {
                rc = double_to_char(ds[i], precisions[j], buf, precisions[j] + sizes[k]);
                printf(rc == true ? "number=%lg, precision=%d, size=%d, buf=%s, strlen(buf)=%d\n" : "number=%lg, precision=%d, size=%d, return false\n",
                    ds[i], precisions[j], precisions[j] + sizes[k], buf, strlen(buf));
            }
        }
    }

    return 0;
}

DaveBirdsall:
Looks like this is ready to merge. I will merge it.

PR #1222, opened 2017-08-31
fixed an issue that caused charsets/TEST014 to loop.

anoopsharma00:
an existing bug causes process loop when infer_charset is ON
and query invalidation needs to remove entries from query cache.
Until this bug is fixed, turn query_cache OFF.

### DIFF
--- a/core/sql/regress/charsets/TEST014
+++ b/core/sql/regress/charsets/TEST014
@@ -28,6 +28,11 @@
 --
 -- @@@ END COPYRIGHT @@@
 
+-- an existing bug causes process loop when infer_charset is ON
+-- and query invalidation needs to remove entries from query cache.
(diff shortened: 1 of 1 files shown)


core/sql/regress/charsets/TEST014
```diff
@@ -28,6 +28,11 @@
 --
 -- @@@ END COPYRIGHT @@@
 
+-- an existing bug causes process loop when infer_charset is ON
```
DaveBirdsall: Is there a JIRA for this bug? Do we know anything about the bug (other than this test runs into it)?

anoopsharma00: jira TRAFODION-2730 tracks this issue.

anoopsharma00:
jenkins retest

PR #1224, opened 2017-09-06
Jdbc fail

xwq:


core/sqf/sqenvcom.sh
```diff
@@ -32,7 +32,7 @@
 #Product version (Trafodion or derivative product)
 export TRAFODION_VER_PROD="Apache Trafodion"
 export TRAFODION_VER_MAJOR=2
-export TRAFODION_VER_MINOR=2
+export TRAFODION_VER_MINOR=3
```
DaveBirdsall: I don't think this is correct. It looks like we are renumbering the release as 2.3, but tihs pull request is to the release2.2 branch. Perhaps a commit was included in this change that should not have been?

xwq:
Hi Dave. Sorry for that I forget to close this one. Actually I have created a anther PR for this https://github.com/apache/incubator-trafodion/pull/1225

PR #1225, opened 2017-09-06
[TRAFODION-2704]close unused statment in JDBC regression

xwq:


core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
```diff
@@ -6489,6 +6506,31 @@ odbc_SQLSrvr_ExtractLob_sme_(
                 exception_->u.SQLError.errorList._buffer = QryLobExtractSrvrStmt->sqlError.errorList._buffer;
                 exception_->u.ParamError.ParamDesc = SQLSVC_EXCEPTION_EXECUTE_FAILED;
             }
+
+            IDL_long_long zeroValue = 0;
+            snprintf(LobExtractQuery, sizeof(LobExtractQuery), "EXTRACT LOBTOBUFFER(LOB'%s', LOCATION %Ld, SIZE %Ld)", lobHandle, (Int64)lobDataValue, &zeroValue);
+
+            retcode = QryLobExtractSrvrStmt->ExecDirect(NULL, LobExtractQuery, EXTERNAL_STMT, TYPE_CALL, SQL_ASYNC_ENABLE_OFF, 0);
+
```
sandhyasun: So who is the caller of this with 'zero' value. It needs to be triggered by the JDBC driver. But I am not seeing any changes in the driver ? The 'zeroLength' can be passed in only when the application decides to close the resultset or the connection. It cannot be called soon after the extract statement is executed. Can you clarify ?

xwq: For now, the lob extract all data in one query, so after this query, we can call the "zeroLength' query directly. 

In the further, I will support the extract the data in multiple times for large data, and then, we need JDBC change.

sandhyasun: Ok this will not work for a large stream of data though. We cannot allocate that much in memory. Are all these limitations being documented   in a JIRA ?

xwq: I will documented it. 

For the large data, they are some logic in JDBC need to be designed, I will add the code when I have the solution.

sandhyasun: This seems like a big restriction since we will allow a max of 512 MB only.  This means the user cannot even havea picture file of 1 G or 2 G in this release ?  How about for update lob ?  Will the caller be able to use "update lob  <lobhandle> append " several times to append data to a single lob ?

xwq: For "update lob append", it can be called multi times to append large data to a single LOB. 

The "extract LOB", maybe I can create another PR for it.

selvaganesang: I am assuming that the streaming concept is implemented in the JDBC/ODBC driver and the server side just fetches the data in bulk at some size, If the fetch bulk size is greater than the user requested size, it would be serviced from the driver itself or when it is lesser multiple chunks will be fetched to satisfy the user requested size. If so, this change looks ok. However, there is a drawback that this change may not be portable to JDBC T2 driver because there can be only one extract query active at a time.

sandhyasun: The multiple check extraction is not implemented yet according to Weiqing's previous comment and it should be. Otherwise the application won't be able to  extract more than the first  512 MB of data.That is our extract chunk size default in SQL.  But he will fix it in with a different JIRA and PR.

sandhyasun: About your comment on T2, currently for this release only T4 support is done for LOBs . None of the other drivers support LOB datatype. They need to be addressed one by one. Weiqing, we need separate  JIRAs for those too.

xwq:
@selvaganesang @sandhyasun 

The part of fetch LOB context is the early work. Now, the JDBC will fetch the whole LOB context once. I planed to improve it. 

The JDBC need implement the  streaming concept. Since the we can't modify the LOB context and only append it. SO for JDBC, it need fetch all the LOB context, and modify it and than write it again.
I need be careful for the code change in JDBC to handle different case.
For example, there are some different case:

For example, The JDBC has already fetch 100 bytes, and the length is 200.
1.  if the application call setBytes(200, data), the JDBC can only append the data.
2.  if the application call setBytes(150, data), the JDBC need fetch anther 50 bytes and update the whole LOB.
2. if the application calls setBytes(50, data), the JDBC can modify the lob context local and then update the whole LOB.

From the example, I think i need handle different case very carefully, so the work need some time to finish.
For now, I think we can tell the use the limitation of the JDBC LOB API.

For anther driver, ODBC Linux is on working.  

JDBC T2 need some rework fro ROWSETS part and then the LOB function will be added.

selvaganesang:
@Sandhya explained to me why the current code causes the change in current behavior.  Currently, without   the lines 6509 - 6514, the JDBC/ODBC application should be able to get the more lob data from the current position. The lines 6509-6514 closes the lob and hence it is not possible for the application to get more lob data.

I would revert back my earlier comment that this change looks ok.  I was under the opinion that the EXTRACT command can be used to extract the lob data from any given position. @Sandhya clarified that EXTRACT command can't work that way if the lob cursor is closed. The lines 6509-6514 closes the lob cursor. Hence this change needs to be refactored

core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
sandhyasun: Ok if this will be reworked as part of TRAFODION-2740, we can keep this as temp code to close the statement - although prematurely. Just make sure there is a comment in the code saying this will be removed when TRAFODION-2740 is addressed.

xwq:
I will create anther PR for TRAFODION-2740 which will also include this PR. We don't need merge this PR anymore.

sandhyasun:
Ok but without the "close" call we will definitely run into the limit for the number of opens being 14.  This may cause issues in production. So we do need to have the close issue addressed.

PR #1227, opened 2017-09-08
[TRAFODION-2732] Add delimited identifier handling to sample table code path

DaveBirdsall:
UPDATE STATISTICS generates several SQL statements internally. When referencing column names, the safest thing to do is to double-quote them, because sometimes we have SQL keywords serving as column names. Many code paths in UPDATE STATISTICS already do this; this particular change is for one that didn't.

### DIFF
--- a/core/sql/ustat/hs_la.cpp
+++ b/core/sql/ustat/hs_la.cpp
@@ -189,6 +189,23 @@ Lng32 HSHbaseTableDef::getColumnNames()
     return retcode;
   }
 
+// local function used by the X::DescribeColumnNames functions
+NAString getAnsiName(NAString & columnName)
+  {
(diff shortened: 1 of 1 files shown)


DaveBirdsall:
I am running full regressions on this one over the weekend. Please do not merge until I've reported the results.

robertamarton:
Looks good to me.  I wonder if there are other parts of SQL where we generate internal syntax that would run into a similar issue.

DaveBirdsall:
That is quite possible. I did check the likely places within the ustat code and I believe they are all now covered.

DaveBirdsall:
I ran full regressions on this overnight last night on a workstation and all tests passed.

PR #1228, opened 2017-09-09
[TRAFODION-2733]  Provide an improved memory quota assignment for big memory operators (BMO)

selvaganesang:
1) Enabled memory quota per node. The CQD BMO_MEMORY_LIMIT_PER_NODE_IN_MB
(renamed from EXE_MEMORY_LIMIT_PER_CPU) is set to 10240 MB by default.

Old attribute | Old value | Renamed Attribute | New value
-- | -- | -- | --
EXE_MEMORY_LIMIT_PER_CPU | 0 | BMO_MEMORY_LIMIT_PER_NODE | 10240
EXE_MEMORY_LIMIT_LOWER_BOUND_HASHGROUPBY | 10 | BMO_MEMORY_LIMIT_LOWER_BOUND_HASHGROUPBY | 25
EXE_MEMORY_LIMIT_LOWER_BOUND_HASHJOIN | 10 | BMO_MEMORY_LIMIT_LOWER_BOUND_HASHJOIN | 25
EXE_MEMORY_LIMIT_LOWER_BOUND_PROBE_CACHE | 10 | BMO_MEMORY_LIMIT_LOWER_BOUND_PROBE_CACHE | 25
EXE_MEMORY_LIMIT_LOWER_BOUND_SORT | 10 | BMO_MEMORY_LIMIT_LOWER_BOU

2) Changes in EXPLAIN
Estimated memory per node for all BMOs at ROOT operator
Estimated memory per instance for every BMO operator
Memory quota per instance for every BMO operator
3) BMO TDB contains the memory quota per esp instance now.
4) Root TDB now contains the limit per node and estimated memory per node.
This can be used by WMS to change the memory allocation during
runtime without compilation. - Not yet implemented.
5) Added a CQD BMO_MEMORY_LIMIT_UPPER_BOUND to cap the memory
consumed by BMO by the same queries with less number of
bmos.
6) The unused memory quota is yielded to other fragments in the process
also.
7) Removed the code to limit the ESPs from being assigned to a fragment
based on the BMO memory quota.
8) Added a new CQD BMO_MEMORY_ESTIMATE_RATIO_CAP to cap the memory
estimate skew by any one BMO operator to 0.7.
9) To disable the memory quota per node, set BMO_MEMORY_LIMIT_PER_NODE_IN_MB to 0.
10) This memory quota is distributed proportionally based on the estimated memory
taking into consideration the number of bmo instances per operator and
the number of nodes available in the cluster to host these instances.
Hence, this memory quota should be valid in multi-fragments independent of the
number of fragments in an ESP.
10) Removed the CQD EXE_MEMORY_LIMIT_NONBMOS_PERCENT and EXE_MEMORY_RESERVED_FOR_MXOSRVR_IN_MB
11) Fixed BMO stats WM to be at least the allocated memory.
12) Changed the sort operator to account the bmo memory correctly.
13)  Introduced a new CQD BMO_MEMORY_EQUAL_QUOTA_SHARE_RATIO
to determine how much of BMO_MEMORY_LIMITE_PER_NODE_IN_MB is divided equally between
BMO operators and the rest would depend upon the memory estimate. This ratio is set
to 0.5 by default. Changing it to 0 will divide the memory quota proportional to the memory estimate of the BMO operator. This change is needed to get better BMO quota allocation when the memory estimate goes awry for certain types of query.


selvaganesang:
jenkins, retest

selvaganesang:
@prashanth-vasudev , @sureshsubbiah @zellerh  @DaveBirdsall Please review the PR #1228

core/sql/cli/Context.h
```diff
@@ -505,6 +519,8 @@ class ContextCli : public ExGod {
   // It points to a list of 'black box' of data allocated by user and is returned
   // back whenever this context is in use.
   HashQueue * trafSElist_;
+  // memory quota allocation given back by BMOs to be used by other BMOs
+  ULng32 unusedBMOsMemoryQuota_;
```
DaveBirdsall: What unit of measurement is this value? Bytes? KB? MB? If Bytes, is ULng32 big enough?

core/sql/comexe/ComTdbSort.h
```diff
@@ -201,11 +201,15 @@ class ComTdbSort : public ComTdb
...
-  Float32 sortMemEstInMbPerCpu_;                        // 64-67
+  Float32 sortMemEstInKBPerNode_;                        // 64-67
   Float32 bmoCitizenshipFactor_;                        // 68-71
   Int32  pMemoryContingencyMB_;                        // 72-75
   UInt16 sortGrowthPercent_;                            // 76-77
-  char fillersComTdbSort_[18];                          // 78-95
+  char filler_1[2];                                     // 78-79
+  Int32 topNThreshold_;                                 // 80-83
+  Float32 estMemoryUsage_;                             // 84-87
+  Float32 bmoQuotaRatio_;                              // 88-92
```
DaveBirdsall: Comment should be 88-91, and on the next line 92-95

core/sql/executor/ex_exe_stmt_globals.h
```diff
@@ -307,18 +307,38 @@ class ExExeStmtGlobals : public ex_globals
 
   inline NABoolean grabMemoryQuotaIfAvailable(ULng32 size)
   { 
+    CliGlobals *cli_globals = GetCliGlobals();
+    if (cli_globals->isEspProcess())
+       return cli_globals->grabMemoryQuotaIfAvailable(size);
     if ( unusedBMOsMemoryQuota_ < size ) return FALSE;
     unusedBMOsMemoryQuota_ -= size ;
```
DaveBirdsall: I don't understand how this code relates to the code in cli/CliGlobals.cpp. There the comments suggest that the memory quota is always stored in a context object. But here we seem to be storing it in statement globals in the Master process case.

core/sql/generator/GenRelJoin.cpp
```diff
@@ -1784,36 +1784,44 @@ short HashJoin::codeGen(Generator * generator) {
...
-        hashj_tdb->setMemoryQuotaMB( UInt16(memQuota) );
+    if ( memQuota < hjMemoryLowbound ) {
+       memQuota = hjMemoryLowbound;
+       memQuotaRatio = BMOQuotaRatio::MIN_QUOTA;
     }
+    else if (memQuota >  memoryUpperbound)
+       memQuota = memoryUpperbound;
+       memQuotaRatio = BMOQuotaRatio::MIN_QUOTA;
+    hashj_tdb->setMemoryQuotaMB( UInt16(memQuota) );
+    hashj_tdb->setBmoQuotaRatio(memQuotaRatio);
```
DaveBirdsall: I take it memQuotaRatio is an "out" parameter at line 1810? Also, why use MIN_QUOTA in both the cases of memQuota being too low and too high? Contrast this code with RelRoot::codeGen where we set MIN_QUOTA only in the too low case.

core/sql/optimizer/RelExpr.cpp
```diff
@@ -1988,77 +1988,62 @@ NABoolean RelExpr::containsNode(OperatorTypeEnum nodeType)
...
-     // usage in the fragment, and evenly distrbuted among BMOs in the fragment.
-     return ((exeMem/2) * (BMOsMemoryUsagePerFragment/totalBMOsMemoryUsage + 
-                        double(numBMOsPerFragment)/totalNumBMOs)
-            ) / numBMOsPerFragment;
+     double bmoMemoryRatio = bmoMemoryUsage / totalBMOsMemoryUsage;
+     double capMemoryRatio = 1; 
+     if (totalNumBMOs > 1) {
+        capMemoryRatio = ActiveSchemaDB()->getDefaults().getAsDouble(BMO_MEMORY_ESTIMATE_RATIO_CAP);
+        if (capMemoryRatio > 0 && capMemoryRatio <=1 && bmoMemoryRatio > capMemoryRatio)
+           bmoMemoryRatio = capMemoryRatio;
```
zellerh: Capping the ratio makes sense, but I wonder how we manage to give the memory we capped to the other BMOs. Example: Two BMOs, one with 10 GB, the other with 10 MB. (99.9% + .1%). Let's say we cap the first BMO at 80%. That works. But, won't we still just give .1% of the available memory to the second BMO, because we didn't adjust totalBMOsMemoryUsage?

core/sql/comexe/ComTdbSort.h
selvaganesang: done

core/sql/executor/ex_exe_stmt_globals.h
DaveBirdsall: See cli/CliGlobals.cpp line 1103

core/sql/generator/GenRelJoin.cpp
selvaganesang: Good Catch. Fixed RelExpr::computeMemoryQuota to return bmoQuotaRatio to return the ratio correctly.  However, this needs to be revisited when the memory quota is reassigned based on runtime settings. See item 4 in the PR commit comment.

core/sql/cli/Globals.cpp
```diff
@@ -1099,6 +1100,52 @@ void CliGlobals::deleteContexts()
...
+  return context->unusedMemoryQuota();
+}
+
+void CliGlobals::yieldMemoryQuota(ULng32 size)
+{
+  ContextCli *context;
+  if (espProcess_)
+     context = defaultContext_;
+  else
+     context = currContext();
```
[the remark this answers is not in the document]
selvaganesang: Yes. However currContext() is expected to return the thread specific context in case of master in case of multi-threaded JDBC T2 application.  In case of ESP, we always want to return the defaultContext so that memory quota is stored in the default context.

core/sql/executor/ex_exe_stmt_globals.h
```diff
@@ -307,18 +307,38 @@ class ExExeStmtGlobals : public ex_globals
...
   }
 
-  inline void resetMemoryQuota() { unusedBMOsMemoryQuota_ = 0 ; }
+  inline void resetMemoryQuota() 
+  {
+    CliGlobals *cli_globals = GetCliGlobals();
+    if (cli_globals->isEspProcess())
+       return cli_globals->resetMemoryQuota();
+    unusedBMOsMemoryQuota_ = 0 ; 
+  }
```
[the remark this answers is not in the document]
selvaganesang: In case of master process the unused memory quota is yielded to others operators within the statement and hence it is stored in statement globals. In case of ESP, the unused memory quota to yielded to operators in other fragments within the process and hence it is stored in the default context.  The part of the code deals with the unused memory quota.

 I am assuming that you want details about the quota assignment itself.  Based on this assumption:

 Each open statement in the master would get its own quota.   If it is non-ESP plan, each BMO operator will be capped to max BMO limit of 1200 MB as opposed to 10240 MB per node. If there are more than one BMO operator in such a plan,  each BMO operator memory quota will be assigned proportionally as per item 14 subjected to min and max limits.

it is a ESP plan, the memory quota will be assigned as per item 14 subjected to min and max limit. 
Currently,  Trafodion won't assign the instances from a same fragment to a process. It also won't assign the ESP to two different queries either from the same context or different contexts.  Because TDB contains the memory quota assignment per instance, different instances within the same process should work irrespective how the instances are assigned.

Yielding the unused memory quota will work fine as long as the ESP is not shared for different ESPs at the same time.  Currently the ESP is released only when the statement is deallocated.  The   current plan is not to share ESPs when it is multi-threaded. It is just that all the fragments and all the instances within the same fragment  from a query will be hosted in the same process via different threads.

Hence,
1) Yes. Each open statement would get its own quota assignment
2) All ESPs and BMO operator instances memory consumption from a query will be limited by BMO_MEMORY_LIMIT_PER_NODE. However, memory consumption won't be exact amount as specified by this CQD because the memory quota assignment is subjected to min and max limits.
3) ESPs are not shared currently.  There is no plan to share in the multi-threaded ESP environment also.
4) . Each statement will have its own quota irrespective of the context to which its belongs

core/sql/generator/GenPreCode.cpp
```diff
@@ -2154,34 +2154,10 @@ RelExpr * RelRoot::preCodeGen(Generator * generator,
...
-         m2 = 0;
-      } else { 
-
-         // nBMOs use more memory than the portion, adjust m2 to 
-         // that of (1-ratio)*m
-         if (m1 > m*ratio )
-           m2 = m*(1-ratio);
-      }
+      // total per node
+      double m = defs.getAsDouble(BMO_MEMORY_LIMIT_PER_NODE) * (1024*1024);
```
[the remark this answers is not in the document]
selvaganesang: Will do

core/sql/cli/Globals.cpp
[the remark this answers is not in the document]
zellerh: Thanks, sounds ok, although I don't quite understand why we wouldn't treat both master and ESP the same. Otherwise, when we switch between serial and parallel plans we may see an unexpected change in behavior? This leads me to a more general comment: I wonder whether it would be better to set an overall limit per connection, not per node. I wonder whether this would be easier to manage, both for users as well as for the SQL engine.

core/sql/optimizer/RelExpr.cpp
```diff
@@ -1988,77 +1988,62 @@ NABoolean RelExpr::containsNode(OperatorTypeEnum nodeType)
...
+     // estimated memory usage to the total estimated memory usage of all BMOs
+   
+     // The ratio can be capped by the CQD
+     double equalQuotaShareRatio = 0;
+     equalQuotaShareRatio = ActiveSchemaDB()->getDefaults().getAsDouble(BMO_MEMORY_EQUAL_QUOTA_SHARE_RATIO);
+/*
+     char *equalQuotaShareRatioStr = getenv("BMO_MEMORY_EQUAL_QUOTA_SHARE_RATIO");
+     if (equalQuotaShareRatioStr != NULL)
+         equalQuotaShareRatio = atof(equalQuotaShareRatioStr);
+*/
```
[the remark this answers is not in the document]
selvaganesang: done

core/sql/executor/ex_exe_stmt_globals.h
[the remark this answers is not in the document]
zellerh: Thanks very much for the detailed explanation! Now I understand that your model will work fine. My hope would be that we will share ESPs more, not less, in the future. A long time ago we designed them to be shared, but a lot of bugs made us share less and less. Maybe something to keep in mind for the future.

core/sql/optimizer/RelExpr.cpp
selvaganesang: Yes. That's true. In that case, we need to do multiple pass or adjust the totalBMOMemoryUsage for the unassigned BMO operator.  Will attempt to do this later

core/sql/generator/GenProbeCache.cpp
```diff
@@ -470,30 +463,38 @@ CostScalar ProbeCache::getEstimatedRunTimeMemoryUsage(NABoolean perCPU)
...
+     totalMemory *= numOfStreams;
   }
-
-  double  memoryLimitPerCpu =
-      ActiveSchemaDB()->getDefaults().getAsLong(EXE_MEMORY_FOR_PROBE_CACHE_IN_MB) * 1024 * 1024;
-  if (totalMemory > memoryLimitPerCpu)
-     totalMemory = memoryLimitPerCpu;          
+  if (numStreams != NULL)
+     *numStreams = numOfStreams;
+  if ( perNode == TRUE )
```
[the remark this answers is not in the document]
selvaganesang: I am curious why is it unsafe? Can you please give some details so that I can modify it accordingly.

core/sql/generator/GenProbeCache.cpp
```diff
@@ -470,30 +463,38 @@ CostScalar ProbeCache::getEstimatedRunTimeMemoryUsage(NABoolean perCPU)
...
   }
-
-  double  memoryLimitPerCpu =
-      ActiveSchemaDB()->getDefaults().getAsLong(EXE_MEMORY_FOR_PROBE_CACHE_IN_MB) * 1024 * 1024;
-  if (totalMemory > memoryLimitPerCpu)
-     totalMemory = memoryLimitPerCpu;          
+  if (numStreams != NULL)
+     *numStreams = numOfStreams;
+  if ( perNode == TRUE ) 
+     totalMemory /= MINOF(MAXOF(((NAClusterInfoLinux*)gpClusterInfo)->getTotalNumberOfCPUs(), 1), numOfStreams);
```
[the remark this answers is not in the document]
selvaganesang: will do

core/sql/generator/GenProbeCache.cpp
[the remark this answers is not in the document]
zellerh: Yes, I guess you didn't write this, it's old code. As far as I understand, TRUE is defined as 1. If perNode happens to be -1, we will not execute the if statement, but we usually want to treat any value other than 0 as true. That's why I would either use if (perNode) or if (perNode != FALSE).

zellerh:
Thank you, appreciate all your replies.

core/sql/generator/GenProbeCache.cpp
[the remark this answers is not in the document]
selvaganesang: While making the change, I realized that I didn't get what is expected. I believe it is already taking the MINOF(numOfStreams, NumberOfNodes())

core/sql/cli/Context.h
selvaganesang: It is in MB.

core/sql/executor/ex_sort.cpp
```diff
@@ -364,12 +365,15 @@ ExSortTcb::ExSortTcb(const ExSortTdb & sort_tdb,
         sortType_.useIterQSForRunGeneration_ = 1;
         break;
   }   
+
+  sortHeap_ = new(getHeap()) NAHeap("Sort Heap", (NAHeap *)getHeap(), 204800);
```
prashanth-vasudev: Allocation of sortHeap_ here seems to be duplicate allocation since sortHeap_ is already allocated at line 317.  This line can be removed.

selvaganesang: Done

DaveBirdsall:
Just to reiterate, looks good to me. Thanks for all the replies.

PR #1230, opened 2017-09-11
Trafodion-2735 :LOB: Drop table/schema returns 8616 error but the drop statement can't be retried

sandhyasun:
LOB drop table is not supported by DTM currently. 
Each LOB tables contains 3 underlying Hbase tables and one or more HDFS data files that contain the LOB data.
During drop the code used to delete all the HDFS files first and then drop the LOB hbase tables.
This PR chnages that by doing 2 things  1. Move the code drop HDFS files below the calls that drop the Hase tables. This way if there are any errors during the Hbase drop tables, the TM will protect them and roll them back. The HDFS files will stay intact. 2. Ignore errors from "File Not Found" errors when dropping a LOB table. 
The issue is that twhen a conflict  occured between 2 sessions and the LOB table lost its HDFS data but the LOB Hbase descriptor tables got rolled back. A subsequenct drop from another session, expected the HDFS files to be there and raised an error. SInce the HDFS files are not protected by TM, there is currently no way to roll them back.   So the current fix doe snot fully addressthe issue but alleviates the issue and allows 2 streams of tests to run and handle conflict errors just fine. 

The real fix need to be in TM where the LOB data files are also  protected by a transaction.

### DIFF
--- a/core/sql/cli/Cli.cpp
+++ b/core/sql/cli/Cli.cpp
@@ -10519,7 +10519,7 @@ Lng32 SQLCLI_LOB_GC_Interface
                
   if (rc )
     {
-      cliRC = 9999; // Warning 
+      cliRC = 1; // Warning 
       ComDiagsArea * da = &diags;
       lobDebugInfo("compactLobDataFile Failed",0,__LINE__,lobTrace);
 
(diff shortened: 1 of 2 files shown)


prashanth-vasudev:
Looks good to me +1

PR #1231, opened 2017-09-13
[Trafodion 2739] RMS semaphore handling need to log the error for easy resolution of the issue

selvaganesang:
RMS semaphore handling had the following:
1) Passed in parameters that are no longer used. It had these parameters for legacy reasons
2) Was returning errno for short return code. This might have caused unexplainable behavior
   at times.
3) The caller was aborting for non-zero return value, but the return value is not easily visible.

All the above problems have been fixed in this PR.

### DIFF
--- a/core/sql/cli/Cli.cpp
+++ b/core/sql/cli/Cli.cpp
@@ -8325,19 +8325,15 @@ Lng32 SQLCLI_RegisterQuery(CliGlobals *cliGlobals,
   StatsGlobals *statsGlobals = cliGlobals->getStatsGlobals();
   if (statsGlobals == NULL)
     return retcode;
-  short error;
-  short savedPriority, savedStopMode;
+  int error;
   error = statsGlobals->getStatsSemaphore(cliGlobals->getSemId(),
-                 cliGlobals->myPin(), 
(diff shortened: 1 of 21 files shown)


core/sql/cli/Cli.cpp
```diff
@@ -8325,19 +8325,15 @@ Lng32 SQLCLI_RegisterQuery(CliGlobals *cliGlobals,
...
     return retcode;
-  short error;
-  short savedPriority, savedStopMode;
+  int error;
   error = statsGlobals->getStatsSemaphore(cliGlobals->getSemId(),
-                 cliGlobals->myPin(), 
-                savedPriority, savedStopMode, FALSE /*shouldTimeout*/);
-  ex_assert(error == 0, "getStatsSemaphore() returned an error");
-
+                 cliGlobals->myPin());
```
DaveBirdsall: Why do we ignore the error?

selvaganesang: Any error would have asserted in StatsGlobals::getStatsSemaphore() method itself

sureshsubbiah:
+1. Looks good. I can find no issues.

PR #1232, opened 2017-09-15
[TRAFODION-2738] Rowset buffer size during insert/upsert should be li…

selvaganesang:
…mited

Rowset insert includes UnpackRows operator in its query tree. UnPackRows
operator was allocating a larger memory based on rowset size and row size
to flow the data to its parent operator.

Introduced a CQD EXE_MEMORY_FOR_UNPACK_ROWS_IN_MB to limit the amount
of memory allocated in this operator. The default value is 100MB.

In addition, streamlined the NAMemory infrastruture to return NULL or
dump core when more than 2 GB of memory is allocated in one chunk.


core/sql/sqlcomp/nadefaults.cpp
```diff
@@ -1336,6 +1336,8 @@ SDDkwd__(EXE_DIAGNOSTIC_EVENTS,		"OFF"),
 
  SDDui___(EXE_MEMORY_FOR_PROBE_CACHE_IN_MB,	"100"),
 
+ SDDui___(EXE_MEMORY_FOR_UNPACK_ROWS_IN_MB,	"100"),
```
sureshsubbiah: A default value of 100 MB maybe too small. I understand that this is good for mixed workloads, but do consider than plans with Unpack, especially when used for insert/upsert are simple. Unpack is always serial and part of master exe. Often there is only one in a query.  The cost of having a low value here seems to be that queue length from unpack could be limited by this when we are loading a table with wide rows (even if actual data in row is not wide). We know from performance runs that queue length is critical to upsert load performance. This is a hunch on my part and could be misguided. It will be good to performance test for upsert of table with wide rows, if you think that is worthwhile.

sureshsubbiah:
+1. Change looks good.

core/sql/sqlcomp/nadefaults.cpp
selvaganesang: Because there is an easy way to change the memory size, we can determine the optimal default at a later date.  I believe the current default of 100 MB should be good enough for most of the use cases.

selvaganesang:
@sureshsubbiah Can you please review the 2nd commit

sureshsubbiah:
+1 Looks good.

PR #1233, opened 2017-09-15
[TRAFODION-2651] Fixed monitor-to-monitor communication by adding

zcorrea:
reconnect logic to better handle various communication glitches which previously
resulted in false node down situations.


sureshsubbiah:
+1

PR #1234, opened 2017-09-16
[TRAFODION-2746] Fixed various problem detected in large clusters (> 30)

zcorrea:
The problems were:

1. A segmentation violation occurred during the Integration phase, when the new 
   monitor is establishing the socket communication paths between itself and 
   the existing monitors.
   a. Information is exchanged between the master (creator) monitor and the
      slave (new) monitor process which tells the new monitor which nodes
      monitor process make up the existing cluster instance. During these
      exchanges, in CCluster::ReceiveSock() one of the messages was large
      enough to require chunking and the logic which kept track of the
      number of bytes received was not calculated correctly which resulted
      in an overwrite past the boundary of the receive buffer. 
2. A second segmentation violation was due to a buffer overwrite during the
   Joining (revive) phase.
   a. In requeue.cxx, when creating the buffer in the master (creator) monitor
      which is populated with the cluster state information to be sent to the
      slave (new) monitor process, the calculation did not properly account for
      the number of logical and physical nodes. So that when the buffer was
      populated, it would overwrite past the allocated buffer.
3. A third problem was also note in the one of the monitor would remain in
   the Joining state and never come out of it.
   a. The problem was in the order of logic when calling 
      CCluster::ResetIntegratingPNid() which triggers the
      CCommAccept::commAcceptorSock() to accept another new node to
      integrate. The invocation to ResetIntegratingPNid() was done before
      resetting the creator flag. Due to kernel scheduling, this resetting
      of the creator flag was happening after another monitor started the
      Integration phase and it was breaking the node integration protocol
      by terminating it too early. So the new monitor would stay in the
      Joining state for ever since the protocol was broken.
4. The last segmentation violation was due to stderr buffer overwrite in
   CRedirectStderr::handleOutput() where the size returned by snprintf() 
   was used to terminate the buffer containing stderr data >= 4096 which
   is the size of the buffer.

### DIFF
--- a/core/sqf/monitor/linux/cluster.cxx
+++ b/core/sqf/monitor/linux/cluster.cxx
@@ -346,17 +346,15 @@ void CCluster::NodeReady( CNode *spareNode )
         lnode->Up();
     }
 
-    ResetIntegratingPNid();
     spareNode->SetActivatingSpare( false );
-
     if ( MyNode->IsCreator() )
     {
         MyNode->SetCreator( false, -1, -1 );
     }
+    ResetIntegratingPNid();
 
     TRACE_EXIT;
 }
-
 
 // Assigns a new TMLeader if given pnid is same as TmLeaderNid 
 // TmLeader is a logical node num. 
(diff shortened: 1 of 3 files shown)


zcorrea:
jenkins, retest

sureshsubbiah:
+1

PR #1240, opened 2017-09-20
[TRAFODION-2740]JDBC extract LOB context in chunks

xwq:


DaveBirdsall:
Just checking: I notice this change is against the release2.2 branch. Is this a planned fix for that branch? @traflm, would you comment?

xwq:
Yes, it's planed for release 2.2.

dcs/src/test/jdbc_test/src/test/java/org/trafodion/jdbc_test/TestForeignKey.java
```diff
@@ -74,12 +74,15 @@
 	public static void doTestSuiteSetup() throws Exception {
 		try{
 			_conn = DriverManager.getConnection(Utils.url, Utils.usr, Utils.pwd);
-			Statement stmt = _conn.createStatement();
-			
-			stmt.execute(strCreatePKTABLE1Query);
-			stmt.execute(strCreatePKTABLE2Query);
-			stmt.execute(strCreateFKTABLE1Query);
-			stmt.execute(strCreateFKTABLE2Query);
+            try (
```
arvind-narain: Thanks for the change Weiqing. Learnt something new related to try-with-resource-statement from your change - it does reduce the earlier code change planned - thanks.

Could you please also take care of replacing DriverMagager.getConnection calls from *Blob*, *Clob*, TestGetIndexInfo and this file ? This would take care random non-existent schema errors seen.

selvaganesang:
Though this fix would take care closing the LOB extract statement, it doesn't implement the streaming concept in true sense.  For streaming concept to work, Blob should be materialized only when the application issues Blob.getOutputStream().getBytes() method. It looks like we read the whole Blob data in chunks() and keeps it around in our driver layer. This will result in large java memory usage on the client side.
 
Also, we should close the LOB only when Blob.free() is called.  If the application fails to call Blob.free() and the Blob data is fetched once, the lob cursor should be closed, when the corresponding Statement or PreparedStatement object of the query that got the LOB column is closed.

However, this problem exists in the driver before this PR.  I would suggest this problem is also taken care of.  Hence, I didn't review this code thoroughly.

xwq:
@selvaganesang For the stream concept. It really difficult to do in JDBC side now since the SQL side only support APPEND operator. As my  comment in another PR, we have to many case to consider. In many case, we need retire the whole data to DRIVER side to modify it and then write them back.

Now, I read all them data once and then we can support ALL the JDBC APIs. In the further, we can improve the internal JDBC to make the process more efficiently.

selvaganesang:
@sandhyasun  Can you please comment this PR? @xwq wants to get this PR merged with the current limitations.

core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
```diff
@@ -6411,8 +6412,34 @@ odbc_SQLSrvr_ExtractLob_sme_(
         exception_->exception_nr = odbc_SQLsrvr_ExtractLob_ParamError_exn_;
         exception_->u.ParamError.ParamDesc = SQLSVC_EXCEPTION_UNABLE_TO_ALLOCATE_SQL_STMT;
     }
+FILE * fp= NULL;
+        fp = fopen("/tmp/closeLOB", "a+");
```
sandhyasun: What is this /tmp/closeLOB used for ?

core/conn/jdbcT4/src/main/java/org/trafodion/jdbc/t4/TrafT4Lob.java
```diff
@@ -91,23 +92,49 @@ public void read() throws SQLException {
...
 
 		T4Connection t4connection = this.connection_.getServerHandle().getT4Connection();
-		LogicalByteArray wbuffer = ExtractLobMessage.marshal(ExtractLobMessage.LOB_EXTRACT_BUFFER, lobHandle_, 1, 0,
+		LogicalByteArray wbuffer = ExtractLobMessage.marshal(ExtractLobMessage.LOB_EXTRACT_LEN, lobHandle_, 1, 0,
 				connection_.ic_);
 
 		LogicalByteArray rbuffer = t4connection.getReadBuffer(TRANSPORT.SRVR_API_EXTRACTLOB, wbuffer);
 		ExtractLobReply reply = new ExtractLobReply(rbuffer, connection_.ic_);
-		data_ = reply.lobDataValue;
+		length = reply.lobLength;
```
sandhyasun: Is this length the entire length of the lob  returned by the SQL layer ?

xwq: Yes, it the entire length of the LOB.

sandhyasun: Ok that may be very large.  I thought we're supposed to return each chunk to application to consume ?  Will you implement that at next phase ?

core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
xwq: Sorry , I will remove them. debug code.

core/conn/jdbcT4/src/main/java/org/trafodion/jdbc/t4/TrafT4Lob.java
xwq: I have talked it with Selva. I think the logic is a little complex to support some APIs like setBytes(). I will add write down the design to discuss with you about the next phase.

sandhyasun: ok

dcs/src/test/jdbc_test/src/test/java/org/trafodion/jdbc_test/TestForeignKey.java
sandhyasun: I see you did a new commit to fix some review comments. Can you also fix this one ? This fix is needed for the daily builds to pass. We often get "non existent schema " errrors with JDBC tests.

sandhyasun:
Merging  this now.

PR #1241, opened 2017-09-20
[TRAFODION-2750] Using function strtod is not enough to convert C_CHAR to DOUBLE

Weixin-Xu:
client app with odbc driver :
DDL : "CREATE TABLE DOUBLE_TEST(A DOUBLE PRECISION)"
buffer =   "-2.22507e-308"
Bind SQL_C_CHAR to SQL_DOUBLE
return :
State: 22003
Native Error: 0
Error: [Trafodion ODBC Driver] NUMERIC VALUE OUT OF RANGE. Incorrect Format or Data. Row: 1 Column: 1

unit test for driver code :
char buffer[] = "-2.22507e-308";
double db;
errno = 0;
db = strtod(buffer,NULL);
return :
db = -0.000000
errno = 34

### DIFF
--- a/core/conn/unixodbc/odbc/odbcclient/unixcli/cli/ctosqlconv.cpp
+++ b/core/conn/unixodbc/odbc/odbcclient/unixcli/cli/ctosqlconv.cpp
@@ -3545,7 +3545,7 @@ unsigned long ODBC::ConvertCharToNumeric(SQLPOINTER srcDataPtr,
     rTrim(cTmpBuf);
     tempLen = strlen(cTmpBuf);
     errno = 0;
-	dTmp = strtod(cTmpBuf, &errorCharPtr);
+	dTmp = strtold(cTmpBuf, &errorCharPtr);
 	if (errno == ERANGE || errorCharPtr < (cTmpBuf + tempLen))
 		return IDS_22_003;
 	return SQL_SUCCESS;


DaveBirdsall:
The test that failed, seabase/TEST031, is not going through ODBC or JDBC so it is a false failure. The remaining tests passed. Even so, I'll try a retest to see if the failure is non-deterministic.

DaveBirdsall:
jenkins, retest

DaveBirdsall:
+1 I think this change is OK. I tried the unit test on a Red Hat Linux development workstation. The behavior I saw there was errno of 34 on both the strtod and strtold calls. I tried it also in Visual Studio 2015 on a Windows 8 box. There, errno was 0 on both calls. I'm guessing that in the environment where this fix was done, the behavior was different for the two calls. "long double" support is platform specific.

xwq:
+ 1. Looks good to me.

core/conn/unixodbc/odbc/odbcclient/unixcli/cli/ctosqlconv.cpp
```diff
@@ -3545,7 +3545,7 @@ unsigned long ODBC::ConvertCharToNumeric(SQLPOINTER srcDataPtr,
     rTrim(cTmpBuf);
     tempLen = strlen(cTmpBuf);
     errno = 0;
-	dTmp = strtod(cTmpBuf, &errorCharPtr);
+	dTmp = strtold(cTmpBuf, &errorCharPtr);
```
selvaganesang: This looks innocuous, but strlod return long double which is 16 byte.  This 16 byte is assigned to 8 byte double dTmp. I am not sure if there is an implicit casting to double from long double can work as expected.

DaveBirdsall: Yeah, when I tried it I found on Red Hat Linux long double was 16 bytes. But on Windows 8, it is just 8 bytes. It depends on the platform. For what it is worth, I found that the implicit cast from long double to double worked fine on both platforms.

xwq:
The value "-2.22507e-308" is very close to the limit of 8 byte double. The issues is that it out of the double precision [https://en.wikipedia.org/wiki/Double-precision_floating-point_format](url).  

If the value if out this range, the value will be not exactly accurate for DOUBLE.

So, I think the change is not needed. We don't need support this value.

core/conn/unixodbc/odbc/odbcclient/unixcli/cli/ctosqlconv.cpp
selvaganesang: Thanks Dave. In that case, this change looks good to me

selvaganesang:
Is this change ready to merge?

DaveBirdsall:
I am OK with it.

DaveBirdsall:
What does the original author, @Weixin-Xu, think?

Weixin-Xu:
The value "-2.22507e-308" can not be stored in double accurately the same as in server. So I think it is not a bug.
For 8 byte double, the function strtold will not bring advantage about converting string to double.

DaveBirdsall:
Thanks, @Weixin-Xu. So, you don't want this to be merged, then? Will you instead close the request?

Thanks.

PR #1242, opened 2017-09-25
[TRAFODION-2753]LOB: filetolob() no longer works in trafci

xwq:


sandhyasun:
+1

sandhyasun:
Once the check test failure in  Phoenix T4 is addressed , this should be good to go.

xwq:
I checked the log, the failure was caused by timeout of "Check HDP service HDFS start status". I will restarted the test.

xwq:
jenkins, retest.

PR #1245, opened 2017-09-26
[TRAFODION-2754] Get statistics cores sqlci or mxosrvr at str_sprintf()

selvaganesang:
str_sprintf is Trafodion specific formatting function. Replaced
str_sprintf with sprint that provides better checking with parameters.
str_sprintf is replaced with sprintf in the statistics output related functions
only. It is advised that the new code should start using the standard formatting
functions.


anoopsharma00:
Would it be better to change str_sprintf to internally call sprintf (or vsprintf)
by passing var args through the use of va_list/va_start ?
That would not require all callers to change the call and will be a simpler change.
It will also be correct in all cases whether caller changes or not.
Or is there a technical reason why that is not feasible?
This would be a call from within str_sprintf but for a short method, it would
not be expensive.

selvaganesang:
When sprintf is used directly, C compiler flags out all the inconsistencies between the format and the parameters.  str_sprinf allowed different format specifiers than the standard ones. In this PR, I had focussed only on the statistics output, the code that uses str_sprintf a lot.  I am planning for another PR where str_sprintf will be defined as sprintf via #define.  This should bring out all the inconsistencies between the format and the parameters.

anoopsharma00:
Could  you provide an example where C compilation catches inconsistencies?
Here is a code fragment that contains inconsistent format and values, but it didn't
return any compilation errors during traf compilation. Should it have?
  double dd;
  char ddbuf[100];
  sprintf(ddbuf, "%1abc%d%s%Ld%f%t%y", "a", dd);

What I was thinking is to change str_sprintf in str.cpp to something like the following. 
Wont this behave the same way as changing all calls to sprintf?
  Int32 str_sprintf(char * buffer, const char * format, ...)
  {
    va_list args;
    va_start(args, format);
    vsprintf(buffer, format, args);
    return 0;
  }

DaveBirdsall:
Another way to achieve what @anoopsharma00 suggests would be to define str_printf as a macro, e.g. #define str_printf sprintf. Then you could do away with the str_printf function altogether.

anoopsharma00:
@selvaganesang had suggested of using #define to map str_sprintf to sprintf.

Few comments regarding that:
-- the reason for traf str_* methods was to have one entry point to str related methods called
  by traf code. Those methods can then be internally optimized, if need be
-- there are other str methods which follow the same design. For example: str_cmp 
  internally calls memcmp, or str_cpy_all which internally calls memmove
-- it makes it slightly inconvenient during debugging as one need to set breakpoint at
   sprintf instead of str_sprintf that is shown in code. It may also cause issues to look at variables inside of
   sprintf or other internal system method
-- it can differentiate between a traf call and a non-traf call. For ex, a breakpoint to str_cpy_all
  or str_sprintf can only be called by traf. If it is defined as another method, one cannot differentiate
  where that was called from without looking at backtrace.
-- over time this design methodology has not been followed consistently but why make it worse.

selvaganesang:
I tried the above code snippet with -Wformat -Werror option. This is how Trafodion is compiled.

[centos@gselva-trafodion-ins gselva]$ g++ -Wformat -Werror simple.c
cc1plus: warnings being treated as errors
simple.c: In function \u2018int main(int, char**)\u2019:
simple.c:8: error: format \u2018%1a\u2019 expects type \u2018double\u2019, but argument 3 has type \u2018const char*\u2019
simple.c:8: error: format \u2018%d\u2019 expects type \u2018int\u2019, but argument 4 has type \u2018double\u2019
simple.c:8: error: too few arguments for format
[centos@gselva-trafodion-ins gselva]$ 

I found some interesting issues when I attempted to change str_sprint as sprint via #define. Changing str_sprintf as suggested by you will work but it won't be able to find these issues at compile time. I guess it might result in errors at runtime.

selvaganesang:
I am not attempting to change other Trafodion specific str functions. str_sprintf has deviated from the standard and it doesn't  ensure that correct format specifiers are given.  This was causing issues at runtime when you don't use the correct format at times.

selvaganesang:
Did I address all your concerns? Is this PR ready to merge?

DaveBirdsall:
I'm happy with it.

selvaganesang:
jenkins, retest

sandhyasun:
+1. Looked at the LOB code related changes and they look fine. The spaces where LOB handle is generated and the LOB descriptor table names are being   derived are the key places. They look fine. A lot of Trace messages used it and they should be fine with sprintf too.

PR #1248, opened 2017-09-28
[trafodion-2728] - bug fix - SHOWSTATS sometimes gives strange messag…

andyyangcn:
…e when stats don't exist

### DIFF
--- a/core/sql/ustat/hs_globals.cpp
+++ b/core/sql/ustat/hs_globals.cpp
@@ -8472,6 +8472,23 @@ Lng32 HSGlobalsClass::groupListFromTable(HSColGroupStruct*& groupList,
 
     // Initialize the pointer to the group list we will build.
     groupList = NULL;
+    
+    // if showstats for a native hbase table, need to check if the schema _HBASESTATS_ exist
+    if (strcmp(hstogram_table->data(), "TRAFODION.\"_HBASESTATS_\".SB_HISTOGRAMS") == 0)
+      {
+        NAString queryStr = "SELECT count(*) FROM \"_MD_\".OBJECTS WHERE SCHEMA_NAME='_HBASESTATS_' "
+                            "AND OBJECT_NAME='__SCHEMA__' AND OBJECT_TYPE='PS';";
+        HSCursor cursor;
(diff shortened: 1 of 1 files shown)


core/sql/ustat/hs_globals.cpp
```diff
@@ -8472,6 +8472,23 @@ Lng32 HSGlobalsClass::groupListFromTable(HSColGroupStruct*& groupList,
...
+        HSCursor cursor;
+        retcode = cursor.prepareQuery(queryStr.data(), 0, 1);
+        HSHandleError(retcode);
+        retcode = cursor.open();
+        HSHandleError(retcode);
+        ULng32 cnt;
+        retcode = cursor.fetch (1, (void *)&cnt);
+        HSHandleError(retcode);
+        if (cnt == 0)
+          return 0;
```
DaveBirdsall: We need to do an LM->StopTimer() before returning.

core/sql/ustat/hs_globals.cpp
```diff
@@ -8472,6 +8472,23 @@ Lng32 HSGlobalsClass::groupListFromTable(HSColGroupStruct*& groupList,
 
     // Initialize the pointer to the group list we will build.
     groupList = NULL;
+    
+    // if showstats for a native hbase table, need to check if the schema _HBASESTATS_ exist
+    if (strcmp(hstogram_table->data(), "TRAFODION.\"_HBASESTATS_\".SB_HISTOGRAMS") == 0)
```
DaveBirdsall: Are there any other examples of this kind of bug besides native HBase access? What happens if we do SHOWSTATS on a native Hive table, for example, but the "_HIVESTATS_" schema has not yet been created? (This kind of question is why I was hoping for a solution where we interpreted the error code from cursor103.open() instead.)

andyyangcn:
For hive table and hbase table, cursor103.prepareQuery() will return -1 if there is no related schemas. And for table under seabase, cursor103.prepareQuery returns 0, but cursor103.open() will return -1. It seems we cannot judge if there is on object. So I add the judgment for the existence of SB_HISTOGRAMS table.

PR #1249, opened 2017-09-30
[TRAFODION-2757] Fix bug JDBC executeQuery() throws exception on the "with ... select" stmt

CoderSong2015:
The jdbcT4 driver can't check the sql statement type succefully if the beginning of the query is 'with ... select'.


selvaganesang:
+1

PR #1250, opened 2017-09-30
[TRAFODION-2758] Sort operator that uses TopN sort at runtime is not …

sureshsubbiah:
…accurately determined at compile time

This commit includes fixes for these three issues. Please see JIRA for a description of each fix.

[TRAFODION-2758] Sort operator that uses TopN sort at runtime is not accurately determined at compile time
[TRAFODION-2759] Use of rowset with large memory leads to various errors at runtime
[TRAFODION-2760] hbase cache blocks is OFF for broad table with narrow index

[TRAFODION-2758]
comexe/ComTdbSort.cpp
comexe/ComTdbSort.h
executor/ex_sort.cpp
generator/GenExplain.cpp
generator/GenPreCode.cpp
generator/GenRelMisc.cpp
optimizer/RelExpr.cpp
[TRAFODION-2759]
bin/SqlciErrors.txt
generator/GenRelPackedRows.cpp
sqlcomp/DefaultConstants.h
sqlcomp/nadefaults.cpp
[TRAFODION-2760]
generator/GenRelScan.cpp

### DIFF
--- a/core/sql/optimizer/RelExpr.cpp
+++ b/core/sql/optimizer/RelExpr.cpp
@@ -1078,7 +1078,33 @@ void RelExpr::pushDownGenericUpdateRootOutputs( const ValueIdSet &outputs)
 //QSTUFF
 
 void RelExpr::needSortedNRows(NABoolean val)
-{
+{ 
+  // The operators listed below can create OR propogate a GET_N
+  // request. Other operatots will turn a GET_N request into GET_ALL
+  // There are a few exceptions like right side of NJ for semi join etc.
+  // but these are not relevant for FirstN sort
+  // This method should only in the generator since we are using 
+  // physical node types.
+  OperatorTypeEnum operatorType = getOperatorType();
+  if ((operatorType != REL_FIRST_N) &&
+      (operatorType != REL_EXCHANGE) &&
+      (operatorType != REL_MERGE_UNION) &&
+      (operatorType != REL_PROBE_CACHE) &&
+      (operatorType != REL_ROOT) && 
+      (operatorType != REL_LEFT_NESTED_JOIN) &&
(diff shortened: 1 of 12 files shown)


core/sql/optimizer/RelExpr.cpp
```diff
@@ -1078,7 +1078,33 @@ void RelExpr::pushDownGenericUpdateRootOutputs( const ValueIdSet &outputs)
...
+  if ((operatorType != REL_FIRST_N) &&
+      (operatorType != REL_EXCHANGE) &&
+      (operatorType != REL_MERGE_UNION) &&
+      (operatorType != REL_PROBE_CACHE) &&
+      (operatorType != REL_ROOT) && 
+      (operatorType != REL_LEFT_NESTED_JOIN) &&
+      (operatorType != REL_LEFT_TSJ) &&
+      (operatorType != REL_MAP_VALUEIDS))
+    return ;
+
```
selvaganesang: This code looks ok.  Do you think it will be easy to understand if this condition is changed to use the operators that turns GET_N request to GET_ALL? First there should be only handful operators that would turn GET_N request to GET_ALL and there is less probable to add an operator that would turn GET_N to GET_ALL

sureshsubbiah: Thank you for the review. I think there are more operators that change GET_N to GET_ALL, than those that preserve a GET_N. I like to think of it as most operators may need more than N rows to produce N rows (say a join or a groupby). There are fewer that can guarantee N rows, every time they receive N rows.

sandhyasun:
+1 Looks good ! Thanks for cleaning up the sortNRows_ deprecated flag and making the code more readable.

PR #1252, opened 2017-10-01
[TRAFODION-2754] Get statistics cores sqlci or mxosrvr at str_sprintf()

selvaganesang:
Fix for executor/TEST130 failure in daily build


anoopsharma00:
fix to the failure is to update the following line in regress/tools/regress-filter-linux:
  SYSKEY='[0-9]\{17,20\}\>'
(change 18 to 17 as unique ids on some systems return a 17 digit number).
This same issue was seen on another branch.

selvaganesang:
Thanks @anoopsharma00  I will update it

PR #1253, opened 2017-10-03
[TRAFODION-1610][TRAFODION-2630] Generate constr exprs on update operators

DaveBirdsall:
This set of changes plugs a hole in our implementation of update/upsert/merge operators. It completes work started by JIRA TRAFODION-1575.

This set of changes adds logic to the generator to generate update constraint check and insert constraint check expressions for update operators. (The update constraint check handles the run-time update path; the insert constraint check handles the run-time insert path for upsert and merge; it is absent for vanilla update and for merges that lack a "when not matched" clause.)

This set of changes also removes temporary code added long ago that transformed an UPDATE to an INSERT + a DELETE when constraint expressions were present. (The INSERT operator had a correct implementation of constraint checking.) This should make many non-key-column UPDATEs faster, and will also eliminate the possibility of HBase timestamp collisions because the INSERT happens within a millisecond of the DELETE.


DaveBirdsall:
FYI: I ran this fix through full regressions last night on a clean instance on a workstation. They pass.

DaveBirdsall:
@selvaganesang and @zellerh, would you please review?

core/sql/generator/GenRelUpdate.cpp
```diff
@@ -292,8 +292,85 @@ static short genUpdExpr(
...
+  ValueIdList savedSourceVIDlist;
+  NAList<Attributes*> savedSourceAttrsList(generator->wHeap());
+
+  for (ValueId sourceValId = constraintColumns.init();
+       constraintColumns.next(sourceValId);
+       constraintColumns.advance(sourceValId))
+    {
+      NAColumn * sourceCol = ((IndexColumn*)sourceValId.getItemExpr())->getNAColumn();
+      ValueId targetValId;
+      NAColumn *targetCol = NULL;
```
DaveBirdsall: Just noticed something that could be improved. It would be better to set targetCol to NULL just inside the "for" loop, so that the "if" at line 329 would not be taken. It's not an observable bug, though, because even if on the second time around, targetCol were pointing to an NAColumn from the first time around, the other comparisons at line 330 onward would come out false. Nevertheless, making the change would make it cleaner.

core/sql/generator/GenRelUpdate.cpp
```diff
@@ -292,8 +292,85 @@ static short genUpdExpr(
...
+        {
+          const ItemExpr *assignExpr = targetRecExprArray[ni].getItemExpr();
+          targetValId = assignExpr->child(0)->castToItemExpr()->getValueId();
+          NAColumn *targetCol = NULL;         
+          if (targetValId.getItemExpr()->getOperatorType() == ITM_BASECOLUMN)
+            targetCol = ((BaseColumn*)targetValId.getItemExpr())->getNAColumn();
+          else if (targetValId.getItemExpr()->getOperatorType() == ITM_INDEXCOLUMN)
+            targetCol = ((IndexColumn*)targetValId.getItemExpr())->getNAColumn();
+                
+          if ((targetCol) &&
```
zellerh: This code to find the corresponding column seems risky to me. Isn't there a way to find the column by position? If the NATables for source and target are the same, that should be possible. If they are not, then matching by name is probably not the right thing anyway.

core/sql/generator/GenRelUpdate.cpp
```diff
@@ -292,8 +292,85 @@ static short genUpdExpr(
...
+
+      if (found)
+	{
+          Attributes * sourceValAttr = (generator->addMapInfo(sourceValId, 0))->getAttr();
+          Attributes * targetValAttr = (generator->getMapInfo(targetValId, 0))->getAttr();
+
+          // Save original location attributes so we can change them back after
+          // generating the update constraint expression
+
+          Attributes * savedValAttr = new(generator->wHeap()) Attributes();
```
zellerh: This is probably existing code that's modified, but I still have a comment: Instead of altering the map tables, it might be better to make a ValueIdMap m for the old and new values and calling m.rewriteValueIdUp(newValId, constrTree->getValueId()).

core/sql/generator/GenRelUpdate.cpp
```diff
@@ -292,8 +292,85 @@ static short genUpdExpr(
...
+
+          sourceValAttr->copyLocationAttrs(targetValAttr);
+        }
+
+    }
+
+  // Now that we have remapped the Attributes for the columns to their values
+  // in the new record, we can generate the update constraint expression.
+
+  expGen->generateExpr(constrTree->getValueId(), ex_expr::exp_SCAN_PRED,
```
zellerh: I don't quite understand what happens when the constraint tree refers to columns that are not in the assign list, or is the assign list targetRecExpr always complete? If it's always complete, there would be an optimization to skip constraints that don't refer to any updated columns.

core/sql/generator/GenRelUpdate.cpp
```diff
@@ -292,8 +292,85 @@ static short genUpdExpr(
...
+                       targetExpr);
+
+  // Now put the Attributes back the way they were.
+
+  for (Lng32 i = 0; i < savedSourceVIDlist.entries(); i++)
+    {
+      ValueId sourceValId = savedSourceVIDlist[i];
+      Attributes * sourceValAttr = (generator->getMapInfo(sourceValId, 0))->getAttr();
+      sourceValAttr->copyLocationAttrs(savedSourceAttrsList[i]);
```
zellerh: If we created a new map table entry on line 343, this code won't delete that entry. As far as I can tell, it would put back the temporary location that was assigned for a new entry? See comment above, I would recommend not modifying map tables. If you want to stay with the MapTable method, though, it might be better to drop those entries that were created on line 343 here.

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -11552,35 +11552,7 @@ RelExpr *Update::bindNode(BindWA *bindWA)
   NABoolean transformUpdateKey = updatesClusteringKeyOrUniqueIndexKey(bindWA);
   if (bindWA->errStatus()) // error occurred in updatesCKOrUniqueIndexKey()
     return this;
-  // To be removed when TRAFODION-1610 is implemented.
-  NABoolean xnsfrmHbaseUpdate = FALSE;
-  if ((hbaseOper()) && (NOT isMerge()))
-    {      
-      if (CmpCommon::getDefault(HBASE_TRANSFORM_UPDATE_TO_DELETE_INSERT) == DF_ON)
```
zellerh: I assume this is the only use of this CQD, so the CQD should be removed?

core/sql/generator/GenRelUpdate.cpp
```diff
@@ -292,8 +292,85 @@ static short genUpdExpr(
...
+  // come from targetRecExprArray.
+
+  ValueIdList savedSourceVIDlist;
+  NAList<Attributes*> savedSourceAttrsList(generator->wHeap());
+
+  for (ValueId sourceValId = constraintColumns.init();
+       constraintColumns.next(sourceValId);
+       constraintColumns.advance(sourceValId))
+    {
+      NAColumn * sourceCol = ((IndexColumn*)sourceValId.getItemExpr())->getNAColumn();
```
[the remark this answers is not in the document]
DaveBirdsall: Thanks. Will add.

core/sql/generator/GenRelUpdate.cpp
DaveBirdsall: I used the same technique that the existing code used. I'm curious how source and target tables could differ... check constraints are always local to a table, correct? That said, I'm happy to consider alternatives. Can you point me to an example?

core/sql/generator/GenRelUpdate.cpp
DaveBirdsall: Happy to consider. Can you point me to an instructive example?

core/sql/generator/GenRelUpdate.cpp
DaveBirdsall: Good question. I will investigate.

core/sql/optimizer/BindRelExpr.cpp
DaveBirdsall: You are correct! Good catch. I will remove it.

core/sql/generator/GenRelUpdate.cpp
zellerh: Thanks. Here is one example, but there are many others: FileScan::preCodeGen() in file GenPreCode.cpp:

```C++
                    ValueIdMap divExprMap;
                    ValueId computedBeg;

                    // If hashJoinBeg is :sysHV1 and the computed column
                    // expression is A/100, then the begin value for
                    // the computed column is :sysHV1/100. Do this
                    // rewrite by using a ValueIdMap
                    divExprMap.addMapEntry(underlyingCol, hashJoinBeg);
                    divExprMap.rewriteValueIdDown(computedColExpr->getValueId(),
                                                  computedBeg);
                    newBeg = computedBeg.getItemExpr();
```

There are also examples in file OptPhysRelExpr where we rewrite expressions on the select side of an insert/select into expressions on the insert side, using the ValueIdMap that can be accessed with Join::updateToSelectMap().

core/sql/generator/GenRelUpdate.cpp
zellerh: Yes, sorry to comment on surround code. If both source and target share the same NATable, then wouldn't the following be sufficient:

```c++
  if ((targetCol) && courceCol == targetCol)
    ...
```

Another option would be to compare their column number:

```c++
   CMPASSERT(sourceCol->getNATable() == targetCol->getNATable());
   if (targetCol && sourceCol->getPosition() == targetCol->getPosition())
      ...
```

One thing might be a problem: I wonder whether in some cases we might suppress caching of the NATable on the insert side and allow it on the select side or vice versa.

I wonder whether @sureshsubbiah has any additional comments on this topic.

core/sql/generator/GenRelUpdate.cpp
DaveBirdsall: I stepped through the code for an example of this. In the example I looked at, it appears the assign list is complete. My example was an unsalted table with three columns, and I did a MERGE that updated one column, with a set clause of C = 'GLAD'. I found that the assign list for the update record had all three columns in it:

(gdb) p targetRecExprArray.entries()
$2 = 3
(gdb) p targetRecExprArray[0].getItemExpr()->display()
(TRAFODION.SCH.T2.A ASSIGN TRAFODION.SCH.T2.A)
$3 = void
(gdb) p targetRecExprArray[1].getItemExpr()->display()
(TRAFODION.SCH.T2.B ASSIGN TRAFODION.SCH.T2.B)
$4 = void
(gdb) p targetRecExprArray[2].getItemExpr()->display()
(TRAFODION.SCH.T2.C ASSIGN %('GLAD'))
$5 = void
(gdb) 

In a nutshell, then, the constraint expression is generated whole, and the columns that aren't affected by the update simply refer to their old values.

It does seem like there is an optimization opportunity here: If the constraint expression doesn't refer to a changed column, we could omit it. If it is not too much trouble, I will do this.

zellerh: Thanks for validating this.

DaveBirdsall:
@zellerh, I've just submitted a rework based on your comments. I took your suggestions on all of them (thanks for the examples). I will run full regressions on a workstation overnight tonight.

core/sql/generator/GenRelUpdate.cpp
```diff
@@ -1718,19 +1806,56 @@ short HbaseUpdate::codeGen(Generator * generator)
...
-      GenAssert(FALSE, "Should not reach here. This update should have been transformed to delete/insert");
-      // To be uncommented when TRAFODION-1610 is implemented
-      // Need to generate insConstraintExpr also
-/*
+      // Generate the update and insert constraint check expressions
+
+      // The attributes for the columns referenced in the constraint expressions
+      // refer to the source values of the columns. We want to evaluate the
+      // constraints aganst the target values, though. So, there is some
+      // Attributes gymnastics that has to happen to generate them.
```
zellerh: Are the Attributes gymnastics gone with the new code? Is it now all couch potatoes?

DaveBirdsall: Oops... I should have deleted the comments from 1811-1814

DaveBirdsall: Or at least edited them

core/sql/generator/GenRelUpdate.cpp
```diff
@@ -1718,19 +1806,56 @@ short HbaseUpdate::codeGen(Generator * generator)
...
+      // Prepare the constraint tree for generation
+
       ItemExpr *constrTree =
         getCheckConstraints().rebuildExprTree(ITM_AND, TRUE, TRUE);
 
       if (getTableDesc()->getNATable()->hasSerializedEncodedColumn())
-        constrTree = generator->addCompDecodeForDerialization(constrTree);
+        constrTree = generator->addCompDecodeForDerialization(constrTree, isAlignedFormat);
+
+      // Generate the update constraint expression, substituting Attributes for
```
DaveBirdsall: Should have edited this comment too

core/sql/generator/GenRelUpdate.cpp
```diff
@@ -1718,19 +1806,56 @@ short HbaseUpdate::codeGen(Generator * generator)
...
+
+      genUpdConstraintExpr(generator,
+                           constrTree,
+                           constraintColumns,
+                           newRecExprArray(),
+                           &updConstraintExpr /* out */);
+
+      if ((isMerge()) && (mergeInsertRecExprArray().entries() > 0))
+        {
+          // Generate the insert constraint expression, substituting Attributes for
```
DaveBirdsall: And this one

PR #1254, opened 2017-10-03
[TRAFODION-2719] Check for truncation and four more fixes

zellerh:
This pull request combines five fixes, each in its own commit.

**[TRAFODION-2719] Check for truncation of character columns**

When inserting into a UTF-8 character column with a fixed number
of characters, do a check whether the source string contains no
more than the allowed number of characters (just checking the
byte length is not enough). This bug does not show when using
sqlci, since sqlci does its own check before calling the
executor. It only happens when using JDBC or trafci.

- core/sql/cli/CliExpExchange.cpp
- core/sql/exp/exp_expr.h
- core/sql/generator/GenExpGenerator.cpp
- core/sql/sqlcomp/DefaultConstants.h
- core/sql/sqlcomp/nadefaults.cpp

**[TRAFODION-2736] Missing predicates on salt columns**

In some cases, when we generate an index join (join of
an alternate index with the clustering index), we lost
the generated predicates on computed columns such as
salt and division. 

- core/sql/optimizer/RelScan.h
- partial: core/sql/optimizer/TransRule.cpp

**[TRAFODION-2751] Unnecessary PROBE_CACHE in an index join**

An index join produces unique matches from the index, so
a probe cache would be a waste of resources. Added a check
to suppress probe cache for index joins.

- core/sql/optimizer/OptLogRelExpr.cpp
- partial: core/sql/optimizer/TransRule.cpp
- core/sql/regress/core/EXPECTED005.SB

**[TRAFODION-2752] Fix error message when UDR method not found**

The error message displayed only the signature, but not the
method name.

- core/sql/regress/udr/EXPECTED100.SB
- core/sql/regress/udr/TEST100
- core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp

**[TRAFODION-2761] Error when all TMUDF columns are eliminated**

Fix a couple of places where we would core dump when all
columns of the input table of a TMUDF were eliminated. Also
removed an unused NAColumnArray method that has a bug.

- core/sql/executor/ExUdr.cpp
- core/sql/generator/GenUdr.cpp
- core/sql/optimizer/NAColumn.cpp
- core/sql/optimizer/NAColumn.h
- core/sql/regress/udr/EXPECTED001
- core/sql/regress/udr/TEST001
- core/sql/regress/udr/TEST001_Sessionize.java

**Other fixes:**

Change the swjdbc script to pick up the Trafodion version from
the current environment, not when we run install_traf_components.

- core/sqf/sql/scripts/install_traf_components


core/sql/optimizer/OptLogRelExpr.cpp
```diff
@@ -2394,7 +2394,7 @@ Join::synthConstraints(NormWA * normWAPtr)
   (void) leftGA.hasCardConstraint(minLeft,maxLeft);
   (void) rightGA.hasCardConstraint(minRight,maxRight);
 
-  if (maxLeft == 1)
+  if (maxLeft == 1 || isIndexJoin_)
```
DaveBirdsall: What happens in the case of joining a non-unique index to a base table? Would we want to have probe caches then?

zellerh: The non-unique index contains one row for each base table row, and it contains the clustering key of that row. So, there is still a 1:1 match between the non-unique index and the base table/clustering index, and a probe cache would not have any cache hits.

zellerh:
jenkins, retest

core/sql/optimizer/OptLogRelExpr.cpp
DaveBirdsall: Thanks. I guess I was thinking about joins to the index from elsewhere. But that's not what this code is dealing with.

zellerh:
There are three failures:

- seabase/TEST130: This looks like the "missing inserts" issue Sandhya is working on.
- core: Two tests, 131 and 162 fail with diffs in hive.hive.* tables, this may be a setup issue on the slave that runs the tests.

PR #1255, opened 2017-10-03
[TRAFODION-2762] Allow UPDATE STATS to create sample tables regardless

DaveBirdsall:
This change causes UPDATE STATISTICS to always set CQD ALLOW_NULLABLE_UNIQUE_KEY_CONSTRAINT 'ON' and CQD CONTROL QUERY DEFAULT CAT_ERROR_ON_NOTNULL_STOREBY 'OFF'.

This allows UPDATE STATISTICS to create a sample table with a nullable primary key (the first CQD) or a nullable STORE BY (the second CQD). In this way, the user does not need to remember to set these CQDs when doing UPDATE STATISTICS on a table that was created with either of these experimental features.


zellerh:
+1
Looks good.

PR #1257, opened 2017-10-04
[TRAFODION-2763] Left join with non-equi join predicate in ON clause …

sureshsubbiah:
…may cause corefiles

This PR also includes a small change for
[TRAFODION-2764] LOAD and UNLOAD statements with LOB columns cause runtime errors
The files affected by each change is different. Please see JIRA for a description of the problem and the fix.

2763
generator/GenPreCode.cpp
2764
bin/SqlciErrors.txt
optimizer/BindRelExpr.cpp


zellerh:
+1
Looks good to me.

PR #1258, opened 2017-10-05
[TRAFODION-2765] Change heuristics so MDAM is considered more often

DaveBirdsall:
This change adds new heuristic code to ScanOptimizer::useSimpleFileScanOptimizer (optimizer/ScanOptimizer.cpp). The new code checks to see if a single subset scan uses all the key predicates in the query. If so, we don't bother to cost MDAM. But if not, we do consider MDAM.

The old code contained some (probably unintended) odd heuristics. Given a salted table with keys A, B and C, and a query SELECT * FROM T WHERE A = 5 AND C = 20, the old code would consider MDAM for serial plans but not for parallel ones! This is because in the serial case, a single subset would be a full table scan (so it is quite reasonable to try MDAM), but in the parallel case, because the "_SALT_" column would be covered by an equality predicate, the single subset SearchKey would have equality predicates on a prefix of the key. It would assume that a single subset with equality on "_SALT_" and A was always better than considering MDAM to column C. However, if the UEC of B is low, and the UEC of C is high, using MDAM to column C can be quite efficient. The new heuristics will allow this possibility to be explored.

### DIFF
--- a/core/sql/optimizer/ScanOptimizer.cpp
+++ b/core/sql/optimizer/ScanOptimizer.cpp
@@ -3001,6 +3001,48 @@ ScanOptimizer::useSimpleFileScanOptimizer(const FileScan& associatedFileScan
                        ,indexDesc
                        );
 
+    if (CmpCommon::getDefault(MDAM_FSO_SIMPLE_RULE) == DF_ON)
+      {
+        // Quickly-computed special cases
+
+        if (searchKey.isUnique())
+          return TRUE;   // unique access, don't need to consider MDAM
+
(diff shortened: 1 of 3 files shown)


DaveBirdsall:
Forgot to mention: The new heuristic is guarded by a new CQD: it is used when CQD MDAM_FSO_SIMPLE_RULE is 'ON', which is the default. The old behavior can be had by setting this CQD to 'OFF'.

DaveBirdsall:
In parallel with Jenkins, I am running full regressions on a workstation overnight. Please do not merge until I report those results.

DaveBirdsall:
Full regressions also passed on my workstation overnight last night.

PR #1260, opened 2017-10-09
[TRAFODION-2768] Make Trafodion code base to compile in RH7

selvaganesang:
RH7 g++ compiler produces more warnings than RH6 g++.

This PR contains the following changes to compile successfully.

1. Embedding macro within the string literals expect a space character after “ as in {“create table “ SEABASE_SCHEMA” } - A space between “ and SEABASE.
2.  g++ is now data typed better. Incompatible casting or assignment will be flagged as errors except NULL can be assigned to pointer or bool or numeric variable due to -Wno-conversion-null.
3. Uninitialized variables are not permitted.
4. Cast the const references as references like ((FileScan &)getFileScan()).getHiveSearchKey() when needed.
4a.	Casting as (FileScan) will result in copying rather than passing in references. This can cause the changes to the object to be lost.
5. Missing return or invalid return type will be flagged as errors.
6. Avoid casting to void * wherever possible. NAHeap and ContextCli objects were passed in void * earlier.
7. str_spintf is now typedef to sprintf. So, invalid format specification will be flagged as error during compilation instead of failing at runtime because of -Wformat compiler option.
7a.	Printf format specification doesn’t truncate the data if it overflows the width specification. For eg %04d will print 5 digit if the number is greater than 9999. str_sprintf was truncating the data earlier.
8. Pointers are expected to be long data type (8 bytes) when stored in a numeric data type. I had to cast int32 or int variable as long in some places of the code to make it compile. It might need to be fixed correctly. Suggest to avoid storing pointers in as a numeric variable in future.


core/sqf/src/tm/tm.cpp
```diff
@@ -460,9 +460,13 @@ void tm_process_req_requestregioninfo(CTmTxMessage * pp_msg)
        TM_Txid_legacy lv_transid;
    } u;
 
-   char tname[2000], ername[50], rname[100], offline[20], regid[200], hostname[200], port[100];
+   char tname[2000];
+   tname[299] = '\0';
```
DaveBirdsall: Odd that we'd set just element 299 of a 2000-byte buffer to null. But I see that this is the way it was before.

selvaganesang:
jenkins, retest

arvind-narain:
jenkins, retest

arvind-narain:
jenkins, retest -- mvn repo errors

selvaganesang:
jenkins, retest

PR #1262, opened 2017-10-11
TRAFODION-2731 CodeCleanup: Phase 5. Cleanup of Platform.h

anoopsharma00:
Removal of unneeded/unused entries from common/Platform.h and
their usage from other files.

Removal of some other defines that are no longer used. There
might be a compilation error if an obsolete define is being used.
If that happens, those defines need to be removed from the code.


core/sql/common/Ipc.cpp
```diff
@@ -4580,44 +4565,21 @@ IpcServerClass::IpcServerClass(IpcEnvironment *env,
...
-	  // communicate via the message system.
-	  // in order to run in an NT only or simulated environment we can set an environment
-	  // variable to override that mechanism.
-	  if (getenv("SQL_NO_NSK_LITE") == NULL)
-	  {allocationMethod_ = IPC_LAUNCH_GUARDIAN_PROCESS;
-	  }
-	  else 
-	  {
-	  allocationMethod_ = IPC_LAUNCH_NT_PROCESS;
+      // NA_WINNT is set and NA_GUARDIAN_IPC is set
```
DaveBirdsall: Perhaps delete this one line of the comment since the defines are no longer used?

core/sql/common/Ipc.cpp
```diff
@@ -4580,44 +4565,21 @@ IpcServerClass::IpcServerClass(IpcEnvironment *env,
...
+      // The standard method on NT is to create a Guardian process
+      // in order to run in an NT only or simulated environment we can set an environment
+      // variable to override that mechanism.
+      if (getenv("SQL_NO_NSK_LITE") == NULL)
```
DaveBirdsall: Do we want to keep the NT code?

anoopsharma00: There are still some parts of code which use some of the NSK and NT constructs.
They need to be looked at carefully before removing them.

But yes, obvious obsolete references like the one you pointed out, will be removed
in a subsequent checkin.

PR #1264, opened 2017-10-12
[TRAFODION-2772] - retrieve a value from Json string got an error: Json value is invalid

andyyangcn:
[TRAFODION-2772] - retrieve a value from Json string got an error: Json value is invalid

### DIFF
--- a/core/sql/exp/exp_function.cpp
+++ b/core/sql/exp/exp_function.cpp
@@ -6506,10 +6506,14 @@ ex_expr::exp_return_type ex_function_json_object_field_text::eval(char *op_data[
 
     char *rltStr = NULL;
     char *jsonStr = new(heap) char[len1+1];
-    memset(jsonStr, 0, len1+1);
-    strncpy(jsonStr, op_data[1], len1);
     char *jsonAttr = new(heap) char[len2+1];
+    if (jsonStr == NULL || jsonAttr == NULL)
+    {
+        return ex_expr::EXPR_ERROR;
+    }
+    memset(jsonStr, 0, len1+1);
     memset(jsonAttr, 0, len2+1);
+    strncpy(jsonStr, op_data[1], len1);
     strncpy(jsonAttr, op_data[2], len2);
     JsonReturnType ret = json_extract_path_text(&rltStr, jsonStr, 1, jsonAttr);
     if (ret != JSON_OK)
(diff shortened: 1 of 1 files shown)


core/sql/exp/exp_function.cpp
```diff
@@ -6503,8 +6503,19 @@ ex_expr::exp_return_type ex_function_json_object_field_text::eval(char *op_data[
...
         len2 = Attributes::trimFillerSpaces( op_data[2], prec2, len2, cs );
     }
+
     char *rltStr = NULL;
-    JsonReturnType ret = json_extract_path_text(&rltStr, op_data[1], 1, op_data[2]);
+    char *jsonStr = new(heap) char[len1+1];
+    char *jsonAttr = new(heap) char[len2+1];
+    if (jsonStr == NULL || jsonAttr == NULL)
+    {
+        return ex_expr::EXPR_ERROR;
```
DaveBirdsall: We should add something to the diagnostic area when making this return. Though I expect that might fail with a memory error as well. Might be worthwhile to see how other functions behave when memory allocations fail.

core/sql/exp/exp_function.cpp
```diff
@@ -6503,8 +6503,19 @@ ex_expr::exp_return_type ex_function_json_object_field_text::eval(char *op_data[
...
+    }
+    memset(jsonStr, 0, len1+1);
+    memset(jsonAttr, 0, len2+1);
```
selvaganesang: Consider setting the null character after strncpy instead

core/sql/exp/exp_function.cpp
andyyangcn: I checked the whole file Exp_function.cpp, and found we haven't check if the pointer is NULL when we allocated memory from heap via new(heap). Does that mean it will always succeed to allocate memory from a memory pool. So, it seems I can remove the check.

core/sql/exp/exp_function.cpp
andyyangcn: strncpy(jsonStr, op_data[1], len1);
jsonStr[len1] = '\0';
Do you mean the change should look like above?

andyyangcn:
create a new post request for this.

core/sql/exp/exp_function.cpp
selvaganesang: Yes. No need to do memset.

PR #1265, opened 2017-10-12
[TRAFODION-2756] JDBC regression test case passed with some exceptions print

orrk:
These changes do the following:
1. Add some try-with-resource in test case to make sure the connection and statement are closed.
2. Add some catch block and fail information to make sure no exception when test pass


DaveBirdsall:
Shouldn't this pull request be against the apache:master branch? (Release 2.2 is about to go out the door.)

PR #1266, opened 2017-10-16
[TRAFODION-2771] Fix lob_guide/pom_xml to copy LOB guide to right place

DaveBirdsall:
The last pull request for this JIRA copied the LOB Guide to the sql_reference directory, overwriting the SQL Reference Manual. This one copies it to the right place on the web site.

### DIFF
--- a/docs/lob_guide/pom.xml
+++ b/docs/lob_guide/pom.xml
@@ -244,15 +244,15 @@
                   - target/docs/<version>/<document> contains the PDF version and the web book. The web book is named index.html
                 --> 
                 <!-- Copy the PDF file to its target directories -->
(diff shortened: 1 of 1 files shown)


DaveBirdsall:
I am merging this immediately because the SQL Reference Guide is presently not available on the web site and this change will fix that.

zellerh:
+1
Looks good to me, sorry I didn't catch that in the original PR. Was wondering why the directory was sql_reference but then thought that this would be a common directory for multiple manuals...

DaveBirdsall:
Thanks. Unfortunately it is hard to test that particular aspect out without actually merging the change. Sigh.

PR #1267, opened 2017-10-18
[TRAFODION-2775] Insert does not raise duplicate row error for hbase …

sureshsubbiah:
…format table with defaulted first column

There was a problem with column being checked during CheckAndPut call for such tables. We were always checking the first column in the row, which may not have existed in the previous version of this rowid. Please see JIRA for a complete explanation.

### DIFF
--- a/core/sql/exp/ExpHbaseInterface.cpp
+++ b/core/sql/exp/ExpHbaseInterface.cpp
@@ -842,7 +842,9 @@ Lng32 ExpHbaseInterface_JNI::insertRow(
     transID = getTransactionIDFromContext();
   retCode_ = client_->insertRow((NAHeap *)heap_, tblName.val, hbs_,
                                 useTRex_, transID, rowID, row, timestamp, 
-                                checkAndPut, asyncOperation, useRegionXn, &htc);
+                                checkAndPut, asyncOperation, useRegionXn, 
+				0, // checkAndPut is false, so colIndexToCheck is not used
+				&htc);
   if (retCode_ != HBC_OK) {
     return -HBASE_ACCESS_ERROR;
   }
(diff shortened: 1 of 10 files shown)


core/sql/exp/ExpHbaseInterface.cpp
```diff
@@ -842,7 +842,9 @@ Lng32 ExpHbaseInterface_JNI::insertRow(
     transID = getTransactionIDFromContext();
   retCode_ = client_->insertRow((NAHeap *)heap_, tblName.val, hbs_,
                                 useTRex_, transID, rowID, row, timestamp, 
-                                checkAndPut, asyncOperation, useRegionXn, &htc);
+                                checkAndPut, asyncOperation, useRegionXn, 
+				0, // checkAndPut is false, so colIndexToCheck is not used
```
DaveBirdsall: How will the insertRow method distinguish between the "not used" case and a column index of zero? (Elsewhere I saw a comment that the column index is 0-based.)

sureshsubbiah: The java code in HTableClient::putRow has this line of code which will ensure that the number passed in for colIndexToCheck is meaningless when checkAndPut is FALSE. -1 would have been easier to read. I can make that change later, if you agree.

if (checkAndPut && colIndex == colIndexToCheck) {

DaveBirdsall: Thanks; code change looks good to me.

selvaganesang:
In the HTableClient.putRow method there is a parameter columnToCheck which expects the column name instead of colIndex. To avoid confusion, please consider removing this parameter.

PR #1268, opened 2017-10-18
[TRAFODION-2776] Mdam plans with more than one disjunct sometimes cause either a compiler core or have an incorrect predicate

sureshsubbiah:
The problem is that during MdamKey::preCodegen we regenerate the disjuncts and expect to get the same number as we saw during optimize phase. This is mostly true, but in some cases not. There is some defensive code to handle this case, but it does not trigger when there is a partKey predicate. The fix removes the partKey predicate check for the defensive code.


sureshsubbiah:
jenkins retest

PR #1269, opened 2017-10-18
[TRAFODION-2777] Fix latent bug unmasked by JIRA TRAFODION-2765 fix

DaveBirdsall:
This fix comments out some code that now causes non-determinstic failures in test compGeneral/TEST006. The essence of the bug is that queries against tables lacking statistics would non-deterministically flip to MDAM. The commented-out code contains several issues; rather than fixing them it seems more efficient to simply continue with the MDAM costing code rewrite which is occurring under JIRA TRAFODION-2645.

### DIFF
--- a/core/sql/optimizer/ScanOptimizer.cpp
+++ b/core/sql/optimizer/ScanOptimizer.cpp
@@ -5702,6 +5702,63 @@ FileScanOptimizer::runMdamTests
 }
 #endif
 
+
+// The following method is now dead code and will soon be replaced functionally.
+// This method attempts to decide whether MDAM should be used on a table that
+// lack statistics. It does so by analyzing the key predicates, considering
+// the upper bounds on UECs implied by them and the number of ranges they
+// define.
+//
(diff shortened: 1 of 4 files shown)


zellerh:
+1
Looks good to me.

DaveBirdsall:
Two regress tests failed, seabase/TEST003 and TEST010. 

For TEST003, a plan switched over from MDAM to single subset, causing a different code path for predicate evaluation. That reveals an existing bug in the LARGEINT UNSIGNED implementation that will go away when all clients support the new data type. Nonetheless, a JIRA TRAFODION-2778 has been written to track that issue. For this pull request, I've added a "known diffs" file for this test.

For TEST010, several plans switched over from MDAM to single subset. The purpose of the test seems to be in part to exercise the MDAM run-time code, so I made changes to the test itself to force MDAM where necessary. So, for this pull request, I've changed the test script and the expected results.

DaveBirdsall:
Jenkins, retest

DaveBirdsall:
Note: I ran full regressions on this change overnight on a workstation. All tests passed.

PR #1271, opened 2017-10-19
[TRAFODION-2770] "Table not found" errors not being handled correctly by AQR for hive access

sandhyasun:
..and misc LOB related minor fixes


DaveBirdsall:
Is the JIRA number in the title correct?

selvaganesang:
Consider using snprintf wherever possible to avoid corruption especially when the printf buffer is in the stack,

sandhyasun:
Changes to use snprintf

PR #1272, opened 2017-10-19
[TRAFODION-2773] When the timeout is set, odb is still hang

SuJinpei:
we needn't call other function which may cause odb hang again when receiving a signal, we just want to exit.

### DIFF
--- a/core/conn/odb/src/odb.c
+++ b/core/conn/odb/src/odb.c
@@ -4156,18 +4156,17 @@ static void sigcatch(int sig)
     exit ( EX_SIGNAL );
 #else 
     if ( tn == 1 ) { /* single threaded */
-        tclean( 0 );
         gclean();
-        exit( EX_SIGNAL );
     } else {
         for ( i = 0 ; i < tn ; i++ ) {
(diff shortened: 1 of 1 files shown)


core/conn/odb/src/odb.c
```diff
@@ -4156,18 +4156,17 @@ static void sigcatch(int sig)
     exit ( EX_SIGNAL );
 #else 
     if ( tn == 1 ) { /* single threaded */
-        tclean( 0 );
         gclean();
-        exit( EX_SIGNAL );
     } else {
         for ( i = 0 ; i < tn ; i++ ) {
-            if ( !pthread_kill(thid[i], 0) ) {  /* If this thread is alive... */
+            if ( pthread_kill(thid[i], 0) ) {  /* If this thread is alive... */
```
selvaganesang: pthread_kill returns 0 when the thread exists. So, earlier code ( ! pthread_kill ...) is a valid one

SuJinpei: Yes, I misunderstood it.

PR #1273, opened 2017-10-20
[TRAFODION-2780] The mxosrvr dumps core when connection idle timer ex…

selvaganesang:
…pires at times

This core is accompanied by mxssmp process core.

Got rid of both connection idle timer and server idle timer, instead use the timeout
mechanism provided by the socket API select.  More code removed.

Tested the connection idle timeout mechanism via a sample JDBC T4 program just to ensure
that the mxosrvr is put into available pool after the connection idle timer expires.

Trafci seems to be resetting the connection idle timer periodically from the client.

It looks like even the connection idle timer is Trafodion specific concept because of
a dedicated process for every database connection to Trafodion SQL engine. From JDBC specification,
the connection inactive timeout needs to be implemented in the connection pooling infrastructure.
The default connection idle timer in Trafodion is 10 mins.

### DIFF
--- a/core/conn/odbc/src/odbc/nsksrvr/SrvrConnect.cpp
+++ b/core/conn/odbc/src/odbc/nsksrvr/SrvrConnect.cpp
@@ -219,20 +219,6 @@ Int32 SQL_EXEC_GetAuthState(
 
 #define MAX_EVAR_VALUE_LENGTH 3900 + 1
 
-#define START_CONN_IDLE_TIMER \
-	if (srvrGlobal != NULL && \
-		srvrGlobal->srvrState == SRVR_CONNECTED && \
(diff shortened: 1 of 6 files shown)


selvaganesang:
@arvind-narain Can you please review it too

core/conn/odbc/src/odbc/nsksrvr/SrvrConnect.cpp
```diff
@@ -1190,10 +1176,6 @@ ImplInit (
    srvrGlobal->receiveThrId = getpid();
    timer_register();
```
arvind-narain: Thanks for the change Selva. 
Could you please undo the change for StatisticsTimerHandle ? Thanks

selvaganesang: Thanks for catching this inadvertent deletion of timer initialization call.

PR #1276, opened 2017-10-24
[TRAFODION-2781] Omit LOB columns from update stats sample tables

DaveBirdsall:
UPDATE STATISTICS does not support statistics on LOB columns. However, it was including LOB columns when it creates sample tables. This is unnecessary overhead, which could be eliminated.

This set of changes eliminates the LOB columns from the sample tables, and eliminates the logic that populates them. There are two parts to the changes:

1. A new clause, WITHOUT LOB COLUMNS, has been added to the CREATE TABLE LIKE syntax. So, one can say, CREATE TABLE T1 LIKE T WITHOUT LOB COLUMNS. The target table contains all the columns of the source table except for LOB columns. UPDATE STATS code was changed to use this new syntax when creating sample tables.

2. Logic in UPDATE STATS code to populate sample tables is now careful to generate an explicit select list rather than relying on "SELECT *". It was already doing this for the case of over-sized character type columns; now it does so always. LOB columns are excluded from the select list.


zellerh:
+1
Looks good to me.

DaveBirdsall:
FYI: I ran full regressions on this change on a workstation overnight and they passed.

PR #1277, opened 2017-10-26
[TRAFODION-2783] jdbc_test_cdh fails at times with type 2 JDBC driver…

selvaganesang:
… dumping core

In case of Type 2 JDBC driver, the Trafodion SQL engine is a library that is dynamic
loaded into the process. Initialization of C++ static objects in the dynamic loaded libraries
are supposed to be done before dlopen returns. But the behavior seems to be nondeterministic when
there are multiple threads and when there are dependent static objects (An static object expects
another to be initialized before it). I think, the order of the initialization is not guaranteed
by the standard.

Refactored the code to initialize static object CharInfo::builtinCollationDB_ in a thread safe manner.

### DIFF
--- a/core/sql/common/charinfo.cpp
+++ b/core/sql/common/charinfo.cpp
@@ -54,6 +54,7 @@
 #include "CmpConnection.h"
 #include "CmpContext.h"
 #include "CmpCommon.h"
+#include "CliSemaphore.h"
 
 using namespace std;
 
@@ -364,7 +365,8 @@ CollationDB::CollationDB(CollHeap *h)
   : CollationDBSupertype(h), heap_(h), refreshNeeded_(TRUE)
 {
     if (this == CharInfo::builtinCollationDB_) return;
-    cmpCurrentContext->getCollationDBList()->insert(this);
+    if (cmpCurrentContext != NULL)
+       cmpCurrentContext->getCollationDBList()->insert(this);
 }
 
 CollationDB::CollationDB(CollHeap *h, const CollationInfo *co, size_t count)
@@ -430,7 +432,8 @@ CollationDB * CollationDB::nextCDB() const
 
 const CollationInfo* CollationDB::getCollationInfo(CharInfo::Collation co) const
 {
-  CollIndex i, n = entries();
+  CollIndex i, n;
+  n = entries();
   for (i = 0; i < n; i++)
     if (co == at(i)->co_)
       return at(i);
@@ -533,7 +536,7 @@ static const CollationInfo mapCOArray[] = {
 
 #define SIZEOF_CO (sizeof(mapCOArray)/sizeof(CollationInfo))
 
-const CollationDB *CharInfo::builtinCollationDB_;
+const CollationDB *CharInfo::builtinCollationDB_ = NULL;
 
 CharInfo::Collation CharInfo::getCollationEnum(const char* name,
 					       NABoolean formatNSK,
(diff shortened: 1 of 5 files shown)


core/sql/common/charinfo.h
```diff
@@ -260,11 +260,14 @@ class CharInfo
...
 
+  static void initBuiltinCollationDB();
+ 
+
 private:
 friend class CollationDB;			// needs to access builtinCDB_
 
    static const char*	const localeCharSet_;
-   static const CollationDB   builtinCollationDB_;
+   static const CollationDB   *builtinCollationDB_;
```
DaveBirdsall: Suggest initializing to NULL. So if you get a thread race condition, you'll get better behavior if one thread accessing this pointer beats the other thread that is initializing it.

core/sql/common/charinfo.cpp
```diff
@@ -363,21 +363,22 @@ void CollationInfo::display() const
 CollationDB::CollationDB(CollHeap *h)
   : CollationDBSupertype(h), heap_(h), refreshNeeded_(TRUE)
 {
-    if (this == &CharInfo::builtinCollationDB_) return;
+    if (this == CharInfo::builtinCollationDB_) return;
```
DaveBirdsall: I'm confused. Is CollationDB a singleton object? Or is there one global version, and other parts of the code might create their own version?

core/sql/common/charinfo.cpp
```diff
@@ -529,8 +530,10 @@ static const CollationInfo mapCOArray[] = {
   CollationInfo(NULL, CharInfo::UNKNOWN_COLLATION,  SQLCOLLATIONSTRING_UNKNOWN,
   			STATIC_NEG)
 };
-static const size_t SIZEOF_CO = sizeof(mapCOArray)/sizeof(CollationInfo);
-const CollationDB CharInfo::builtinCollationDB_(NULL, mapCOArray, SIZEOF_CO);
+
+#define SIZEOF_CO (sizeof(mapCOArray)/sizeof(CollationInfo))
+
+const CollationDB *CharInfo::builtinCollationDB_;
```
DaveBirdsall: Suggest initializing this global pointer to null, so you get better behavior in race conditions.

core/sql/common/charinfo.cpp
```diff
@@ -693,3 +696,9 @@ Int32 CharInfo::getMaxConvertedLenInBytes(CharSet sourceCS,
   return ((sourceLenInBytes/minBytesPerChar(sourceCS)) *
           maxBytesPerChar(targetCS));
 }
+
+void CharInfo::initBuiltinCollationDB()
+{
+   builtinCollationDB_ = new CollationDB(NULL, mapCOArray, SIZEOF_CO);
```
DaveBirdsall: Is mapCOArray another global object? If it is const it is OK. If not, how do we know that it has been constructed already?

core/sql/common/charinfo.h
DaveBirdsall: Oh, sorry; this is the declaration rather than the definition. Ignore my comment.

selvaganesang:
Thanks Dave. I am pushing another commit after more testing in my work space. This code seem to be confusing to me too. But, what I gather is that there is a static member variable CharInfo::builtinCollationDB_ that needs to constructed in a thread safe manner and it is dependent on another static const object.
So, I let C++ prologue to initialize the dependent static object and construct this CharInfo::builtinCollationDB_ in a thread safe manner. In earlier commits I had piggy backed on the CLI code to construct this static variable and just tested if sqlci and some select worked fine. It turned out that arkcmp and other trafodion processes require this static variable to be constructed without CLI

core/sql/common/charinfo.cpp
selvaganesang: Yes. C++ prologue ensures that it is initialized correctly. In master process, it will be initialized before dlOpen and in other processes it should be done before main is called.

core/sql/common/charinfo.cpp
selvaganesang: It is required in the rework and it is initialized to NULL

core/sql/common/charinfo.cpp
selvaganesang: CharInfo::builtinCollationDB_ is a singleton object. It looks like there could other CollationDB objects, but it will have all the data in the builtinCollationDB object too

PR #1278, opened 2017-10-27
TRAFODION-2784 trafci help has no description for jline

gpj1987:
when do trafci.sh -h there is no description for how to use jline


svarnau:
jenkins, ok to test

PR #1279, opened 2017-10-27
[TRAFODION-2420] RMS Enhancements

selvaganesang:
Refactored the ESP memory management so that all the SQL memory both
heap and space objects are accounted correctly in RMS

With this change, the total SQL memory used by all processes involved
in the query is represented in the counter "SQL Heap WM"


selvaganesang:
@zellerh Can you please review this PR

PR #1282, opened 2017-10-30
Trafodion-2705 user has "SHOW" privilege can't do 'showddl user'

robertamarton:
Showddl code now checks to see if current user matches the user name specified
in the showddl command.  It also verifies that the user has SHOW component
privilege.

Simplified privs1/TEST141 as part of goal to make privilege tests run faster.


selvaganesang:
Changes Looks good. It is good to see exception is being changed to passing error via return code because Trafodion has rarely used exception concepts. It makes it less confusing in Trafodion code base.

selvaganesang:
I have comment on the usage of assert in some of the methods called in by this change like ComUser::getCurrentUser.  The assert API may be no-op in Trafodion because some of the cpp file and Makefile used in Trafodion sets in the macro NDEBUG. This can cause non-deterministic behavior when these methods relies on assert but it becomes no-op.

robertamarton:
I will look at the use of assert and return internal error, when it make sense.  Thanks for the info.

PR #1285, opened 2017-10-31
TRAFODION-2789 jdbc does not implement NetworkTimeout related api

gpj1987:
implement networkTimeout relate api


svarnau:
jenkins, ok to test

selvaganesang:
Can you please explain how the network timeout is implemented in JDBC driver/server.  If there is no underlying support, it might give false impression that it is supported

selvaganesang:
Looked at the T4 driver code and it looks like the network timeout is being used in the socket layer. However, I didn't review the code in detail to figure out if it kicks in.

So I would say +1 for this change

PR #1289, opened 2017-11-02
[TRAFODION-2792] Single character change to stop leaking temp files

svarnau:


selvaganesang:
+1 This should be the minimum change one could do and still get the desired results

PR #1299, opened 2017-11-15
[TRAFODION-2806] optimize 'hbase classpath' cmd in sqenvcom.sh

mkby:
cache the output of 'hbase classpath' to a file under $TRAF_VAR/hbase_classpath, then sqenvcom.sh will get the hbase lib from this file if exists.

### DIFF
--- a/core/sqf/sqenvcom.sh
+++ b/core/sqf/sqenvcom.sh
@@ -286,6 +286,17 @@ else
   export LOC_JVMLIBS=$JAVA_HOME/jre/lib/i386/server
 fi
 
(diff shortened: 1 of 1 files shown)


svarnau:
Looks good.

svarnau:
I think there is a problem with this change. It is not sufficient stand-alone.

I have recently tested installation, and trafstart failed because the classpath did not get refreshed after installation of TRX files in HBase.

I suggest that a good time to remove the cached file (to update the cache) is not necessarily at installation time, but at trafodion start time.  Just in case HBase changed something since we previously started.

PR #1300, opened 2017-11-16
[TRAFODION-2807] We need support ODBC windows 32 bit driver build.

SuJinpei:
add a bat script to build 32-bit driver.
add iss script to generate 32-bit driver installer.
add 32-bit driver build steps to README


SuJinpei:
Jenkins retest

PR #1301, opened 2017-11-16
new COMMENT-ON SQL statement: resolve merge conflict

EEDY:
1. resolve enum type conflict while merging to esgyn

Compiled and test.


DaveBirdsall:
What JIRA is this associated with? (It's OK to use the same one as your previous COMMENT ON request)

PR #1306, opened 2017-11-20
[TRAFODION-2813] Ignore leading salt columns when checking index compatibility

DaveBirdsall:
See the JIRA for fix details.

### DIFF
--- a/core/sql/optimizer/IndexDesc.cpp
+++ b/core/sql/optimizer/IndexDesc.cpp
@@ -823,7 +823,7 @@ IndexProperty::compareIndexPromise(const IndexProperty *ixProp) const
             else if ( indexCol->getDefinition() != otherIndexCol->getDefinition() )
               return INCOMPATIBLE;
             else
-              done = TRUE;  // leading column (ignoring "_SALT_" if both have "_SALT_") is different
+              done = TRUE;  // leading column (ignoring "_SALT_" if both have "_SALT_") is the same; do Kb checks
           }
       }
 
--- a/core/sql/optimizer/IndexDesc.cpp
+++ b/core/sql/optimizer/IndexDesc.cpp
@@ -802,10 +802,30 @@ IndexProperty::compareIndexPromise(const IndexProperty *ixProp) const
 
     const IndexDesc * index = getIndexDesc();
     const IndexDesc * otherIndex = ixProp->getIndexDesc();
-    if ( ((IndexColumn *)(index->getIndexKey()[0]).getItemExpr())->getDefinition() != 
-         ((IndexColumn *)(otherIndex->getIndexKey()[0]).getItemExpr())->getDefinition() )
-
-      return INCOMPATIBLE;
+    
+    // If the two indexes have differing leading columns, consider them incompatible.
+    // For this check, we ignore the "_SALT_" column if both are salted.
+    CollIndex columnToCheck = 0;
+    NABoolean done = FALSE;
+    while (!done)
+      {
+        if (columnToCheck >= index->getIndexKey().entries())
+          return INCOMPATIBLE;  // must be one of the indexes is just "_SALT_" (seems unlikely actually)
+        else if (columnToCheck >= otherIndex->getIndexKey().entries())
+          return INCOMPATIBLE;  // must be one of the indexes is just "_SALT_" (seems unlikely actually)
+        else
+          {
+            IndexColumn * indexCol = (IndexColumn *)(index->getIndexKey()[columnToCheck]).getItemExpr();
+            IndexColumn * otherIndexCol = (IndexColumn *)(otherIndex->getIndexKey()[columnToCheck]).getItemExpr();
+            if ( indexCol->getNAColumn()->isSaltColumn() &&
+                 otherIndexCol->getNAColumn()->isSaltColumn() )
+              columnToCheck++;
+            else if ( indexCol->getDefinition() != otherIndexCol->getDefinition() )
+              return INCOMPATIBLE;
+            else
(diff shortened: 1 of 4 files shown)


DaveBirdsall:
Do not merge yet; I will be running full regressions overnight. (There may be some plan changes in regression tests due to this change.)

core/sql/optimizer/IndexDesc.cpp
```diff
@@ -802,10 +802,30 @@ IndexProperty::compareIndexPromise(const IndexProperty *ixProp) const
...
+          {
+            IndexColumn * indexCol = (IndexColumn *)(index->getIndexKey()[columnToCheck]).getItemExpr();
+            IndexColumn * otherIndexCol = (IndexColumn *)(otherIndex->getIndexKey()[columnToCheck]).getItemExpr();
+            if ( indexCol->getNAColumn()->isSaltColumn() &&
+                 otherIndexCol->getNAColumn()->isSaltColumn() )
+              columnToCheck++;
+            else if ( indexCol->getDefinition() != otherIndexCol->getDefinition() )
+              return INCOMPATIBLE;
+            else
+              done = TRUE;  // leading column (ignoring "_SALT_" if both have "_SALT_") is different
```
DaveBirdsall: This comment is incorrect... it should read "leading column ... is the same"

core/sql/optimizer/IndexDesc.cpp
```diff
@@ -802,10 +802,30 @@ IndexProperty::compareIndexPromise(const IndexProperty *ixProp) const
...
+        if (columnToCheck >= index->getIndexKey().entries())
+          return INCOMPATIBLE;  // must be one of the indexes is just "_SALT_" (seems unlikely actually)
+        else if (columnToCheck >= otherIndex->getIndexKey().entries())
+          return INCOMPATIBLE;  // must be one of the indexes is just "_SALT_" (seems unlikely actually)
+        else
+          {
+            IndexColumn * indexCol = (IndexColumn *)(index->getIndexKey()[columnToCheck]).getItemExpr();
+            IndexColumn * otherIndexCol = (IndexColumn *)(otherIndex->getIndexKey()[columnToCheck]).getItemExpr();
+            if ( indexCol->getNAColumn()->isSaltColumn() &&
+                 otherIndexCol->getNAColumn()->isSaltColumn() )
```
zellerh: Looks good. I am wondering whether we should do the same thing with divisioning columns (method NAColumn::isDivisioningColumn()).

DaveBirdsall: I considered that. I think the intent of the code in IndexProperty::compareIndexPromise is, for each user column, to retain the "best" (in some sense) index that begins with that column. The salt column is not a user column in any sense; with the exception of pre-partitioning rows across regions, a user would never have a predicate on salt. A divisioning column is a fuzzy area. It is a system-generated column, but it reflects true user content. A user predicate may well result in a predicate on the divisioning column. So, I thought it was more in tune with the spirit of IndexProperty::compareIndexPromise to treat divisioning columns like user columns. But I do not feel strongly about it; a case could be made the other way.

DaveBirdsall:
Full regressions passed on a workstation last night, so there were no plan changes in the regress tests.

core/sql/optimizer/IndexDesc.cpp
zellerh: Just wanted to bring this up, I can also live with both solutions. The computation of predicates for salt and division columns is similar, both are computed from user columns. The main difference is that the expressions of divisioning columns are monotonically increasing. I would have thought that the following indexes should be treated as incompatible:

create index ix1 on t(a) division like table;
create index ix2 on t(b) division like table;

DaveBirdsall: It depends. Suppose the DIVISION BY clause is on C. Should we think of ix1 as an index on C, A or just on A? If I have a query, select c from table where c = value, then ix1 and ix2 are compatible for that query. Today, IndexProperty::compareIndexPromise doesn't take the select predicates into account; it is just looking for the "best" index on C. (And on A. And on B.)

zellerh: IMHO the user should consider this a partitioned/divisioned index on A and not an index on (C, A), and so should we.

DaveBirdsall: OK. Now a question: Can an index have a different divisioning column than the base table? Or can I assume that if NAColumn::isDivisioningColumn() returns true for both index and otherIndex that they are the same column?

zellerh: Right now, all we allow is DIVISION LIKE TABLE, the index must have the same divisioning columns as the base table, if any. So, it's the same restriction we have for salting.

DaveBirdsall:
Do not merge this yet. I intend to change it to address the divisioning column also.

DaveBirdsall:
The rework contains two more changes:

1. I implemented @zellerh 's suggestion to skip divisioning columns also when computing index compatibility.
2. The unit tests for this encountered a core in GenPreCode.cpp, where the ValueId for the divisioning column was not in the MapTable when the plan was generated against an index. @nonstop-qfchen encountered this bug also and suggested a fix; @zellerh reworked this fix, offering a change to optimizer/ItemExpr.cpp. This rework includes the latter change, along with changes to regression test compGeneral/TEST071 to exercise this code path. Thanks to @zellerh and @nonstop-qfchen for this change.

I will re-run full regressions overnight on a workstation before merging this change.

zellerh:
+1
Looks good to me

DaveBirdsall:
Full regressions run on a workstation passed overnight.

PR #1307, opened 2017-11-21
[TRAFODION-2791] 'Not casespecific' column comparison returns wrong

duoluodexiaokeke:
'Not casespecific' column comparison returns wrongresult,  now like and t.c1=t.c2 not support

### DIFF
--- a/core/sql/optimizer/ItemColRef.h
+++ b/core/sql/optimizer/ItemColRef.h
@@ -350,6 +350,14 @@ public:
            NAMemory * outHeap = CmpCommon::statementHeap()
         );
 
+  ConstValue(const NAString& strval,
+           NABoolean isCaseInSensitive,
+           enum CharInfo::CharSet charSet=CharInfo::DefaultCharSet,
+           enum CharInfo::Collation collation=CharInfo::DefaultCollation,
+           enum CharInfo::Coercibility coercibility=CharInfo::COERCIBLE,
+           NAMemory * outHeap = CmpCommon::statementHeap()
+        );
+
   // constructor for a wide (unicode) string constant
   ConstValue(const NAWString& strval,
            enum CharInfo::CharSet charSet=CharInfo::UNICODE,
@@ -358,6 +366,16 @@ public:
            NAMemory * outHeap = CmpCommon::statementHeap(),
            enum CharInfo::CharSet strLitPrefixCharSet=CharInfo::UnknownCharSet
         );
+
+  ConstValue(const NAWString& strval,
+           NABoolean isCaseInSensitive,
+           enum CharInfo::CharSet charSet=CharInfo::UNICODE,
+           enum CharInfo::Collation collation=CharInfo::DefaultCollation,
+           enum CharInfo::Coercibility coercibility=CharInfo::COERCIBLE,
+           NAMemory * outHeap = CmpCommon::statementHeap(),
+           enum CharInfo::CharSet strLitPrefixCharSet=CharInfo::UnknownCharSet
+        );
+
 
 // constructor for a string constant with unknown charset (both the
 // single-byte and double-byte string values are known)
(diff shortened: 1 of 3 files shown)


svarnau:
jenkins, ok to test

core/sql/optimizer/OptRange.cpp
```diff
@@ -503,7 +503,13 @@ ConstValue* OptRangeSpec::getConstOperand(ItemExpr* predExpr, Lng32 constInx)
   // currently support. Predicates involving types not yet supported will be
   // treated as residual predicates.
   if (QRDescGenerator::typeSupported(static_cast<ConstValue*>(right)->getType()))
+  {
```
zellerh: This adds a case-insensitive flag to the type in the RangeSpec, but I don't think RangeSpecs are written to handle case-insensitive comparisons. Take a look at the methods that deal with comparisons when building RangeSpecs, in file Range.cpp. So, I think you would have to do one of two things: a) disable the RangeSpec transformation for case-insensitive comparison operators (the easy way) or change the RangeSpec methods to handle case-insensitive comparisons.

core/sql/optimizer/ItemColRef.h
```diff
@@ -350,6 +350,14 @@ class ConstValue : public ItemExpr
            NAMemory * outHeap = CmpCommon::statementHeap()
         );
 
+  ConstValue(const NAString& strval,
+           NABoolean isCaseInSensitive,
+           enum CharInfo::CharSet charSet=CharInfo::DefaultCharSet,
+           enum CharInfo::Collation collation=CharInfo::DefaultCollation,
+           enum CharInfo::Coercibility coercibility=CharInfo::COERCIBLE,
+           NAMemory * outHeap = CmpCommon::statementHeap()
+        );
```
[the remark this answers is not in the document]
duoluodexiaokeke: first, I think isCaseInSensitive is a import attribute,
second, i don't think add this argument to the end of constructor is a good way.
if add this argument in the second position, it will change some code where call it.
so i add a new constructor.

core/sql/optimizer/OptRange.cpp
duoluodexiaokeke: I do those operators in order to add case-insensitive to new constvalue.
i don't understand the a
for b, case-insensitive only use in 
             if (parentOfStart)
                parentOfStart->child(1) =
                    new(mvqrHeap_) ConstValue(charSubrange->start,
 if add a RangeSpec methods about case-insensitive , we need add this attribute to parentOfStart->child(1) after operating above.
so i don't think this is a good way

core/sql/optimizer/ItemColRef.h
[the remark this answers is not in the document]
zellerh: Yes, you need to add it as the last argument of the existing constructors.

core/sql/optimizer/OptRange.cpp
zellerh: RangeSpecs keep a list of points and intervals and these intervals need to be ordered, otherwise RangeSpec doesn't work. If we have case-insensitive comparison, that ordering needs to be different, but as far as I could tell, there is no code at the moment that would do a case-insensitive ordering of the values.

zellerh:
-1
You may need to study RangeSpecs a bit more before solving this problem.

zellerh:
Hi again, here is a suggestion: Change method OptRangeSpec::buildRange() and add a check whether a comparison operation (ITM_EQUAL, ITM_LESS, etc.) involves a case-inspecific comparison. If so, then set the local variable isRange to FALSE. This should avoid getting into this problem, as there will be no more RangeSpecs for case-inspecific comparisons.

DaveBirdsall:
core/TEST032 and core/TEST038 are failing with compiler assertion errors. Those need to be fixed before we can merge this change.

PR #1312, opened 2017-11-27
[TRAFODION-2819] Fix 4247 error on drop table when certain CQDs are present

DaveBirdsall:
When CQD TRAF_MAX_CHARACTER_COL_LENGTH is set to a value lower than '500', then DROP TABLE will fail when it tries to do a DELETE under the covers on the SB_HISTOGRAMS table. The latter fails with error 4247, raised in CmpSeabaseDDL::getColInfo (sqlcomp/CmpSeabaseDDLcommon.cpp).

The fix is to change CmpSeabaseDDL::getColInfo to exempt metadata, histogram and repository tables from the maximum column length check.


DaveBirdsall:
FYI: I ran full regressions on this change overnight on a clean workstation instance. All passed.

PR #1314, opened 2017-11-28
[TRAFODION-2821] Trafodion core code base needs to be thread safe

selvaganesang:
The method Ids of the different Java classes are now initialized
in a thread safe manner in JNI layer.

The missing code to initialize the thread specific JNIEnv are added.

### DIFF
--- a/core/sql/executor/OrcFileReader.cpp
+++ b/core/sql/executor/OrcFileReader.cpp
@@ -285,9 +285,10 @@ OFR_RetCode OrcFileReader::isEOF(bool& isEOF)
 //OFR_RetCode OrcFileReader::fetchNextRow(Int64 stopOffset, char* buffer)
 OFR_RetCode OrcFileReader::fetchNextRow(char * buffer, long& array_length, long& rowNumber, int& num_columns)
 {
-/*
+
   if (initJNIEnv() != JOI_OK)
      return OFR_ERROR_FETCHROW_EXCEPTION;
+/*
   // java.lang.String fetchNextRow(long stopOffset);
   jstring jresult = (jstring)jenv_->CallObjectMethod(javaObj_, JavaMethods_[JM_FETCHROW2].methodID, stopOffset);
   if (jresult==NULL && getLastError()) 
@@ -376,7 +377,6 @@ OFR_RetCode OrcFileReader::fetchNextRow(char * buffer, long& array_length, long&
         }
 
 	jenv_->GetByteArrayRegion(jrow, 0, row_length, (jbyte*)buffer);
- 	jenv_->DeleteLocalRef(jrow);  
 
   jenv_->PopLocalFrame(NULL);
   return (OFR_OK);
@@ -494,5 +494,5 @@ Removed until implemented
   QRLogger::log(CAT_SQL_HDFS_ORC_FILE_READER, LL_DEBUG, "  =>Returning %d, read %ld bytes in %d rows.", retCode, bytesRead, rowsRead);
   return retCode;
 */
-  return OFR_OK;
+  return (OFR_NOMORE);
 }
--- a/core/sql/executor/OrcFileReader.cpp
+++ b/core/sql/executor/OrcFileReader.cpp
@@ -30,6 +30,8 @@
 
 JavaMethodInit* OrcFileReader::JavaMethods_ = NULL;
 jclass OrcFileReader::javaClass_ = 0;
+bool OrcFileReader::javaMethodsInitialized_ = false;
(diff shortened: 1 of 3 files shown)


core/sql/executor/OrcFileReader.cpp
```diff
@@ -111,9 +133,15 @@ OFR_RetCode OrcFileReader::init()
...
     JavaMethods_[JM_CLOSE     ].jm_name      = "close";
     JavaMethods_[JM_CLOSE     ].jm_signature = "()Ljava/lang/String;";
-   
-    return (OFR_RetCode)JavaObjectInterface::init(className, javaClass_, JavaMethods_, (Int32)JM_LAST, FALSE);
+    
+    lv_retcode = (OFR_RetCode)JavaObjectInterface::init(className,
+							javaClass_,
+							JavaMethods_,
+							(Int32)JM_LAST, javaMethodsInitialized_);
+    javaMethodsInitialized_ = TRUE;
```
DaveBirdsall: Nit. This should be "true" instead of "TRUE" since this is a bool instead of an NABoolean. But it's an unobservable bug.

core/sql/executor/OrcFileReader.cpp
```diff
@@ -122,24 +150,28 @@ OFR_RetCode OrcFileReader::init()
 OFR_RetCode OrcFileReader::open(const char* path)
 {
   QRLogger::log(CAT_SQL_HDFS_ORC_FILE_READER, LL_DEBUG, "OrcFileReader::open(%s) called.", path);
+
+  if (initJNIEnv() != JOI_OK)
+     return OFR_ERROR_OPEN_PARAM;
   jstring js_path = jenv_->NewStringUTF(path);
   if (js_path == NULL) 
+  {
+    jenv_->PopLocalFrame(NULL);
```
DaveBirdsall: I'm a little confused about the rules for when a call to PopLocalFrame is needed. Is it needed whenever we have previously done a call into any jenv_ method? (in this case NewStringUTF).

core/sql/executor/OrcFileReader.cpp
```diff
@@ -236,58 +314,71 @@ OFR_RetCode OrcFileReader::fetchNextRow(char * buffer, long& array_length, long&
...
+        if (jrow == NULL)
+        {
+           jenv_->PopLocalFrame(NULL);
+           return (OFR_ERROR_FETCHROW_EXCEPTION);
+        }
 
-		jenv_->GetByteArrayRegion(jrow, 0, row_length, (jbyte*)buffer);
- 		jenv_->DeleteLocalRef(jrow);  
+	jenv_->GetByteArrayRegion(jrow, 0, row_length, (jbyte*)buffer);
+ 	jenv_->DeleteLocalRef(jrow);
```
DaveBirdsall: I assume the DeleteLocalRef call is no longer needed (since we do a PopLocalFrame after this) but it is harmless?

core/sql/executor/SequenceFileReader.cpp
```diff
@@ -189,9 +192,10 @@ SFR_RetCode SequenceFileReader::open(const char* path)
   if (jresult != NULL)
   {
     logError(CAT_SQL_HDFS_SEQ_FILE_READER, "SequenceFileReader::open()", jresult);
+    jenv_->PopLocalFrame(NULL);
```
DaveBirdsall: The DeleteLocalRef at line 190 is no longer needed, but harmless, right?

core/sql/executor/OrcFileReader.cpp
selvaganesang: PopLocalFrame should be called before returning from the method after initJNIEnv() is called.

core/sql/executor/OrcFileReader.cpp
selvaganesang: Yes. Will remove

core/sql/executor/SequenceFileReader.cpp
selvaganesang: Yes. Will remove

core/sql/executor/OrcFileReader.cpp
selvaganesang: TRUE is defined in the standard include file glib.h as

#define	FALSE	(0)
#define	TRUE	(!FALSE)

I would think it should be ok to initialize bool datatype as TRUE.

DaveBirdsall: That's fine.

core/sql/executor/OrcFileReader.cpp
DaveBirdsall: Thanks.

PR #1316, opened 2017-11-29
[TRAFODION-2822] Make [first n] views non-updatable; prevent bad MERGE plans

DaveBirdsall:
This pull request contains two changes:

1. Views defined using [first n] or [any n] are now marked as not updatable and not insertable. So, when a MERGE statement is attempted on a new view of this type, we will get a compile time error that the view is not insertable or updatable (instead of getting an incorrect result).

2. MERGE statements that have a WHEN NOT MATCHED INSERT action are prevented from undergoing a tuple substitution transformation (that is, the TSJRule and TSJFlowRule will not fire on a merge node possessing a WHEN NOT MATCHED INSERT action). This is necessary because the run-time implementation of insert actions takes place in the merge node itself; that is, scanning has to happen in the merge node. These two rules would take the scanning out of the merge node into a separate scan node.

Note: Existing [first n] / [any n] views remain marked as updatable and insertable. (These attributes are calculated at CREATE VIEW time and stored in the metadata.) The second change above will cause MERGE statements having WHEN NOT MATCHED INSERT actions to fail at compile time against such views with an error 2235 (Optimizer could not produce a plan for the statement). While this is a non-obvious error from a user perspective it is far better than allowing execution and getting an incorrect result.

### DIFF
--- a/core/sql/optimizer/NormRelExpr.cpp
+++ b/core/sql/optimizer/NormRelExpr.cpp
@@ -6995,11 +6995,6 @@ NABoolean RelRoot::isUpdatableBasic(NABoolean isView,
   // QSTUFF
 
     {
-      // if child is a FirstN node, skip it.
-      if ((child(0)->castToRelExpr()->getOperatorType() == REL_FIRST_N) &&
-	  (child(0)->child(0)))
-	scan = (Scan *)child(0)->child(0)->castToRelExpr();
-      else
 	scan = (Scan *)child(0)->castToRelExpr();
     }
 
--- a/core/sql/optimizer/RelExpr.cpp
+++ b/core/sql/optimizer/RelExpr.cpp
@@ -13535,6 +13535,12 @@ MergeUpdate::MergeUpdate(const CorrName &name,
   setCacheableNode(CmpMain::BIND);
   
   setIsMergeUpdate(TRUE);
+
+  // if there is a WHERE NOT MATCHED INSERT action, then the scan
+  // has to take place in the merge node at run time, so we have
+  // to suppress the TSJ transformation on this node
+  if (insertValues)
+    setNoFlow(TRUE);
 }
 
 MergeUpdate::~MergeUpdate() {}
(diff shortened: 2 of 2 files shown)


zellerh:
+1
Looks good to me, but I'm not familiar enough with merge statements to really review the second change. Maybe @sureshsubbiah or @sandhyasun can also take a look.

DaveBirdsall:
FYI: I ran full regressions on this change last night on a workstation. All tests passed.

sandhyasun:
Has this been tested on various  test cases where the table has multiple indexes and does it work in those cases where there is no view ?

DaveBirdsall:
Yes, @sandhyasun. It passed regression test executor/TEST015, which has examples of tables with multiple indexes and also no views.

core/sql/optimizer/NormRelExpr.cpp
```diff
@@ -6995,11 +6995,6 @@ NABoolean RelRoot::isUpdatableBasic(NABoolean isView,
   // QSTUFF
 
     {
-      // if child is a FirstN node, skip it.
-      if ((child(0)->castToRelExpr()->getOperatorType() == REL_FIRST_N) &&
-	  (child(0)->child(0)))
-	scan = (Scan *)child(0)->child(0)->castToRelExpr();
-      else
 	scan = (Scan *)child(0)->castToRelExpr();
```
sureshsubbiah: There may be a few isolated cases where the FirstN node is added  during preCodeGen. Please see GenPreCode.cpp RelRoot::preCodeGen(). The example given there about Order by where sort is added in optimizer, or a FirstN where the N value is to be specified with a param seem to be cases where we would add the FirstN later. Will current change cause such views to marked as updateable? This is minor and could be resolved later, if any change is necessary.

sureshsubbiah:
+1. Changes look great. Thank you for explaining the change to me.

core/sql/optimizer/NormRelExpr.cpp
DaveBirdsall: I will look into it.

DaveBirdsall: Thanks, Suresh, for the pointer. While trying to go through the RelRoot::preCodeGen() logic and cause a FirstN node to be inserted, I ran into this bug:

1. I created a table t1 with six rows.
2. I did "create view v1 as select [first 5] * from t1 order by a;"  (I wanted to see if the order by clause was permitted and if so would it circumvent the updatable view check. It did!)
3. I did "update v1 set b = 6". It updated 6 rows.

So, it appears that the updatable view test is not as complete as it needs to be.

I have written a new JIRA, https://issues.apache.org/jira/browse/TRAFODION-2840, to track this problem.

PR #1318, opened 2017-11-30
TRAFODION-2825 add jdbc user case

gpj1987:
some users indicate that they have no experience on how to use jdbc, they require some jdbc demos


selvaganesang:
Thanks for doing this. I will ,merge this. Check-PR or daily build doesn't exercise the sample programs shipped with Trafodion.

PR #1319, opened 2017-11-30
TRAFODION-2824

gpj1987:
there are some codes and comments contain HP in trafodion, like HPT4Connection，HPT4ResultSet

### DIFF
--- a/core/conn/jdbc_type2/src/main/java/org/apache/trafodion/jdbc/t2/SQLMXPreparedStatement.java
+++ b/core/conn/jdbc_type2/src/main/java/org/apache/trafodion/jdbc/t2/SQLMXPreparedStatement.java
@@ -68,7 +68,7 @@ import java.util.Locale;
  * so on) for setting IN parameter must specify types that are compatible with
  * the defined SQL type of the input parameter. For instance, if the IN
  * parameter has SQL type <tt>INTEGER</tt>, then the method <tt>setInt</tt>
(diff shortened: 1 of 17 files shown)


core/conn/jdbc_type2/src/main/java/org/apache/trafodion/jdbc/t2/SQLMXPreparedStatement.java
```diff
@@ -2772,7 +2772,7 @@ private void setLong(int parameterIndex) throws SQLException {
 	/**
 	 * Sets the designated parameter to SQL <tt>NULL</tt>.
 	 * <p>
-	 * <B>Note:</B> You must specify the parameters SQL type (although HP SQL/MX
+	 * <B>Note:</B> You must specify the parameters SQL type (although Trafodion SQL/MX
```
DaveBirdsall: Actually, SQL/MX is a reference to the predecessor HP product as well. If you wish, consider changing "Trafodion SQL/MX" to just "Trafodion"

DaveBirdsall:
I merged this change and rebuilt the manuals on the Trafodion web site just now. Thanks for this change.

selvaganesang:
The file sql/langman//LmSQLMXDriver.java is obsolete and it is not used. Please remove this file from git. The T4 related changes in LmUtility.java might be used if anyone changes the default jdbc driver in stored procedure to jdbc T4 driver. By default, stored procedures use T2 driver.

PR #1324, opened 2017-12-05
TRAFODION-2829 the examples of python is running

liyuanyuan-1009:
First, Make sure that your environment is installed with Jython, then All Python examples can run when you use the rigth command. To modify the Installer.java-tmpl file to add all the jar dependent paths to the Session.py,and add the first line of each Python file.


svarnau:
jenkins, add user

PR #1326, opened 2017-12-06
[TRAFODION-2827] Turn update costing code on by default

DaveBirdsall:
This pull request changes the default for CQD HBASE_UPDATE_COSTING to 'ON'.

The former setting ('OFF') used stub code for update costing, which assigned a zero cost to all update operations. Among other effects, it would cause update statements to always get serial plans.

The new setting uses a different set of stub code. That code is a copy of the delete costing code. I call it "stub" code because the theory of operation of update is somewhat different than that for delete. Nevertheless, this code should give much better plans than the former stub.

Also included in this pull request are updated expected files for five regression tests. In all these cases, an update plan flipped from an update node to a tuple_flow(scan,update) plan. Except for unique row accesses, the latter plan should always be superior.


zellerh:
+1
Looking forward to better plans!

PR #1327, opened 2017-12-07
[TRAFODION-2803] add regression test for COMMENT-ON statement

EEDY:
add regression test for COMMENT-ON statement
[https://issues.apache.org/jira/browse/TRAFODION-2803](url)


EEDY:
Hi @robertamarton ,
This is regression test of COMMENT-ON statement, can you please help review this? Thanks in advance.

robertamarton:
Looks good to me.  Thanks for create the regress job.

PR #1328, opened 2017-12-07
[TRAFODION-2831] Add project DOAP file for inclusion in ASF project list

svarnau:
The project DOAP file was created from template at:
  https://projects.apache.org/create.html

That page also says it should not be included in source/binary packages,
hence it is excluded from git export used to create source product.


svarnau:
The test failure is spurious, since only docs were changed in this PR.

PR #1331, opened 2017-12-08
[TRAFODION-2835] Add script to aid in merge conflict resolution

DaveBirdsall:
Very often a conflict occurs because a test has been updated in two branches, and that test has output containing DDL timestamps or plan IDs or other unimportant information that varies on every regress run. When these are the only differences, the conflict resolution is simple: Delete all the HEAD stuff and accept all the new stuff. This script does exactly that.

The script looks for lines of the following form:

  ```
<<<<<<< HEAD
  -- Definition current  Wed Nov 29 08:53:46 2017
  =======
  -- Definition current  Wed Dec  6 19:32:34 2017
  >>>>>>> b149874... [TRAFODION-2827] Turn update costing code on by default
```

When it finds this, it removes the lines from <<<<<< HEAD to ======, and removes the final >>>>>> line, leaving just:

`  -- Definition current  Wed Dec  6 19:32:34 2017`

### DIFF
--- a/core/sqf/sql/scripts/beheadMergeConflict.py
+++ b/core/sqf/sql/scripts/beheadMergeConflict.py
@@ -1,0 +1,112 @@
+# @@@ START COPYRIGHT @@@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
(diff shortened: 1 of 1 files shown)


svarnau:
Your script ran afoul of the jenkins check that aims to make sure people do not submit un-merged code.
Maybe the check could be changed to only flag the merge markers at beginning of the line.

svarnau:
The very small issue I have is that this is being put in the sql/scripts directory, which is a dumping ground for end-user (admin) tools as well as developer tools. Everything there will be added to the binary deliverable package.  I hate to see developer stuff in the installed product..

selvaganesang:
Have you tried out kdiff3 to aid you in merging. It is a GUI based diff tool that helps in resolving the conflicts easily.

DaveBirdsall:
@selvaganesang, thanks for the tip. I'll give kdiff3 a try.

@svarnau, I'm open to putting developer scripts somewhere else. Can you suggest a better place? Also, I'm happy to change the jenkins check too. Can you point me to it?

svarnau:
I have made the change to the jenkins check.

I looked but I don't see a better place to put it. It would be a separate JIRA to separate out the developer and end-user scripts, so don't let my gripe slow down your change.

DaveBirdsall:
jenkins, retest

PR #1334, opened 2017-12-11
[TRAFODION-2804] Limit clause support in statements LOAD/INSERT/CREATE-TABLE + query

EEDY:
add "LIMIT" clause support for statements create-as-select, insert-into-select and load-into-select

### DIFF
--- a/core/sql/regress/compGeneral/FILTER045
+++ b/core/sql/regress/compGeneral/FILTER045
@@ -1,0 +1,17 @@
+#! /bin/sh
+
+# Specialized filter for  project to filter out
+# 1. Syskey values
+# 2. "Funny names" for index columns appearing multiple times
+#    (done twice, since it may appear multiple times)
+# 3. refreshed elapsed time
+
+fil=$1
+if [ "$fil" = "" ]; then
+  echo "Usage: $0 filename"
+  exit 1
+fi
+
+sed "
+s/\(Time:\) .*$/\1/g
+" $fil
--- a/core/sql/regress/compGeneral/TEST045
+++ b/core/sql/regress/compGeneral/TEST045
@@ -534,9 +534,9 @@ insert NOMVLOG into sub_limit_02 (a, b) select a, b from sub_limit_01 order by a
 
 load into sub_limit_03 select * from sub_limit_01 limit 16;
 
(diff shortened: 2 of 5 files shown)


anoopsharma00:
Have tests been added to dev regressions to validate this change?

Also make sure that shift/reduce and reduce/reduce parser conflicts do not increase
with this change.

EEDY:
There is no shift/reduce and reduce/reduce parser conflicts increase with this change.
And i just add regression test for this syntax change.

selvaganesang:
jenkins, retest

core/sql/regress/compGeneral/FILTER045
```diff
@@ -0,0 +1,17 @@
+#! /bin/sh
+
+# Specialized filter for  project to filter out
+# 1. Syskey values
```
zellerh: This filter doesn't really change syskey and funny names.

core/sql/regress/compGeneral/TEST045
```diff
@@ -498,8 +512,36 @@ update statistics for table date_dim on every column;
 update statistics for table date_dim on (d_qoy, d_year);
 update statistics for table store_sales on every column;
 
+
+
+--------------------------------------------------------------------
+?section common_subexpr
```
zellerh: Why did you name this section "common_subexpr"? This regression test is for a feature called common subexpressions, but this section has nothing to do at all with common subexpressions, so that is really misleading.

EEDY: Sorry, didn't realize "common subexpressions" is name of a feature. Will find a better place to put these tests.

zellerh: Yes, I often found it pretty difficult to find a good place to put new tests. Maybe compGeneralTEST015 might work? If you don't find a better place you can also leave the test in TEST045 and we can rename the section to something else, like insert_with_limit_clause.

zellerh: Note the change is already merged, maybe you or I can do this in a future checkin when it's convenient.

EEDY: OK, understood.

PR #1337, opened 2017-12-12
[TRAFODION-2838] Add error check to NATableDB::get

DaveBirdsall:
This will prevent cores in  NATable::updateExtTableAttrs when an error occurs under the NATable constructor for a Hive table that has a Trafodion external table definition.


DaveBirdsall:
@zellerh, a zero value for getColumnCount() occurs because an error has been detected in one of the routines that the NATable constructor calls. The diags area has been populated. The added check here merely causes us to go to the end of the method, where we mark the BindWA with error status. So it isn't that we are ignoring the external table declaration, rather it is the case that an error has already been detected.

zellerh:
Thanks for the explanation!

PR #1340, opened 2017-12-13
[TRAFODION-2843] Fix outer join on aggregate bug

DaveBirdsall:
Formerly, the code was allowing an uncovered aggregate expression to be generated as a before join predicate. This lead to a Generator assert. The test for uncovered expressions in Join::preCodeGen has been tightened up so this does not happen.


DaveBirdsall:
FYI: I am running full regressions on a clean workstation instance overnight tonight.

DaveBirdsall:
I ran core, compGeneral and executor regression suites last night on a workstation and they pass. These suites contain all the known full outer join tests. So, I will merge this change now.

PR #1341, opened 2017-12-14
[TRAFODION-2733] BMO quota changes

selvaganesang:
To ensure BMO quota assignment works even when the estimates are way off.

Introduced a CQD BMO_MEMORY_ESTIMATE_OUTLIER_FACTOR. When a BMO memory
estimate exceeds BMO_MEMORY_LIMIT_PER_NODE_IN_MB by this factor,
the BMO memory estimate for the operator is reset to
BMO_MEMORY_LIMIT_PER_NODE_IN_MB * BMO_MEMORY_ESTIMATE_OUTLIER_FACTOR.
Then BMO memory estimate and the quota is assigned based on the
revised estimate.


zellerh:
+1
Looks good to me. The max cardinality needs some changes, but since right now max cardinality isn't used in the algorithm, that can wait.

PR #1344, opened 2017-12-14
TRAFODION-2844 add strategy to dcsserver for restart moxsrvr

mashengchen:
When mxosrvr down, DCS server will restart it , and if mxosrvr down a lot of times(6 default) in a period of time(30 min default), there should reject the restart behavior.
if this down time is more than 30 min (default) to 6 times ago (default), DCS server should allow the restart

### DIFF
--- a/dcs/src/main/java/org/trafodion/dcs/server/ServerManager.java
+++ b/dcs/src/main/java/org/trafodion/dcs/server/ServerManager.java
@@ -22,7 +22,6 @@
 **********************************************************************/
 package org.trafodion.dcs.server;
 
-import java.net.InetAddress;
 import java.io.*;
 import java.util.*;
 import java.util.concurrent.Callable;
@@ -31,9 +30,8 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ExecutorCompletionService;
 import java.util.concurrent.CompletionService;
 import java.util.concurrent.Future;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.CountDownLatch;
-import java.text.DateFormat;
+import java.util.concurrent.TimeUnit;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -46,7 +44,6 @@ import org.apache.zookeeper.WatchedEvent;
 import org.apache.zookeeper.KeeperException;
 import org.trafodion.dcs.Constants;
 import org.trafodion.dcs.util.Bytes;
-import org.trafodion.dcs.util.DcsConfiguration;
 import org.trafodion.dcs.util.DcsNetworkConfiguration;
 import org.trafodion.dcs.util.RetryCounter;
 import org.trafodion.dcs.util.RetryCounterFactory;
@@ -56,32 +53,32 @@ import org.trafodion.dcs.script.ScriptContext;
 
 public final class ServerManager implements Callable {
     private static final Log LOG = LogFactory.getLog(ServerManager.class);
-    private static Configuration conf;
-    private static ZkClient zkc;
-    private static boolean userProgEnabled;
-    private static String userProgramHome;
-    private static String userProgramCommand;
-    private static String hostName;
-    private static String masterHostName;
-    private static long masterStartTime;
-    private static int port;
-    private static int portRange;
-    private static DcsNetworkConfiguration netConf;
-    private static int instance;
-    private static int childServers;
-    private static String parentZnode;
-    private static int connectingTimeout;
-    private static int zkSessionTimeout;
-    private static int userProgExitAfterDisconnect;
-    private static int infoPort;
-    private static int maxHeapPctExit;
-    private static int statisticsIntervalTime;
-    private static int statisticsLimitTime;
-    private static String statisticsType;
-    private static String statisticsEnable;
-    private static String sqlplanEnable;
-    private static int userProgPortMapToSecs;
-    private static int userProgPortBindToSecs;
+    private Configuration conf;
+    private ZkClient zkc;
+    private boolean userProgEnabled;
+    private String userProgramHome;
+    private String userProgramCommand;
+    private String hostName;
+    private String masterHostName;
+    private long masterStartTime;
+    private int port;
+    private int portRange;
+    private DcsNetworkConfiguration netConf;
+    private int instance;
+    private int childServers;
+    private String parentZnode;
+    private int connectingTimeout;
+    private int zkSessionTimeout;
+    private int userProgExitAfterDisconnect;
+    private int infoPort;
+    private int maxHeapPctExit;
+    private int statisticsIntervalTime;
+    private int statisticsLimitTime;
+    private String statisticsType;
+    private String statisticsEnable;
+    private String sqlplanEnable;
+    private int userProgPortMapToSecs;
+    private int userProgPortBindToSecs;
     private ServerHandler[] serverHandlers;
     private int maxRestartAttempts;
     private int retryIntervalMillis;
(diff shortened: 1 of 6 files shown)


dcs/src/main/java/org/trafodion/dcs/server/ServerManager.java
```diff
@@ -255,33 +259,17 @@ private void cleanupZk() {
...
-            }
-        }
-
         public ServerHandler(Configuration conf ,int childInstance) {
             int maxRestartAttempts = conf.getInt(Constants.DCS_SERVER_USER_PROGRAM_RESTART_HANDLER_ATTEMPTS,
                     Constants.DEFAULT_DCS_SERVER_USER_PROGRAM_RESTART_HANDLER_ATTEMPTS);
-            int retryIntervalMillis = conf.getInt(
-                    Constants.DCS_SERVER_USER_PROGRAM_RESTART_HANDLER_RETRY_INTERVAL_MILLIS,
-                    Constants.DEFAULT_DCS_SERVER_USER_PROGRAM_RESTART_HANDLER_RETRY_INTERVAL_MILLIS);
+            int retryIntervalMinutes = conf.getInt(
```
hegdean: Why did we change from millis to minutes. Millis is more granular and you can achieve mins from millis

dcs/src/main/java/org/trafodion/dcs/util/RetryCounter.java
```diff
@@ -20,54 +20,79 @@
...
+    private Queue<Long> queue;
 
-  public RetryCounter(int maxRetries, 
-  int retryIntervalMillis, TimeUnit timeUnit) {
-    this.maxRetries = maxRetries;
-    this.retriesRemaining = maxRetries;
-    this.retryIntervalMillis = retryIntervalMillis;
-    this.timeUnit = timeUnit;
-  }
+    public RetryCounter(int maxRetries, int retryInterval) {
```
kevinxu021: Not very reasonable to remove last parameter.

traflm:
@kevinxu021 could you confirm the change now?

kevinxu021:
Good to go!

dcs/src/main/java/org/trafodion/dcs/util/RetryCounter.java
```diff
@@ -20,54 +20,94 @@
...
-  }
+    /**
+     * Sleep for a exponentially back off time
+     * @throws InterruptedException
+     */
+    public void sleepUntilNextRetry() throws InterruptedException {
+        int attempts = getAttemptTimes();
+        long sleepTime = (long) (retryInterval * Math.log(attempts + 15));
+        LOG.info("Sleeping " + sleepTime + "ms before retry #" + attempts + "...");
+        timeUnit.sleep(sleepTime);
```
selvaganesang: Good catch It looks like earlier the sleep time was incorrect if the TimeUnit wasn't in milliseconds.

mashengchen:
@selvaganesang 
If the mxosrvr dies quickly due to some issue (say 6 times within a minute), then dcs server would not restart it again, because we think it is the vicious behavior that kill the mxosrvr.
If the mxosrvr dies slowly (say 5 times in 25 minutes and 6th time after many hours), dcs server will restart the mxosrvr.

dcs/src/main/java/org/trafodion/dcs/util/RetryCounter.java
mashengchen: the timeUnit type dependents on what programmer give, if gives TimeUnit.minute, it will be minute level.
so here do you means the sleep is always in milliseconds level?

selvaganesang: No. I meant this is a good change

selvaganesang:
Thanks @mashengchen.  I understand that if the  method RetryCounter::shouldRetryInnerMinutes() returns false, the mxosrvr won't be restarted;, If it returns true the mxosrvr would be started.

In this method, the peek would get the head of the entry rather than the last entry. For the 2nd case (say 5 times in 25 mins and 6th time after many hours), shouldn't we look at the last entry via peekLast().  offer() would add at the last

mashengchen:
@selvaganesang 
No, the design is that it always compares the first time against the last time, here in your case , that will be a many hours+25mins, we consider the sixth time is an accident down, not a man-made. so there will restart the dcs server.

selvaganesang:
I overlooked the fact you are sending the difference between the current time stamp and the first. This change looks ok for most cases.  With the current defaults, there might be always a case where the strategy didn't start the  mxosrvr, but you would want it to be started. Consider a case where there were 5 failures within a minute, but 6th failure after 20 minutes.  I think this woudn't start the mxosrvr based on current defaults. This may not be considered as a vicious failure  as you have said, while six failures in a minute could be considered as a vicious.  Hopefully, we can tickle the DCS server some way externally to start the mxosrvr in such a case instead of waiting for 30 more minutes.

mashengchen:
agree , it's will have several situation from your suggestion, i need to think about it.

selvaganesang:
My suggestion would be to change the defaults. If there are N failures in 5 minutes, stop restarting the mxosrvr for 30 minutes. If the current failure is more than X minutes away from the last failure, restart the mxosrvr.  N and X can be made configurable. 5 minutes was assumed that it takes around a minute to start the mxosrvr and register etc.  30 minutes is the breathing time needed to get back the node to normal state externally.  If it doesn't , then you would say 6 cores every 30 minutes.  Choosing N is no brainier, it should be a little more than it takes mxosrvr to start and register etc. Default value of 6 for N seems reasonable.  Choosing X is a bit hard.   If the mxosrvr dies quite often due to queries from the application, this time is dependent upon how much time the application can wait to get another connection. This would matter when the mxosrvr resource is tight otherwise the application can always find another server to connect.  Default value of 5 or 10 minutes may be fine.

selvaganesang:
BTW, I liked the way you have coded to implement some strategy

mashengchen:
@selvaganesang 
I thought your suggestion, it's considerate, but it may endless and will be very complex.
In your case , say the 20min, it may be a common down( means not a vicious ), but 18min also can be a common down , 10 min also, so here we set 30min as a default interval, users can set "dcs.server.startup.mxosrvr.user.program.restart.handler.retry.timeout.minutes" a smaller value , maybe 20min.

mashengchen:
i though your suggection, and it's considerate, but it may be complex to cover all circumstance
discuss two situations
1:
if mxosrvr starts successful, say passed N step, then for a long time , it failure, this failure is more than X mins, and should restart, then while restarting it failure for some reason , maybe network error or something else i'm not sure, and this failure maybe a 1 min failure, i think it should restart again, but conflict with the strategy .
2:
if mxosrvr starts successful, say passed N step, then for a long time , it failure, this failure is more than X mins, and should restart, then restart successful , then users connect to this mxosrvr and do query (just exec values (1);) for 100 times, then mxosrvr down for the code logic, and the interval may be a 1-5mins, this can't restart either, for conflict with the strategy.

my idea is use my strategy ,and set "dcs.server.startup.mxosrvr.user.program.restart.handler.retry.timeout.minutes" a smaller value.
but this is not enough, there still need add a timer strategy, that is restart the "never restart mxosrvr" regularly, maybe one or two times a day.
and i think i can add the timer strategy in another PR. so how do you think this?

mashengchen:
I had do a calc, that it will cost near 4 mins to do 600 queries (just values(1);) for one mxosrvr , so i use 5mins * 6 times = 30mins as the default for "dcs.server.startup.mxosrvr.user.program.restart.handler.retry.timeout.minutes"

mashengchen:
change the time from 30 to 6

traflm:
@selvaganesang Could you review it again now? Is this good to be merged now?

selvaganesang:
+0

The recent commit doesn't address the issues I have raised. However, I am not giving -1 because I am assuming that some testing has been done with this strategy and It is better than the earlier one and solves some issue. Neither the jira [TRAFODION-2844] nor this PR explained what issue it solved. It just provided the solution

PR #1345, opened 2017-12-14
[TRAFODION-2839] Change Incr UPDATE STATS to not use CBFs by default

DaveBirdsall:
Change the default for CQD USTAT_INCREMENTAL_UPDATE_STATISTICS from 'ON' to 'SAMPLE'. This causes the incremental update statistics logic to skip using Counting Bloom Filters, but instead always rebuild the histograms from the incremental sample. By doing so, we avoid accumulating CBF files in the $HOME/cbfs directories. As a side benefit, it turns out there is an elapsed time improvement to incremental update statistics in doing this. (Making this change was suggested by @blfritch a year ago actually, for performance reasons.)


zellerh:
+1
Looks good to me.

PR #1353, opened 2017-12-18
[TRAFODION-2802] Prepare the build environment with one command

sumory:


svarnau:
Please do not open new pull-requests, when an existing pull-request exists for the same change/jira. It is better to just push new commits onto the same branch so that we can see previous commits & changes.

If it is absolutely necessary to do so, then it is good idea to put comment in the older one and close it.

svarnau:
jenkins, ok to test

sumory:
The old pull-request is already closed with related comment, thanks.

install/traf_checkset_env.sh
```diff
@@ -0,0 +1,243 @@
...
+fi
+
+javahome=`grep JAVA_HOME ~/.bashrc | wc -l`
+if [ "x${javahome}" = "x0" ]; then
+    echo -en "export JAVA_HOME=${javapath}\n" >> $HOME/.bashrc
+    echo -en "export PATH=\$PATH:\$JAVA_HOME/bin\n" >> $HOME/.bashrc
+fi
+
+# check and set locale
+local_locale=`locale | grep -v "xen_US.UTF-8"`
```
zellerh: This code did not work for me when I tried it on my workstation. I saw several issues: a) the extra "x" in "xen_US" on line 197, the comparison to "x" in line 198 and the fact that my locale command returns multiple lines. The following worked for me:

```
local_locale=`locale | grep  "LANG="`
if [ "${local_locale}" != "LANG=en_US.UTF-8" ]; then
```

install/traf_checkset_env.sh
```diff
@@ -0,0 +1,243 @@
...
+    if [ "x${lang_set}" = "x0" ]; then
+	echo -en "export LANG=\"en_US.UTF-8\"\n" >> $HOME/.bashrc
+    fi
+    lc_all_set=`grep LC_ALL ~/.bashrc | wc -l`
+    if [ "x${lc_all_set}" = "x0" ]; then
+	echo -en "export LC_ALL=\"en_US.UTF-8\"\n" >> $HOME/.bashrc
+    fi
+fi
+
+# check error
```
zellerh: See above.

install/traf_checkset_env.sh
```diff
@@ -0,0 +1,243 @@
...
+# specific language governing permissions and limitations
+# under the License.
+#
+# @@@ END COPYRIGHT @@@
+#
+# make sure the environment can build the code of trafodion or not.
+# must configure the yum repo right before execute this script.
+# run this script with normal user, while must has sudo permission.
+
+#default path
```
zellerh: This command needs a help text that should mention the following:

- Who should use this script (people who want to set up a single node
  development system on a machine where they have sudo privileges)
- What does the script do (e.g. yum install, build/setup tools).
- A link to the Contributor guide wiki page that describes these steps
  (https://cwiki.apache.org/confluence/display/TRAFODION/Create+Build+Environment)
- What environment variables must be set before calling this script
- What optional environment variables can be set
- What directories are created by the script, and how they can be customized

Once this PR is merged, we can also update the wiki page and mention this command.

install/traf_checkset_env.sh
```diff
@@ -0,0 +1,243 @@
...
+MY_JVM_PATH=${MY_JVM_PATH-"/usr/lib/jvm"}
+MY_JAVA_VER=${MY_JAVA_VER-"java-1.7.0-openjdk"}
+MY_SUDO=${MY_SUDO-"sudo"}
+MY_YUM=${MY_YUM-"${MY_SUDO} yum -y"}
+
+# for setup tools
+MY_LOCAL_SW_DIST=${MY_LOCAL_SW_DIST-${HOME}/local_software_tools}
```
zellerh: It's pretty common that the root directory of a server is small (at least on the OpenStack and Amazon VMs I have seen). Maybe add yet another variable like this one:

```
MY_SW_HOME=${MY_SW_HOME-${HOME}}
```

Then use $MY_SW_HOME instead of $HOME below.

install/traf_checkset_env.sh
```diff
@@ -0,0 +1,243 @@
...
+	faillibs=("${faillibs[@]}" "${dependlib}")
+	echo "ERROR: install dependent library [${dependlib}] fail"
+    else
+	echo "INFO: install dependent library [${dependlib}] success"
+    fi
+done
+echo "INFO: install dependent library finish"
+
+# remove pdsh and qt-dev
+echo "INFO: remove pdsh and qt-dev trafodion scripts get confused."
```
zellerh: Maybe say "INFO: remove pdsh and qt-dev commands to avoid problems with trafodion scripts"

install/traf_checkset_env.sh
```diff
@@ -0,0 +1,243 @@
...
+dirs="ASSEMBLY_EXCEPTION  bin  include  jre  jre-abrt  lib  LICENSE" \
+				     " tapset  THIRD_PARTY_README"
+if [ ! -d ${javapath} ]; then
+     echo "ERROR: java dir [${javapath}] isn't right"
+     exit 1
+fi
+
+javahome=`grep JAVA_HOME ~/.bashrc | wc -l`
+if [ "x${javahome}" = "x0" ]; then
+    echo -en "export JAVA_HOME=${javapath}\n" >> $HOME/.bashrc
```
zellerh: When writing into the users .bashrc, I would recommend adding a comment that shows what tool wrote this. Something like:

```
# Lines added by traf_checkset_env.sh
exprt JAVA_HOME=...
```

install/traf_checkset_env.sh
```diff
@@ -0,0 +1,243 @@
...
+    echo -en "export PATH=\$PATH:\$JAVA_HOME/bin\n" >> $HOME/.bashrc
+fi
```
zellerh: It would probably be good to add a test here that runs

```
$JAVA_HOME/bin/java -version
```

and checks that we get an acceptable version.

install/traf_checkset_env.sh
```diff
@@ -0,0 +1,243 @@
...
+	    openssl-devel.i686 perl-Config-IniFiles perl-Config-Tiny	\
+	    perl-DBD-SQLite perl-Expect perl-IO-Tty perl-Math-Calc-Units \
+	    perl-Params-Validate perl-Parse-RecDescent perl-TermReadKey	\
+	    perl-Time-HiRes protobuf-compiler protobuf-devel		\
+	    readline-devel rpm-build saslwrapper sqlite-devel unixODBC	\
+	    unixODBC-devel uuid-perl wget xerces-c-devel xinetd		\
+	    ncurses-devel)
+faillibs=()
+echo "INFO: install dependent library start"
+for dependlib in ${dependlibs[@]}
```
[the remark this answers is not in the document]
xiaozhongwang: When we ware creating build environment recently, there are a lot error caused by installing software by one command. there are no error messages when installing, but make an error when compiling. so it's necessary I think.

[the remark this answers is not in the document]
zellerh: Sounds good, we can always change that later if needed.

install/traf_checkset_env.sh
```diff
@@ -0,0 +1,243 @@
...
+cpuarch=`uname -m`
+echo "INFO: os is [${osname}], cpu architecture is [${cpuarch}]"
+
+osdistributor=`lsb_release -i | cut -d : -f 2-`
+case ${osdistributor} in
+    *RedHat*|*CentOS*)
+	echo "INFO: os distributor id is [${osdistributor}]"
+	;;
+    *)
+	echo "ERROR: The OS distribution is [${osdistributor}], but Trafodion only supports RedHat and CentOS presently."
```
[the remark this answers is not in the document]
sumory: I don‘t have the environment as you say, if the user has this environment, user can installed by modifying the environment variable MY_YUM

DaveBirdsall:
Is this ready to merge now? @zellerh, what do you think?

zellerh:
+1
Yes, looks good to me. Thank you for adding the help printout and sorry for taking a while to respond.

DaveBirdsall:
Thanks for this contribution. Sorry it's taken so long to merge.

PR #1354, opened 2017-12-18
[TRAFODION-2850] Add a Note for EMPTY_BLOB() and EMPTY_CLOB() for *Trafodion SQL Large Objects Guide*

liuyu000:


docs/lob_guide/src/asciidoc/_chapters/working_with_lob.adoc
```diff
@@ -235,6 +235,8 @@ The source for inserting into a LOB can be any of the following:
 An unnamed parameter can be used to prepare a statement and then during an execution, either a function or a simple string parameter can be passed in which will be converted to LOB data.
 
 * `EMPTY_BLOB()` or `EMPTY_CLOB()` 
++
+NOTE: If you want to insert `EMPTY_BLOB()` or `EMPTY_CLOB()` into a lob column, the CQD `TRAF_CLOB_AS_VARCHAR` which is *ON* by default must be turned *OFF* before creating the table.
```
DaveBirdsall: You may want to elaborate: The CQD must be 'ON' at the time the table with the BLOB / CLOB column is created. (Otherwise the table is created with a VARCHAR.)

liuyu000: Hi @DaveBirdsall, 
1. The CQD  must be turned **'OFF'** before creating a table, otherwise an error will be raised if `EMPTY_BLOB()` or `EMPTY_CLOB()` is inserted into a lob column, and the **column definition** of lob column is **VARCHAR** (see the first attachment).

2. Correct behaviour:
Set the CQD 'OFF', then create a table and insert `EMPTY_BLOB()` or `EMPTY_CLOB()`, it will succeed and the column definition is **CLOB** rather than VARCHAR (see the second attachment).

3. The default value of this CQD is ON. 

Thanks for your reminding, may I change this note to the following?

Note: If you want to insert `EMPTY_BLOB()` or `EMPTY_CLOB()` into a lob column, the CQD `TRAF_CLOB_AS_VARCHAR` which is **ON** by default must be turned **OFF** before creating a table, otherwise an error will be raised and the column definition of the lob column is VARCHAR.

turn the cqd off after creating a table
set the cqd off before creating a table

DaveBirdsall: Oops... I meant to say, "The CQD must be 'OFF' at the time the table with the BLOB / CLOB column is created. (Otherwise the table is created with a VARCHAR.)" Apologies. 

My fault for typing that comment quickly before rushing off to an appointment.

DaveBirdsall: And yes, your suggested comment looks right to me.

liuyu000: No worries. Everyone makes mistakes, that's why there is an eraser on every pencil.
Thanks for your remiding, Dave :)

PR #1364, opened 2018-01-02
[TRAFODION-2854] Load encounter Operating system error 201

selvaganesang:
When load returns an error during insert, the row is reconstructed
to log in the same format as the source. During the reconstruction process
a wrong tuple format was used.


sureshsubbiah:
+1. Looks good to me.

PR #1365, opened 2018-01-03
TRAFODION-2873. Cleanup of LobGlobals and LobLoad classes

sandhyasun:
This fix also avoids a core that was very occasionally caused during inserts into LOB tables. The fix involves cleaning up the LOBGlibals and eliminating unneeded datamemebers and keeping only the ExLobGlobals members. Also cleaned up all LobLoad related code since it is obsolete and worked only with DP2.


core/sql/exp/exp_clause.cpp
```diff
@@ -574,8 +574,8 @@ ex_clause::ex_clause(clause_type type,
 	case ITM_LOBCONVERTHANDLE:
 	  setClassID(LOB_CONVERTHANDLE);
 	  break;
-	case ITM_LOBLOAD:
-	  setClassID(LOB_LOAD);
+          /*	case ITM_LOBLOAD:
```
DaveBirdsall: Maybe delete these lines (and the break after them) instead of commenting out?

DaveBirdsall:
Seabase suite build failed for unknown reasons... starting a retest

DaveBirdsall:
jenkins, retest

selvaganesang:
Is ExLobGlobals pertain to the Statement or ExGlobals or the ContextCli or the process?

sandhyasun:
The ExLobGlobals is created whenever a statement or internal API needs to access the LOB functions. It maintains the hdfs connection info, the various lob pointers it needs to access (based on name)  .It is currently a member of ex_globals.It's initialized once per TCB.

selvaganesang:
If so, won't it better to create it using heap_ associated with ex_globals. If there is a connection info that can be shared across context or ESP fragments s, separate them out and tore it in cli_globals and pass that info to ExLobGlobals in ex_globals.  It is strange to see the ExGlobals being constructed from system heap.  And I couldn't find where is it being deleted.  There is a possibility of memory leak.

sandhyasun:
The ExLobGlobals is deleted in  ExpLOBinterfaceCleanup. Will look at getting rid of system new.  There was some issue with using the Map template class with an internal heap. The lobPtrs are stored as a standard template class within the ExLobGlobals.

PR #1367, opened 2018-01-04
[TRAFODION-2875] Fix inaccuracies in comparison in analyzeMessageGuide.py

DaveBirdsall:
Ignore trailing spaces in message text and documentation text.

Take into account multiple substitution fields in message text and documentation text by replacing them with '.elided.' (before they were simply removed).


DaveBirdsall:
jenkins, retest

PR #1369, opened 2018-01-04
TRAFODION-2877 can't restart DCSServer after switch to backup-master

mashengchen:
DcsServer does not be restarted by DcsMaster on HDP.
Steps To Reproduce
1. check Hadoop and instance status
2. kill active DcsMaster
3. repeat Step 1
4. kill DcsServer on the active DcsMaster node
5. repeat Step 1
6. DcsServer never been restarted

### DIFF
--- a/dcs/src/main/java/org/trafodion/dcs/master/ServerManager.java
+++ b/dcs/src/main/java/org/trafodion/dcs/master/ServerManager.java
@@ -240,14 +240,17 @@ public class ServerManager implements Callable {
                         }
                     }
                 } else {
-                    LOG.info("No restart for "
-                            + znodePath
-                            + "\nbecause DcsServer start time ["
-                            + DateFormat.getDateTimeInstance().format(
-                                    new Date(serverStartTimestamp))
-                            + "] was before DcsMaster start time ["
-                            + DateFormat.getDateTimeInstance().format(
-                                    new Date(startupTimestamp)) + "]");
+                    StringBuffer sb = new StringBuffer();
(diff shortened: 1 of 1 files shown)


dcs/src/main/java/org/trafodion/dcs/master/ServerManager.java
```diff
@@ -240,15 +240,14 @@ public ScriptContext call() throws Exception {
...
-                    if (LOG.isDebugEnabled())
-                        LOG.debug("No restart for "
-                                + znodePath
-                                + "\nbecause DcsServer start time ["
-                                + DateFormat.getDateTimeInstance().format(
-                                        new Date(serverStartTimestamp))
-                                + "] was before DcsMaster start time ["
-                                + DateFormat.getDateTimeInstance().format(
-                                        new Date(startupTimestamp)) + "]");
+                    LOG.info("No restart for "
```
DaveBirdsall: Looks like there is another reason for no restart, namely (master.isFollower() && runningServers.size() < configuredServers.size()) ?

mashengchen: yes . if backup-master take over from the failed master, then the 'isFollower' is true, the result will be the code does not run the if conditions, and dcs server not restart

DaveBirdsall: RIght... My point was that the text in the logging seems to need updating, because the 'if' conditions have changed. What do you think?

mashengchen: You are right, I should change the log, thanks.

PR #1372, opened 2018-01-05
TRAFODION-2885 dcs server cant be restart while switching master

mashengchen:
Check all configured servers, if the server is not watched by zookeeper during dcs master or backup master starting, add to the restart queue, then dcs master will try to restart it.


selvaganesang:
Looks good

PR #1380, opened 2018-01-09
[TRAFODION-2886] fix the nullpointer error scanned by TScanCode

xiaozhongwang:
I scanned code of Trafodion by TScanCode, found many nullpointer error, and I fixed them here.


core/sql/executor/cluster.cpp
```diff
@@ -2395,7 +2395,7 @@ NABoolean Cluster::checkAndSplit(ExeErrorCode * rc)
 				         rc);
   
   if ( !next_ || *rc ) {
-    if ( ! rc ) * rc = EXE_NO_MEM_TO_EXEC;
+    * rc = EXE_NO_MEM_TO_EXEC;
```
selvaganesang: The deleted code is correct.  In the new code rc could be NULL

xiaozhongwang: if rc is NULL,  core at line 2397
so rc must be not NULL.

or
make sure rc is not NULL at line 2397

selvaganesang: Then the condition should be
  if (!next_ || rc) 
     if ( !rc ) *rc = EXE_NO_MEM_TO_EXEC;

xiaozhongwang:
I re-read the relevant code. I fully understood its meaning:
1. rc is from checkAndSplit param, and set the value to EXE_OK by: 
    * rc = EXE_OK;
    so the rc must be not NULL, I think this may be an agreement
2. the method Cluster maybe set result to rc, may be not because fail when call new
    so we must check the *rc is set or not.
    don't need to check the rc is NULL or not.
So I think it should be wrote like this:
  if ( !next_ || *rc ) {
    if ( ! (*rc) )  * rc = EXE_NO_MEM_TO_EXEC;
    return TRUE;
  }

what do you think?

selvaganesang:
If there was an initialization *rc = EXE_OK already, then this code should be

if (!next_) {
   *rc = EXE_NO_MEM_TO_EXEC;
    return TRUE;
}

Sorry for not reviewing the surround code earlier.

xiaozhongwang:
I had modified and committed  as review

xiaozhongwang:
Your suggestion is wrong. I had already explained before
the method Cluster maybe set error to rc, may be not. because may fail when call new
so we must check the * rc is set or not. If I write as following, will rewrite the error set by Cluster method:
`if (! next_) {`
`* rc = EXE_NO_MEM_TO_EXEC;`
`return TRUE;`
`}`

selvaganesang:
In that case you are making an assumption that EXE_OK is 0

It should be  if (*rc == EXE_OK)  *rc = EXE_NO_MEM_TO_EXEC;

xiaozhongwang:
Yes, this is right!
I change it.

selvaganesang:
Please change it as follows. Making this suggestion to avoid another round of review

if (!next_) {
   if (*rc == EXE_OK) 
     *rc = EXE_NO_MEM_TO_EXEC;
}

core/sql/executor/cluster.cpp
```diff
@@ -2395,7 +2395,7 @@ NABoolean Cluster::checkAndSplit(ExeErrorCode * rc)
 				         rc);
   
   if ( !next_ || *rc ) {
```
selvaganesang: Oops.  Please remove the || *rc,

xiaozhongwang: I cann't understand.
Why remove the || *rc,
I think there are two type results：
1、 First, if there are no memory, next_ will get NULL value.
2、 Second, there are some wrong happened in Cluster::Cluster
In the second case, next_ will get a value, but *rc will is not EXE_OK.
If we remove the || *rc,  this check will pass, but there was an error happened.

My understanding is wrong?

selvaganesang:
+1

PR #1384, opened 2018-01-09
[TRAFODION-2879] Core dump ...plus four more fixes, see below

zellerh:
This is a PR for four fixes, submitted as separate commits. I would recommend reviewing each commit separately, as the fixes are not related.

Summary (also in the commit text for each fix):

**[TRAFODION-2879] Core dump due to reference to deallocated memory for EstLogProp object**

In the optimizer, we often use a define to refer to an "empty input
logical properties" object. The problem was that this object was
allocated on the statement heap but it was owned by the CmpContext
object, which lives longer than the statement heap. During debugging,
we have seen cases where this led to references of deleted memory.

The fix is to move the empty logical properties to the CmpStatement
object.

Files: arkcmp/*
       RelExpr.h

**[TRAFODION-2896] Internal error in stored procedures when a warning is generated in SQL**

This problem was seen by Any on a large cluster and Selva found this issue
during debugging. At this point we don't have an easy testcase to reproduce,
but we agree that the code change is something we should do.

udrserv/*

**[TRAFODION-2897] Invalid name in MySQL log file when using install_local_hadoop**

Bug reported on the dev list.

install_local_hadoop

**Other changes:**

- Fixes in the OSIM tool related to external Hive tables. Issue an unregister
  command before creating a new Hive table in OSIM simulation mode.
  Also get the REGISTER commands from the OSIM files and send them to
  Trafodion, not to Hive, as before. Finally, make some minor changes
  to showddl (avoid extra commas, fix comments that showed REGISTER
  as a Hive command).

CmpDescribe.cpp
OptimizerSimulator.cpp

- Cosmetic changes to JDBC TMUDF

JDBCUDR.java


sureshsubbiah:
All changes look good to me. Thank you.

PR #1386, opened 2018-01-10
[TRAFODION-2899] Catalog API SQLColumns support ODBC2.x

Weixin-Xu:
When using ODBC2.x to get description of columns, failure occurs but no error returns.


Weixin-Xu:
For these ODBC2.x, the result of SQLColumns is six columns less than ODBC3.x, other than not supported.
In the tables, the version of every datatype is 3.

sandhyasun:
@arvind-narain  Do these changes look good to you ?

DaveBirdsall:
I will merge this by the end of the day if I don't see any objections before then.

PR #1387, opened 2018-01-10
[TRAFODION-2900] Fix bug could insert more than the maximum number of…

CoderSong2015:
When insert 40 Chinese char to char(30) , of which the character is utf8 , it gets success while it should return fail.
SQL return the max length of the column which is calculated by bytes . The returned length of 30 chars of utf8 character is 120 bytes. However the length of Chinese chars in utf8 maybe is 3 bytes or 4 bytes. So 40 Chinese chars could be inserted into the database. But the char(30) represent that 30 chars regardless of how many bytes a char costs. So I modify the T4 to judge char numbers of the string instead of bytes of the string.
@kevinxu021 @Weixin-Xu @selvaganesang please review the code.

### DIFF
--- a/docs/src/site/markdown/download.md
+++ b/docs/src/site/markdown/download.md
@@ -36,22 +36,22 @@ To build Trafodion from source code, see the [Trafodion Contributor Guide](https
     * [Ambari Plugin][ap220]  -  [PGP][appgp220] [SHA1][apsha220]
 * [Documentation](documentation.html#220_Release)
 
-[src220]: http://www.apache.org/dist/trafodion/apache-trafodion-2.2.0/src/apache-trafodion-2.2.0-src.tar.gz
+[src220]: http://www.apache.org/dyn/closer.lua/trafodion/apache-trafodion-2.2.0/src/apache-trafodion-2.2.0-src.tar.gz
 [pgp220]: http://www.apache.org/dist/trafodion/apache-trafodion-2.2.0/src/apache-trafodion-2.2.0-src.tar.gz.asc
 [sha220]: http://www.apache.org/dist/trafodion/apache-trafodion-2.2.0/src/apache-trafodion-2.2.0-src.tar.gz.sha
-[pins220]: http://www.apache.org/dist/trafodion/apache-trafodion-2.2.0/bin/apache-trafodion_pyinstaller-2.2.0.tar.gz
+[pins220]: http://www.apache.org/dyn/closer.lua/trafodion/apache-trafodion-2.2.0/bin/apache-trafodion_pyinstaller-2.2.0.tar.gz
 [pinpgp220]: http://www.apache.org/dist/trafodion/apache-trafodion-2.2.0/bin/apache-trafodion_pyinstaller-2.2.0.tar.gz.asc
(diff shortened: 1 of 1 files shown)


CoderSong2015:
@xwq please review the code.

CoderSong2015:
I have tested it with sqlci and it returned like below:
>>prepare dd from insert into chtest values(9,?); 

--- SQL command prepared.
>>execute dd using '123456789012345678901234567890123456789087866';

** WARNING[8402] A string overflow occurred during the evaluation of a character expression. **

The string could be insert into the database but will be cutoff by the max chars the column could store, it  
is right to do so. But when using T4 (trafci) to insert the string , it could be insert into database as long as it's numbers of bytes less than the column length, like this

C_ID        C_CHAR                                                                                                                  
----------- ------------------------------------------------------------------------------------------------------------------------
          1 1234567890123456789012345678                                                                                            
          6 123456789012345678901234567890                                                                                          
          8 壹六五柒八伍伍贰叁一一三玖伍拾九捌六肆陆贰三陆三壹拾二七六+七六七六七六七六七六                                                                                
          9 123456789012345678901234567890                                                                                          
         12 123456789012345678901234567890123456789087866

The third row(id is 8), is inserted by trafci,  chars numbers of it obviously  larger than 30, while the fourth row is inserted by sqlci, and the query is  "execute dd using'1234567890123456789012345678901234\
56789087866';" , same as the third row, and the string is inserted correctly(?It should not be correct)!  This is really strange behavior of executor, right?

I think this behavior of sql should be checked carefully, and my commit could avoid it in the  driver  level .

selvaganesang:
I agree with @DaveBirdsall. This should be handled at the executor layer

CoderSong2015:
So is there any document to help me learn how to debug at the executor? If anyone could solve the problem, could you tell me the way you debug at the executor?  Thanks very very much! @selvaganesang  @DaveBirdsall

CoderSong2015:
> numOfChar = new String(tmpBarray, "utf-8").length();

It will get the length of the String instead of the string itselt. @xiaozhongwang

traflm:
@CoderSong2015 I think you can take a look at : ex_conv_clause::eval
By default, ANSI SQL behavior is to raise WARNING instead of ERROR, so I think trafci should catch the SQL Warning in this case.
From the code comments 
// result of the conversion, an error should be issued instead of a warning.
// This is according to ANSI SQL92
So this is correct behavior, not a defect in executor. But trafci should catch the WARNING, instead of calculating the char length in driver. I think we should change the trafci here, not the JDBC driver.

traflm:
@CoderSong2015 no, I think the best fix is to fix in trafci to catch the WARNING, since in this case, SQL should already return warning. sqlci works correct, but trafci not report WARNING.
If the reporter think ERROR is required, you should change the code I listed above to ERROR instead of WARNING. There are several places in that method, some return ERROR, some return WARNING. So I think you may need to do some debug to check where to change.
I didn't do any debug, so just some suggestions. 
And I recommend to change trafci to return warning in this case, but I don't see what is the real ANSI definition here.

CoderSong2015:
Thanks very much . I will try to solve the problem and create another pull request.

PR #1391, opened 2018-01-11
TRAFODION-2905 query id contain null

mashengchen:
if stmtLabel is "null" string, then it will skip driver's conditional judgment (if (stmtName ==null) throw exception).


selvaganesang:
What makes the stmtLabel to be "null" string and why?

mashengchen:
jenkins, retest

mashengchen:
see DatabaseQuery.java Line228.  prepareCallStmt(null)
when the query's sessionAutoPrepare is false , there will pass null as param.

selvaganesang:
Requesting @hegdean to take a look at why Trafci doesn't take care of  null stmtName

mashengchen:
Can someone else review or merge this PR?

PR #1394, opened 2018-01-12
[TRAFODION-2891] fix the bufoverrun Critical error checked by TScanCode

xiaozhongwang:
fix the bufoverrun Critical error checked by TScanCode

### DIFF
--- a/core/conn/odb/src/odb.c
+++ b/core/conn/odb/src/odb.c
@@ -5313,7 +5313,7 @@ static void etabadd(char type, char *run, int id)
                     }
                 }
                 if ( etab[no].type == 'e' ) { /* name & create output file */
-                    for ( i = j = 0; etab[no].tgt[i] && i < sizeof(buff); i++ ) {
+                    for ( i = j = 0; i < sizeof(buff) && etab[no].tgt[i]; i++ ) {
                         switch ( etab[no].tgt[i] ) {
                         case '%':
                             switch ( etab[no].tgt[++i] ) {
@@ -5570,7 +5570,7 @@ static void etabadd(char type, char *run, int id)
                         strmcat ( etab[no].src , (char *)Ostn[2] , ll , 0 );
                         etab[no].Ocso[2] = &Ostn[2][0]; /* update table name pointer with expanded tab name */
                         memset ( buff, 0, sizeof(buff));
-                        for ( i = j = 0; etab[no].tgt[i] && i < sizeof(buff); i++ ) {
+                        for ( i = j = 0; i < sizeof(buff) && etab[no].tgt[i]; i++ ) {
                             switch ( etab[no].tgt[i] ) {
                             case '%':
                                 switch ( etab[no].tgt[++i] ) {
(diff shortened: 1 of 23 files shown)


core/conn/odb/src/odb.c
```diff
@@ -5313,7 +5313,7 @@ static void etabadd(char type, char *run, int id)
                     }
                 }
                 if ( etab[no].type == 'e' ) { /* name & create output file */
-                    for ( i = j = 0; etab[no].tgt[i] && i < sizeof(buff); i++ ) {
+                    for ( i = j = 0; i < sizeof(buff) && etab[no].tgt[i]; i++ ) {
```
traflm: I don't really get this fully understand.
If i is out of bound for arry tgt, why this change will prevent the tag[i] to be executed?

core/conn/odbc/src/odbc/Common/linux/sqmem.cpp
```diff
@@ -400,7 +400,7 @@ void TestPool(void * membase, int length)
    long j, index;
    short error;
    long pass;
-   void * pool_ptrs[256];
+   void * pool_ptrs[256 + 1];
```
traflm: very good catch!

core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
```diff
@@ -5153,7 +5153,7 @@ odbc_SQLSvc_GetSQLCatalogs_sme_(
                  {
                     ERROR_DESC_def *p_buffer = QryCatalogSrvrStmt->sqlError.errorList._buffer;
                     strncpy(RequestError, p_buffer->errorText,sizeof(RequestError) -1);
-                    RequestError[sizeof(RequestError)] = '\0';
+                    RequestError[sizeof(RequestError) - 1] = '\0';
```
traflm: very good catch!

core/sqf/src/seatrans/tm/hbasetmlib2/hbasetm.h
```diff
@@ -139,9 +139,9 @@ class CHbaseTM : public JavaObjectInterfaceTM
       JM_REGTRUNCABORT,
       JM_DROPTABLE,
       JM_RQREGINFO,
-      JM_LAST
+      JM_TMLAST
    };
-   JavaMethodInit JavaMethods_[JM_LAST];
+   JavaMethodInit JavaMethods_[JM_TMLAST];
```
traflm: This is critical!

core/conn/odb/src/odb.c
xiaozhongwang: Most of time, they are equivalent.  There are exception in a special case 
Most of compiler judge the condition from left to right, so if the tgt is at the end of memory, and the offset i is equal to sizeof(buff), the access to memory goes beyond memory.
This will make a coredump, and cann't be repeatable.

core/sqf/src/seatrans/tm/hbasetmlib2/hbasetm.h
xiaozhongwang: There is  another define of JM_LAST in this file, and their value is different.

core/conn/odb/src/odb.c
DaveBirdsall: I tried to make sense of this too. It looks like etab[no].tgt is populated in function parseopt from some string buffer; I can't tell if it is buff (which is a static variable declared as a char[256]) or something else. My suspicion is that the test "i < sizeof(buff)" has nothing to do with etab[no].tgt but rather some other buffer being copied to. In any case, reversing the order of these tests seems harmless.

PR #1395, opened 2018-01-12
[TRAFODION-2903] correct Column_Size fetched from mxosrvr

Weixin-Xu:
DATE precision expect: 10 and actual: 11
TIME precision expect: 8 and actual: 9
TIMESTAMP precision expect: 19 and actual: 20


Weixin-Xu:
@traflm ,the precision expected is set up in Microsoft.
 Do we need to agree on Microsoft?

sandhyasun:
@traflm,  Are you satisfied with the explanations for this PR ?

traflm:
+1, it is good for me

PR #1397, opened 2018-01-13
[TRAFODION-2904] optionally add leading space when get a number colum…

traflm:
…n to display

### DIFF
--- a/core/sql/optimizer/BindItemExpr.cpp
+++ b/core/sql/optimizer/BindItemExpr.cpp
@@ -3654,12 +3654,10 @@ ItemExpr *Concat::bindNode(BindWA *bindWA)
         else if (convType == 2)
           {
             Parser parser(bindWA->currentCmpContext());
-            char buf[1000];
+            char buf[128];
             
-            // right justify the string representation of numeric operand 
-            // and then do the concat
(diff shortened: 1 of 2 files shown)


traflm:
I will update the regression test expected file asap

core/sql/optimizer/BindItemExpr.cpp
```diff
@@ -3654,12 +3654,10 @@ ItemExpr *Concat::bindNode(BindWA *bindWA)
...
           {
             Parser parser(bindWA->currentCmpContext());
-            char buf[1000];
+            char buf[128];
             
-            // right justify the string representation of numeric operand 
-            // and then do the concat
-            sprintf(buf, "CAST(SPACE(%d - CHAR_LENGTH(CAST(@A1 AS VARCHAR(%d)))) || CAST(@A1 AS VARCHAR(%d)) AS VARCHAR(%d))",
-                    dLen, dLen, dLen, dLen);
+            sprintf(buf, "CAST(CAST(@A1 AS VARCHAR(%d)) AS VARCHAR(%d))",
```
DaveBirdsall: Just curious: Why do we need two casts? Would one be enough? (Looks like we use the same length for both.)

traflm: That is my question as well :-) 
I don't really understand the syntax here, if without the out CAST, the parser will fail to parse this string.
There are many such cases, I wish to understand the special syntax here later. But this is what it required for now.

PR #1401, opened 2018-01-16
[TRAFODION-2911] Catalog Api gives wrong values about NON_UNIQUE column

Weixin-Xu:
Table TRAFODION.\"_MD_\".INDEXES only has column "is_unique", but the output column is NON_UNIQUE.


Weixin-Xu:
The failure of jenkins is caused by jdbc_test.
The index is normal, but the expected value of NON_UNIQUE column is false, it is wrong

Weixin-Xu:
jenkins, retest

PR #1402, opened 2018-01-16
[TRAFODION-2857] Remove remaining "incubator" path on downloads page

svarnau:
As Seth pointed out on the user list, I missed updating paths on the
downloads web page.  The old release names still have "incubating" in
the dir and file names, but the whole trafodion dist site has moved
out of the incubator directory.


zellerh:
+1

PR #1404, opened 2018-01-18
Trafodion 2916

SuJinpei:
**the root cause**:  the allocating size of str is determined by max field length + 128(here is 1500+128), but during loading, str will be filled with actual field data exceeding this size(here maybe 2000 characters) if str overflow the client app will crash. 
**solution**: increase the size of str to the size of one-time fread(default is 262144). there is still a risk of a crash when a user provides an incredibly large size of an invalid field.
**additional fix**: also fix other same type potential bugs in Oload2 and OloadJason.

### DIFF
--- a/core/conn/odb/src/odb.c
+++ b/core/conn/odb/src/odb.c
@@ -2314,7 +2314,6 @@ int main(int ac, char *av[])
             (void)signal(SIGINT, sigcatch);                     /* Keyboard Ctrl-C */
             (void)signal(SIGTERM, sigcatch);                    /* Software termination (kill) */
             WaitForMultipleObjects(i, thhn, TRUE, INFINITE);    /* wait threads */
-            Sleep(1000 * 10);
             for ( i = 0 ; i < tn ; i++)                         /* close thread handles */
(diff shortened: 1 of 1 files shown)


core/conn/odb/src/odb.c
```diff
@@ -7080,7 +7080,7 @@ static void Oload(int eid)
             mfl = (size_t) etab[eid].td[i].Osize;
 
     /* Allocate field buffer */
-    if ( (str = (char *)malloc (mfl + 128)) == (void *)NULL ) {
+    if ( (str = (char *)calloc (1, etab[eid].buffsz + 1)) == (void *)NULL ) {
```
sureshsubbiah: Is the buffer str used to read in source data for string column and reused for the next row/column? I am anxious that we do not do this large allocation repeatedly. I feel this is addressed already, the question is more of a sanity check as I cannot trace the variable str back far enough.

SuJinpei: Hi Suresh
      str will only be allocated one time for each loading thread.

PR #1405, opened 2018-01-18
[TRAFODION-2873] Removed usage of system new from LobGlobals and other classes

sandhyasun:
ExpOBInterfaceInit will take a parent heap and derive a lobHeap from it. ExLobGlobals will get created from this heap and store this heap pointer for use by other classes.
ExpLOBinterfaceCleanup  destroys the LOB globals and will utilize the heap pointer stored within it. 
In cli/Cli.cpp , Removed the initialization/destruction  of LOB glabals from the ::createLOB interface and moved it out before the interface is called to be consistent with other LOB API usage.


core/sql/exp/ExpLOBaccess.h
```diff
@@ -589,7 +589,7 @@ class ExLobGlobals
 {
   public :
   
-    ExLobGlobals(); 
+    ExLobGlobals(NAHeap *lobHeap=NULL);
```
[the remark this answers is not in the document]
DaveBirdsall: To @selvaganesang's point, perhaps we should leave this without a default, and then code NULL explicitly in ExpLobProcess.cpp?

core/sql/cli/Cli.cpp
```diff
@@ -9611,12 +9611,28 @@ Lng32 SQLCLI_LOBddlInterface
 	    
 	  } // for
 
+        //Initialize LOB interface 
+        
+        Int32 rc= ExpLOBoper::initLOBglobal(exLobGlob,currContext.exHeap(),&currContext,hdfsServer,hdfsPort);
+        if (rc)
+          {
+            {
```
DaveBirdsall: It doesn't hurt anything, but I'm curious: Why two sets of braces? Looks like one would do?

core/sql/cli/Cli.cpp
```diff
@@ -9726,10 +9742,26 @@ Lng32 SQLCLI_LOBddlInterface
         //lob data files.  Note that if there is an error in the drop of the 
         //descriptor tables above , the transaction will restore each of the 
         //above tables . 
+        //Initialize LOB interface 
+       
+        Int32 rc= ExpLOBoper::initLOBglobal(exLobGlob,currContext.exHeap(),&currContext,hdfsServer,hdfsPort);
+        if (rc)
+          {
+            {
```
DaveBirdsall: Same comment here

core/sql/cli/Cli.cpp
```diff
@@ -9765,12 +9797,27 @@ Lng32 SQLCLI_LOBddlInterface
 	    
 	    goto error_return;
 	  }
-	
+	//Initialize LOB interface 
+        
+        Int32 rc= ExpLOBoper::initLOBglobal(exLobGlob,currContext.exHeap(),&currContext,hdfsServer,hdfsPort);
+        if (rc)
+          {
+            {
```
DaveBirdsall: And here

core/sql/exp/ExpLOBaccess.h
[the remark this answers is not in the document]
sandhyasun: Sure that's another way. Will do that.

core/sql/exp/ExpLOBprocess.cpp
```diff
@@ -575,7 +575,7 @@ Lng32 main(Lng32 argc, char *argv[])
     // setup log4cxx
     QRLogger::initLog4cxx(QRLogger::QRL_LOB);
     // initialize lob globals
-    lobGlobals = new ExLobGlobals();
+    lobGlobals = new ExLobGlobals(NULL);
```
selvaganesang: Thanks for avoiding the default value for the param. But, I think sending heap as NULL might make this program to fail. Have you tested this program? If you have tested this program already, then this PR is ready to be merged.

sandhyasun: The ExpLOBprocess.cpp is dormant code for now. If we enable it and create a LOB process to offload some operations like GC etc, we will need to create and setup globals , send messages form the master process to it etc .  For now I just had to ensure it compiled.

PR #1407, opened 2018-01-19
Add hibernate dialect support.

xlloveyun:


tools/hibernate-dialect/pom.xml
```diff
@@ -0,0 +1,30 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
kevinxu021: License head is required for every file.

kevinxu021:
+1 Good!

svarnau:
I have a concern here. Maven POM is set to pull in the hibernate-core dependency, which appears to be licensed under LGPL.  That is not something we can bundle.

DaveBirdsall:
@svarnau, did the latest change answer your concern?

svarnau:
No, my concern is not addressed. There is still license issue with the hibernate dependency, as far as I can see.

kevinxu021:
I don't familiar with the license issue. Is there any document saying we can not use? As i know if we want to make any changes on the source code, it should involve license issue. We also did use Python which is GPL license in DCS pom.xml. What do you think or any body we could ask?

kevinxu021:
Looked a pom.xml in Apache Spark and some of the dependencies are not from Apache. So I think it might be ok( correct me if i am wrong).

PR #1408, opened 2018-01-19
[TRAFODION-2921] Add Syntax and Examples of *UPSERT USING LOAD* for *LOAD Statement* for *Trafodion SQL Reference Manual*

liuyu000:


sureshsubbiah:
I too would prefer that we do not document this feature. I am considering removing it. I am so sorry that effort was spent in documenting it.

The syntax for this statement is present in the parser. As present in code one could make a case that LOAD section is an appropriate place to document it. The feature grew out of a requirement that UPSERT  USING LOAD has no transactional protection in case of failure, while LOAD has protection against failure using snapshot mechanism. With this feature one can use the LOAD framework in code, to do UPSERT USING LOAD, and thereby use snapshot for protection against failure.

I would like to deprecate this syntax because
1) It is rather confusing. It is not clear what exactly it is, LOAD, UPSERT or UPSERT USING LOAD
2) UPSERT USING LOAD is typically used for performance reasons. With the overhead of using the utility LOAD framework and creating snapshots, the performance benefit is lost except for the largest upserts.

liuyu000:
@selvaganesang  @sureshsubbiah 
Sorry I read your comments just now, thank you for your reviews and detailed comments :)
I've moved the example of `UPSERT USING LOAD` to `UPSERT Statement` and created a PR for it (Click [Here](https://github.com/apache/trafodion/pull/1411)).
I'll close this PR.

PR #1410, opened 2018-01-22
TRAFODION-2922 fix return incorrect affect row count

partnation:
The batchRowCount array is used to store rowcount for each batch sql comman,it should be the return value of ist_.getRowCount() . BatchRowCount_new twice, apply for extra memory.


kevinxu021:
Good to go!

sureshsubbiah:
jenkins, add user

sandhyasun:
+1 Looks like this is ready to be merged.

PR #1412, opened 2018-01-23
TRAFODION-2885 enhancement for dcs server restart

mashengchen:
this is the enhancement for TRAFODION-2885


mashengchen:
jenkins , retest

PR #1413, opened 2018-01-24
[TRAFODION-2802 ] source the .bashrc before access JAVA_HOME

xiaozhongwang:
modify for running on a new server

### DIFF
--- a/install/traf_checkset_env.sh
+++ b/install/traf_checkset_env.sh
@@ -244,11 +244,14 @@ if [ ! -d ${javapath} ]; then
      exit 1
 fi
 
+source $HOME/.bashrc
 javahome=`grep JAVA_HOME ~/.bashrc | wc -l`
 if [ "x${javahome}" = "x0" ]; then
     echo -en "\n# Added by traf_checkset_env.sh of trafodion\n" >> $HOME/.bashrc
     echo -en "export JAVA_HOME=${javapath}\n" >> $HOME/.bashrc
     echo -en "export PATH=\$PATH:\$JAVA_HOME/bin\n" >> $HOME/.bashrc
+    export JAVA_HOME=${javapath}
+    export PATH=$PATH:$JAVA_HOME/bin
 else
     java_version=`${JAVA_HOME}/bin/java -version 2>&1 | awk 'NR==1{ gsub(/"/,""); print $3}'`
     case ${java_version} in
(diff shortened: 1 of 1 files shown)


install/traf_checkset_env.sh
```diff
@@ -244,11 +244,14 @@ if [ ! -d ${javapath} ]; then
      exit 1
 fi
 
+source $HOME/.bashrc
```
zellerh: Sourcing ~/.bashrc and looking into this file is not a very good solution, IMHO. There are so many other ways we can set JAVA_HOME. For example: /etc/bashrc, ~/.trafodion, $TRAF_HOME/.sqenvcom.sh.

I would suggest to just check whether JAVA_HOME is set, otherwise duplicate the code in $TRAF_HOME/.sqenvcom.sh that sets JAVA_HOME. If JAVA_HOME is not already set, then you can then add a line 

export JAVA_HOME=...

To the file ~/.trafodion.

install/traf_checkset_env.sh
```diff
@@ -244,11 +244,14 @@ if [ ! -d ${javapath} ]; then
...
 javahome=`grep JAVA_HOME ~/.bashrc | wc -l`
```
zellerh: See comment above

install/traf_checkset_env.sh
```diff
@@ -244,11 +244,14 @@ if [ ! -d ${javapath} ]; then
...
 if [ "x${javahome}" = "x0" ]; then
     echo -en "\n# Added by traf_checkset_env.sh of trafodion\n" >> $HOME/.bashrc
     echo -en "export JAVA_HOME=${javapath}\n" >> $HOME/.bashrc
     echo -en "export PATH=\$PATH:\$JAVA_HOME/bin\n" >> $HOME/.bashrc
```
zellerh: See comment above. Writing into the user's .bashrc is probably not a very good idea. You can create a file ~/.trafodion instead.

install/traf_checkset_env.sh
xiaozhongwang: Before, we must prepare environment step by step as following document: [https://cwiki.apache.org/confluence/display/TRAFODION/Create+Build+Environment]
I only write them in one command. 
Yes. I think this is not a good way too.

xiaozhongwang:
@DaveBirdsall I build environment on a new server.
Because this cause a error when run first time, I can rerun this script instead, while I choose to modify the script.

PR #1417, opened 2018-01-26
[TRAFODION-2917] Refactor Trafodion implementation of hdfs scan for t…

selvaganesang:
…ext formatted hive tables

Part-1 changes.

Created a new class org.trafodion.sql.HDFSClient. Any direct HDFS access
will be routed to this class via JNI instead of using libhdfs.

Modified the existing code expect for the following to route the HDFS request via this class
1. LOB access
2. Direct HDFS scan of the table
3. Sample data creation during update stats

Added a new class org.trafodio.sql.HdfsScan for scanning one or many ranges of a Hive
text formatted table. This class will be used for Direct HDFS scan in near future.


core/sql/executor/ExFastTransport.h
```diff
@@ -407,6 +408,7 @@ class ExHdfsFastExtractTcb : public ExFastExtractTcb
                           
   NABoolean isSequenceFile();
   void createSequenceFileError(Int32 sfwRetCode);
+  void createHdfsClientFileError(Int32 sfwRetCode);
```
eowhadi: should be hdfsClientRetCode instead of sfwRetCode

core/sql/executor/ExHbaseAccess.cpp
```diff
@@ -502,6 +506,8 @@ void ExHbaseAccessTcb::freeResources()
      NADELETEBASIC(directRowBuffer_, getHeap());
   if (colVal_.val != NULL)
      NADELETEBASIC(colVal_.val, getHeap());
+  if (hdfsClient_ != NULL) 
+     NADELETE(hdfsClient_, HdfsClient, getHeap());
 }
```
eowhadi: I am wondering if there is a need to delete the new introduces loggingFileName_ here?

selvaganesang: Yes. It is a good catch. I will fix this too in my next commit.  This won't cause memory leak as such because the heap is destroyed.

core/sql/executor/ExFastTransport.h
selvaganesang: Will do in my next commit though it is a just parameter name in the method declaration.

selvaganesang:
Read paths are not yet affected. The affected write path are
1. Log error rows creation in the LOAD command
2. UNLOAD command that extracts the trafodion data to hdfs files.

I thought that it would be clear from the modified files, hence I didn't add it in the description.

core/sql/executor/ExFastTransport.cpp
```diff
@@ -1248,6 +1279,23 @@ void ExHdfsFastExtractTcb::createSequenceFileError(Int32 sfwRetCode)
...
+
+  ComDiagsArea * diagsArea = NULL;
+  char* errorMsg = hdfsClient_->getErrorText((HDFS_Client_RetCode)hdfsClientRetCode);
+  ExRaiseSqlError(getHeap(),
+                  &diagsArea,
+                  (ExeErrorCode)(8447),
+                  NULL, NULL, NULL, NULL,
+                  errorMsg,
+                (char *)currContext->getJniErrorStr().data());
+  //ex_queue_entry *pentry_down = qParent_.down->getHeadEntry();
```
sureshsubbiah: Consider removing these 2 lines of comments in next commit. There are similar lines in the createSequenceFileError() method that could be removed.

core/sql/executor/ExHbaseIUD.cpp
```diff
@@ -1158,7 +1158,7 @@ ExHbaseAccessBulkLoadPrepSQTcb::ExHbaseAccessBulkLoadPrepSQTcb(
                       "traf_upsert_err",
                       fileNum,
                       loggingFileName_);
-   LoggingFileCreated_ = FALSE;
+   loggingFileCreated_ = FALSE;
```
sureshsubbiah: Is this line necessary now? loggingFileCreated_ belongs to the base class now I gather. Its constructor would have done this already.

core/sql/executor/ExHdfsScan.cpp
```diff
@@ -283,6 +285,8 @@ void ExHdfsScanTcb::freeResources()
      ExpLOBinterfaceCleanup(lobGlob_, (NAHeap *)getGlobals()->getDefaultHeap());
      lobGlob_ = NULL;
   }
+  if (hdfsClient_ != NULL) 
+     NADELETE(hdfsClient_, HdfsClient, getHeap());
```
sureshsubbiah: Same comment as Eric on ExHBaseAccess::freeResources(). Should we release loggingFileName_ here? Constructor guarantees it is never null.

core/sql/executor/ExHdfsScan.cpp
```diff
@@ -1948,6 +1948,54 @@ short ExHdfsScanTcb::handleDone(ExWorkProcRetcode &rc)
   return 0;
 }
 
+void ExHdfsScanTcb::handleException(NAHeap *heap,
```
sureshsubbiah: Makes me wish we had a common base class for HBaseAccess, HdfsClient and maybe SequenceFileRead/Write. Any think that requiresJava/JNI to go to another file format. Could some of this logic be shared in that common base class? This is not a request for change, simply a question that will help me understand future refactoring choices.

core/sql/executor/HdfsClient_JNI.cpp
```diff
@@ -0,0 +1,452 @@
...
+// ===== Class HdfsScan
+// ===========================================================================
+
+JavaMethodInit* HdfsScan::JavaMethods_ = NULL;
+jclass HdfsScan::javaClass_ = 0;
+bool HdfsScan::javaMethodsInitialized_ = false;
+pthread_mutex_t HdfsScan::javaMethodsInitMutex_ = PTHREAD_MUTEX_INITIALIZER;
+
+static const char* const hdfsScanErrorEnumStr[] = 
+{
```
sureshsubbiah: I am surprised that this is empty. Is that because HdfsScan java side is now in preview and error handling has not been introduced yet?

core/sql/executor/ExHdfsScan.cpp
selvaganesang: JavaObjectInterface is the base class for all these classes.

core/sql/executor/HdfsClient_JNI.cpp
selvaganesang: This will be populated when HdfsScan JNI methods are introduced

core/sql/executor/ExHbaseIUD.cpp
selvaganesang: Yes. I will confirm and remove it

core/sql/executor/HdfsClient_JNI.cpp
```diff
@@ -0,0 +1,452 @@
...
+  }
+  return rc;
+}
+        
+char* HdfsScan::getErrorText(HDFS_Scan_RetCode errEnum)
+{
+  if (errEnum < (HDFS_Scan_RetCode)JOI_LAST)
+    return JavaObjectInterface::getErrorText((JOI_RetCode)errEnum);
+  else
+    return (char*)hdfsScanErrorEnumStr[errEnum-HDFS_SCAN_FIRST-1];
```
sureshsubbiah: I wonder why there is a "-1" here, but no equivalent "-1"  in HdfsClient::getErrorText()

core/sql/src/main/java/org/trafodion/sql/HDFSClient.java
```diff
@@ -0,0 +1,319 @@
...
+  
+        Path dstPath = new Path(dstPathStr);
+        dstPath = dstPath.makeQualified(dstPath.toUri(), null);
+        FileSystem dstFs = FileSystem.get(dstPath.toUri(),config_);
+        
+        if (dstFs.exists(dstPath))
+        {
+          if (logger_.isDebugEnabled()) logger_.debug("HDFSClient.hdfsMergeFiles() - destination files exists" );
+          // for this prototype we just delete the file-- will change in next code drops
+          dstFs.delete(dstPath, false);
```
sureshsubbiah: Maybe this, and the following two lines of comments should just be removed? Are we supposed to silently delete existing file and overwrite with new merged file?

core/sql/src/main/java/org/trafodion/sql/HDFSClient.java
```diff
@@ -0,0 +1,319 @@
...
+      executorService_ = Executors.newCachedThreadPool();
+      try {
+         defaultFs_ = FileSystem.get(config_);
+      }
+      catch (IOException ioe) {
+         throw new RuntimeException("Exception in HDFSClient static block", ioe);
+      }
+   }
+
+   class HDFSRead implements Callable
```
sureshsubbiah: Could you please explain how the classes HdfsClient, HdfsClient.HDFSRead and HdfsScan are related? Thank you for the nice comments in HdfsScan.java. I took HdfsClient to be the class that contains all the methods that we removed from SequenceFileWriter.  If that is true, why do we need a HDFSRead subclass? Is this for the future or is there some functionality that I missed. For error row logging do we need read?

core/sql/src/main/java/org/trafodion/sql/HDFSClient.java
```diff
@@ -0,0 +1,319 @@
...
+      Path filePath = null;
+      if (!compress || (compress && fname.endsWith(".gz")))
+        filePath = new Path(fname);
+      else
+        filePath = new Path(fname + ".gz");
+        
+      FileSystem fs = FileSystem.get(filePath.toUri(),config_);
+      FSDataOutputStream fsOut = fs.create(filePath, true);
+      
+      if (compress) {
```
sureshsubbiah: Since adding a new compression type takes only a few lines of code, it will be good to look into increasing the types of compressed unload files we can write to. I can file a JIRA if there is interest. It is unrelated to focus of this change.

core/sql/src/main/java/org/trafodion/sql/HDFSClient.java
selvaganesang: HdfsClient is meant for both read and write to hdfs text files. Need to check if we can use for other format files too. HdfsClient is also used to create Hdfs files..

I have made 2nd commit that might help to understand it better with some comments in ExHdfsScan.h

selvaganesang:
@zellerh Can you please review this change

selvaganesang:
Thanks @sureshsubbiah for review the first commit. I have pushed 3 more commits that completes the refactoring of HDFS scan via java APIs except for Hdfs files with compression.  This change is not yet enabled. It can be enabled for testing via  cqd use_libhdfs_scan 'off'. I am still working on the failures seen with hive test suite regressions with this change.

selvaganesang:
Friendly reminder to @sureshsubbiah  to review this change again

sureshsubbiah:
Thank you @selvaganesang for patiently waiting for and for encouraging review. I promise to complete mine in the next 24 hrs.

core/sql/src/main/java/org/trafodion/sql/HDFSClient.java
```diff
@@ -0,0 +1,319 @@
...
+      {
+      }
+ 
+      public Object call() throws IOException 
+      {
+         int bytesRead;
+         int totalBytesRead = 0;
+         if (! buf_.hasArray()) {
+            try {
+              fsdis_.seek(pos_);
```
sureshsubbiah: Should we set readahead on this stream?

core/sql/src/main/java/org/trafodion/sql/HdfsScan.java
```diff
@@ -0,0 +1,289 @@
...
+      for (int i = 0; i < 2 ; i++) {
+          if (buf_[i].hasArray())
+             bufLen_[i] = buf_[i].array().length;
+          else
+             bufLen_[i] = buf_[i].capacity();
+      }
+      hdfsClient_ = new HDFSClient[2];
+      hdfsScanRanges_ = new HdfsScanRange[filename.length]; 
+      for (int i = 0; i < filename.length; i++) {
+         hdfsScanRanges_[i] = new HdfsScanRange(filename[i], pos[i], len[i], rangeNum[i]);
```
sureshsubbiah: An interesting will be to create ranges, (i.e. files) with 0 bytes. We know that sqoop sometimes creates such files.

core/sql/executor/ExHdfsScan.cpp
```diff
@@ -380,10 +407,12 @@ ExWorkProcRetcode ExHdfsScanTcb::work()
   HdfsFileInfo *hdfo = NULL;
   Lng32 openType = 0;
   int changedLen = 0;
- ContextCli *currContext = getGlobals()->castToExExeStmtGlobals()->getCliGlobals()->currContext();
-   hdfsFS hdfs = currContext->getHdfsServerConnection(hdfsScanTdb().hostName_,hdfsScanTdb().port_);
-   hdfsFileInfo *dirInfo = NULL;
-   Int32 hdfsErrorDetail = 0;//this is errno returned form underlying hdfsOpenFile call.
+  ContextCli *currContext = getGlobals()->castToExExeStmtGlobals()->getCliGlobals()->currContext();
+  hdfsFS hdfs = currContext->getHdfsServerConnection(hdfsScanTdb().hostName_,hdfsScanTdb().port_);
```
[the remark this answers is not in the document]
selvaganesang: Didn't make any change in this line except for aligning it

core/sql/executor/ExHdfsScan.cpp
```diff
@@ -514,11 +541,108 @@ ExWorkProcRetcode ExHdfsScanTcb::work()
...
+             bufLogicalEnd_ = hdfsScanBuf_[retArray_[BUF_NO]].buf_ + retArray_[BYTES_COMPLETED] - extraBytesRead_;
+             prevRangeNum_ = retArray_[RANGE_NO];
+             headRoomCopied_ = 0;
+             if (recordSkip_) {
+		hdfsBufNextRow_ = hdfs_strchr((char *)bufBegin_,
+			      hdfsScanTdb().recordDelimiter_, 
+                              (char *)bufEnd_,
+			      checkRangeDelimiter_, 
+			      hdfsScanTdb().getHiveScanMode(), &changedLen);
+                if (hdfsBufNextRow_ == NULL) {
```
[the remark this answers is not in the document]
selvaganesang: Yes

core/sql/executor/ExHdfsScan.cpp
```diff
@@ -1149,8 +1297,12 @@ ExWorkProcRetcode ExHdfsScanTcb::work()
 	        if (hdfsScanTdb().continueOnError())
 	        {
 	          if ((pentry_down->downState.request == ex_queue::GET_N) &&
-	              (pentry_down->downState.requestValue == matches_))
-	            step_ = CLOSE_FILE;
+	              (pentry_down->downState.requestValue == matches_)) {
+                     if (useLibhdfsScan_)
+                        step_ = CLOSE_HDFS_CURSOR;
```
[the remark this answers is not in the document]
selvaganesang: No Thanks for catching it.. Will change it back to CLOSE_HDFS_CURSOR

core/sql/executor/ExExeUtilGet.cpp
```diff
@@ -3521,13 +3521,9 @@ ExExeUtilGetHbaseObjectsTcb::ExExeUtilGetHbaseObjectsTcb(
...
      : ExExeUtilGetMetadataInfoTcb( exe_util_tdb, glob)
 {
-  int jniDebugPort = 0;
-  int jniDebugTimeout = 0;
   ehi_ = ExpHbaseInterface::newInstance(glob->getDefaultHeap(),
 					(char*)exe_util_tdb.server(), 
-					(char*)exe_util_tdb.zkPort(),
-                                        jniDebugPort,
-                                        jniDebugTimeout);
+					(char*)exe_util_tdb.zkPort());
```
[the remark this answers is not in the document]
selvaganesang: DebugPort needs to be initialized as part of JVM invocation. See JavaObjectInterface:;createJVM. This debugPort is never used. Because it is per process, it needs be be made static

core/sql/src/main/java/org/trafodion/sql/HDFSClient.java
```diff
@@ -0,0 +1,319 @@
...
+   {
+      HDFSRead() 
+      {
+      }
+ 
+      public Object call() throws IOException 
+      {
+         int bytesRead;
+         int totalBytesRead = 0;
+         if (! buf_.hasArray()) {
```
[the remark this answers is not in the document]
selvaganesang: HdfsScan supports both direct and non-direct buffers.  Currently we use Direct byte buffers wherein the ByteBuffer is wrapping the native array in C++. However, it is possible for hdfsScan to use non-direct ByteBuffer which is wrapper for byteArray in jVM memory

core/sql/executor/ExHdfsScan.cpp
```diff
@@ -118,15 +119,39 @@ ExHdfsScanTcb::ExHdfsScanTcb(
...
-  hdfsScanBuffer_[readBufSize] = '\0';
-
+  hdfsScanBufMaxSize_ = hdfsScanTdb.hdfsBufSize_;
+  headRoom_ = (Int32)hdfsScanTdb.rangeTailIOSize_;
+
+  if (useLibhdfsScan_) {
+     hdfsScanBuffer_ = new(heap) char[ hdfsScanBufMaxSize_ + 1 ]; 
+     hdfsScanBuffer_[hdfsScanBufMaxSize_] = '\0';
+  } else {
+     hdfsScanBufBacking_[0] = new (heap) BYTE[hdfsScanBufMaxSize_ + 2 * (headRoom_)];
```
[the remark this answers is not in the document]
selvaganesang: This behavior remains unchanged. Let me check if there is a regression test suite already for it

core/sql/generator/GenRelScan.cpp
```diff
@@ -1391,6 +1391,9 @@ if (hTabStats->isOrcFile())
   hdfsscan_tdb->setUseCif(useCIF);
   hdfsscan_tdb->setUseCifDefrag(useCIFDegrag);
 
+  if (CmpCommon::getDefault(USE_LIBHDFS_SCAN) == DF_ON)
```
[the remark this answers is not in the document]
selvaganesang: I need to confirm if compression is already supported.  Are there any tests for compression in hive regressions already? If so, it is handled already.

selvaganesang:
jenkins, retest

PR #1420, opened 2018-01-29
[TRAFODION-2912] Better handling of non-deterministic scalar UDFs

zellerh:
Fix some issues found by Andy Yang and others while writing a
non-deterministic scalar UDF (a random generator in this case).

This UDF was transformed into a hash join, which executes the UDF
only once and not once per row. Another problem is the probe cache,
which can also lead to a single execution instead of once per row.

The fix records the non-deterministic UDF attribute in the group
attributes and it adds checks in the normalizer to suppress the
conversion from a TSJ to a non-TSJ when non-deterministic UDFs are
present. The probe cache logic already had this check, so all that was
needed was to set the attribute.

Note that there may be some more complex queries where we still won't
execute the UDF once per row. In general, there is no absolute
guarantee that a non-deterministic scalar UDF is executed once per row
(of the cartesian product of all the tables joined??). However, in
simple cases like the added test we should try to call the UDF for
every row that satisfies the join predicates.

### DIFF
--- a/core/sql/regress/udr/EXPECTED103
+++ b/core/sql/regress/udr/EXPECTED103
@@ -209,6 +209,68 @@ CREATE LIBRARY TRAFODION.UDR103SCH.FUNCTIONSFORTEST103 FILE '/mnt2/ansharma/ansh
 
 --- SQL operation complete.
 >>
+>>create function nonDeterministicRandom
++>  () returns (r integer)
++>  language c parameter style sql external name 'nonDeterministicRandom'
++>  library functionsForTest103
++>  not deterministic no sql final call allow any parallelism state area size 1024 ;
+
+--- SQL operation complete.
+>>
+>>cqd nested_joins 'off';
+
+--- SQL operation complete.
+>>cqd merge_joins 'off';
+
+--- SQL operation complete.
+>>cqd join_order_by_user 'on';
+
+--- SQL operation complete.
+>>prepare s from
++>select a, nonDeterministicRandom() r1,
++>       generateRandomNumber(a, 4) r2, generateRandomNumber(123, 4) r3
++>from (values (1), (2), (3)) T(a);
+
+--- SQL command prepared.
+>>explain options 'f' s;
+
+LC   RC   OP   OPERATOR              OPT       DESCRIPTION           CARD
+---- ---- ---- --------------------  --------  --------------------  ---------
+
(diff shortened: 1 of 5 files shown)


core/sql/regress/udr/TEST103_functions.cpp
```diff
@@ -117,11 +117,41 @@ SQLUDR_LIBFUNC SQLUDR_INT32 genRandomNumber(SQLUDR_INT32 *in1,
...
+  int *my_state = (int *) statearea->stmt_data.data;
+
+  if (calltype == SQLUDR_CALLTYPE_INITIAL && *my_state == 0)
+    {
+      *my_state = 555;
+    }
+  else
+    // Use a simple linear congruential generator, we
+    // want deterministic regression results, despite
+    // the name of this function. Note that a
```
sureshsubbiah: Typo. Extra "a". Smallest nit possible :)

core/sql/regress/udr/EXPECTED103
```diff
@@ -396,6 +458,68 @@ CREATE LIBRARY TRAFODION.UDR103SCH.FUNCTIONSFORTEST103 FILE '/mnt2/ansharma/ansh
...
+>>cqd join_order_by_user 'on';
+
+--- SQL operation complete.
+>>prepare s from
++>select a, nonDeterministicRandom() r1,
++>       generateRandomNumber(a, 4) r2, generateRandomNumber(123, 4) r3
++>from (values (1), (2), (3)) T(a);
+
+--- SQL command prepared.
+>>explain options 'f' s;
```
sureshsubbiah: Will we get the same join order every time? I wonder what in the compiler affects the join order for such a query. This is a general question, not specific to this change.

core/sql/regress/udr/TEST103_functions.cpp
zellerh: Thanks, I think I probably fixed the issues I wanted to comment on with some CQDs. Will delete the half sentence.

core/sql/regress/udr/EXPECTED103
zellerh: Yes, I was concerned about that as well and I saw different join orders. For that reason I used a bunch of CQDs - HJ, MJ off and JOBU, to prevent random plan changes.

PR #1422, opened 2018-01-30
pyinstaller fix - kerberos related issues while executing hdfs commands

mkby:


robertamarton:
Thanks for making this change.

DaveBirdsall:
Could you add a JIRA for this? Also, I notice this is set up for release 2.2. That release is presently under vote. Did you intend this to go into release 2.2? (@traflm, do you approve this change for release 2.2?)

robertamarton:
See https://issues.apache.org/jira/browse/TRAFODION-2935

     Roberta

DaveBirdsall:
Hi,

So, it sounds like we need to reference [TRAFODION-2935] in this pull request.

Is the consensus then that we'll do another release candidate build for release2.2 and another vote once this is applied?

Dave

sandhyasun:
@mkby   I think this PR should go against Trafodion  2.3.   @traflm  Do you agree ?

DaveBirdsall:
@sandhyasun, there is an e-mail on the dev list where @traflm lists this JIRA (TRAFODION-2935) as a must-fix for the next RC3 build of Release 2.2.

sandhyasun:
Ok. Th JIRA is not marked as a blocker and testing with kerberos wasn't part of the goal for the 2.2 release I thought. But will let the release manager decide and mark it as needed. Thanks !

robertamarton:
Actually, it is a regression.  Kerberos has been available on Trafodion for the last several releases.   I test it before every release.
This bug will not allow any installation that has Kerberos enabled to work.

If no one is using Kerberos or plans to use Kerberos, then it won’t be an issue.  However, our documentation states that Kerberos is supported.

     Roberta

DaveBirdsall:
That suggests that this is a "Must Fix".  +1 to including it in R2.2 RC3.

sandhyasun:
+1 for 2.2

PR #1425, opened 2018-01-31
[Trafodion-2937] the actual copied data is less than expected when copying data from MySQL to Trafodion

SuJinpei:
**root cause:** During building extract command, odb tool will try to split the task into multithread if parallel, but the parallel is equal 1 here,  so it should still go to single thread branch. 
**fixed by:**  only if parallel bigger than 1, odb go splitting branch.
**Aditional fix:**  make code cleaner and odb support g++ 4.8.5

### DIFF
--- a/core/conn/odb/src/odb.c
+++ b/core/conn/odb/src/odb.c
@@ -3382,6 +3382,7 @@ Please note the fixed length '6' in strmicmp: SELECT/UPDATET/DELETE/INSERT have 
         case SQL_WCHAR:
             if ( etab[eid].dbt != VERTICA ) /* Vertica's CHAR field length is in bytes (not chars) */
                 Ors[j] *= etab[eid].bpwc;
+            /* FALLTHRU */
         case SQL_CHAR:
             if ( etab[eid].dbt != VERTICA ) /* Vertica's CHAR field length is in bytes (not chars) */
                 Ors[j] *= etab[eid].bpc;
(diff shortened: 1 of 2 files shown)


core/conn/odb/src/odb.c
```diff
@@ -10867,7 +10874,7 @@ static int Ocopy(int eid)
     }
 
     /* Build Extract Command */
-    if ( etab[eid].ps ) {
+    if ( etab[eid].ps > 1 ) {
```
sureshsubbiah: Could you please explain either here or in the JIRA how this change will address the bug. I am sorry I am familiar enough with the code to understand without comments.

SuJinpei: Hi Suresh,  if ps > 1, odb will split select job to number of ps thread, for example, if we have [1, 12] range of data and ps == 3, the data will first be divided to 3 group: [1, 4), [4, 7), [7, 10), then change last group to [7, 12),  the coding deal with last group in line 5830, but there make assumption ps > 1.  So for this bug, data range is [6, 10], will be split to [6, 10), because of ps == 1, this group has no chance to be changed. so data == 10 lost. I found many expression "if (etab[no].ps) " make an assumption that ps > 1 and below line(5835) contains a potential bug. so I will make another commit to fix this.  thank you for making me aware of these issues.

etab[no].sbmax = ( j + 1 ) == etab[l].ps ? (sbmax + 1 ) : etab[no].sbmin + d ;

SuJinpei: This change will address the bug because the default ps == 0 in fact means ps == 1,  so odb ought to go default branch if ps == 1.

sureshsubbiah:
+1. Thank you @SuJinpei for explaining. Thank you Dave for checking.

PR #1427, opened 2018-01-31
TRAFODION-2940 In HA env, one node lose network, when recover, trafci can't use

mashengchen:
when there loses network for a long time ,and then network recover, there will trigger zookeeper session expired, at this time ,check whether current dcsmaster is leader, if not unbind this node's floating ip, and make dcsmaster init status, then rerun dcs master.


dcs/src/main/java/org/trafodion/dcs/master/DcsMaster.java
```diff
@@ -111,11 +104,59 @@ public DcsMaster(String[] args) {
...
+                        while (true) {
+                            try {
+                                ZkClient tmpZkc = new ZkClient();
+                                tmpZkc.connect();
+                                tmpZkc.close();
+                                tmpZkc = null;
+                                LOG.info("Connected to ZooKeeper successful, restart DCS Master.");
+                                // reset lock
+                                isLeader = new CountDownLatch(1);
+                                break;
```
DaveBirdsall: I'm not sure I understand this logic. Do we sit inside one of these method calls during normal processing? Or does tmpZkc.connect() and close() complete immediately? If so, it looks like we just loop around the while loop and do it again, over and over.

mashengchen: this logic is for when dcsmaster return with network erro situation.
in the logic , it will try to connect to zk
if it can't conn ( tmpZkc.connect(); ) , there will in catch block and do retry
if it connect to zk, then dcsmaster will run call() method again, in the time dcsmaster rework ,there must hava another backup-master working ( because there must one dcs master work and current master lose network ,then backupmaster take over the role) ,  so when dcsmaster rework , it will set value in zk /rootpath/dcs/leader/ then hang by lock "isLeader = new CountDownLatch(1);"

traflm:
Hope some DCS experts to take a review. @mashengchen  please invite proper DCS experts. I cannot understand these changes well.

mashengchen:
@arvind-narain @kevinxu021 can you help to take a look

hegdean:
Can you please describe with more details regarding the original issue in the JIRA.  It is unclear to me the scenario with 2 floating IPs. Let's hold off on the merge until we understand the original issue that will help us understand the changes done

mashengchen:
1. config ha env
2. sqstart
3. use iptables to down master node's network, (iptables -I INPUT -s hostname -j DROP)
4. sleep for 300 seconds
5. backup-master take over the master role
6. recover network (iptables -I INPUT -s hostname -d DROP)
7. pdsh $MY_NODES ifconfig|grep 23400 will have 2 results, one is the down dcsmaster and another is the backup-master
actually , the old dcsmaster is still in while loop when network recover, so there will have 2 working dcsmaster.

dcs/bin/scripts/dcsunbind.sh
```diff
@@ -51,6 +51,31 @@ function check_node {
...
+        if [ $status -eq 0 ]; then
+            tempinterface=`echo $gv_float_interface:$gv_port`
+            # check if another interface is bound to this virtual ip address
+            echo "$myifport" | grep "$tempinterface"  > /dev/null
+            if [ $? -eq 0 ]; then
+                unbindip=`echo "$myifport" | awk '{print $2}'`
+                unbindlb=`echo "$myifport"|awk '{print $NF}'`
+                echo "Virtual ip $gv_float_external_ip is in use on node $HOSTNAME bound to interface $myinterface($unbindlb) - unbinding..."
+                sudo /sbin/ip addr del $unbindip dev $myinterface
+                status=$?
```
svarnau: It seems unnecessary to duplicate entire function just to change whether commands run locally or remotely. That can be done with a variable, so that we don't have two copies of this code to maintain.

mashengchen: this function is a little different from the one using PDSH,  the ouput of `$SQ_PDSH $MY_NODES /sbin/ip addr show` is different from `/sbin/ip link show` , so the cut step after the " | " (pipe) is different.

mashengchen:
@svarnau yes , I had done the test, and it solved the duplicate IP problem

dcs/src/main/java/org/trafodion/dcs/master/FloatingIp.java
```diff
@@ -61,6 +61,43 @@ public boolean isEnabled() {
...
+            sb.append(", stdout [" + scriptContext.getStdOut().toString() + "]");
+        if (!scriptContext.getStdErr().toString().isEmpty())
+            sb.append(", stderr [" + scriptContext.getStdErr().toString() + "]");
+        if (LOG.isErrorEnabled())
+            LOG.info(sb.toString());
+
+        if (scriptContext.getExitCode() == 0)
+            LOG.info("Unbind Floating IP successful");
+        else
+            LOG.error("Unbind Floating IP failed");
```
kevinxu021: It's better to print exit code.

dcs/src/main/java/org/trafodion/dcs/master/listener/ListenerService.java
```diff
@@ -263,14 +264,20 @@ else if ((key.interestOps() & SelectionKey.OP_WRITE) == SelectionKey.OP_WRITE) {
                         }
                     }
                 }
+                if (this.isInterrupted()) {
+                    throw new InterruptedException();
```
kevinxu021: Add some specific message for this exception.

dcs/src/main/java/org/trafodion/dcs/master/listener/ListenerService.java
```diff
@@ -263,14 +264,20 @@ else if ((key.interestOps() & SelectionKey.OP_WRITE) == SelectionKey.OP_WRITE) {
...
+                }
                 //gc();
             }
+        } catch (InterruptedException e) {
```
kevinxu021: Log what happens by info level at least.

dcs/src/main/java/org/trafodion/dcs/master/listener/ListenerService.java
```diff
@@ -263,14 +264,20 @@ else if ((key.interestOps() & SelectionKey.OP_WRITE) == SelectionKey.OP_WRITE) {
...
         } catch (IOException e) {
             LOG.error(e);
```
kevinxu021: it's better to use 2 parameters API for LOG.error

dcs/src/main/java/org/trafodion/dcs/master/listener/ListenerService.java
```diff
@@ -263,14 +264,20 @@ else if ((key.interestOps() & SelectionKey.OP_WRITE) == SelectionKey.OP_WRITE) {
...
             System.exit(1);
         } finally {
+            LOG.info("close ServerSocketChannel...");
             if (server != null) {
                 try {
+                    server.socket().close();
                     server.close();
                 } catch (IOException e) {
                     e.printStackTrace();
```
kevinxu021: it's better to print the message into log file.

dcs/src/main/java/org/trafodion/dcs/master/listener/ListenerWorker.java
```diff
@@ -93,6 +93,7 @@ public void run() {
                         try {
                             queue.wait();
                         } catch (InterruptedException e) {
+                            return;
```
kevinxu021: Log message for what happened by debug level.

dcs/src/main/java/org/trafodion/dcs/zookeeper/ZkClient.java
```diff
@@ -151,8 +163,28 @@ public void connect() throws IOException, InterruptedException {
...
+                    Stat stat = zk.exists(checkPath, false);
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug("stat = [" + stat + "].");
+                    }
+                    if (stat == null) {
+                        // this means master has change.
+                        setSessionRecoverSuccessful(false);
+                    }
+                } catch (KeeperException e) {
+                    e.printStackTrace();
```
kevinxu021: Log messages in file.

dcs/src/main/java/org/trafodion/dcs/zookeeper/ZkClient.java
```diff
@@ -176,7 +208,19 @@ public ZooKeeper getZk() {
...
-		}
+		} else if (event.getState() == Watcher.Event.KeeperState.Expired) {
+            LOG.info("session expired. now rebuilding");
+            // session expired, may be never happending. but if it happen there
+            // need to close old client and rebuild new client
+            try {
+                connect(true);
+            } catch (IOException e) {
+                setSessionRecoverSuccessful(false);
+                LOG.error("session expired and throw IOException while do reconnect: " + e.getMessage());
```
kevinxu021: Use 2 params API for log. If it doesn't matter for whatever the message is, it should be LOG.warn.

dcs/src/main/java/org/trafodion/dcs/zookeeper/ZkClient.java
```diff
@@ -176,7 +208,19 @@ public ZooKeeper getZk() {
...
+            } catch (InterruptedException e) {
+                LOG.error("session expired and throw InterruptedException while do reconnect: " + e.getMessage());
```
kevinxu021: The same as above.

dcs/src/main/java/org/trafodion/dcs/zookeeper/ZkClient.java
```diff
@@ -176,7 +208,19 @@ public ZooKeeper getZk() {
 	public void process(WatchedEvent event) {
 		if(event.getState() == Watcher.Event.KeeperState.SyncConnected) {
 			connectedSignal.countDown();
-		}
+		} else if (event.getState() == Watcher.Event.KeeperState.Expired) {
+            LOG.info("session expired. now rebuilding");
+            // session expired, may be never happending. but if it happen there
+            // need to close old client and rebuild new client
+            try {
+                connect(true);
```
kevinxu021: In connect(...), I saw it was catch for zookeeper exception, how can get exception here?

hegdean:
@arvind can u please review this

arvind:
I think the wrong Arvind was tagged :)

PR #1428, opened 2018-01-31
[TRAFODION-2874] New syntax to retrieve the LOB HDFS filename, offset for a LOBhandle

sandhyasun:
for both external and internal LOBs . Also added syntax to return starting offset of a particular LOB handle in the LOB Hdfs data file.


sandhyasun:
Looks like a parser change that's part of this PR is causing a clash in SPJ syntax. Investigating....

core/sql/cli/Cli.cpp
```diff
@@ -9160,11 +9160,45 @@ Lng32 SQLCLI_LOBcliInterface
...
+      {
+	
+	//Retrive offset of the first chunk
+	str_sprintf(query, "select  c.dataOffset from table(ghost table %s) h, table(ghost table %s) c where h.descPartnKey = c.descPartnKey and h.syskey = c.descSyskey and h.descPartnKey = %ld and h.syskey = %ld and c.chunkNum = 1 for read committed access",
+		    lobDescHandleName, lobDescChunksName, 
+		    descPartnKey, inDescSyskey);
+
+        lobDebugInfo(query,0,__LINE__,lobTrace);
+	// set parserflags to allow ghost table
+	currContext.setSqlParserFlags(0x1);
```
DaveBirdsall: Is there code somewhere to revert the parser flags setting?

core/sql/exp/ExpLOBaccess.cpp
```diff
@@ -870,6 +870,67 @@ Ex_Lob_Error ExLob::getLength(char *handleIn, Int32 handleInLen,Int64 &outLobLen
       }
   return err;
 }
+Ex_Lob_Error ExLob::getOffset(char *handleIn, Int32 handleInLen,Int64 &outLobOffset,LobsSubOper so, Int64 transId)
+{
+  char logBuf[4096];
+  Int32 cliErr = 0;
+  Ex_Lob_Error err=LOB_OPER_OK; 
+  char *blackBox = new(getLobGlobalHeap()) char[MAX_LOB_FILE_NAME_LEN+6];
```
DaveBirdsall: It doesn't look like this is used anywhere. Maybe we can delete this?

core/sql/exp/ExpLOBaccess.cpp
```diff
@@ -870,6 +870,67 @@ Ex_Lob_Error ExLob::getLength(char *handleIn, Int32 handleInLen,Int64 &outLobLen
...
+{
+  char logBuf[4096];
+  Int32 cliErr = 0;
+  Ex_Lob_Error err=LOB_OPER_OK; 
+  Int64 dummy = 0;
+  Int32 dummy2 = 0;
+  if (so != Lob_External_File)
+    {
+      //Derive the filename from the LOB handle and return
+      str_cpy_all(outFileName, (char *)lobDataFile_.data(),lobDataFile_.length());
```
DaveBirdsall: Do we know that the outFileName buffer is long enough?

core/sql/cli/Cli.cpp
sandhyasun: Good catch Dave ! It's missing here and from the code I cloned from the block above. Will fix both.

core/sql/exp/ExpLOBaccess.cpp
sandhyasun: Internally we allocate enough to hold the name. But when the final result is copied out, it's copied out to a caller  provided address (64 bit address). So we are assuming the caller (mxosrvr)  has allocated enough space.

sandhyasun: On further thought, maybe there needs to be an extra check in the sscp. in ::processLobRequest where we should check if another process on another node has already propgated/set the same lock into the local shared segment and if so error out and ensure that the current  SetLobkLock call fails .  This way only one of them succeeds. There can be cases where both processes repeatedly fail but with AQR and erroring out seems like a safer thing to do. Will set @selvaganesang  comment on this issue.

PR #1429, opened 2018-02-01
[TRAFODION-2927] Refactor and improve UPDATE STATISTICS logging

DaveBirdsall:
This set of changes achieves two things:

1. It provides a mechanism for UPDATE STATISTICS to capture log information for each command, and retain that log in the event of a severe error. (Otherwise the log is deleted.)

2. It refactors the UPDATE STATISTICS logging facility to use the QRLogger/CommonLogger/log4cxx mechanism used by the rest of the SQL engine.

Externals:

The UPDATE STATISTICS LOG statement has been extended as follows:

UPDATE STATISTICS LOG ON -- Starts continuous logging of all UPDATE STATISTICS and SHOWSTATS activity into one log file. Logging continues until the session has ended or a different UPDATE STATISTICS LOG command is entered. This is the same behavior as today, except the log file will be written to the $TRAF_HOME/logs directory like other SQL log files.

UPDATE STATISTICS LOG OFF -- Stops logging. Nothing further is logged until a different UPDATE STATISTICS LOG command is entered. This is the same behavior as today.

UPDATE STATISTICS LOG SYSTEM -- The "SYSTEM" option is new. When specified, each UPDATE STATISTICS statement (but not SHOWSTATS) is logged to a separate log file. If a severe error occurs (e.g. an internal error), the log file is retained. Otherwise it is deleted at the end of the UPDATE STATISTICS command.

A CQD has been added. CQD USTAT_AUTOMATIC_LOGGING, having values of 'ON' and 'OFF'. If 'ON', then UPDATE STATISTICS LOG SYSTEM semantics are turned on by default. If 'OFF', then UPDATE STATISTICS LOG OFF semantics are turned on by default. (Note that no equivalent of UPDATE STATISTICS LOG ON is provided as that is remarkably noisy; if one really wants to log continuously, then one must use UPDATE STATISTICS LOG ON directly.) 

The default for this CQD is 'OFF'. So by default no automatic logging is done. On large systems (e.g. scalability testing, or production systems with large tables), it is recommended to set this CQD to 'ON' in the system defaults table. In this way, logs will be captured for long-running UPDATE STATISTICS operations that fail with severe errors.

In terms of design: The existing QRLogger/CommonLogger routines assume a model of one log file per category, which doesn't meet the needs for UPDATE STATISTICS. So, QRLogger has been extended to add methods to explicitly configure log files on a per process, even per command basis. For such files, we don't bother logging the process ID and node number in the individual log messages, as all of the messages are coming from one process. Instead, UPDATE STATISTICS puts the node number, PID and Linux timestamp of file creation into the log file name itself.


core/sql/ustat/hs_log.cpp
```diff
@@ -389,44 +371,64 @@ NABoolean HSLogMan::GetLogFile(NAString & path, const char* cqd_value)
...
+            // strip off any directory path name; we will always use the logs directory
+            // as configured via QRLogger
+            while ((fileNameStem > ustatLog) && (*(fileNameStem - 1) != '/'))
+              fileNameStem--;
+          }
+
+        logFileName += fileNameStem;
+        logFileName += qualifiers;
+
+        NABoolean logStarted = QRLogger::startLogFile(CAT_SQL_USTAT,logFileName.c_str());
```
zellerh: This could allow a user to create a file anywhere the trafodion user has write access. For UDFs, we are creating a sandbox in $TRAF_HOME/udr for different types of users, beginning with a public area $TRAF_HOME/udr/public. I wonder whether it would be better here to sandbox the log files as well, e.g. into something like $TRAF_HOME/logs/ustat_logs. In the UDF code we also try to catch names like a/../../../export/lib/myfile that would escape the sandbox.

A related comment/question (a non-issue), maybe I can bring it up here as well, as it's related to the choice of log file directory: As far as I understand, these log files don't conform to the format we are using for executor, TM, DCS, etc. Therefore they can't be read by the event_log_reader UDF and usually they shouldn't be, unless they get placed into one of the directories where this UDF looks (e.g. $TRAF_HOME/logs) and they have a name prefix that matches the list supported by this UDF (e.g. "master_exec_" or "mon"). So, there should be no issue. To be safe, I would recommend keeping the update stats log files in a separate directory (maybe a subdirectory), so that the event_log_reader UDF doesn't have to look at them each time it is called. Also, this avoids errors for all users if somebody accidentally chooses one of the special prefixes like "mon" for their ustat log files.

DaveBirdsall: Re: your first paragraph. The code at line 394 gets the directory of the SQL catagory (which is presently configured as $TRAF_HOME/logs). The code from lines 398-408 strips off any directory path that a user might have specified in CQD USTAT_LOG. So, no matter what, files should always be created in $TRAF_HOME/logs. Regarding the idea of sandboxing in a sub-directory, I'll consider that; that would avoid the issue in your second paragraph.

zellerh: Sorry, I was thinking that QRLogger::getRootLogDirectory() might return an empty string and thought that lines 406 and 407 only remove the last name part of the USTAT_LOG directory.

Would this be another way to overwrite an arbitrary file: CQD USTAT_LOG '../../export/lib'?

zellerh: Sorry, never mind, my apologies for being slow, I realize my mistake now. I was looking at the part that got removed instead of the part that gets retained in fileNameStem.

DaveBirdsall:
When I made the original set of changes, I was largely ignorant of the event_log_reader UDF. Hans' comments brings this up. After thinking about it, it seems to me that in the spirit of integrating the UPDATE STATISTICS logs with the other SQL logs, I should make them compatible with the event_log_reader as well. So, in the rework, I have done so. The rework has three changes:

1. The row format is now the same as other SQL logs, so event_log_reader should be able to process them.
2. The file names have been restricted so now they always begin with the ustat prefix. The function of the CQD USTAT_LOG has been modified somewhat to achieve that goal. The file name portion of that CQD will now be used as a second qualifier in the ULOG name. (The default has been changed to the empty string, so by default there is no added second qualifier.) So, for example, if I have CQD USTAT_LOG 'birdsall' in place, any ULOGs produced by me will have file name ustat.birdsall.<node>.<pid>.<timestamp>.log. This addresses the concern about accidentally naming a ULOG "mon.<something>"; it will no longer be possible to do this.
3. I've updated the analyzeULOG.py script to deal with the new line format. It can still handle the old too; but one will have to supply the parameter --ULOGFormat old in order for it to do so.

core/sql/ustat/hs_log.cpp
zellerh: +1
Looks good to me. Nice idea to make the format similar to the other log files, so we could easily change the event_log_reader UDF to pick up these ustat logs as well.

zellerh:
+1
Looks good to me. Nice idea to make the format similar to the other log files, so we could easily change the event_log_reader UDF to pick up these ustat logs as well.

I'm entering this comment twice, since the first time I did so for the commit, and that comment on the commit doesn't seem to show up on the PR page.

DaveBirdsall:
Please do not merge yet. I will be making one more change to this pull request.

DaveBirdsall:
I changed the default for CQD USTAT_AUTOMATIC_LOGGING to "ON". I did this after doing a test to measure the overhead of having automatic logging on. My data showed that the cost was about 0.32 seconds of elapsed time per UPDATE STATISTICS command on average. This is in the realm of noise, as the elapsed time of repeated UPDATE STATISTICS commands on the same table varies more than this.

zellerh:
+1

PR #1433, opened 2018-02-01
[TRAFODION-2939] odb copied data exceeds the max size specified by max option when copying data from oracle to Trafodion

SuJinpei:
**root cause:** max rows were divided into each thread may cause etab[eid].mr < etab[eid].r, so even only one-time fetch, the actual fetched rows will exceed max row of each thread.
**fixed by:** limit etab[eid].r less or equal to etab[eid].mr.

### DIFF
--- a/core/conn/odb/src/odb.c
+++ b/core/conn/odb/src/odb.c
@@ -5491,7 +5491,7 @@ static void etabadd(char type, char *run, int id)
                     }
                 } else {                            /* not a load job */
                     etab[no].k = no;                /* record grandparent for copy/diff ops */
-                    if ( etab[no].ps ){
+                    if ( etab[no].ps > 1 ){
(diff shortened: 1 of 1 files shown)


core/conn/odb/src/odb.c
```diff
@@ -5491,9 +5491,11 @@ static void etabadd(char type, char *run, int id)
                     }
                 } else {                            /* not a load job */
                     etab[no].k = no;                /* record grandparent for copy/diff ops */
-                    if ( etab[no].ps ){
+                    if ( etab[no].ps > 1 ){
```
sureshsubbiah: Can you please explain why we don't need this IF block when etab[no].ps  =1 ? Is it it because for a extract/copy when we have one stream comparing rowset size to max number of rows is done elsewhere in this file?

PR #1434, opened 2018-02-01
[TRAFODION-2945] Remove unused file core/sql/optimizer/RoutineInvocation.h

wenjun-zhu:
File core/sql/optimizer/RoutineInvocation.h is not used anywhere in Trafodion.


sureshsubbiah:
+1. I don't recall when or why we added this file. This is good change unless someone had added the file for a future use.

PR #1439, opened 2018-02-05
[TRAFODION-2772] - retrieve a value from Json string got an error

andyyangcn:


core/sql/exp/exp_function.cpp
```diff
@@ -6503,8 +6503,15 @@ ex_expr::exp_return_type ex_function_json_object_field_text::eval(char *op_data[
         Int32 prec2 = ((SimpleType *)getOperand(2))->getPrecision();
         len2 = Attributes::trimFillerSpaces( op_data[2], prec2, len2, cs );
     }
+
     char *rltStr = NULL;
-    JsonReturnType ret = json_extract_path_text(&rltStr, op_data[1], 1, op_data[2]);
+    char *jsonStr = new(heap) char[len1+1];
+    char *jsonAttr = new(heap) char[len2+1];
```
DaveBirdsall: I'm wondering if we need to delete jsonStr and jsonAttr after the json_extract_path_text call to avoid unnecessary heap pressure. Though if json_extract_path_text itself does new's on the same heap, we'd get heap fragmentation.

Another approach would be to allocate these on the stack instead, avoiding both concerns:

char jsonStr[len1+1];
char jsonAttr[len2+1];
...

andyyangcn: Since the value of len1 can only be determined when executing the query, we cannot define jsonStr as an Array.
Will release the memory manually.

DaveBirdsall: But we can! C++ allows dynamically-sized arrays on the stack. Try it!

DaveBirdsall: Here's an example. Compile this program:

// Demonstrate dynamically-sized arrays on C++ stack

#include <iostream>

using namespace std;

int main(int argc, char * argv[])
{
for (size_t i = 5; i < 10; i++)
  {
    char temp[i+1];
    cout << "sizeof(temp) = " << sizeof(temp) << endl;
  }

return 0;
}

When you run it, you'll get this output:

[birdsall@edev08 dynChar]$ ./dynChar.exe
sizeof(temp) = 6
sizeof(temp) = 7
sizeof(temp) = 8
sizeof(temp) = 9
sizeof(temp) = 10
[birdsall@edev08 dynChar]$

PR #1441, opened 2018-02-05
[TRAFODION-2908] create table with wrong char length hang and crash

traflm:
add a check if the key length is too long report error. Otherwise, SQL will go into a very long process to build the encoded key and finally failed. And in some cases crash, it makes little sense to check why it crash, so just add check and return error when key is too long.
HBase max rowkey len is 32K, so it is the hard limit for now.


core/sql/bin/SqlciErrors.txt
```diff
@@ -140,7 +140,7 @@
 1138 ZZZZZ 99999 ADVANCED CRTCL DIALOUT --- unused ---
 1139 ZZZZZ 99999 BEGINNER MAJOR DBADMIN System-generated column $0~ColumnName of base table $1~TableName cannot appear in the search condition of a check constraint definition.
 1140 ZZZZZ 99999 BEGINNER MAJOR DBADMIN Row-length $0~int0 exceeds the maximum allowed row-length of $1~int1 for table $2~TableName.
-1141 ZZZZZ 99999 BEGINNER MAJOR DBADMIN --- unused ---
+1141 ZZZZZ 99999 BEGINNER MAJOR DBADMIN Key-length $0~int0 exceeds the maximum allowed rowkey length $1~int1.
```
DaveBirdsall: Consider rewording: "Key length $0-int0 exceeds the maximum allowed key length of $1~int1".

DaveBirdsall: Also, would you please add the new message to the Messages Guide? Thanks.

traflm: Thanks Dave for the review, I will update teh Message Guide and modify the workding

traflm:
Thanks Dave for the review.

PR #1442, opened 2018-02-06
[TRAFODION-2955] odb crash when executing sql statements in a file and the file contains space only line. 

SuJinpei:
**root cause:** original odb didn't consider blank line.
**fixed by:** deal with the blank line.
**verifying:** run a regression test and found this bug had been fixed and no new issues were introduced.


DaveBirdsall:
jenkins, retest

PR #1444, opened 2018-02-15
[TRAFODION-2954] add MySQL function unix_timestamp,uuid,sleep

traflm:
This is to add three MySQL compatible functions into Trafodion. 
It is required for another JIRA to add unix_timestamp, uuid as default value of a column definition.
sleep is simple to add, so add with this change as well.
The document will be updated for a separate JIRA asap.


traflm:
jenkins, retest

core/sql/exp/exp_function.cpp
```diff
@@ -2501,6 +2518,100 @@ ex_expr::exp_return_type ExFunctionReverseStr::eval(char *op_data[],
...
+						   CollHeap* heap,
+						   ComDiagsArea** diagsArea)
+{
+  char *opData = op_data[1];
+  //if there is input value
+  if(opData[0] != 0 &&  getNumOperands() == 2)
+  {
+    struct tm* ptr;
+    char* r = strptime(opData, "%Y-%m-%d %H:%M:%S", ptr);
+    if( r == NULL)
```
DaveBirdsall: Should we also test for *r pointing to a null character? See the man page for strptime. If there is garbage data after the seconds field, r will point to the first byte of the garbage. I'm guessing we want to raise an error in that case.  Consider a test of the form "if ((r == NULL) || (*r != '\0'))"

traflm: Thanks Dave, you are correct. We want to be very strict here. I will add this checking.

zellerh:
Could you please add some queries that use these functions to some regression test? Don't sleep too long, though :-) You may need to add a filter to mask out the random digits in the UUIDs.

Also, you probably found this out already, Trafodion does not give any guarantees over where it places these functions. Let's say you have a table t with 1000 rows and you issue this query:

select uuid(), sleep(1) from t;

There is no guarantee that you get 1000 different uuids back or 1000 clones of the same one. Also, the query may sleep some amount other than 1 or 1000 units. The outcome depends on the query and maybe other factors such as CQDs. This makes these functions not as useful as they seem and that's why the designers of the SQL language have stayed away from non-deterministic functions and functions with side-effects.

If you create scalar UDFs for such functions you have some limited control via the "NOT DETERMINISTIC" attribute, as you probably have already seen in TRAFODION-2912. That is the better approach, but it comes with a performance penalty.

Here is a use case that many people may want to have, so it would be good to allow it:

create table t(id char(36) not null default uuid());

Again, it is not easy to guarantee that every row inserted into the table will get its own uuid. So, at second thought, don't allow this unless you ensure that you get a unique column this way.

zellerh:
Another example:
```
select * from t1 join t2 on t1.ts = t2.ts
where t1.ts between unix_timestamp() - 1000 and unix_timestamp() - 800;
```
Queries like this may give you complete nonsense back. This is because the Trafodion compiler assumes all ItemExprs are deterministic.

As long as you are willing to live with these problems that's probably ok. Many times you will get what you want. When you update the SQL Reference Manual you should make it very clear that these functions have undefined behavior.

traflm:
Thanks Hans, yes, the uuid() and sleep() may have issues as you described. The major usage of uuid is to be used as the default function for column definition. But you are right, it could be used in a select list or anywhere a function is allowed and the output is deterministic. Let me add these in the documentation and add some regression test. The reason I didn't add tests is because they are random, but as you suggested, the output can be filtered out.
For unix_timestamp(), the purpose is to have them using same value id in a single query, and each occurrence of them should invoke once and represent identical value. But I didn't strictly test it. Let me add test case, and do more check on this.
These are very excellent and knowledgeable comments. I never thought in this way. And it is very good for me to learn the design reasons behind all these. :-) 
Let me try again, and since they are implemented in MySQL, Let me try to make it as much same as in MySQL and document it clearly.

traflm:
hi, Hans,

I just tested , and it seems uuid() and sleep() together works as expected for a simple 4 rows table.
But let me test it more:

>>>>select uuid(), sleep(1)  from nullfirsttest;

(EXPR)                                (EXPR)
------------------------------------  -----------

99254720-6fb5-4b92-8267-bdb9f8850891            1
1715766e-7d90-4399-9a9d-9741be601ca1            1
28385e62-eeae-4ae5-8566-c5cdf28bf5b4            1
a97231b8-ad20-4cf1-8333-63c8a047bd30            1

--- 4 row(s) selected.
>>


Thanks,
Ming

traflm:
one more test shows that unix_timestamp() stays same, and uuid varies for each row. The testing table has 1000 rows. Here is sample output:
>>select 'uuidrow', uuid(), unix_timestamp() from T002T1K;

(EXPR)   (EXPR)                                (EXPR)
-------  ------------------------------------  --------------------

uuidrow  cd94fff8-98e8-4ce3-97b7-628476e2bcc3            1518771963
uuidrow  6fbd22b9-d110-4dcc-952b-850edef2c812            1518771963
uuidrow  789442dd-001f-480f-b1f9-0db9690c8c4e            1518771963
uuidrow  f5fbfd59-3b0b-4ede-8a45-56f268b117da            1518771963
uuidrow  5629dbcb-5c68-4f0c-94ef-1b154a26ee2c            1518771963
uuidrow  15f7e2c8-a468-43e6-b6e2-a07b85dae9bb            1518771963
uuidrow  0c01bd73-7ac4-4dc2-82d2-7fc36b5f2d16            1518771963
uuidrow  394cdd66-15e4-4ffb-af24-4dce5fc94dc1            1518771963
uuidrow  bb09ccb7-3dfe-4381-b26b-f66e1dd38496            1518771963
uuidrow  d870364c-1fc9-4915-b114-b1a92f34944f            1518771963
uuidrow  18a15daf-c037-4e61-a80d-392c72cfbd54            1518771963
uuidrow  31fec3a3-34c5-4b24-93b8-2f0129a49334            1518771963
uuidrow  98eaa09a-e0ae-4a80-bba2-9932f801b06d            1518771963
uuidrow  5557e5e3-bf77-4479-82c8-ef08eb02ef0a            1518771963
uuidrow  87a28386-d2a2-4848-8c69-56d1b8297ec8            1518771963
uuidrow  42448c13-297b-4302-9991-bbb6c7d2c2de            1518771963
uuidrow  1d502632-b177-4f8f-8ce9-b66309f895e4            1518771963

zellerh:
Yes, as I expected, the results will vary.

traflm:
Hi, Hans,

You are right!
Let me change the uuid() implementation and make sure it will be same in a single query. And do a new commit.

And there is another requirement for random() function, which is by nature need to vary for each occurrence, but we should call it random(), not uuid(). So let me fix uuid.

Thanks,
Ming

traflm:
Hi, @zellerh , I have updated the uuid code to make it evaluate once per query. So the 1000 rows will have same uuid() if in a single query , and also:

select uuid(), uuid() from dual;
(EXPR)                                (EXPR)
------------------------------------  ------------------------------------
760a6cc2-f3da-44b0-b21a-a8e9a975e11e  760a6cc2-f3da-44b0-b21a-a8e9a975e11e
--- 1 row(s) selected.

Ming

traflm:
Maybe I still not correctly understand the issue here. But here is what I want to make sure:
Not implemented yet for default clause, but I wish it:
create table tt1 (c1 int, c2 timestamp default current, c3 varchar(100) default uuid() );

select * from NULLFIRSTTEST;
C1
-----------
          1
          2
          ?
          ?


upsert into tt1(c1) select c1 from NULLFIRSTTEST;

select * from tt1;
C1           C2      C3
-----------  -------------------------- --------------------------
          1  2018-02-16 23:06:58.732655  1d502632-b177-4f8f-8ce9-b66309f895e4
          2  2018-02-16 23:06:58.732655  1d502632-b177-4f8f-8ce9-b66309f895e4
          ?  2018-02-16 23:06:58.732655  1d502632-b177-4f8f-8ce9-b66309f895e4
          ?  2018-02-16 23:06:58.732655  1d502632-b177-4f8f-8ce9-b66309f895e4


Current code should already support this behavior when uuid is allowed for default clause.

Ming

core/sql/regress/executor/TEST002
```diff
@@ -1188,6 +1188,32 @@ select * from regexp_test where c1 regexp '^(25[0-5]|2[0-4][0-9]|[0-1]{1}[0-9]{2
...
+0,
+X01
+from (values(0)) t
+transpose 0,1,2,3,4,5,6,7,8,9 as x10
+transpose 0,1,2,3,4,5,6,7,8,9 as x1
+transpose 0,1,2,3,4,5,6,7,8,9 as X01;
+
+select sleep(5) from dual;
+select 'unixtimestamp',unix_timestamp() from dual;
+select 'uuidrow', uuid(), unix_timestamp() from T002T1K;
```
zellerh: If you want to ensure that we see different uuids and not the same one repeated, you could add a query select count(distinct uuid()) from T002T1K.

core/sql/optimizer/GroupAttr.cpp
```diff
@@ -1793,6 +1793,8 @@ void GroupAttributes::resolveCharacteristicInputs(const ValueIdSet& externalInpu
 	     ItemExpr * vidExpr = vid.getItemExpr();
 	     if ((vidExpr->getOperatorType() == ITM_CURRENT_USER) ||
 		 (vidExpr->getOperatorType() == ITM_CURRENT_TIMESTAMP) ||
+		 (vidExpr->getOperatorType() == ITM_UNIQUE_SHORT_ID) ||
+		 (vidExpr->getOperatorType() == ITM_UNIQUE_ID) ||
```
zellerh: This comment applies to all the changes in this commit: Don't these code changes do the opposite of what you want to achieve? Functions like CURRENT_TIMESTAMP and CURRENT_USER are carefully designed such that we execute them only once per query, in the master. You want to make sure your functions are evaluated once per row. Did you find that these changes helped fix the issues you were seeing?

zellerh:
Hi @traflm ,

About your question: No, I'm not worried that different calls to uuid() will return the same value (that might not happen in my lifetime :-). My main concern is that methods like RelExpr::pushdownCoveredExpr() (the main method that decides where to evaluate an expression) may decide to evaluate the uuid() function on the parent rather than push it down. The uuid() function requires no inputs, so it can be evaluated anywhere, and the normalizer is completely free to choose a location for the evaluation. This is all done based on heuristics, and the code gets very complicated, especially when VEGies are involved.

Another problem comes from how the SQL standard is defined. SQL is a descriptive language and leaves the optimizer lots of choices. Let's say I have a table T with n rows:

```
select *
from T x, T y
where x.a = y.a + sleep(1)
```

The standard says that "effectively", we form the Cartesian product of the tables (T and T, in this case) and evaluate the where clause on it. If you take this literally, we would have to sleep n**2 seconds for this query. However, we generally assume ItemExprs like + and = and also sleep() to be deterministic and without side-effects, so we use something like a hash join that evaluates the where clause much less often than n**2 times. But, a function with side-effects like sleep() violates the definition of "effectively". This is just a very simple example.

Here is another:

```
with x as (select uuid() from T)
select count(distinct *)
from (x union all x) x2
```

If table T has n rows, what's the answer to this query, n or 2*n? As you probably know, with Trafodion you probably get 2*n by default, but set some CQDs and suddenly you may get n.

These are somewhat esoteric examples, but if you get into more complex queries like TPC-DS, you may run into similar issues quite frequently.

Again, I'm fine with these functions as long as we make it very clear that they are not well-defined and should not be used beyond very simple queries.

What we would need to do this right are "not deterministic" and "external action" attributes of functions. We have "not deterministic" for UDFs, but we don't honor this attribute in every case.

DaveBirdsall:
To @zellerh's comments: Thinking about sleep(), I wonder if maybe the desired semantics are to execute it just once per query? If so, treating it like current_timestamp is appropriate. Evaluate it in the master, at the beginning of execution.

traflm:
Yes, Dave, I think sleep() once per query is what I should change. 
Thanks Hans for the very excellent explanation and details! And I will revert the uuid() back to evaluate multiple times per query and make it clear in the documents, these functions should only be used in very few cases with careful notice. Let me describe these in the document. And I will update the regress test to make it make more sense ..

core/sql/optimizer/GroupAttr.cpp
traflm: You are right, Hans, I also think in the opposite, I just simply follow the current_timestamp behavior, and this really is not what I originally want. So I will change uuid() to be evaluated once per row.

traflm:
Hi, Hans,
I tested on MySQL 5.7 UUID() is evaluated once per row(), And in Oracle sys_guid() is evaluated once per query. Since this is non-ANSI functions, we have to choose the behavior by ourselves.
Since this JIRA is to simulate the MySQL, so I would like to evaluate uuid() once per row, it makes more sense if the original idea is to use it as default value of a column, and uniqueness will be a desired result.
But I will make it clear in the documentation that the usage of uuid() is 'non-deterministic' and should not appear in JOIN, UNION ect. As you mentioned, the non-deterministic flag is now only for UDR, so I will leave the functions as it is for now, and we can change it when Trafodion has evolved to check non-deterministic flag everywhere , and I will have a new JIRA for the documentation of these.

Thanks,
Ming

traflm:
hi, @zellerh , @DaveBirdsall 
I have updated the code again.
This time: 
make the sleep() as input value, and can only be evaluated once. Make sleep() only accept literal input. 
update test002, make sure sleep() only eval once per query, not per row.
update test002, make sure uuid will be distinct and match the row number, so make sure it is eval once per row and no conflict.

create JIRA for later documentation: https://issues.apache.org/jira/browse/TRAFODION-2968
make sure these functions can only be used in very simple queries as they designed for.

Please help to review again. Thanks.

core/sql/regress/executor/TEST002
```diff
@@ -1188,6 +1188,38 @@ select * from regexp_test where c1 regexp '^(25[0-5]|2[0-4][0-9]|[0-1]{1}[0-9]{2
...
+transpose 0,1,2,3,4,5,6,7,8,9 as x1
+transpose 0,1,2,3,4,5,6,7,8,9 as X01;
+
+create table t002timert (c0 int, c1 int, c2 largeint);
+create table t002tmp1 (c1 int);
+insert into t002tmp1 values(1),(2),(3);
+
+insert into t002timert select 1, sleep(5) , unix_timestamp() from t002tmp1;
+insert into t002timert select 2, sleep(5) , unix_timestamp() from t002tmp1;
+select 'sleeptimetest002', di from (select ( max(c2) - min(c2)) as di from t002timert ) where di between 5 and 9;
```
traflm: above two INSERTS will insert two different unix_timestamp() , the difference is the second between their run, so if the sleep(5) eval once, the diff should be around 5 seconds, and less than 10. So the query is using 'between 5 and 9'. If something wrong, this test query will return 0 rows and fail the test.

zellerh:
+1
Looks good to me. Thanks for your patience and for all your work!

traflm:
Thanks @zellerh @DaveBirdsall , I learned a lot in this review process.

PR #1447, opened 2018-02-16
[TRAFODION-2727] Memory leak in the compiler part of the code in Traf…

selvaganesang:
…odion

It was found that CmpMessageReplyCode was not getting deleted at all times.
But, the handling of CmpMessageDescribe messages in the compiler code has some
mysterious issue that corrupts the memory causing unpredictable behavior when
CmpMessageReplyCode is deleted.  To circumvent this problem, the CmpStatement heap
is used for embedded compiler to allocate CmpMessageReplyCode and the Context heap
is used in case of standalone compiler. The handling of CmpMessageDescribe will
continue to leak memory in standalone compiler.


traflm:
+1 looks good for me as well.

PR #1449, opened 2018-02-16
RMS infrastructure now reads the configured pid Max from /proc/sys/ke…

selvaganesang:
…rnel/pid_max

but limits it to a maximum of 128K. It can be overridden by "PID_MAX" environment
variable in ms.env.

If the PID_MAX variable is set to 0, 64K is assumed to be configured pid max.
If the PID_MAX variable is set to less than 32K, then 32K is assumed to be configured pid max.

When a SQL process has a pin greater than the configured pid max, an error message
"Pid , is higher than the configured pid max <configured_pid>" is written to
the SQL log file and disables RMS for those processes. Node won't be brought down.

"Get Statistics for rms " shows the the configured pid max.

The environment variable to set the size of RMS shared segment size is now
changed to RMS_SHARED_SEG_SIZE_MB.

genms change for controlling pid_max and rms shared segment size


core/sql/cli/Globals.cpp
```diff
@@ -176,10 +176,21 @@ void CliGlobals::init( NABoolean espProcess,
...
+    NABoolean reportError = FALSE;
+    char msg[256];;
+    if (statsGlobals_ != NULL && myPin_ >= statsGlobals_->getConfiguredPidMax())
+       reportError = TRUE;
+    if ((statsGlobals_ == NULL)
+      || ((statsGlobals_ != NULL) && 
+        ((statsGlobals_->getVersion() != StatsGlobals::CURRENT_SHARED_OBJECTS_VERSION_) ||
+        (myPin_ >= statsGlobals_->getConfiguredPidMax()))))
     {
+      if (reportError) {
```
DaveBirdsall: Could this lead to a log storm? Imagine that thousands of SQL processes with Pids that are too high come and go. (If there is one, there is likely to be many.) Could we blow up our logs?

core/sql/executor/ex_esp_frag_dir.cpp
```diff
@@ -101,11 +101,23 @@ ExEspFragInstanceDir::ExEspFragInstanceDir(CliGlobals *cliGlobals,
...
+  NABoolean reportError = FALSE;
+  char msg[256];;
+  if (statsGlobals_ != NULL && pid_ >= statsGlobals_->getConfiguredPidMax())
+     reportError = TRUE;
+  if ((statsGlobals_ == NULL)
+     || ((statsGlobals_ != NULL) && 
+        ((statsGlobals_->getVersion() != StatsGlobals::CURRENT_SHARED_OBJECTS_VERSION_) ||
+        (pid_ >= statsGlobals_->getConfiguredPidMax()))))
   {
+    if (reportError) {
```
DaveBirdsall: Same comment as earlier.

selvaganesang:
The idea of limiting PID_MAX in RMS infrastructure serves two purposes
a) To limit the pre-allocated pid array size
b)  To make the installer aware that the unusually large pid_max is configured on the system

 Trafodion is made to tolerate the unusually large number of pd_max configuration without bringing the node down. However, these processes don't participate in RMS making devoid of the benefits RMS brings to them like detecting what queries are run from these process, cancel the query etc. , Hence, it  is recommended that PID_MAX configuration in RMS is in sync with the /proc/sys/kernel/pid_max.  It is expected the /proc/sys/kernel/pid_max is changed to reasonable number.   

So, I don't expect the log storm to happen.

zellerh:
Ideally we should not see cases where RMS and /proc/sys/kernel/pid_max differ, but in practice I would assume that this does happen occasionally. If we create thousands of ESPs, won't that be a problem, as Dave pointed out? I like Dave's solution of putting some flag into the shared segment, to avoid too many log messages. You could limit it to something like 100, so people examining the log would still notice the first 100 as an unusual event. I assume that "get statistics for rms" will report a problem when RMS is not configured correctly for the pid_max on at least one of the nodes? This could be another reason for marking the shared segment.

selvaganesang:
In Esgyn's internal testing, it is felt that this restriction of limiting PID_MAX in RMS infrastructure needs to be removed entirely. I will be working to remove this restriction soon.

zellerh:
Great, that's even better! My gut feeling is the same. People aren't changing PID_MAX for fun, and when they (have to) do it, they do not expect EsgynDB to have big issues with such a change. Sorry you have to go from a simple array to something more complicated.

selvaganesang:
However, I believe configuring a node with more than 128K active processes is a stretch by itself even with larger memory configuration.  It could be because some application is  running amok producing many threads or process or the application is having problems with pid recycling.  I added this comment just to record by opinion, but I will work to remove this restriction soon.

DaveBirdsall:
Is there a JIRA for this change? Could you please add it? Thanks.

selvaganesang:
Yes. Forgot to add the JIRA in the title. Here is the JIRA
https://issues.apache.org/jira/browse/TRAFODION-2963

selvaganesang:
Added code to avoid log storm instead of allowing RMS infrastructure to relax the restriction on the maximum pid number allowed. When time permits, I will use the different collection mechanism instead of simple array at a later date.

selvaganesang:
@zellerh Can you please review 0f04ef9784

PR #1451, opened 2018-02-21
[TRAFODION-2965] Fix row count stats for partial groupby

zellerh:
Hash partial groupbys now report their row count in operator-level
statistics. These are not considered BMOs, so they need to use the
generic stats entry, not the BMO stats, to report the row count.

### DIFF
--- a/core/sql/executor/ex_hash_grby.cpp
+++ b/core/sql/executor/ex_hash_grby.cpp
@@ -1027,8 +1027,10 @@ void ex_hash_grby_tcb::returnResultCurrentRow(HashRow * dataPointer)
   upParentEntry->upState.downIndex = parentQueue_.down->getHeadIndex();
   
   // if stats are to be collected, collect them.
-  if (bmoStats_)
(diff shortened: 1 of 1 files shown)


core/sql/executor/ex_hash_grby.cpp
```diff
@@ -1027,8 +1027,10 @@ void ex_hash_grby_tcb::returnResultCurrentRow(HashRow * dataPointer)
   upParentEntry->upState.downIndex = parentQueue_.down->getHeadIndex();
   
   // if stats are to be collected, collect them.
-  if (bmoStats_)
-    bmoStats_->incActualRowsReturned();
+  ExOperStats *statsEntry = getStatsEntry();
```
DaveBirdsall: Is the member bmoStats_ no longer used then? Should it be deleted?

zellerh: bmoStats_ is also used for reporting memory-related counters, so we still need it.

PR #1452, opened 2018-02-21
[TRAFODION-2967] Fix CommonLogger::buildMsgBuffer core in UPDATE STATS

DaveBirdsall:
Instead of passing raw data as a message template to the logging layers, pass "%s", and then pass the raw data as a parameter. That prevents misinterpretation of substrings that look like printf format strings, avoiding cores.


zellerh:
+1
Looks good to me.

PR #1454, opened 2018-02-27
[TRAFODION-2969] Fix interaction of [first n] etc. with subqueries

DaveBirdsall:
Formerly, a [first n] in a subquery in a SELECT list resulted in an internal error in the Normalizer. The problem was that the Normalizer method Subquery::transformToRelExpr was not expecting a FirstN node at the top of the subquery tree.

Besides the fact that the Normalizer doesn't like it, having FirstN above the GroupByAgg node that enforces the one-row cardinality rule doesn't make sense.

In this set of changes, the FirstN node is placed underneath the GroupByAgg node when it matters. For [first n] / [any n] where n > 1, we can simply throw the [first n] / [any n] away, since the subquery already enforces at most one row semantics. For [first 1] and [last 1], we need to insert the FirstN node below the GroupByAgg node so that we narrow the result set down to one (or zero) rows before GroupByAgg enforces the cardinality rule. So, adding [first 1] to a subquery avoids cardinality violations. For [last 0] there is a subtlety. [last 0] by definition returns no rows, so we want the subquery to return null. The way we chose to do this is to force the creation of a one-row GroupByAgg node always, and put the FirstN node under that. In this way, GroupByAgg will see zero rows and produce a null result.

Regression test core/TEST002 has had many [first n] - subquery interaction tests added.

### DIFF
--- a/core/sql/optimizer/RelExpr.cpp
+++ b/core/sql/optimizer/RelExpr.cpp
@@ -11390,7 +11390,7 @@ void RelRoot::setMvBindContext(MvBindContext * pMvBindContext)
 
 NABoolean RelRoot::addOneRowAggregates(BindWA* bindWA, NABoolean forceGroupByAgg)
 {
-  NABoolean GroupByAggNodeAdded = FALSE;
+  NABoolean groupByAggNodeAdded = FALSE;
   RelExpr * childOfRoot = child(0);
   GroupByAgg *aggNode = NULL;
   // If the One Row Subquery is already enforced by a scalar aggregate
@@ -11426,7 +11426,7 @@ NABoolean RelRoot::addOneRowAggregates(BindWA* bindWA, NABoolean forceGroupByAgg
 
     }
   if (aggNode)
-    return GroupByAggNodeAdded;
+    return groupByAggNodeAdded;
 
   const RETDesc *oldTable = getRETDesc();
   RETDesc *resultTable = new(bindWA->wHeap()) RETDesc(bindWA);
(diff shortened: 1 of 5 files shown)


core/sql/optimizer/RelExpr.cpp
```diff
@@ -11388,8 +11388,9 @@ void RelRoot::setMvBindContext(MvBindContext * pMvBindContext)
   pMvBindContextForScope_ = pMvBindContext;
 }
 
-void RelRoot::addOneRowAggregates(BindWA* bindWA)
+NABoolean RelRoot::addOneRowAggregates(BindWA* bindWA, NABoolean forceGroupByAgg)
 {
+  NABoolean GroupByAggNodeAdded = FALSE;
```
zellerh: A small nit: This variable should start with a lower case letter.

core/sql/regress/core/EXPECTED002.LINUX
```diff
@@ -464,22 +478,16 @@ A            AMAX         B                   BMAX                C            C
...
-+>  );
-
-(EXPR)
-------
-
-     1
-     1
-
---- 2 row(s) selected.
+>>-- This test disabled since it is non -deterministic.
```
zellerh: Unless you plan to work on this right away it might be good to file a JIRA to keep track of this.

DaveBirdsall: Sure, I can do that.

DaveBirdsall: JIRA filed. See https://issues.apache.org/jira/browse/TRAFODION-2972

core/sql/optimizer/RelExpr.cpp
DaveBirdsall: Done; changed it to start with lower case letter

zellerh:
+1

PR #1455, opened 2018-02-27
[TRAFODION-2971] Add examples for *CREATE EXTERNAL TABLE Statement* in *Trafodion SQL Reference Manual*

liuyu000:


docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
```diff
@@ -3499,6 +3499,152 @@ CREATE TABLE t1 (c1 int not null primary key, c2 char(50));
...
+  , CA_SUITE_NUMBER                  string
+  , CA_CITY                          string
+  , CA_COUNTY                        string
+  , CA_STATE                         string
+  , CA_ZIP                           string
+  , CA_COUNTRY                       string
+  , CA_GMT_OFFSET                    float
+  , CA_LOCATION_TYPE                 string
+  )
+  stored as orc
```
DaveBirdsall: My understanding is that the support in Trafodion for Hive ORC files is incomplete and at best experimental at this point. I'm not sure we should use ORC in our examples until the feature is product quality.

docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
```diff
@@ -3499,6 +3499,152 @@ CREATE TABLE t1 (c1 int not null primary key, c2 char(50));
...
+--- 0 row(s) selected. 
+
+Start Time             2018/01/31 02:21:55.409105
+End Time               2018/01/31 02:22:09.299215
+Elapsed Time                      00:00:13.890110
+Compile Time                      00:00:00.264101
+Execution Time                    00:00:13.625538
+```
+
+** This is the definition of the table _customer_address2_, it has same structure and size as the table _customer_address1_, but it does not have an external table.
```
DaveBirdsall: It might be good to talk about why performance is better when there is an external table. I can think of two reasons: 1. Trafodion statistics exist for the table. (Which seems to be the case in this example.) And 2. The Trafodion definition supplies upper bounds for varchar lengths, which may lead to better plans and/or run-time behavior. It is possible to have 2. without having 1. (As one can issue a CREATE EXTERNAL TABLE statement without also doing UPDATE STATISTICS.) So this is a key point: To get full performance benefit, one must also run UPDATE STATISTICS.

docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
liuyu000: Thanks for your explainations! 
I’ll close this pull request and add another example (not ORC) showing that the performance of trafodion external table is better. :)

PR #1456, opened 2018-02-27
[TRAFODION-2966] Fix new MDAM costing code to respect CQS forces

DaveBirdsall:
The new MDAM costing code was not respecting CONTROL QUERY SHAPE directives to force MDAM. This has been fixed.


zellerh:
+1
Looks good to me (looking forward to the optimizer listening to my CQS instead of ignoring me :-)

DaveBirdsall:
@zellerh, the old MDAM costing code (which is on by default) does respect the CQS to my knowlege. Have you found a bug in that area?

PR #1460, opened 2018-03-01
[TRAFODION-2974] Make event log reader and JDBC real TMUDFs

zellerh:
Roberta pointed out that we have two predefined UDFs, EVENT_LOG_READER
and JDBC, where the system administrator should have the ability to
control who can execute these functions.

To do this, these two UDFs cannot be "predefined" UDFs anymore, since
those don't have the metadata that's required for doing grant and
revoke.

Roberta also pointed out that the JDBC UDF should refuse to connect to
the T2 driver, for security reasons.

The fix leaves the predefined TMUDFs in place, for now, they will be
removed in R2.4 (see TRAFODION-2975).

The new "real" TMUDFs are in the "_LIBMGR_" schema, mostly for
convenience, as this schema has other UDFs that are created when
Trafodion is initialized.

Note for reviewers: The new JDBCUDR.java file is the same as the existing file, except for the package name and for the note added to the old file.

### DIFF
--- a/core/sql/bin/SqlciErrors.txt
+++ b/core/sql/bin/SqlciErrors.txt
@@ -1331,6 +1331,7 @@ $1~String1 --------------------------------
 4320 ZZZZZ 99999 BEGINNER MAJOR DBADMIN Stream access is not allowed on multi-partitioned table or index, when flag ATTEMPT_ASYNCHRONOUS_ACCESS is set to OFF. Object in scope: $0~TableName.
 4321 ZZZZZ 99999 BEGINNER MAJOR DBADMIN An embedded update/delete is not allowed on a partitioned table, when flag ATTEMPT_ASYNCHRONOUS_ACCESS is set to OFF. Object in scope: $0~TableName.
 4322 0A000 99999 BEGINNER MAJOR DBADMIN A column with BLOB datatype cannot be used in this clause or function.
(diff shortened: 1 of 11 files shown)


core/sql/bin/SqlciErrors.txt
```diff
@@ -1331,6 +1331,7 @@ $1~String1 --------------------------------
 4320 ZZZZZ 99999 BEGINNER MAJOR DBADMIN Stream access is not allowed on multi-partitioned table or index, when flag ATTEMPT_ASYNCHRONOUS_ACCESS is set to OFF. Object in scope: $0~TableName.
 4321 ZZZZZ 99999 BEGINNER MAJOR DBADMIN An embedded update/delete is not allowed on a partitioned table, when flag ATTEMPT_ASYNCHRONOUS_ACCESS is set to OFF. Object in scope: $0~TableName.
 4322 0A000 99999 BEGINNER MAJOR DBADMIN A column with BLOB datatype cannot be used in this clause or function.
+4323 ZZZZZ 99999 BEGINNER MAJOR DBADMIN Use of predefined UDF $0~String0 is deprecated and this function will be removed in a future release. Please use the function with the same name in schema TRAFODION."_LIBMGR_" instead. You may need to issue this command first: INITIALIZE TRAFODION, UPGRADE LIBRARY MANAGEMENT.
```
DaveBirdsall: Consider adding this message to the Trafodion Messages Guide.

DaveBirdsall:
Looks good to me. Just one comment.

robertamarton:
Thanks Hans for making this change.  Change looks good.

PR #1461, opened 2018-03-01
 [TRA[TRAFODION-2917] Refactor Trafodion implementation of hdfs scan for text formatted hive tables

selvaganesang:
Removed the dependency on libhdfs in Trafodion code in some more portions of the code.
    
    Also, introduced HDFS Scan to use the java layer to read HDFS data providing the following features:
    a) Prefetch the data using double buffering concept
    b) Avoid unnecessary data copy
    c) Ensure that the data read is initiated to be read into the other buffer in java layer, while it is being processed in one buffer in the native side, without any hiccups.
    
    This HDFS Scan is still being tested and hence it is turned off by default.
    
    Also changed the code to obtain millisecond-resolution modification timestamp for HDFS files


zellerh:
+1
Looks good to me. Nice to see the move from libhdfs to JNI!

PR #1464, opened 2018-03-07
TRAFODION-2981 mxosrvr crashes during explain using EXPLAIN_QID

mashengchen:
SELECT * FROM TABLE(explain(null, 'EXPLAIN_QID=MXID11000007019212384153894533416000000002206U3333308T150010100_923_SQL_CUR_2'));
this is related to TRAFODION-1755

### DIFF
--- a/core/sql/executor/ExExplain.cpp
+++ b/core/sql/executor/ExExplain.cpp
@@ -1724,9 +1724,9 @@ short ExExplainTcb::getExplainFromRepos(char * qid, Lng32 qidLen)
   if (vi->get(0, ptr, len))
     goto label_error2;
   
-  explainFragLen_ = str_decoded_len(len); // remove trailing null terminator
+  explainFragLen_ = str_decoded_len(len - 1); // remove trailing null terminator
   explainFrag_ = new(getHeap()) char[explainFragLen_];
-  if (str_decode(explainFrag_, explainFragLen_, ptr, len) < 0)
+  if (str_decode(explainFrag_, explainFragLen_, ptr, len - 1) < 0)
     {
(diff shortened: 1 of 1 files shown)


mashengchen:
@anoopsharma00 can you help to review this?

core/sql/executor/ExExplain.cpp
```diff
@@ -1724,9 +1724,9 @@ short ExExplainTcb::getExplainFromRepos(char * qid, Lng32 qidLen)
   if (vi->get(0, ptr, len))
     goto label_error2;
   
-  explainFragLen_ = str_decoded_len(len); // remove trailing null terminator
+  explainFragLen_ = str_decoded_len(len - 1); // remove trailing null terminator
```
sandhyasun: The JIRA isn't showing  trace fileor debugging info. Can you explain what the issue was with passingin 'len' into str_decoded_len ? Seems like the explainFragLen_ would have had sufficient length to allocate the explain fragment ? Some explanation in the PR or JIRA will help.

mashengchen: for the '-1' , i think it should remove trailinng null terminator as the comment.
but last time in anoop 's commit, the '-1' was removed, but the method (getExplainFromRepos) is for EXPLAIN_QID,  not for EXPLAIN_STMT (see TRAFODION-1755), so i guess he changed the place there is no need to change.

DaveBirdsall: @sandhyasun, are you content with this change now? Can I merge it?

sandhyasun:
+1

PR #1469, opened 2018-03-12
[TRAFODION-2990] Fix CREATE TABLE LIKE having long numeric default constant

DaveBirdsall:
The routine LineBreakSqlText (common/NAString.cpp) was inserting a line break after the decimal point in a numeric constant, making one token into two. The code has been changed to distinguish between a dot that separates two identifiers (a legitimate place to insert line breaks) and decimal points.

### DIFF
--- a/core/sql/common/NAString.cpp
+++ b/core/sql/common/NAString.cpp
@@ -1815,7 +1815,8 @@ size_t LineBreakSqlText(NAString &sqlText,
 
 	  } // unquoted space
 
-	else if (*s == '.')
+	else if ((*s == '.') &&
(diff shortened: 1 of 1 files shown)


DaveBirdsall:
I am also running full regressions on a workstation. This change could cause changes in the output formatting of INVOKE and SHOWDDL, which might require updating expected results. Please do not merge until I report back on the regressions run.

zellerh:
+1
Looks good to me.

DaveBirdsall:
Full regressions passed. So no changes to expected files are needed.

PR #1470, opened 2018-03-12
[TRAFODION-2853] memory leak of ComDiagsArea in CmpContext heap of mxosrvr

selvaganesang:
The ComDiagsArea is allocated in many places and from different heaps in Trafodion, making it difficult to  detect the source of the leak. Hence, a different approach is taken to fix this issue.
    
Currently, ComDiagsArea is allocated in many places unconditionally even when SQL statement completes execution without any error or warnings. Then it is deallocated. Changed this strategy and allocate ComDiagsArea only when there is an error or warning while compiling or executing the SQL statement.
    
This should help the product in two ways
    1) To reduce the path length.  The smaller query execution would benefit the most by chopping of few more microseconds.
    2) Reduce the memory growth due to leaked ComDiagsArea

### DIFF
--- a/core/sql/executor/ex_root.cpp
+++ b/core/sql/executor/ex_root.cpp
@@ -2235,6 +2235,7 @@ Int32 ex_root_tcb::cancel(ExExeStmtGlobals * glob, ComDiagsArea *&diagsArea,
             }
           else 
             {
+/*
               // redrive the scheduler.
               // Fix for CR 6701 - some ExExeUtil operators call back
               // in to the CLI and explicitly clear the curr context
@@ -2246,13 +2247,15 @@ Int32 ex_root_tcb::cancel(ExExeStmtGlobals * glob, ComDiagsArea *&diagsArea,
               ContextCli *context = statement->getContext();
               ComDiagsArea *savedContextDiags = context->diags().copy();
               context->diags().clear();
+*/
 
               schedRetcode = glob->getScheduler()->work();
-
+/*
               savedContextDiags->mergeAfter(context->diags());
               context->diags().clear();
               context->diags().mergeAfter(*savedContextDiags);
               savedContextDiags->decrRefCount();
+*/
             }
         }
       if (!getQueueDiags)
@@ -2641,6 +2644,7 @@ void ex_root_tcb::registerCB(ComDiagsArea *&diagsArea)
   SessionDefaults *sessionDefaults = context->getSessionDefaults();
   if (sessionDefaults)
   {
+  
     // Note that it will be required that if a session does not
     // allow queries to be canceled, then it also will not be 
     // possible to suspend the queries.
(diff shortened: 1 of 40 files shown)


zellerh:
+1
Looks good to me.

core/sql/executor/ex_root.cpp
```diff
@@ -2659,33 +2663,21 @@ void ex_root_tcb::registerCB(ComDiagsArea *&diagsArea)
...
   if (cbServer_ == NULL || cbServer_->getControlConnection() == NULL)		
   {		
       // We could not get a phandle for the cancel broker.  However,		
       // let the query run (on the assumption that it will not need to 		
       // be canceled) and convert any error conditions to warnings.		
-
-      // tbd - figure a way retry registration later, as the query progresses.		
-      if (diagsArea != NULL)		
-         NegateAllErrors(diagsArea);		
+      diagsArea->negateErrors(fromCond);
```
DaveBirdsall: The method's caller worries about deallocating the diagsArea in this code path, right?

core/sql/export/ComDiags.h
```diff
@@ -837,7 +837,7 @@ class ComDiagsArea : public IpcMessageObj {
 
   // These members provide set and get operations on the data
   // of a ComDiagsArea that is defined in ANSI table 21, in subclause
-  // 18.1.  See also, ``Creating Errors Korrectly.''
+  // 18.1.  See also, ``Creating Errors Correctly.''
```
DaveBirdsall: I think "Creating Errors Korrectly" is a reference to a paper Bill Rassieur wrote way back in HP days. (And yes he intentionally mis-spelled "Correctly".) I don't know if a copy is up on the Trafodion web site or not. Might be nice to put it there if not.

core/sql/sqlci/SqlCmd.cpp
```diff
@@ -440,25 +461,50 @@ void handleLocalError(ComDiagsArea &diags, SqlciEnv *sqlci_env)
...
+{
+   Int32 rc;
+   Int64 diagsCondCount;
+   rc = SQL_EXEC_GetDiagnosticsStmtInfo2(stmt,
+                           SQLDIAG_NUMBER, &diagsCondCount,
+                           NULL, 0, NULL);
+   if (rc == 0)
+      return 0; 
+   else
+      return diagsCondCount;
```
DaveBirdsall: This is bizzarre. So if SQL_EXEC_GetDiagnosticsStmtInfo2 returns an error, we return diagsCondCount?

core/sql/executor/ex_root.cpp
selvaganesang: Yes.  The caller is responsible to deallocate it

zellerh:
+1 on the third commit (already gave my +1 for the rest)

core/sql/sqlci/SqlCmd.cpp
selvaganesang: I believe the return value is context()->diags()->mainSQLCode()  which would be 100 from the previous CLI call.   Yes. It is bizarre.  This change was done to get the warnings when some regressions failed.

PR #1472, opened 2018-03-13
[TRAFODION-2958] Consolidate the master and backup files in dcs conf folder to a single file (masters)

hegdean:


core/sqf/sql/scripts/dcscheck
```diff
@@ -73,14 +72,14 @@ if ( [ $sq_stat == 0 ] || [ $sq_stat == 1 ] ); then
 
 ### Get the configured primary node for DcsMaster
   if [ -s ${DCS_INSTALL_DIR}/conf/master ]; then
```
DaveBirdsall: Should this statement be changed to conf/masters? If not, should this "if" be deleted?

hegdean: Yes this should be deleted. will take care of it . Thanks

PR #1473, opened 2018-03-13
[TRAFODION-2977] Fix issues with maximum key length detection

DaveBirdsall:
This set of changes fixes several minor bugs:

1. The key length limit that Trafodion imposes differed from the HBase limit. They are now the same.

2. CREATE INDEX logic lacked a check on key length. This has now been added.

3. There was a typo in the message text for message 1141 preventing the computed key length from being displayed. This has been fixed.

4. There was a bug in the generator where, if CREATE INDEX were allowed to have a long key, we generated plans to populate that index that would overrun their buffers, causing ESPs to core. This has been fixed; we now guarantee that buffers are long enough for at least one row.


robertamarton:
Looks good

PR #1474, opened 2018-03-14
[TRAFODION-2988] prepare query returns 2006 internal error

kakaxi3019:
add deal with OR OperatorType

### DIFF
--- a/core/sql/optimizer/mdam.cpp
+++ b/core/sql/optimizer/mdam.cpp
@@ -2002,7 +2002,8 @@ void Disjuncts::computeCommonPredicates()
 	      inVidset.next(predId);
 	      inVidset.advance(predId) )
 	 {
-            if(predId.getItemExpr()->getOperatorType() == ITM_RANGE_SPEC_FUNC )
+            //TRAFODION-2988
+            /*if(predId.getItemExpr()->getOperatorType() == ITM_RANGE_SPEC_FUNC )
             {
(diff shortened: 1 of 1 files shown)


core/sql/optimizer/mdam.cpp
```diff
@@ -2002,7 +2002,8 @@ void Disjuncts::computeCommonPredicates()
 	      inVidset.next(predId);
 	      inVidset.advance(predId) )
 	 {
-            if(predId.getItemExpr()->getOperatorType() == ITM_RANGE_SPEC_FUNC )
+            //TRAFODION-2988
+            /*if(predId.getItemExpr()->getOperatorType() == ITM_RANGE_SPEC_FUNC )
```
DaveBirdsall: One cosmetic issue. It is better to delete code you are replacing rather than commenting it out. The only reason to comment out code is if we are temporarily disabling it in hopes of putting it back it later.

DaveBirdsall:
I noticed that this change didn't get automatically noted in the JIRA. If you begin the subject line with "[TRAFODION-2988]" (capital letters, and without a word in front of it) then github will automatically put a note in the JIRA about the change.

core/sql/optimizer/mdam.cpp
kakaxi3019: Thanks Dave,your suggestion is right.I will fix it and push again.

PR #1477, opened 2018-03-15
[TRAFODION-2985] Revert 2.2.0 download links to mirror pointers

svarnau:


svarnau:
The policy is a bit more complicated.  The checksum files should be on dist, so that the main files on the mirrors can be verified.  But only the most recent release should be on the mirrors. The older releases should be on the archive site.

PierreSmits:
We can suffice with using the following for the download location:
https://www.apache.org/dyn/closer.lua/trafodion

This makes it future proof as it does not point to a specific release. The selected mirror will show the correct version..

traflm:
+1
Just understand all these rules.

gstein:
http://www.apache.org/legal/release-policy.html#host-GA

DaveBirdsall:
Thanks for the policy pointer, Greg.

PR #1479, opened 2018-03-15
[TRAFODION-2993] Add *SYSDATE* in *Trafodion SQL Reference Manual*

liuyu000:


zellerh:
Does CURRENT_DATE really return the date of the client machine, not that of the server? I'm surprised how we could do that.

zellerh:
+1
Sorry about the earlier comment, I think Anoop may have explained this to me in the past.

PR #1481, opened 2018-03-19
[Trafodion-2335] support functions in the column default definition

traflm:
This PR add the support in column default definition the expansion to use following functions:
uuid
unix_timestamp
to_char

To support to_char, it saves the original text of full to_char function, and save it as DEFAULT_VALUE in "_MD_".COLUMNS. 
And the INSERT binder will get this original string, and parse into the ItemExpr tree.

The semantic check is done in setDefaultAttribute.

UUID/UNIX_TIMESTAMP follow the current_timestamp implementation.


traflm:
I will check why two regression test core dump

traflm:
jenkins, retest

traflm:
jenkins, retest

core/sql/parser/sqlparser.y
```diff
@@ -8734,6 +8738,30 @@ datetime_value_function : TOK_CURDATE '(' ')'
...
+                                 NAString * ves= unicodeToChar
+                                   (ToTokvalPlusYYText(&$3)->yytext,
+                                   ToTokvalPlusYYText(&$3)->yyleng,
+                                   (CharInfo::CharSet) (
+                                          ComGetNameInterfaceCharSet() // CharInfo::UTF8
+                                          ),
+                                   PARSERHEAP()); 
+                                 //save the original text
+                                 NAString fullstr;
+                                 fullstr  += "TO_CHAR(";
```
DaveBirdsall: I'm curious why you needed to save the original text. Was the unparse function on the DateFormat node not adequate?

core/sql/parser/ElemDDLColDef.h
```diff
@@ -309,6 +312,8 @@ class ElemDDLColDef : public ElemDDLNode
   NABoolean seabaseSerialized_;
 
   NABoolean isColDefaultSpec_;
+
+  Int32 errCode_;
```
traflm: This is leftover code, I will remove it. When there are more comments, and change together

traflm:
The first test failure seems random

core/sql/regress/seabase/EXPECTED020
```diff
@@ -3945,4 +3945,35 @@ AAAA                       ?            2
...
+>>--negative tests
+>>--the function is not variable-free, so should fail
+>>create table test020t45(a  largeint not null default unix_timestamp(),
++>                        b  varchar(10) default to_char(test020t45.c,'YYYYMMDD'),
++>                        e  int );
+
+*** ERROR[1084] An invalid default value was specified for column B.
+
+*** ERROR[8822] The statement was not prepared.
+
```
selvaganesang: Can you please add tests to test upsert with default function definition when these columns are omitted in upsert command. Upsert can be transformed into merge or plain upsert. It is based on the TRAF_UPSERT_MODE.

core/sql/common/ComSmallDefs.h
```diff
@@ -609,47 +609,59 @@ enum ComColumnClass { COM_UNKNOWN_CLASS
 #define COM_ALTERED_USER_COLUMN_LIT         "C "
 
 enum ComColumnDefaultClass { COM_CURRENT_DEFAULT
+                           , COM_CURRENT_UT_DEFAULT
```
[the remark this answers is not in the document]
traflm: @robertamarton Could you help to provide some comments here?
@selvaganesang  I think new values will NOT cause problems, the MD table structure is not changed. I don't find any checking code relying on values in this column. But Roberta should have more insight here.

core/sql/regress/seabase/EXPECTED020
traflm: Thanks Selva, Let me add such a test case.

selvaganesang: @traflm Thank for adding the test for upsert.  This test just ensures that the upsert command worked. But, it doesn't check if the added default values are correct. You can take some cue from earlier tests in the same test file seabase/TEST020.  Search for cqd traf_upsert_mode

core/sql/common/ComSmallDefs.h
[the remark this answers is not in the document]
anoopsharma00: that is a good point. If a new value is being added to an existing enum in common/ComSmallDefs.h and that value will be stored in metadata, then it cannot be added in the middle of that enum. It should be added at the end.
Better yet, assign a number to  the enums so they dont cause an issue if a new
value is added in the middle.

Something like:
enum ComColumnDefaultClass { COM_CURRENT_DEFAULT                   = 0
                           , COM_NO_DEFAULT                        = 1
                           , COM_NULL_DEFAULT                      = 2
                           , COM_USER_DEFINED_DEFAULT              = 3
                           , COM_USER_FUNCTION_DEFAULT             = 4
                           , COM_IDENTITY_GENERATED_BY_DEFAULT     = 5
                           , COM_IDENTITY_GENERATED_ALWAYS         = 6
                           , COM_ALWAYS_COMPUTE_COMPUTED_COLUMN_DEFAULT = 7
                           , COM_ALWAYS_DEFAULT_COMPUTED_COLUMN_DEFAULT = 8
                           };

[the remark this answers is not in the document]
zellerh: I'm glad you found this. I would recommend to add a big comment saying that these values are stored in the metadata (COLUMNS.DEFAULT_CLASS) and cannot be changed (this is in addition to actually assigning the numbers, as Anoop said). There are only few such enums and it would be good if those were very clearly marked.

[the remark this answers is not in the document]
traflm: Thanks all, I now understand what is the issue! If there are old metadata, and there is a upgrade, then things are messed up.
But I think this enum can be changed by adding new value at the end, instead of "cannot be changed". So let me add them at the end, and add big comments.

core/sql/regress/seabase/EXPECTED020
traflm: @selvaganesang By adding new test case for upsert mode 'merge', the missing column of unix_timestamp() default value is updated. So still something wrong, I need to fix it. Thanks for reminding this.

traflm:
@selvaganesang the upsert mode 'merge' issue was fixed. Could you take a review again?

core/sql/regress/seabase/EXPECTED020
selvaganesang: Please consider doing the same tests for aligned format table in a later check-in

selvaganesang:
+1. I just looked at the new tests added and to confirm if the enum is set

traflm:
Thanks @selvaganesang , I will merge this and continue next JIRA to support SEQUENCE as default val.

PR #1482, opened 2018-03-19
[TRAFODION-2998] sleep execution code cannot build on CentOS 7

traflm:
Wrong data type conversion , and wrong usage of labs


traflm:
wrong fix...

PR #1483, opened 2018-03-19
[TRAFODION-2998] sleep execution code cannot build on CentOS 7

traflm:
Wrong implementation in the sleep runtime code. And made it not able to be built on CentOS 7 because of bad data type conversion.


selvaganesang:
+1 Looks good to me. This could have caused memory corruption in some part of the code.

PR #1484, opened 2018-03-19
[TRAFODION-3000] Fix binder issue with index maintenance on MERGE DELETE

DaveBirdsall:
The inlining code for index maintenance in one place was assuming a record descriptor (RETDesc) for new values existed for a MERGE DELETE statement lacking a "when not matched" clause. But no such record descriptor exists, since there are no values being updated or inserted. That place has been changed to use the old record descriptor.


DaveBirdsall:
I will run full regressions on this change overnight on a workstation. Please do not merge until I report results.

zellerh:
+1
Looks good to me.

DaveBirdsall:
Full regressions passed aside from compGeneral/TEST042 which is failing even in the master branch; the failure in my run was the same as the failure in the master branch run.

PR #1485, opened 2018-03-20
[TRAFODION-2999] make UUID non reserved keyword

traflm:


traflm:
@selvaganesang could you please help to review this change, it is for uuid keyword.
There is requirement to create table name as 'uuid'

core/sql/common/ComResWords.cpp
```diff
@@ -395,8 +395,6 @@ const ComResWord ComResWords::resWords_[] = {
   ComResWord("USAGE",            ANS_|RESWORD_),
   ComResWord("USER",             ANS_|RESWORD_),
   ComResWord("USING",            ANS_|RESWORD_),
-  ComResWord("UUID",            ANS_|RESWORD_),
-  ComResWord("VALUE",            ANS_|RESWORD_),
```
DaveBirdsall: Looks like we accidentally removed VALUE from the reserved words list? This is incorrect.

traflm: Yes, @DaveBirdsall , Let me revert the changes to 'VALUE'.
The reason is we found 'VALUE' is also a popular string for column/table names and there is no strict reason to keep it as reserved word. 
And we changed it in a commercial branch and test and no side effect. So I want to change it together with 'UUID' ( UUID for sure is very popular in the naming of table/columns )
But for clarity, let me change it, unless you change your mind and allow this :-)

DaveBirdsall: It is better to leave VALUE reserved, I think. It is reserved in ISO SQL-2003 (which is the latest copy I have available to me); we may want to implement a feature in the future that uses this keyword. If we do not reserve it, that may be difficult to do.

traflm:
jenkins, retest

core/sql/common/ComResWords.cpp
```diff
@@ -395,8 +395,7 @@ const ComResWord ComResWords::resWords_[] = {
...
+  ComResWord("VALUE",           ANS_|RESWORD_|MPWORD_),
```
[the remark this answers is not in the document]
traflm: Thanks Dave, let me remove this line.

PR #1487, opened 2018-03-20
[TRAFODION-3003]Trafodion keepalive support

CoderSong2015:
Keepalive could be configured by modifying file src/main/java/org/trafodion/dcs/Constants.java or src/main/resources/dcs-default.xml

Modify variable DCS_SERVER_PROGRAM_KEEPALIVE_OPT;

It's a string with parameter ENABLE IDLETIME INTERTIME RETRYCNT (Seconds)
ENABLE=enable/unenable/default ,IDLETIME= number ,INTERTIME=number,RETRYCNT=number
e.g:
ENABLE=enable,IDLETIME=300,INTERTIME=5,RETRYCNT=3

Default value is 1,3600,3,20(Only effective when value configured is set incorrectly)

The value will be read in when mxosrvr start. Mxosrvr will set the socket after getting a connection.

### DIFF
--- a/core/conn/odbc/src/odbc/Common/Global.h
+++ b/core/conn/odbc/src/odbc/Common/Global.h
@@ -139,9 +139,15 @@ class ODBCMXTraceMsg;
 #define DEFAULT_REFRESH_RATE_SECS		60
 #define DEFAULT_SRVR_IDLE_TIMEOUT		0
 #define DEFAULT_CONN_IDLE_TIMEOUT		0
+#define DEFAULT_KEEPALIVE               1     //OPEN KEEPALIVE
+#define DEFAULT_KEEPALIVE_TIMESEC       3600
+#define DEFAULT_KEEPALIVE_COUNT         3
+#define DEFAULT_KEEPALIVE_INTVL         20
 #define INFINITE_SRVR_IDLE_TIMEOUT		-1
 #define INFINITE_CONN_IDLE_TIMEOUT		-1
 #define STATE_TRANSITION_TIMEOUT_SECS	300
+
+#define CLIENT_KEEPALIVE_ATTR_TIMEOUT   3001
 
 #define JDBC_ATTR_CONN_IDLE_TIMEOUT		3000
 #define JDBC_DATASOURCE_CONN_IDLE_TIMEOUT -1L
@@ -935,7 +941,10 @@ typedef struct _SRVR_GLOBAL_Def
 		bzero(m_ProcName,sizeof(m_ProcName));
 		m_bNewConnection = false;
 		m_bNewService = false;
-
+               bzero(clientKeepaliveStatus, sizeof(clientKeepaliveStatus));
+               clientKeepaliveIdletime = 0;
+               clientKeepaliveIntervaltime = 0;
+               clientKeepaliveRetrycount = 0;
 		m_rule_wms_off = false;		// perf
 		m_rule_endstats_off = false;// perf
 
@@ -1053,6 +1062,10 @@ typedef struct _SRVR_GLOBAL_Def
 
 	tip_handle_t		tip_gateway;
 
+    char    clientKeepaliveStatus[64];
+    int     clientKeepaliveIdletime;
+    int     clientKeepaliveIntervaltime;
+    int     clientKeepaliveRetrycount;
 	char				*pxid_url;
 	IDL_long_long		local_xid;
 	UINT				xid_length;
--- a/core/conn/odbc/src/odbc/nsksrvr/Interface/Listener_srvr.cpp
+++ b/core/conn/odbc/src/odbc/nsksrvr/Interface/Listener_srvr.cpp
@@ -27,7 +27,7 @@
 
 #include "Global.h"
 
-//extern SRVR_GLOBAL_Def *srvrGlobal;  // needed in the platform specific implementation file
+extern SRVR_GLOBAL_Def *srvrGlobal;  // needed in the platform specific implementation file
 //extern void flushCollectors();       // needed in the platform specific implementation file
 
 CNSKListenerSrvr::CNSKListenerSrvr()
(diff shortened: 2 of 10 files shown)


selvaganesang:
How do you handle the situation when KEEPALIVE mechanism determines that the client has gone way on the server side. How do you handle it on the client side when the server is gone.  I tried to read surround code to check if this is handled already. I might have missed it. Can you please point me to the code where these situations are handled.

dcs/src/main/resources/dcs-default.xml
```diff
@@ -386,4 +386,12 @@
         Timeout minutes between first and max times (6 default) DCS Server startup MXOSRVR.
     </description>
   </property>
+  <property>
+    <name>dcs.server.user.program.keepalive.opt</name>
+    <value>ENABLE=enable,IDLETIME=300,INTERTIME=5,RETRYCNT=3</value>
```
hegdean: Can we split this into multiple property/values and  change the property name 
 dcs.server.user.program.tcp.keepalive.status
dcs.server.user.program.tcp.keepalive.idletime
dcs.server.user.program.tcp.keepalive.intervaltime
dcs.server.user.program.tcp.keepalive.retrycount

dcs/src/main/java/org/trafodion/dcs/Constants.java
```diff
@@ -92,7 +92,7 @@
     public static final String DCS_SERVER_USER_PROGRAM_COMMAND = "dcs.server.user.program.command";
 
     /** Default value for DCS server user program command */
-    public static final String DEFAULT_DCS_SERVER_USER_PROGRAM_COMMAND = "cd ${dcs.user.program.home};. ./sqenv.sh;mxosrvr -ZKHOST -RZ -ZKPNODE -CNGTO -ZKSTO -EADSCO -TCPADD -MAXHEAPPCT -STATISTICSINTERVAL -STATISTICSLIMIT -STATISTICSTYPE -STATISTICSENABLE -SQLPLAN -PORTMAPTOSECS -PORTBINDTOSECS";
+    public static final String DEFAULT_DCS_SERVER_USER_PROGRAM_COMMAND = "cd ${dcs.user.program.home};. ./sqenv.sh;mxosrvr -ZKHOST -RZ -ZKPNODE -CNGTO -ZKSTO -EADSCO -TCPADD -MAXHEAPPCT -STATISTICSINTERVAL -STATISTICSLIMIT -STATISTICSTYPE -STATISTICSENABLE -SQLPLAN -PORTMAPTOSECS -PORTBINDTOSECS -KEEPALIVEOPT";
```
hegdean: Instead of KEEPALIVEOPT change this to TCPKEEPALIVESTATUS, TCPKEEPALIVEIDLETIME, TCPKEEPALIVEINTERVAL, TCPKEEPALIVERETRYLIMIT

DaveBirdsall:
Could you add a JIRA to this pull request? Thanks.

CoderSong2015:
@DaveBirdsall  OK

dcs/src/main/java/org/trafodion/dcs/Constants.java
CoderSong2015: Got it.

CoderSong2015:
@hegdean @DaveBirdsall Could you review the code again? Thanks.

core/conn/odbc/src/odbc/nsksrvr/Interface/Listener_srvr.cpp
```diff
@@ -81,4 +81,46 @@ void CNSKListenerSrvr::TCP_PROCESSNAME_PORT(FILE* fp)
...
+                keepaliveOpt.keepCount = DEFAULT_KEEPALIVE_COUNT;
+    }else
+    if(0 == strcmp(keepaliveStatus, "enable")){
+        keepaliveOpt.isKeepalive = 1;
+        keepaliveOpt.keepaliveIdle = idleTime;
+        keepaliveOpt.keepaliveInterval = intervalTime;
+        keepaliveOpt.keepCount = retryCount;
+
+    }else{
+        keepaliveOpt.isKeepalive = 0;
```
DaveBirdsall: Why are the other members (keepaliveIdle and so on) not set in this code path? (I assume it matters since we set them in the "unenable" code path.)

core/conn/odbc/src/odbc/Common/Global.h
```diff
@@ -935,7 +941,10 @@ typedef struct _SRVR_GLOBAL_Def
 		bzero(m_ProcName,sizeof(m_ProcName));
 		m_bNewConnection = false;
 		m_bNewService = false;
-
+        bzero(clientKeepaliveStatus, sizeof(clientKeepaliveStatus));
```
DaveBirdsall: Minor point: If we could keep the indentation consistent, that will make the code easier to read in the future. (Perhaps the new code uses tabs and the old code doesn't? Or maybe the old code uses tabs but the new code doesn't? In the past, we've tried to avoid tabs because they lead to these kinds of problems but we have not been successful in keeping them out.)

core/conn/odbc/src/odbc/nsksrvr/SrvrMain.cpp
```diff
@@ -1427,7 +1437,59 @@ BOOL getInitParamSrvr(int argc, char *argv[], SRVR_INIT_PARAM_Def &initParam, ch
...
 			}
-		}
+		}else
+        if (strcmp(arg, "-TCPKEEPALIVESTATUS") == 0){
+            if (++count < argc && argv[count][0] != '-')
+            {
+                if (strlen(argv[count]) < sizeof(keepaliveStatus) - 1)
+                {
+                    memset(keepaliveStatus, 0, sizeof(keepaliveStatus));
+                    strncpy(keepaliveStatus, argv[count], sizeof(keepaliveStatus));
```
DaveBirdsall: I think you want "sizeof(keepaliveStatus)-1" here. If argv[count] has a strlen longer than keepaliveStatus, the first sizeof(keepaliveStatus) bytes will be copied here, overwriting all of the bytes previously set to null. (I don't know what we do for other parameters. Maybe we tolerate this situation today.)

core/conn/odbc/src/odbc/nsksrvr/SrvrMain.cpp
```diff
@@ -1427,7 +1437,59 @@ BOOL getInitParamSrvr(int argc, char *argv[], SRVR_INIT_PARAM_Def &initParam, ch
...
+            else
+            {
+                argEmpty = TRUE;
+                break;
+            }
+        }else
+        if (strcmp(arg, "-TCPKEEPALIVEIDLETIME") == 0){
+			if (++count < argc )
+			{
+				keepaliveIdletime = atoi(argv[count]);
```
DaveBirdsall: If argv[count] isn't numeric, you'll get zero here. Or if it is a string like '123xyz456', you will get  123 here. Is this OK? Or would you rather have something more robust? Similar comments apply to the other parameters below.

core/conn/odbc/src/odbc/Common/Global.h
CoderSong2015: That because old code uses tab which equals 8 blanks but i use 4 blanks directly（so the code aligned on my vim since I set one tab equals 4 blanks.） I will modify it to keep the indentation consistent, on github.

core/conn/odbc/src/odbc/nsksrvr/SrvrMain.cpp
CoderSong2015: I just used the code above as a reference. I think I need to modify the all relevant codes to keep the robustness. Thanks very much for your all suggestions.

CoderSong2015:
@selvaganesang Sorry I haven't seen  your question before.

In tcpip_listener it will handle the revc error:

> GTransport.m_TCPIPSystemSrvr_list->del_node(pnode->m_nSocketFnum);
SET_ERROR((long)0, NSK, TCPIP, UNKNOWN_API, E_SERVER,"tcpip_listener", O_SELECT, F_SELECT,errno, pnode->m_nSocketFnum);
> SRVR::BreakDialogue(NULL);

CoderSong2015:
@selvaganesang Sorry I haven't seen  your question before.
I think mxosrvr and client  just need to treat it like other socket recv errors. keepalive just ensures that the connection will not disconnect due to tcp timeout.

CoderSong2015:
Why my new changes didn't be updated on pull request. Should I open another pull request?

PR #1488, opened 2018-03-20
[TRAFODION-3002] Fix UPDATE STATS for certain INTERVAL data types

DaveBirdsall:
Formerly, UPDATE STATS would fail with an internal error when processing an INTERVAL DAY(13) TO SECOND(0) column. It would also fail with a core when processing an INTERVAL DAY(8) TO SECOND(0) column. This has been fixed.

### DIFF
--- a/core/sql/ustat/hs_globals.cpp
+++ b/core/sql/ustat/hs_globals.cpp
@@ -5052,7 +5052,7 @@ static void mapInternalSortTypes(HSColGroupStruct *groupList, NABoolean forHive 
           }
         else
           {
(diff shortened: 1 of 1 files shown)


DaveBirdsall:
Just to explain the fix: Formerly, when using SECOND(0), the code would cast the DAY(n) TO SECOND(0) to a SECOND(m) datatype. However, INTERVAL SECOND(m) is actually INTERVAL SECOND(m,6) (that is, the fractional part defaults to 6). But in these examples, there is no fractional part. So, the code has been changed so the cast is to SECOND(m,0) instead.

PR #1489, opened 2018-03-20
[TRAFODION-2853] memory leak of ComDiagsArea in CmpContext heap of mx…

selvaganesang:
…osrvr

Fixed a memory leak in ComDiagsArea that was missed out earlier. Also
took care of the review comments from the earlier PR


selvaganesang:
@DaveBirdsall,  I assumed that you meant similar to the way the ExExeUtilGetRTSStatisticsTcb handles the error.  I doubt it because I didn't see errors being repeated other than core/TESTRTS test during my full regression test run.

zellerh:
+1
Looks good to me.

PR #1492, opened 2018-03-22
[TRAFODION-3007] update statistics report error on volatile table

andyyangcn:


core/sql/ustat/hs_parser.cpp
```diff
@@ -243,13 +243,6 @@ Lng32 AddTableName( const hs_table_type type
...
           {
-            // This is for UPDATE STATISTICS; the volatile schema name exists.
-            // For now, UPDATE STATISTICS is not supported. (See also JIRA Trafodion-2004.)
-
-            HSFuncMergeDiags(-UERR_VOLATILE_TABLES_NOT_SUPPORTED);
-            retcode = -1;
-            HSHandleError(retcode); // causes a return from this function
-
             // The code below is old code that will be needed once we turn on
             // support for UPDATE STATISTICS on volatile tables. We leave it here
```
DaveBirdsall: We could delete the comments "The code below is old code..." since we are turning support back on.

andyyangcn: Thanks Dave.  SB_HISTOGRAM* table in the volatile schema gets dropped when the schema gets dropped. I had tested that.

PR #1493, opened 2018-03-22
[TRAFODION-2821] Trafodion core code base needs to be thread safe

selvaganesang:
Java exceptions thrown while calling the java methods from JNI in
Trafodion are stored in the current context. However in a multi-threaded
ESP environment, these exceptions should be stored in thread specific
variable to enable error handling to be thread safe. Otherwise, the JNI
errors could be overwritten by the JNI errors from another thread.

Also streamlined JNI error handling by extending the getExceptionDetails()
method to log the errors also.

Incorporated error handling in SequenceFileReader JNI methods.

### DIFF
--- a/core/sql/executor/HBaseClient_JNI.cpp
+++ b/core/sql/executor/HBaseClient_JNI.cpp
@@ -430,6 +430,7 @@ HBC_RetCode HBaseClient_JNI::releaseHTableClient(HTableClient_JNI* htc)
       jenv_->CallVoidMethod(javaObj_, JavaMethods_[JM_REL_HTC].methodID, j_htc);
       if (jenv_->ExceptionCheck()) {
          getExceptionDetails(__FILE__, __LINE__, "HBaseClient_JNI::releaseHTableClient()");
+         jenv_->PopLocalFrame(NULL);
          return HBC_ERROR_REL_HTC_EXCEPTION;
       }
   }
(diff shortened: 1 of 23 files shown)


selvaganesang:
@robertamarton Can you please review this change

zellerh:
+1
Looks good to me.

core/sql/executor/HBaseClient_JNI.cpp
```diff
@@ -434,10 +429,7 @@ HBC_RetCode HBaseClient_JNI::releaseHTableClient(HTableClient_JNI* htc)
       tsRecentJMFromJNI = JavaMethods_[JM_REL_HTC].jm_full_name;
       jenv_->CallVoidMethod(javaObj_, JavaMethods_[JM_REL_HTC].methodID, j_htc);
       if (jenv_->ExceptionCheck()) {
-         getExceptionDetails();
-         logError(CAT_SQL_HBASE, __FILE__, __LINE__);
-         logError(CAT_SQL_HBASE, "HBaseClient_JNI::releaseHTableClient()", getLastError());
-         jenv_->PopLocalFrame(NULL);
```
DaveBirdsall: Did you mean to delete the jenv_->PopLocalFrame call here?

selvaganesang: Yes. I will add it back in my next check-in

PR #1495, opened 2018-03-23
[TRAFODION-2970] COMMENT-ON bug fix : support single quote in comment string

EEDY:
1. fix error when single quote in comment string
2. add escape quote for DDL statement in SHOWDDL


robertamarton:
Looks good to me.

PR #1496, opened 2018-03-23
[TRAFODION-3008] support a new TMUDF to generate a series of Number

andyyangcn:


sureshsubbiah:
+1. This is a nice UDF to have. I wonder if we can somehow get a parallel plan for this UDF, which may cause repeated values. This is a minor and likely unnecessary concern.

core/sql/sqludr/SqlUdrPredefSeries.cpp
```diff
@@ -0,0 +1,112 @@
...
+extern "C" UDR * SERIES()
+{
+  return new series();
+}
+
+/*  Step 3: Write the actual UDF code*/
+void series::describeParamsAndColumns(
+                           UDRInvocationInfo &info)
+{
+  // We always expect one input parameter
```
DaveBirdsall: The comment is incorrect; from the code it looks like we expect either two or three input parameters.

core/sql/sqludr/SqlUdrPredefSeries.cpp
```diff
@@ -0,0 +1,112 @@
...
+    throw UDRException(38001,
+                       "Expecting two or three input parameters");
+
+  if (info.par().getType(0).getSQLTypeClass() != info.par().getType(1).getSQLTypeClass())
+    throw UDRException(38001,
+                       "The data type of the first two parameters should be same");
+
+  if (info.par().getType(0).getSQLTypeClass() != TypeInfo::NUMERIC_TYPE)
+    throw UDRException(38001,
+                       "Only support to generate series for exact numerics type");
```
DaveBirdsall: Slightly better English might be "Only exact numeric types are supported"

core/sql/sqludr/SqlUdrPredefSeries.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+  for (int i=0; i<paramCount; i++)
+    info.addFormalParameter(info.par().getColumn(i));
+
+  info.out().addLongColumn("ELEMENT");
+
+  // Set the function type, sessionize behaves like
+  // a reducer in MapReduce. Session ids are local
```
DaveBirdsall: Comments should be updated.

core/sql/sqludr/SqlUdrPredefSeries.cpp
```diff
@@ -0,0 +1,112 @@
...
+    long end = info.par().getLong(1);
+    long step;
+    if (info.par().getNumColumns() == 2)
+      step = 1;
+    else
+      step = info.par().getLong(2);
+
+    if (0 == step)
+      throw UDRException(38001,
+                       "Step cannot be a zero");
```
DaveBirdsall: Slightly better English would be "Step cannot be zero"

PR #1499, opened 2018-03-27
[TRAFODION-3003]Trafodion keepalive support

CoderSong2015:
Keepalive could be configured by modifying file src/main/java/org/trafodion/dcs/Constants.java

Modify variable DCS_SERVER_PROGRAM_TCP_KEEPALIVE_STATUS/IDLETIME/INTERVAL/RETRYCOUNT;

DCS_SERVER_PROGRAM_TCP_KEEPALIVE_STATUS has three value:enable,default,unenable;

Default value is enable,300,3,20(Only effective when value configured is set incorrectly)

The value will be read in when mxosrvr start. Mxosrvr will set the socket after getting a connection.

### DIFF
--- a/core/conn/odbc/src/odbc/Common/Global.h
+++ b/core/conn/odbc/src/odbc/Common/Global.h
@@ -139,9 +139,15 @@ class ODBCMXTraceMsg;
 #define DEFAULT_REFRESH_RATE_SECS		60
 #define DEFAULT_SRVR_IDLE_TIMEOUT		0
 #define DEFAULT_CONN_IDLE_TIMEOUT		0
+#define DEFAULT_KEEPALIVE               1     //OPEN KEEPALIVE
+#define DEFAULT_KEEPALIVE_TIMESEC       3600
+#define DEFAULT_KEEPALIVE_COUNT         3
+#define DEFAULT_KEEPALIVE_INTVL         20
 #define INFINITE_SRVR_IDLE_TIMEOUT		-1
 #define INFINITE_CONN_IDLE_TIMEOUT		-1
 #define STATE_TRANSITION_TIMEOUT_SECS	300
+
+#define CLIENT_KEEPALIVE_ATTR_TIMEOUT   3001
 
 #define JDBC_ATTR_CONN_IDLE_TIMEOUT		3000
 #define JDBC_DATASOURCE_CONN_IDLE_TIMEOUT -1L
@@ -935,7 +941,10 @@ typedef struct _SRVR_GLOBAL_Def
 		bzero(m_ProcName,sizeof(m_ProcName));
 		m_bNewConnection = false;
 		m_bNewService = false;
-
+               bzero(clientKeepaliveStatus, sizeof(clientKeepaliveStatus));
+               clientKeepaliveIdletime = 0;
+               clientKeepaliveIntervaltime = 0;
+               clientKeepaliveRetrycount = 0;
 		m_rule_wms_off = false;		// perf
 		m_rule_endstats_off = false;// perf
 
@@ -1053,6 +1062,10 @@ typedef struct _SRVR_GLOBAL_Def
 
 	tip_handle_t		tip_gateway;
 
+    char    clientKeepaliveStatus[64];
+    int     clientKeepaliveIdletime;
+    int     clientKeepaliveIntervaltime;
+    int     clientKeepaliveRetrycount;
 	char				*pxid_url;
 	IDL_long_long		local_xid;
 	UINT				xid_length;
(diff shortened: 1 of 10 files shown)


CoderSong2015:
I have made some changes to make the code more robust and easier to understand.

core/conn/odbc/src/odbc/nsksrvr/Interface/Listener_srvr.cpp
```diff
@@ -81,4 +81,49 @@ void CNSKListenerSrvr::TCP_PROCESSNAME_PORT(FILE* fp)
...
+                                        char *keepaliveStatus,
+                                        int idleTime,
+                                        int intervalTime,
+                                        int retryCount)
+{
+    //all need to be configured
+    if(NULL == keepaliveStatus){
+        return;
+    }
+    if(0 == strcmp(keepaliveStatus,"default")){
```
hegdean: Can the status be boolean (TRUE or FALSE)  and default value can be set FALSE

dcs/src/main/resources/dcs-default.xml
```diff
@@ -386,4 +386,33 @@
         Timeout minutes between first and max times (6 default) DCS Server startup MXOSRVR.
     </description>
   </property>
+  <property>
+    <name>dcs.server.user.program.tcp.keepalive.status</name>
+    <value>enable</value>
+    <description>
+        Used in  mxosrvr keepalive , parameter is ENABLE IDLETIME INTERTIME RETRYCNT.
```
hegdean: Description should be specific to the property name/value..  specify the default for the property.  Look at how the description for statistics enable is done. This applies to the other tcp keepalive properties as well

dcs/src/main/resources/dcs-default.xml
```diff
@@ -386,4 +386,33 @@
...
+    <value>300</value>
+    <description>
+        Used in  mxosrvr keepalive , parameter is ENABLE IDLETIME INTERTIME RETRYCNT.
+    </description>
+  </property>
+  <property>
+    <name>dcs.server.user.program.tcp.keepalive.intervaltime</name>
+    <value>5</value>
+    <description>
+        Used in  mxosrvr keepalive , parameter is ENABLE IDLETIME INTERTIME RETRYCNT.
```
hegdean: description should be specific to the property. Also would be good to indicate the unit of time  mins, secs or msec and so on

dcs/src/main/java/org/trafodion/dcs/server/ServerManager.java
```diff
@@ -83,6 +83,10 @@
     private int maxRestartAttempts;
     private int retryIntervalMillis;
     private String nid = null;
+    private static String mxosrvrKeepaliveStatus;
```
hegdean: can u change mxosrvr to userProg just like the other properties already defined

CoderSong2015: OK.

CoderSong2015:
@hegdean @selvaganesang Could you review the code?

core/conn/odbc/src/odbc/Common/Global.h
```diff
@@ -139,10 +139,16 @@ class ODBCMXTraceMsg;
 #define DEFAULT_REFRESH_RATE_SECS		60
 #define DEFAULT_SRVR_IDLE_TIMEOUT		0
 #define DEFAULT_CONN_IDLE_TIMEOUT		0
+#define DEFAULT_KEEPALIVE               1     //OPEN KEEPALIVE
+#define DEFAULT_KEEPALIVE_TIMESEC       3600
+#define DEFAULT_KEEPALIVE_COUNT         3
```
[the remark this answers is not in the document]
CoderSong2015: I think 3 times and 5 times are both enough here. Can your tell me why it is better to retry more times?

core/conn/odbc/src/odbc/nsksrvr/Interface/linux/Listener_srvr_ps.cpp
```diff
@@ -263,7 +263,11 @@ void* CNSKListenerSrvr::OpenTCPIPSession()
 //LCOV_EXCL_STOP
 	}
 
-
+    TCP_SetKeepalive(nSocketFnum,
```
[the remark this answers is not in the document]
CoderSong2015: I think you are right, but I have seen the same name style in mxosrvr a lot of times, like in SrvrConnect.cpp and odbcas_drvr.cpp. So I think they are all allowed here.

mkby:
I will commit the changes if there're no more comments in the next 24 hours.

PR #1501, opened 2018-03-28
[TRAFODION-3016] Windows ODBC do not support output with zh-cn.utf8

Weixin-Xu:
The Chinese encode of windows is GBK.
So driver will convert the Chinese characters to GB2312.
But when UTF-8 is needed, it is not supported.
And "SET NAMES 'UTF8'" is used by mysql and PostgreSQL also.

### DIFF
--- a/win-odbc64/odbcclient/drvr35/cstmt.cpp
+++ b/win-odbc64/odbcclient/drvr35/cstmt.cpp
@@ -1063,7 +1063,20 @@ SQLRETURN CStmt::SendSQLCommand(BOOL SkipProcess, SQLCHAR *StatementText,
             else
                 m_StmtType = TYPE_UNKNOWN;
         }
-
+        else if (strcmp(token, "SET") == 0)
+        {
+            token = strtok(NULL, delimiters);
+            if (token != NULL && strcmp(token, "NAMES") == 0)
+            {
+                token = strtok(NULL, delimiters);
+                if (token != NULL && strcmp(token, "'UTF8'") == 0)
+                {
+                    m_ConnectHandle->m_DSValue.setDSCharSet(1);
+                    return SQL_SUCCESS;
+                }
+            }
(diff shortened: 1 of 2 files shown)


win-odbc64/odbcclient/drvr35/cstmt.cpp
```diff
@@ -1063,7 +1063,20 @@ SQLRETURN CStmt::SendSQLCommand(BOOL SkipProcess, SQLCHAR *StatementText,
             else
                 m_StmtType = TYPE_UNKNOWN;
         }
-
+		else if (strcmp(token, "SET") == 0)
```
xiaozhongwang: use space pls!
else the format looks not good.

Weixin-Xu: Thanks for your suggestion, I have corrected this problem.

win-odbc64/odbcclient/drvr35/cstmt.cpp
```diff
@@ -1063,7 +1063,20 @@ SQLRETURN CStmt::SendSQLCommand(BOOL SkipProcess, SQLCHAR *StatementText,
...
                 m_StmtType = TYPE_UNKNOWN;
         }
-
+        else if (strcmp(token, "SET") == 0)
+        {
+            token = strtok(NULL, delimiters);
+            if (token != NULL && strcmp(token, "NAMES") == 0)
+            {
+                token = strtok(NULL, delimiters);
+                if (token != NULL && strcmp(token, "'UTF8'") == 0)
```
DaveBirdsall: I'm wondering if there should be an "else" for this "if"? Suppose I say, "SET NAMES 'UTF9'", what happens?

Weixin-Xu: The buffer will be considered as a query and sent to server.

PR #1502, opened 2018-04-02
various fixes, details in pull request

anoopsharma00:
-- JIRA 2980 support for INSTR function.
   INSTR(source_string, pattern, startPos, occurrence)
   Search for pattern in source_string.
   Start at startPos'th character. Optional, default 1
   Return occurrence'th occurrence. Optional, default 1
   (exp/exp_function.cpp/h, exp/ExpPCodeClauseGen.cpp,
    generator/GenItemFunc.cpp, optimizer/BindItemExpr.cpp, ItemExpr.cpp,
    optimizer/ItemFunc.h, SynthType.cpp, parser/sqlparser.y)

-- hivemd returns precision, scale, display_datatype
  (comexe/ComTdbExeUtil.h, common/BaseTypes.cpp,NAType.cpp/h,
   executor/ExExeUtilGet.cpp, ExExeUtil.h)

-- hive decimal type is treated as sql NUMERIC for performance.
  (common/NAType.cpp)

-- hive select from hivemd with prepare and multiple executes work
  (executor/ExExeUtilGet.cpp)

-- fix for crash in sql buffer pool when allocating large size rows.
   (executor/sql_buffer.cpp)

-- SIGN(op) now returns error if op is not numeric
   (BindItemExpr.cpp)

-- strings of length zero are no longer cached. Caching them causes problem
   during backpatching.
   (optimizer/ItemCache.cpp)

--- compile time predicates on constants were not being created correctly
    in some cases. That has been fixed.
   (ValueDesc.cpp)

-- trim now supported on scaled numeric, float and bignum datatypes
   (BindItemExpr.cpp)

-- CASE is now supported on scaled numerics, float, bignum
   (BindItemExpr.cpp)

-- char_length is now supported on numerics
   (ItemFunc.h, BindItemExpr.cpp)

-- DAYOFMONTH now allowed only on datetime datatypes
   (common/OperTypeEnum.h, sqlparser.y, BindItemExpr.cpp, ItemExpr.cpp)

-- TO_TIME now allowed only on 'datetime with time' or character datatypes.
   (BindItemExpr.cpp, bin/SqlciErrors.txt)

-- updated expected file for compGeneral/test042


DaveBirdsall:
If you put [TRAFODION-2980] in the subject line of this pull request, then this pull request will be automatically recorded in that JIRA.

core/sql/common/NAType.h
```diff
@@ -562,7 +562,8 @@ class NAType : public NABasicObject
 				 short displayCaseSpecific = 0);
 
   short getMyTypeAsText(NAString * outputStr, 
-			NABoolean addNullability = TRUE) const; // output
+			NABoolean addNullability = TRUE,
+                        NABoolean addCollation = TRUE) const; // output
```
DaveBirdsall: Looks like the "// output" comment migrated to the wrong place. Maybe it belongs on "NAString * outputStr"? Or we could just delete it.

core/sql/executor/ExExeUtilGet.cpp
```diff
@@ -5795,40 +5833,51 @@ short ExExeUtilHiveMDaccessTcb::work()
...
-            infoCol->fsDatatype = 
-              getFSTypeFromHiveColType(hcd ? hcd->type_ : hpd->type_);
-
-	    if (infoCol->fsDatatype < 0)
+            Lng32 fstype = -1;
+            Lng32 length = -1;
+            Lng32 precision = -1;
+            Lng32 scale = -1;
+            char sqlType[50];
+            char displayType[100];
```
DaveBirdsall: I get a little worried when I see hard-coded lengths. Elsewhere for example the displayType seems to be 96. Seems like if we are not checking buffer lengths at run time we should be using some enum or const ints for these items or risk buffer overrun problems down the line when someone else is too casual about these lengths.

core/sql/executor/ExExeUtilGet.cpp
```diff
@@ -5795,40 +5833,51 @@ short ExExeUtilHiveMDaccessTcb::work()
...
+                                          sqlType, displayType, charset);
+                 
+	    if (retcode < 0)
 	      {
+                // add a warning and continue.
 		char strP[300];
-		sprintf(strP, "Datatype %s is not supported.", 
-                        (hcd ? hcd->type_ : hpd->type_));
-		*diags << DgSqlCode(-CLI_GET_METADATA_INFO_ERROR)
+		sprintf(strP, "Datatype %s for column '%s' in table %s.%s.%s is not supported. This table will be ignored.",
```
DaveBirdsall: 300 is not long enough. The column name, schema name and table name can each be up to 256 characters.

core/sql/executor/ExExeUtilGet.cpp
```diff
@@ -5795,40 +5833,51 @@ short ExExeUtilHiveMDaccessTcb::work()
...
+                        (hcd ? hcd->type_ : hpd->type_),
+                        (hcd ? hcd->name_ : hpd->name_),
+                        hiveCat_, hiveSch_, htd->tblName_);
+		*diags << DgSqlCode(CLI_GET_METADATA_INFO_ERROR)
 		       << DgString0(strP);
-		
-		step_ = HANDLE_ERROR_;
-		break;
+
+                step_ = ADVANCE_ROW_;
```
DaveBirdsall: getTypeAttrsFromHiveColType only initializes some of the output fields when it returns an error. Since we are converting the error to a warning now, it seems like it should initialize all of the output fields, or we should initialize them here.

core/sql/optimizer/ValueDesc.cpp
```diff
@@ -6381,35 +6380,47 @@ void ValueIdList::convertToTextKey(const ValueIdList& keyList, NAString& result)
...
-		   newVal += val(vLen - constType->getScale(), constType->getScale());
-		 }
+               // See how many positions the result will take in the display
+               Lng32 t = constType->getDisplayLength(constType->getFSDatatype(),
+                                               constType->getNominalSize(),
+                                               constType->getPrecision(),
+                                               constType->getScale(),
+                                               0);
+
+               char strval[100];
```
DaveBirdsall: If you make this "char strval[t+1]" you will be guaranteed that you will never have a buffer overrun here.

core/sql/executor/ExExeUtilGet.cpp
zellerh: Why not use snprintf(strP, sizeof(strP), ...)?

selvaganesang: In #1504 I am changing this type of populating error to use ExRaiseSqlError. This needs to be changed to use ExRaiseSqlError because diags could be null

core/sql/executor/ExExeUtilGet.cpp
selvaganesang: Need to change it into ExRaiseSqlError because diags could be null

core/sql/common/NAType.h
anoopsharma00: thanks, will fix that.

core/sql/executor/ExExeUtilGet.cpp
anoopsharma00: will fix that by creating enums for these lengths.

core/sql/executor/ExExeUtilGet.cpp
anoopsharma00: If that error condition is reached, we dont return that row. It is skipped  and we move
to the next row through step ADVANCE_ROW_.
The warning is maintained in the diags area which is returned when the next row
or EOD is returned.

core/sql/optimizer/ValueDesc.cpp
anoopsharma00: good idea.

PR #1505, opened 2018-04-04
TRAFODION-3019 pstmt.setByte cannot support string.getbyte well

mashengchen:
fill in blank when give value to a char() type.
support Chinese value, when column type is UCS2


core/conn/jdbc_type2/src/main/java/org/apache/trafodion/jdbc/t2/SQLMXDesc.java
```diff
@@ -424,7 +424,7 @@ void checkValidNumericConversion(Locale locale) throws SQLException
     public static final int SQLCHARSETCODE_GB2312           = 17;
 
 	public static final String SQLCHARSETSTRING_UNKNOWN     = "UNKNOWN";
-	public static final String SQLCHARSETSTRING_ISO88591    = "ISO88591";
+	public static final String SQLCHARSETSTRING_ISO88591    = "ISO-8859-1";
```
DaveBirdsall: I am curious why this has been changed. In the SQL engine itself, we use "ISO88591". Will this cause a problem?

PR #1506, opened 2018-04-04
[TRAFODION-3018] Fix issue with UPPER on an UPSHIFT column

DaveBirdsall:
The BuiltinFunction::bindNode method inserts Cast nodes with the matchChildType flag on on top of the operands of any function that can be affected by substituting another equal value. For example, OCTET_LENGTH is sensitive to this: While 'ax' = 'ax   ' from a SQL point of view, the two values have different octet lengths so we want to force the data type of any substitution to whatever was there originally. The Upper::bindNode method removes the Upper node when its operand already has the Upshift attribute. The bug is, Upper::bindNode wasn't also removing the Cast inserted by BuiltinFunction::bindNode on its behalf. This lead to trouble at code generation time when an UPPER function on an upshifted column was used in an equi-join predicate on a hash join.

The fix is to change Upper::bindNode to also remove any such inserted Cast node.

### DIFF
--- a/core/sql/optimizer/BindItemExpr.cpp
+++ b/core/sql/optimizer/BindItemExpr.cpp
@@ -3494,9 +3494,25 @@ ItemExpr *Upper::bindNode(BindWA *bindWA)
   if (bindWA->errStatus()) return this;
 
   // If our child is already upshifted, we can remove this Upper from the tree.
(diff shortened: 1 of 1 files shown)


zellerh:
+1
I'm glad you found a great fix for this problem!

PR #1512, opened 2018-04-10
[TRAFODION-2175] Get statement enhancements

robertamarton:
Added support and privilege checks for the following commands:

  get functions for library
  get procedures for library
  get table_mapping functions for library

  get indexes on table
  get objects on table
  get views on table
  get views on view

  get libraries in schema
  get objects in view
  get tables in view
  get views in view

  get indexes for user
  get tables for user
  get libraries for user
  get views for user

Changed "get libraries for schema" to include libraries where the current user
has execute privilege on one of the libraries routines (functions, procedures,
or table_mapping functions).

Addressed a performance issue when determining if the user has column level
privileges. If the user has granted privileges against native Hive tables
through EsgynDB, we need to get the column name from Hive. The call to get the
column, by calling hivemd, is very expensive. This change checks to see if the
requested user has been granted any column level privileges on a hive table.
If so, we will go ahead and do the mapping (call hivemd). If not, then we will
not include the hivemd fragment for the query.  Since we are scanning the column
privileges table anyway, we also see if the requested user (or their roles) has
been granted any privileges. If so, we include the column privileges check in
the query.

Commented out get statements that we do not support at this time.


DaveBirdsall:
Is there a JIRA for this?

robertamarton:
TRAFODION-2175

DaveBirdsall:
You may want to edit the subject line to add [TRAFODION-2175] at the beginning.

core/sql/executor/ExExeUtilGet.cpp
```diff
@@ -1642,6 +1672,119 @@ NABoolean ExExeUtilGetMetadataInfoTcb::checkUserPrivs(
...
+    return -1;
+  }
+
+  bool hasHive = false;
+  bool hasGrants = false;
+  infoList_->position();
+  OutputInfo * vi = (OutputInfo*)infoList_->getCurr();
+  if (vi && vi->get(0))
+  {
+    if (*(Lng32*)vi->get(0) > 0)
```
DaveBirdsall: I experimented with a SUM aggregate and found that the output is a 64-bit integer. So, *(Lng32*) might give a wrong answer if we ever have someone with 2^32 privileges (admittedly unlikely).

robertamarton: I will change these to int64 for a future delivery.

PR #1518, opened 2018-04-14
[TRAFODION-3021] union, intersect, except should be cacheable

traflm:
Make UNION, INTERSECT, EXCEPT cacheable.


traflm:
jenkins, retest

DaveBirdsall:
Looks like core/TEST005 failed. There are two queries in that test that look at the query cache. That behavior might have changed due to your change. If this is true, you'll need to update the expected results for that test also.

traflm:
Thanks Dave, I run this on my workstation, it passed the test005. but run again, yes, it failed. Let me update the expect file, and have a new try.

PR #1520, opened 2018-04-16
[TRAFODION-2600] Unable to create view ... but user has SELECT privilege

robertamarton:
Query invalidation is not resetting the role list when a user is granted a role.
For DML operations, we always retry the request once, and between retries, the
role list is reset.  So DML works on a retry.
However, DDL operations are not retried, so the role list is not reset and the
create view fails.

An analogous issue exists when the role is revoked from a user and the role
list is not reset.  In this case, the user can still create views even though
they no longer have the privilege.

Changes:
- Grant role: sends a new query invalidation key
- Revoke role: forces a query invalidation check even if the key is not present
- Displays query invalidation keys when debug option DBUSER_DEBUG is set, e.g:
   set envvar DBUSER_DEBUG 1;


robertamarton:
jenkins, retest

PR #1522, opened 2018-04-17
[TRAFODION-3009] Streamline error handling in Executor utility commands

selvaganesang:
Changes to avoid allocation of ComDiagsArea in some more places unless it is
needed to pass errors or warnings.

[TRAFODION-2917] Refactor Trafodion implementation of hdfs scan for text formatted hive tables

Disabling USE_LIBHDFS_SCAN by default. The new implementation of Hdfs Scan is now used to
scan the text type hive files.

Hdfs Scan is now stopped gracefully when the hive scan is cancelled. This avoids the random
core seen with new implementation.

### DIFF
--- a/core/sql/executor/ExExeUtilCli.cpp
+++ b/core/sql/executor/ExExeUtilCli.cpp
@@ -1094,7 +1094,7 @@ Lng32 ExeCliInterface::executeImmediate(const char * stmtStr,
   Lng32 retcode = 0;
 
   ComDiagsArea * tempDiags = NULL;
-  if (globalDiags != NULL && *globalDiags != NULL)
+  if (globalDiags != NULL && *globalDiags != NULL && (*globalDiags)->getNumber() > 0)
   {
      tempDiags = ComDiagsArea::allocate(heap_);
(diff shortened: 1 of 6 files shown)


core/sql/executor/ExExeUtilCli.cpp
```diff
@@ -1119,7 +1119,7 @@ Lng32 ExeCliInterface::executeImmediate(const char * stmtStr,
     {
       // Allocate the diagnostics area if needed
       // and populate the diagnostics conditions
-      if (*globalDiags == NULL && retcode != 0) {
+      if (*globalDiags == NULL && retcode != 0 && retcode != 100) {
```
DaveBirdsall: I take it in the old code we would throw away the 100 warning somewhere else?

selvaganesang: Returning retcode with a value of 100 is ok and should be treated as end of result by the callers. This additional check just avoids creating a ComDiagsArea if the return code is just 100.

selvaganesang:
Please don't merge this yet. I am working on seabase/TEST031 failure seen with this change

PR #1525, opened 2018-04-18
[TRAFODION-3028] Support CONTROL QUERY SHAPE (CQS) for Hive insert

zellerh:
Adding fast_extract() and hive_insert() CQS operator syntax to be able to force shapes of Hive insert statements.

Example:

control query shape hive_insert(cut);
explain insert into hive.hive.t values(1,1);

Note: I don't think this fix works for UNLOAD, since that seems to be
an "ExeUtil" statement that gets executed in two phases.

### DIFF
--- a/core/sql/common/ExprNode.cpp
+++ b/core/sql/common/ExprNode.cpp
@@ -737,6 +737,16 @@ NABoolean OperatorType::match(OperatorTypeEnum wildcard) const
 	      return FALSE;
 	    }
 
+        case REL_ANY_EXTRACT:
+	  switch (op_)
+	    {
+	    case REL_FAST_EXTRACT:
+	    case REL_HIVE_INSERT:
+	      return TRUE;
+	    default:
+	      return FALSE;
+	    }
+
 	case ANY_REL_OR_ITM_OP:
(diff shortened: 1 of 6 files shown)


core/sql/common/ExprNode.cpp
```diff
@@ -737,6 +737,16 @@ NABoolean OperatorType::match(OperatorTypeEnum wildcard) const
 	      return FALSE;
 	    }
 
+        case REL_ANY_EXTRACT:
+	  switch (op_)
+	    {
+	    case REL_FAST_EXTRACT:
```
DaveBirdsall: It's counterintuitive to me that a Hive Insert (or any Insert) is considered an Extract. "Extract" to me has the connotation of pulling something out, while "Insert" to me has the connotation of putting something in. Perhaps "Extract" in this context is in reference to a Trafodion object? But would a Hive Insert always be taking data from a Trafodion object? Or might we support Insert/Select from one Hive object to another?

DaveBirdsall: Thank you for the explanation. It's fine as it is.

PR #1527, opened 2018-04-19
[TRAFODION-3030] Add *RAND Function* in Trafodion SQL Reference Manual

liuyu000:


zellerh:
Unfortunately, non-deterministic functions like RAND() don't work very well in SQL. The documentation needs to mention that. From the current text, it sounds like we guarantee a certain behavior, such as that select RAND(), RAND(), RAND() will give independent values. That may not be true in all situations.

Another example: select * from t where a < RAND(). I can't define what this query means. Do we evaluate RAND() once per query or once per row?

At the least we need a warning in this section that RAND() does not always behaves in predictable ways.

We had this problem with other functions as well. See also TRAFODION-2954, https://github.com/apache/trafodion/pull/1444, TRAFODION-2912.

liuyu000:
@zellerh Thanks for your review :)

1. I've run a test and found that the RAND function is evaluated per row (see example below). 
Can I add this to manual?

```
SQL>SELECT * FROM t1;
C1          C2         
----------- -----------
          1           2
          2           2
          3           2
          4           2
          5           2
--- 5 row(s) selected.

SQL>SELECT c1, c2 FROM t1 WHERE c2 < MOD(RAND(),5);
C1          C2         
----------- -----------
          2           2
          5           2
--- 2 row(s) selected.

SQL>SELECT c1, c2 FROM t1 WHERE c2 < MOD(RAND(),5);
C1          C2         
----------- -----------
          2           2
          4           2
--- 2 row(s) selected.

SQL>SELECT c1, c2 FROM t1 WHERE c2 < MOD(RAND(),5);
--- 0 row(s) selected.

SQL>SELECT c1, c2 FROM t1 WHERE c2 < MOD(RAND(),5);
C1          C2         
----------- -----------
          3           2
          4           2
--- 2 row(s) selected.
```

2. Yes, I found that some results are not same as expected, they are regularly changed (see example below), so I delete the first example in the manual and add a NOTE based on your comments.
```
SQL>SELECT RAND(1), RAND(10), RAND(50), RAND(100), RAND(150), RAND(200), RAND(1000) FROM DUAL;

(EXPR)     (EXPR)     (EXPR)     (EXPR)     (EXPR)     (EXPR)     (EXPR)    
---------- ---------- ---------- ---------- ---------- ---------- ----------
     16807     168070     840350    1680700    2521050    3361400   16807000

--- 1 row(s) selected.
```

traflm:
This is similar as UUID(), I will add documentation of UUID asap. So @liuyu000  can refer to my documentation of UUID, if you can wait for a while. 
Or I think one note can be added here , is it OK @zellerh ? Or maybe some better words for Liu Yu?

Rand() function is non-deterministic and can cause unexpected side effect. It can evaluate once or per row depending on the optimizer. So it can only be used in very few places, please do not use it in a predicate.

zellerh:
Yes, something like your note would work. It is probably a good idea to use similar words for both uuid and rand.

liuyu000:
@zellerh @traflm thanks for your reivew :)
I've incorporated your NOTE @traflm, and will keep an eye on your documentation of UUID.

liuyu000:
@zellerh Herzlichen Dank für ihre Hilfe :)

PR #1530, opened 2018-04-19
[TRAFODION-3031] Fix two issues with nested subquery plans

DaveBirdsall:
See the JIRA for a discussion of the problem.

This set of changes contains two fixes:

1. The heuristics that decide whether to do the semi-join to join + group-by have been modified. First, if the group by would reduce the number of rows in the inner table by a factor of 5.0 or more, the transformation will be enabled. The rationale is that this may reduce the size of the inner hash table if hash join is chosen. Also it allows the join to be commuted, which may lead to a better plan. Second, if the inner table has only a small number of rows (less than 100.0), the transformation will be enabled. The rationale here is that it allows the join to be commuted which may lead to an efficient nested join plan. Both of the constants here are controlled by new CQDs. (Note: Thanks go to @sureshsubbiah for the first of these ideas and the code to do it.)

2. There was a bug in Scan::addIndexInfo, where we would sometimes overlook a useful index. We would mistakenly think that another index provided the same or better information when in fact one index would satisfy a join predicate that the other would not. This bug has been fixed.


DaveBirdsall:
Please do not merge this in yet; I will be running full regressions on a workstation with these changes overnight.

core/sql/optimizer/NormRelExpr.cpp
```diff
@@ -2767,25 +2767,26 @@ Here t2.a is a unique key of table t2.
...
 where t1.b in (1,2,3,4,...,101) ;
 
 
-  Semi Join {pred : t1.b = t2.a}          Join {pred : t1.b = InList.col} 
+  Semi Join {pred : t1.b = InList.col}  Join {pred : t1.b = InList.col}
  /         \                   ------->  /    \
 /           \                           /      \
-Scan t1     Scan t2                 Scan t1     GroupBy {group cols: InList.col}
+Scan t1   TupleList                 Scan t1   GroupBy {group cols: InList.col}
                                                   |
```
zellerh: Nice to make the picture consistent, but from the code it looks like we do this for things other than TupleList, so maybe "Scan t2" or "Q2" would be a better name for the child?

core/sql/optimizer/NormRelExpr.cpp
```diff
@@ -2826,18 +2827,39 @@ RelExpr* Join::transformSemiJoin(NormWA& normWARef)
...
+  is small, nested join might be profitably chosen in that case. */
 
       ValueIdSet preds ;
       preds += joinPred();
       preds += selectionPred();
       preds -= getEquiJoinPredicates() ;
 
+      EstLogPropSharedPtr innerEstLogProp = child(1)->getGroupAttr()->outputLogProp((*GLOBAL_EMPTY_INPUT_LOGPROP));
+      CostScalar innerRowCount = innerEstLogProp->getResultCardinality(); 
+      CostScalar innerUec = innerEstLogProp->getAggregateUec(equiJoinCols1);
```
zellerh: One question about this: This UEC can be very unreliable, if we don't have multi-column statistics. So, could it happen that we have a semi-join with a very large inner table, where our UEC is underestimated? The extra groupby could be a bad choice in that situation.

On the other hand, I think we usually overestimate multi-column UECs, so maybe this is not so much of a concern.

DaveBirdsall:
Do not merge. Both the Jenkins run and my full regression run show some plan changes that need to be analyzed. I will likely submit additional changes.

DaveBirdsall:
The reworks do the following:

1. Update expected results for compGeneral/TEST015. One query plan changed from hash semi join to hash join + group by. I did a performance test of both plans; their execution time average is almost the same.

2. Change the small inner table criterion to fire only if statistics exist on the inner table. (I had meant to do this originally and simply forgot.)

3. Update comments in Join::transformSemiJoin per @zellerh 's suggestions.

zellerh:
+1
Still looks good to me :-)

robertamarton:
Looks good to me

PR #1532, opened 2018-04-21
[TRAFODION-3026] add create option storage policy

traflm:
Some user want Trafodion to support HSM provided by Hadoop 2.6.0. Tables can be stored on different type of storage hardware (SSD/HDD). Hadoop provides HSM in HDFS, but Hbase doesn't provide the support until HBase 2.0. There is HBASE-19858 and plan to support HSM in HBase 1.5.0. But still not released.
So before that HBase feature availible, we now use HDFS API to directly set the storage policy on Hbase table path.
Trafodion adds new HBASE_OPTIONS to support this feature. 
So when creating a Trafodion table, it is possible to specify the storage policy (by default, HDFS provide hot/warm/cold/one_ssd/all_ssd etc)
This provide a basic support of user's requirement.


traflm:
jenkins, extra tests

traflm:
jenkins, extra tests

traflm:
extra test failed at unrelated tests, should not be introduced by this change:
>>create table tlob130ts2 (c1 int not null, c2 blob, primary key (c1));

So I will do full regression by hand in my dev machine.

@DaveBirdsall @selvaganesang could you help to review the code?

traflm:
jenkins, retest

DaveBirdsall:
Yes, I will review tomorrow.

core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
```diff
@@ -552,10 +578,84 @@ public boolean createk(String tblName, Object[] tableOptions,
...
+          ret = "No access to setStoragePolicy on FileSystem from the SecurityManager";
+          m = null; // could happen on setAccessible() or getDeclaredMethod()
+        }
+        if (m != null) {
+          try {
+            m.invoke(fs, path, storagePolicy);
+            if (logger.isDebugEnabled()) {
+              logger.debug("Set storagePolicy=" + storagePolicy + " for path=" + path);
+            }
+          } catch (Exception e) {
```
selvaganesang: Please try to avoid catching generic Exception if possible

core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
```diff
@@ -552,10 +578,84 @@ public boolean createk(String tblName, Object[] tableOptions,
...
+               logger.error("invoke set storage policy error : " + e);
```
selvaganesang: To preserve the exception stack, call logger.error("message", e);

core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
```diff
@@ -552,10 +578,84 @@ public boolean createk(String tblName, Object[] tableOptions,
...
+               ret = "invoke set storage policy error : " + e.getMessage();
```
selvaganesang: To preserve the call stack, either let the method to throw all the exceptions.  If the generic Exception needs to be caught due to underlying API, throw new IOException(e). Let the caller also throw this IOException to be handled in the JNI layer

traflm:
@selvaganesang thanks for your review comments, I add some fix, please help to review again.

traflm:
It turns out permission denied. 
I have to go through the coprocessor I think. 
Will rework.

traflm:
@selvaganesang thanks for your review comments, I add some fix, please help to review again.
In my own workstation, all components Trafodion and HBase running using same userID. In the regression test and in real use case, they are running in different userID.
So I changed the code to use Coporocessor, Trafodion send request to HBase Region Server to change the storage policy of the data path, so running under HBase userID.
Others no change

DaveBirdsall:
@selvaganesang, does this look good to you now? Should we merge it?

selvaganesang:
Sorry for the late response. Earlier, I reviewed the change without considering the overall impact of this feature.  On further thinking, I have the following question.

If we can't modify the storage policy at hbase table level and choose to do it at hdfs level, how do you plan to take care of new data files created via data ingestion and the compaction at Hbase level.

traflm:
hi @selvaganesang , the storage policy is now set at the top level data dir for a given table, so new data files should be inherit it
For compaction, the final data file is still under that data dir, so it is same. But the intermediate files generated I need to check, but I feel it is allowed.

selvaganesang:
@traflm Can you also check what happens if there is a region split?

traflm:
Yes, Selva, I will create a big table and split it and see what will happen

traflm:
@selvaganesang I create a table and force the split, after the split, all new region files are under the top dir of that table. So I think split is not a problem.

selvaganesang:
Thanks for testing the split @traflm.   This change looks ok. But, if you look at the other co-processor service requests, there is a concept of retryable exception. Do we need such concept for this coprocessor service call too?

traflm:
Yes, you are right, let me enhance the coprocessor code to do retry

traflm:
@selvaganesang I added the retry logic, if the coprocessor fail, retry until reach the retry limit. Please help to check again :-) Thank you.

traflm:
I believe this is good to merge, I will wait for current regression issue solved and then merge it.

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionManager.java
```diff
@@ -3226,5 +3260,65 @@ public RecoveryRequestResponse call(TrxRegionService instance) throws IOExceptio
...
+            throw new IOException("coprocessor not response");
+          }
+          if (retry) 
+              retrySleep = retry(retrySleep);
+        } while (retry && retryCount++ < RETRY_ATTEMPTS);
+      }
+      catch (Exception e) {
+         throw new IOException(e);
+      }
+  }
```
selvaganesang: retry is never set to true. So, it is not clear when the request would be retried

selvaganesang: Also, can you please consider coding the LOG.error as follows with 2 arguments
LOG.error(<message>, <Exception>)

Note the  ',' instead of '+'

LOG.error(<nessage + e>), you would only the text of the error message and will not report the stack trace of the exception

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/TrxRegionEndpoint.java.tmpl
```diff
@@ -2404,6 +2410,72 @@ CoprocessorService, Coprocessor {
...
+
+  private static void invokeSetStoragePolicy(final FileSystem fs, final String pathstr,
+      final String storagePolicy)
+       throws IOException {
+        Path path = new Path(pathstr);
+        Method m = null;
+        try {
+            m = fs.getClass().getDeclaredMethod("setStoragePolicy",
+            new Class<?>[] { Path.class, String.class });
+            m.setAccessible(true);
```
selvaganesang: To avoid increased path length, consider issue reflection API to get the method Id once in static block

selvaganesang:
@prashanth-vasudev  Please can you also review this change?

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionManager.java
```diff
@@ -3226,5 +3260,65 @@ public RecoveryRequestResponse call(TrxRegionService instance) throws IOExceptio
...
+        org.apache.hadoop.hbase.coprocessor.transactional.generated.TrxRegionProtos.TrxRegionService.BlockingInterface service =
+          org.apache.hadoop.hbase.coprocessor.transactional.generated.TrxRegionProtos.TrxRegionService.newBlockingStub(channel);
+        org.apache.hadoop.hbase.coprocessor.transactional.generated.TrxRegionProtos.TrafSetStoragePolicyRequest.Builder request =
+         org.apache.hadoop.hbase.coprocessor.transactional.generated.TrxRegionProtos.TrafSetStoragePolicyRequest.newBuilder();
+        String hbaseRoot = config.get("hbase.rootdir");
+        FileSystem fs = FileSystem.get(config);
+      //Construct the HDFS dir
+      //find out if namespace is there
+      String[] parts = tblName.split(":");
+      String namespacestr="";
```
prashanth-vasudev: One concern here is we make assumptions of hbase internal directory structure for the table. This assumption may not hold good between releases. Is there any public api that can be used or some uniform way?

core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
```diff
@@ -538,7 +564,21 @@ public boolean createk(String tblName, Object[] tableOptions,
...
+            if(setDescRet!= null)
+            {
+              if(setDescRet.storagePolicyChanged())
+              {
+                 Object tableOptionsStoragePolicy[] = new Object[HBASE_HDFS_STORAGE_POLICY+1];
+                 for(int i=0; i<HBASE_HDFS_STORAGE_POLICY; i++)
+                   tableOptionsStoragePolicy[i]="";
+                 tableOptionsStoragePolicy[HBASE_HDFS_STORAGE_POLICY]=(String)setDescRet.storagePolicy_ ;
+                 tableOptionsStoragePolicy[HBASE_NAME]=(String)tblName;
+                 alter(tblName,tableOptionsStoragePolicy,transID);
```
prashanth-vasudev: Transactional Alter is not supported yet. This is due to additional changes required to support transactional rollback of a table.  Can you make coprocessor call outside of transactional alter?

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionManager.java
```diff
@@ -2996,6 +3022,11 @@ else if (tableOption.equalsIgnoreCase("USE_DEFAULT"))
                    (Long.parseLong(tableOption));
                returnStatus.setTableDescriptorChanged();
                break ;
+           case HBASE_HDFS_STORAGE_POLICY:
```
prashanth-vasudev: General question.  Is it too late to change (alter ) storage policy of a table after it is created. Should this is part of create?

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionManager.java
traflm: The API is only available at HBase 2.0.  But I don't know when Trafodion will support HBase 2.0. And this feature is required by a customer, so we cannnot wait for months. This is a temp solution, once we move to HBase 2.0, most of these code can be eliminated and using HBase public API.

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/client/transactional/TransactionManager.java
traflm: thanks Selva! Sorry, I should be more careful, I never understand the LOG and exception clear, will change this.

core/sqf/src/seatrans/hbase-trx/src/main/java/org/apache/hadoop/hbase/coprocessor/transactional/TrxRegionEndpoint.java.tmpl
traflm: Yes, Selva, this is a good suggestion.

core/sql/src/main/java/org/trafodion/sql/HBaseClient.java
traflm: I am not farmiliar with DDL transaction, need some time to understand this and make changes here.
But maybe we can make this work in different phases, in order to have it basically working.
I don't think this action need to do rollback, since if the CREATE fail, the path will be removed, so no need to set the storage policy back and then remove.
Removing the file should be part of the original CREATE TABLE process, but maybe I am misunderstanding here.

prashanth-vasudev: Thanks for the responses.  Please verify if transactional alter side of the code is functioning. We can enhance the alter in steps.   Thank you.

selvaganesang:
@traflm  Sorry for the delay.  I believe, when you make a coprocessor call it can raise two exceptions
ServiceException and Throwable.   My understanding is the ServiceException is retryable, while Throwable exception is the wrapper for the exception raised in the coprocessor code, which may or may not be retryable.  The coprocessor code can also return an exception in the response as a text. See TransactionManager.java for details. 

I haven't tested this understanding fully. 

Are you assuming that ret would be null in case of ServiceException? In that case, the ServiceException would be caught and would never be retried.

traflm:
Hi, Selva, I didn't get very clear understanding of the exceptions here. You mentioned there are two kinds of exceptions, and ServiceException is retryable; while Throwable is not retryable.
So I can catch these two exceptions and assign the retry flag. But your last statement is unclear to me: 'the ServiceException would be caught and would never be retired"
I assume if there is RPC error (retryable), the ret will be null, but it is very hard to simulate, and I cannot find descriptions in HBase manuals about this. 
I never saw throwable exception description in https://hbase.apache.org/devapidocs/org/apache/hadoop/hbase/master/MasterRpcServices.html , so not sure how to handle that exception. I assume we just return error to caller for throwable exception.
Let me make this change.

traflm:
@selvaganesang Please help to take a review again. Sorry for fix issues one by one, I am really not quite familiar with Java programming and this logic. Any new suggestions will be very appreciated, and thanks for your patient review again and again, this helps me greatly!

traflm:
Hi, Selva,
I think it is OK to merge it now. The retry logic is not the key. Let me wait for a few days, if you don't have objections, I will merge it. This is required by some users.

traflm:
Thank you Selva for so many valuable  review, I need to better understanding the Java exception handling, I will make that change in the next phase of this work.

PR #1534, opened 2018-04-23
[TRAFODION-3039] SendEventMsg is used in a wrong way

xiaozhongwang:
SendEventMsg use dynamic parameter.
And there is a parameter control the total, all of dynamic parameters are string data type.
But some place input a number parameter for it.
I checked it again and found two such mistakes.

### DIFF
--- a/core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
+++ b/core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
@@ -6533,6 +6533,8 @@ odbc_SQLSrvr_ExtractLob_sme_(
         if (retcode == SQL_ERROR)
         {
             ERROR_DESC_def *p_buffer = QryLobExtractSrvrStmt->sqlError.errorList._buffer;
+            char             errNumStr[128];
+            sprintf(errNumStr, "%d", p_buffer->sqlcode);
             strncpy(RequestError, p_buffer->errorText, sizeof(RequestError) - 1);
 
             SendEventMsg(MSG_SQL_ERROR,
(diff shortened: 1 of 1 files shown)


core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
```diff
@@ -6533,6 +6533,8 @@ odbc_SQLSrvr_ExtractLob_sme_(
         if (retcode == SQL_ERROR)
         {
             ERROR_DESC_def *p_buffer = QryLobExtractSrvrStmt->sqlError.errorList._buffer;
+            char             errNumStr[128];
+            sprintf(errNumStr, "%d", p_buffer->sqlcode);
             strncpy(RequestError, p_buffer->errorText, sizeof(RequestError) - 1);
```
selvaganesang: Surround code issue.  RequestError may not be null terminated when the RequestError size is less than the length of the string in p_buffer->errorText. Also, this can cause core dump due to segment violation if length of errorText is less than the size of RequestBuffer.

arvind-narain: @selvaganesang  since RequestError is initialized at line 6489 and 6585, is there still a concern?
     char RequestError[200] = {0}; 

@xiaozhongwang if you want you could also take care of line 5254 ? 
Also do check regarding %d vs.  %ld for sqlcode.

Sorry for the delay in review. Thanks for the change.

xiaozhongwang: I don't look for this fix a long time. the branch have been deleted.
So I pull a new request for this, modified as your suggestion.

PR #1536, opened 2018-04-24
[TRAFODION-2542] Grantor is not correct when granting privileges for …

robertamarton:
…a role

When granting privileges and the authorization ID is not the current user but
one of roles granted to the current user, then the "granted by" clause is
required.  In addition, the grantor of the privileges becomes the role specified
in the grant statement instead of the current user.

Added a CQD ALLOW_WGO_FOR_ROLES that will return an error if the user tries to
grant a privilege as a role.

Added error message (1194) when a component operation is not defined.

Added a check to not allow the WITH GRANT OPTION when granting privileges
to public

### DIFF
--- a/core/sql/sqlcomp/PrivMgrPrivileges.cpp
+++ b/core/sql/sqlcomp/PrivMgrPrivileges.cpp
@@ -4531,7 +4531,7 @@ PrivStatus PrivMgrPrivileges::getRolesToCheck(
     {
       // just return what getAuthNameFromAuthID returns
       ComUser::getAuthNameFromAuthID(roleIDs[r],roleName, sizeof(roleName),length);
-      if (r > 0)
+      if (rolesWithPrivs.size() > 0)
         rolesWithPrivs += ", ";
       rolesWithPrivs += roleName;
     }
--- a/docs/messages_guide/src/asciidoc/_chapters/ddl_msgs.adoc
+++ b/docs/messages_guide/src/asciidoc/_chapters/ddl_msgs.adoc
@@ -195,8 +195,13 @@ This is currently not supported.
 == SQL 1012
 
 ```
-No privileges were granted. You lack grant option on the specified privileges.
+No privileges were granted.  <auth_id> lacks grant option on the specified privileges. <details>.
 ```
+
+Where <auth-id> is a {project-name} user name.
+
+Where <details> suggests roles you can specify in order for the grant to succeed.
+If <auth_id> has not been granted any relevant roles, <details> is omitted.
 
 *Cause:* You attempted to grant privileges for which you do not have
 grant options.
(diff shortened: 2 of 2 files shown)


core/sql/bin/SqlciErrors.txt
```diff
@@ -11,7 +11,7 @@
 1009 ZZZZZ 99999 BEGINNER MINOR DBADMIN Column $0~ColumnName does not exist in the specified table.
 1010 0A000 99999 ADVANCED MINOR DBADMIN The statement just entered is currently not supported.
 1011 ZZZZZ 99999 ADVANCED MINOR DBADMIN Only one grantee per grant or revoke is allowed.
-1012 01007 99999 BEGINNER MAJOR DBADMIN No privileges were granted.  You lack grant option on the specified privileges.
+1012 01007 99999 BEGINNER MAJOR DBADMIN No privileges were granted.  $0~String0 lacks grant option on the specified privileges. $1~String1
```
DaveBirdsall: Please update the description of this message in the Messages Guide as well.

core/sql/bin/SqlciErrors.txt
```diff
@@ -193,7 +193,7 @@
 1191 ZZZZZ 99999 BEGINNER MAJOR DBADMIN SERIALIZE option is not yet supported for $0~string0 datatype.
 1192 ZZZZZ 99999 BEGINNER MAJOR DBADMIN Failed to retrieve data from Hive metastore.  Call to $0~string0 returned error $1~string1($0~int0). Cause: $2~string2.
 1193 ZZZZZ 99999 UUUUUUUU UUUUU UUUUUUU The $0~string0 specified in the $1~string1 clause must be identical to the primary key for a Trafodion table.
-1194 ZZZZZ 99999 ADVANCED MAJOR DIALOUT --- unused ---
+1194 ZZZZZ 99999 UUUUUUUU UUUUU UUUUUUU Component operation $0~string0 does not exist for component $1~string1
```
DaveBirdsall: Please add this message to the Messages Guide.

core/sql/sqlcomp/PrivMgrComponentPrivileges.cpp
```diff
@@ -840,11 +840,12 @@ PrivStatus PrivMgrComponentPrivileges::grantPrivilege(
       // more items in the list fail and in cases of "ALL".                                             
       if (!componentOperations.nameExists(componentUID,operationName))
       {
-         *pDiags_ << DgSqlCode(-CAT_TABLE_DOES_NOT_EXIST_ERROR)
-                  << DgTableName(operationName.c_str());
+         *pDiags_ << DgSqlCode(-CAT_INVALID_COMPONENT_PRIVILEGE)
```
DaveBirdsall: Looks like we were using the wrong error message before? Thanks for fixing this.

core/sql/sqlcomp/PrivMgrPrivileges.cpp
```diff
@@ -4445,6 +4503,50 @@ PrivStatus PrivMgrPrivileges::getPrivsFromAllGrantors(
...
+    PrivMgrDesc privsOfTheRole(roleIDs[r],true);
+    if (getUserPrivs(objectType, roleIDs[r], emptyRoleIDs, privsOfTheRole,
+                     hasManagePrivPriv) != STATUS_GOOD)
+      return STATUS_ERROR;
+
+    if (!privsOfTheRole.isNull())
+    {
+      // just return what getAuthNameFromAuthID returns
+      ComUser::getAuthNameFromAuthID(roleIDs[r],roleName, sizeof(roleName),length);
+      if (r > 0)
```
DaveBirdsall: This doesn't look correct. I think you want a comma if you've already added a role to rolesWithPrivs. r > 0 means only that you've already processed one of the roleIDs array members. I think this code gives the wrong result if privsOfTheRole.isNull() is true for roleIDs[0].

robertamarton:
Thanks Dave for the pointing out formatting issue and for the message reminder.

PR #1537, opened 2018-04-24
[TRAFODION-3041] Support watchdog query cache

CoderSong2015:
To improvement performance when every query executed needs to be written into repository.

The time when the queries cached will be published into repository is determined by cache time or number of queries. You can configure it in the file dcs/conf/dcs-site.xml.

### DIFF
--- a/core/conn/odbc/src/odbc/nsksrvr/SrvrConnect.cpp
+++ b/core/conn/odbc/src/odbc/nsksrvr/SrvrConnect.cpp
@@ -559,14 +559,16 @@ static void* SessionWatchDog(void* arg)
         int time_limit = aggrInterval;
         //0:None 1:update 2:insert/upsert cache limit 3:achieve timeline
         int execute_flag = REPOS_EXECUTE_NONE;
-        clock_t time_start = clock();
-        clock_t time_end= clock();
+        int sleep_count = 0;
 
         REPOS_STATS repos_stats;
         while(!record_session_done && okToGo)
         {
-            time_start = clock();
-            while(repos_queue.isEmpty() && (((time_end = clock()) - time_start) / 1000000 < time_limit));
+            sleep_count = 0;
+            while(repos_queue.isEmpty() && (sleep_count < time_limit)){
+                sleep(1);
+                sleep_count++;
+            }
 
             if(!repos_queue.isEmpty()){
                 repos_stats = repos_queue.get_task();
--- a/core/conn/odbc/src/odbc/nsksrvr/SrvrConnect.cpp
+++ b/core/conn/odbc/src/odbc/nsksrvr/SrvrConnect.cpp
@@ -119,6 +119,7 @@ extern int sdconn;
 extern int clientConnTimeOut;
 extern short stopOnDisconnect;
 extern int aggrInterval;
+extern int statisticsCacheSize;
 extern int queryPubThreshold;
 extern statistics_type statisticsPubType;
 extern bool bStatisticsEnabled;
@@ -342,10 +343,21 @@ public:
     {
         return my_queue.size();
     }
+
+    bool isEmpty(){
+        return my_queue.empty();
+    }
 };
 
 static Repos_Queue<REPOS_STATS> repos_queue;
 static bool record_session_done = true;
+
+//0:None 1:update 2:insert/upsert cache limit 3:achieve timeline
+
+#define REPOS_EXECUTE_NONE       0
+#define REPOS_EXECUTE_UPDATE     1
+#define REPOS_EXECUTE_CACHELIMIT 2
+#define REPOS_EXECUTE_TIMELINE   3
 
 static void* SessionWatchDog(void* arg)
 {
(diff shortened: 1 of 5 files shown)


core/conn/odbc/src/odbc/nsksrvr/SrvrConnect.cpp
```diff
@@ -530,11 +542,39 @@ static void* SessionWatchDog(void* arg)
...
+        //0:None 1:update 2:insert/upsert cache limit 3:achieve timeline
+        int execute_flag = REPOS_EXECUTE_NONE;
+        clock_t time_start = clock();
+        clock_t time_end= clock();
+
+        REPOS_STATS repos_stats;
+        while(!record_session_done && okToGo)
+        {
+            time_start = clock();
+            while(repos_queue.isEmpty() && (((time_end = clock()) - time_start) / 1000000 < time_limit));
```
DaveBirdsall: I'm not sure I understand this logic. The "while" loop looks like it will spin until the time limit expires or repos_queue is non-empty. This raises two concerns: 1. Spin loops burn CPU needlessly. It would be better to put a sleep call inside the loop. 2. In the logic below, you check to see if the queue is non-empty and then dequeue something. This does not look thread-safe. Another thread might dequeue something after the repos_queue.isEmpty() call and before the repos_queue.get_task() call.

CoderSong2015: Thanks.  I think you are right that spin loops will burn CPU needlessly. As for your second point, every mxosrvr will have only  one watchdog thread, and only this function will call repos_queue.get_task(). Besides, the whole watchdog function  will be locked at the beginning and unlocked at the end so that I think it is thread-safe and doesn't need to add lock there. Thanks again and look forward to your insight. @DaveBirdsall

venkat1m: If you set the statisticsCacheSize to 1, would we preserve the current behavior where mxosrvr publishes the repository records when it is time. For example session start is published as soon as session starts and session end is updated as soon as session ends. Like wise we need aggregate stats every 1 minute. Will the existing behavior be retained by setting statisticsCacheSize to 1 ?

CoderSong2015: Thanks. Of course not. The existing behavior is that as long as there is data in the queue， it will be published into repository immediately. statisticsCacheSize just determines the number of data cached in the queue. if it is set to 1, the data will be sent immediately once it goes into the queue, like  'The existing behavior'. @venkat1m

selvaganesang:
Are we preparing the query every time with literals? Would it be better to prepare the query once and execute it via a rowset and/or one row at a time?

CoderSong2015:
I will need to maintain 5 internal statements if I do so. Besides, it is difficult to use prepare and execute api of internal statement(data need to be organized again). So I'm not sure if it is better that use rowset. Another question, will compiler and executor handle query with same structure at second time faster? If so, I think it is better to do above than moving all data again(use rowset). This is my opinion, welcome to correct me. Thanks for your question. @selvaganesang

selvaganesang:
The current change may be functionally good.  JIRA mentioned something like "query performance will degrade".

If the query in the JIRA refers to the user query, then repository writing is in a different thread and context that of the user query. Hence, it is not in the user query execution path.  

If the query performance refers to the repository queries, I am not sure batching the query to execdirect at once may not improve the sub-micro second execution of the insert statement.. It might  turn it out to be a drag to add the individual SQL string into a vector and process it one by one. There is no process hop other than writing the data to hbase.

Each SQL statement converts the numeric data into a string to form the SQL string at the repository thread and the SQL engine would reverse it to numeric data to be pushed to hbase layer.   Rowset concept could avoid these conversions.

However, if you could demonstrate the stated goal in the JIRA,, please go ahead and merge this PR.

venkat1m:
If a mxosrvr aborts, we would the entire cache, not just the last query stats. I would suggest to change the default value for statisticsCacheSize to be 1 to preserve the current behavior where repository updates are done as and when things change. Explicitly override the default to larger value to enable caching.

CoderSong2015:
I made a simple test, using jdbct4 to execute query 'insert into song values(i)' 10k times. MXOSRVR needs to spend more than about 40% (11mins vs 17mins)time writing statistics into repository. I just used one user thread to insert. 
ENV: centos6.7 16GMEM  2 MXOSRVR .

Type    | Cache size|row in METRIC_QUERY_TABLE|ueser query CPU (AVG) | free memory percent|repos CPU |start time| user query finish time | repos finish time(all finish)|
----|---------------|-----------|----------------------------------|--------------------|----------|------|---|-----------------------|
Cache   | 100 |      43101->53111 |    35-38  | 62.4%->60.2% |   24-28 |10.16:11 |  10:19:03 | 10:27:54|
No Cache| 1   |      53124->63142 |    35-38  | 58.0%->52.4% |   24-29 |10.39:47  | 10:42:13 | 10:56:22|
Cache   | 500 |      63161->73150 |    35-38  | 54.1%->50.5% |   24-31 |11.03:02  | 11:05:42 | 11:13:30|

@selvaganesang 
It is only a simple test.

CoderSong2015:
@selvaganesang I have written an email to trafodion community and gotten the answer why there is improvement in performance. Your second question that why speed of writing the statistics is so slow deserves continuing research.  I will create another JIRA to find it out.

kevinxu021:
+1 Go to go, I think.

PR #1538, opened 2018-04-24
[TRAFODION-2949] Improve LOB Extract/Update Performance 

sandhyasun:
Removing unneeded CLI call to get lob location and instead using the  CQD value - LOB_STORAGE_FILE_DIR to get/save it in Tdb for use later at runtime.


sandhyasun:
jenkins retest

sandhyasun:
No we don't allow for the lcoation to change between compiletime and  runtime. We'll just get an error at  runtime saying the LOB data was not found. Yes the initial design allowed for changes or even each LOB handle's data being stored in potentially different locations. But right now we have jsut one location since the LOB data is stored in  HDFS . If it was Linux and if we go back to storing LOBs on mounted linux drives, we may want to allow for different locations again. But for now this extra CLI call is slowing down each extract or Update call especially during streaming access.

PR #1539, opened 2018-04-24
[TRAFODION-3042] RAND() function is not always random

zellerh:
When called without a seed, we use a seed based on the system timestamp. The random generator we use generates similar output values for seeds that are close together. Adding another randomization step to avoid that.

Also switched from a microsecond to a nanosecond-based timestamp.

### DIFF
--- a/core/sql/exp/exp_function.cpp
+++ b/core/sql/exp/exp_function.cpp
@@ -5953,7 +5953,7 @@ void ExFunctionRandomNum::initSeed(char *op_data[])
       // we call this method again within a short time. The eval() method
       // below doesn't handle seed values that are close to each other
       // very well.
-      seed_ = (((Int64) seed_) * 1664525L + 1664525L) % 2147483648;
+      seed_ = (((Int64) seed_) * 1664525L + 1013904223L) % 2147483648;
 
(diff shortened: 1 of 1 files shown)


core/sql/exp/exp_function.cpp
```diff
@@ -5937,45 +5938,22 @@ void ExFunctionRandomNum::initSeed(char *op_data[])
...
+      seed_  = (Int32) (seedTime.tv_sec  % 2147483648);
+      seed_ ^= (Int32) (seedTime.tv_nsec % 2147483648L);
+
+      // Go through one step of a linear congruential random generator.
+      // (https://en.wikipedia.org/wiki/Linear_congruential_generator).
+      // This is to avoid seed values that are close to each other when
+      // we call this method again within a short time. The eval() method
+      // below doesn't handle seed values that are close to each other
+      // very well.
+      seed_ = (((Int64) seed_) * 1664525L + 1664525L) % 2147483648;
```
DaveBirdsall: Did you mean, (((Int64) seed_) * 1664525L + 1013904223L) & 2147483648? (These are the parameters suggested by the wiki article.)

PR #1540, opened 2018-04-25
[TRAFODION-2953] initial implementation of small table in one Hbase Region

traflm:
This is the very first code drop for TRAFODION-2953, the major purpose is to get feedback to see if the design is OK and shall we continue with this effort or not.
Please check https://issues.apache.org/jira/browse/TRAFODION-2953, I attached a design document there.

This PR contains initial prototype of the design. It still has a major code-refactor to do, but I would like to hear from community first, before I spend more time on it, in case there are something fundamentally wrong.

The major code-refactor I mentioned above is for this prototype code is to add a new PartitionFunction, instead of modifying the SinglePartitionPartitioningFunction directly. I will inherit a new SinglePartitionWithSmallTablePartitioningFunction from SinglePartitionPartitioningFunction next.

One can check the TEST035 to get a demo of how it works.

There are many other tasks following this initial code drop. For example, index/alter/create-super-table etc.

The major idea to support this is to add a new SYSTEM column _TBLNM_, so in a single Region, rows belong to different tables can be distinguished. When doing IUDS, always add predicate _TBLNM_ = 'theNameOfTheTable' to filter out the correct rows. 

Please help to review and let me know your comments, many thanks. 

The requirement is strong. In some use cases, there must be many small tables.

### DIFF
--- a/core/sql/regress/tools/sbdefs
+++ b/core/sql/regress/tools/sbdefs
@@ -22,7 +22,7 @@
 -- defs included during seabase regr run
 
 #ifdef SEABASE_REGRESS
-cqd mode_seabase 'ON';
+--cqd mode_seabase 'ON';
 cqd seabase_volatile_tables 'ON';
 cqd hbase_async_drop_table 'OFF';
 cqd hbase_serialization 'ON';
--- a/core/sql/comexe/ComTdb.h
+++ b/core/sql/comexe/ComTdb.h
@@ -821,6 +821,7 @@ class ComTdbVirtTableTableInfo  : public ComTdbVirtTableBase
   const char * allColFams;
   Int64 objectFlags;  // flags from OBJECTS table
   Int64 tablesFlags;  // flags from TABLES table
+  const char * superTable;
 };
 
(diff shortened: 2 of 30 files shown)


traflm:
something wrong with my PR, most changes not in the PR. Close it.

traflm:
Recovered, PR is no OK...
@anoopsharma00 @DaveBirdsall @zellerh , please help to take a review

core/sql/regress/tools/sbdefs
```diff
@@ -22,7 +22,7 @@
 -- defs included during seabase regr run
 
 #ifdef SEABASE_REGRESS
-cqd mode_seabase 'ON';
+--cqd mode_seabase 'ON';
```
traflm: typo, will fix this in next commit. Sorry

traflm:
This PR is not ready, far from ready, sorry, let me close it.

traflm:
reopen just for some tests, Do not merge

traflm:
jenkins, retest

PR #1542, opened 2018-04-26
[TRAFODION-3044] Add Syntax and Examples of *ENABLE/DISABLE INDEX (ALTER TABLE Statement)* in Trafodion SQL Reference Manual

liuyu000:


docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
```diff
@@ -442,6 +446,10 @@ ref-spec is:
 
 column-list is:
    column-name[, column-name]...
+
+index-name is:
```
DaveBirdsall: I don't think I would make this distinction in the syntax. There is no syntactic difference between what you're calling a "normal index" and a unique index. And all index names are in the same name space, whether or not they are unique. Consider deleting these lines.

docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
```diff
@@ -576,6 +584,59 @@ the two columns or sets of columns must have the same characteristics (data type
...
+<<<
+* ENABLE ALL [UNIQUE] INDEXES
+
+** ENABLE ALL UNIQUE INDEXES
++
+enables all unique indexes on the table.
+
+** ENABLE ALL INDEXES
++
+enables all indexes (including normal indexes and unique indexes) on the table.
```
DaveBirdsall: I have not seen the term "normal indexes" anywhere else. When we feel the need to distinguish an index that is not unique, we usually just say, "non-unique indexes". Suggested wordsmith: "enables all indexes (including both non-unique and unique indexes) on the table".

docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
```diff
@@ -576,6 +584,59 @@ the two columns or sets of columns must have the same characteristics (data type
...
+
+* _index-name_
```
DaveBirdsall: Again, I would not make a distinction at this level between unique and non-unique indexes. Suggested wordsmith: "* _index-name_   <blank lines> is an SQL identifier that specifies the simple name for the index. You cannot qualify index-name... <the rest of your text is fine>"

docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
```diff
@@ -576,6 +584,59 @@ the two columns or sets of columns must have the same characteristics (data type
...
++
+is equal to _index-name_, which is an SQL identifier that specifies the simple name for the index. You cannot qualify index-name with its schema name. 
+Indexes have their own name space within a schema, 
+so an index name might be the same as a table or constraint name. However, no two indexes in a schema can have the same name.
+
+** _unique-index-name_
++
+is the name of the unique index.
+
+* ENABLE INDEX _normal-index-name_
```
DaveBirdsall: Again, I would not distinguish between unique and non-unique indexes here. Simply say "ENABLE INDEX _index-name_ <blank lines> enables an index on the table" is sufficient.

docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
```diff
@@ -576,6 +584,59 @@ the two columns or sets of columns must have the same characteristics (data type
...
+enables an unique index on the table.
+
+* DISABLE ALL [UNIQUE] INDEXES
+** DISABLE ALL UNIQUE INDEXES
++
+disables all unique indexes on the table.
+
+** DISABLE ALL INDEXES
++
+disables all indexes (including normal indexes and unique indexes) on the table.
```
DaveBirdsall: Similar remarks here

docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
```diff
@@ -576,6 +584,59 @@ the two columns or sets of columns must have the same characteristics (data type
...
+
+* DISABLE INDEX _index-name_
+
+** DISABLE INDEX _normal-index-name_
```
DaveBirdsall: Similar remarks here

docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
```diff
@@ -716,6 +777,222 @@ Alter TABLE PRODUCT
 ALTER COLUMN vend_id RENAME TO cstm_id;
 ```
 
+* The following table _orders_ has two normal indexes
```
DaveBirdsall: Suggest "non-unique" instead of "normal"

docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
```diff
@@ -716,6 +777,222 @@ Alter TABLE PRODUCT
...
+ ATTRIBUTES ALIGNED FORMAT 
+;
+ 
+-- GRANT SELECT, INSERT, DELETE, UPDATE, REFERENCES ON TRAFODION.SEABASE.ORDERS TO DB__ROOT WITH GRANT OPTION;
+
+--- SQL operation complete.
+```
+
++
+Disable all indexes and verify that all indexes (including normal indexes and unique indexes) are disabled.
```
DaveBirdsall: Here too

docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
```diff
@@ -716,6 +777,222 @@ Alter TABLE PRODUCT
...
+    CUSTNUM ASC
+  )
+ ATTRIBUTES ALIGNED FORMAT 
+;
+ 
+-- GRANT SELECT, INSERT, DELETE, UPDATE, REFERENCES ON TRAFODION.SEABASE.ORDERS TO DB__ROOT WITH GRANT OPTION;
+
+--- SQL operation complete.
+```
+
```
DaveBirdsall: The examples look great!

docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
liuyu000: Thanks @DaveBirdsall, I've deleted. :)

docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
liuyu000: Thanks @DaveBirdsall, agree with you that the word "non-unique" is more accurate. I've changed that. :)

However, I read the term "normal indexes" in: 

* [Oracle SQL Reference Manual](https://docs.oracle.com/cd/B19306_01/server.102/b14200/statements_5010.htm): 

> An index is a schema object that contains an entry for each value that appears in the indexed column(s) of the table or cluster and provides direct, fast access to rows. Oracle Database supports several types of index:
Normal indexes. (By default, Oracle Database creates B-tree indexes.)
Bitmap indexes, which store rowids associated with a key value as a bitmap
...

> B-tree indexes
These indexes are the standard index type. They are excellent for primary key and highly-selective indexes. Used as concatenated indexes, B-tree indexes can retrieve data sorted by the indexed columns.
...

* [MySQL Reference Manual](https://dev.mysql.com/doc/refman/8.0/en/optimizing-spatial-analysis.html):

> It is also possible to create normal indexes on spatial columns. In a non-SPATIAL index, you must declare a prefix for any spatial column except for POINT columns. 
...

* [PostgreSQL Documentation](https://www.postgresql.org/docs/9.5/static/routine-vacuuming.html):
> Second, it allows PostgreSQL to answer some queries using only the index, without reference to the underlying table. Since PostgreSQL indexes don't contain tuple visibility information, a normal index scan fetches the heap tuple for each matching index entry, to check whether it should be seen by the current transaction.
...

docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
liuyu000: Thanks @DaveBirdsall, I've udpated :)

docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
liuyu000: Thanks @DaveBirdsall, I've udpated :)

docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
liuyu000: Thanks @DaveBirdsall, I've udpated :)

docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
liuyu000: Thanks @DaveBirdsall, I've udpated :)

docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
liuyu000: Thanks @DaveBirdsall, I've udpated :)

docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
liuyu000: Thanks @DaveBirdsall, I've udpated :)

docs/sql_reference/src/asciidoc/_chapters/sql_statements.adoc
liuyu000: Thanks @DaveBirdsall :)

PR #1546, opened 2018-04-30
[TRAFODION-3050] Change default for CQD COMP_BOOL_158 to 'ON'

DaveBirdsall:
This pull request changes the default for CQD COMP_BOOL_158 to 'ON'.

This causes equi-join predicates on VARCHAR/VARCHAR and CHAR/VARCHAR pairs of columns to be rewritten as VEGPreds, enabling transitive closure on them. It also makes the predicate available to the scan index elimination logic (Scan::addIndexInfo) so that certain queries that formerly did not make use of certain indexes will now consider them. This can lead to very efficient nested join plans in some cases.

There is a side effect to turning this CQD on, however. By allowing free use of VEGPreds for such equi-join predicates, the compiler may substitute references to one column with another, if the two columns compare as equal in a join predicate. 

This side effect is thought to be less undesirable at this time than getting bad query plans when such equi-join predicates are present.

Nevertheless, a separate JIRA, TRAFODION-3051, has been written to address this side effect. When that one is fixed, we will be able to have our cake and eat it too.


zellerh:
+1
IMHO this is a good default behavior and I hope there will be very few people who need to revert back to the OFF setting of this CQD.

PR #1548, opened 2018-05-04
[Trafodion-3058] Linux ODBC add Ansi function support

Weixin-Xu:
Because the lack of Ansi function support of Linux ODBC, you should export AppUnicodeType when you use "isql" to connect Trafodion.

But actully it is a wrong action because "isql" calls Ansi APIs and then DriverManager mapping to Unicode APIs.

The environment variable AppUnicodeType should be only used for application which is compiled with defining Unciode.


Weixin-Xu:
The Jenkins failure is caused by pyodbc test, I will fix it ASAP

Weixin-Xu:
@DaveBirdsall  
Hi Dava, I have mixed Ansi and Unicode functions, it can make Jenkins pass.
But I'm not sure whether it is feasible, please review chandle.cpp again after Jenkins passed.

PR #1550, opened 2018-05-06
[TRAFODION-3059] remove "mt_" column family

traflm:
All Trafodion tables have a second column family named "mt_", which was originally designed for SSCC.
SSCC is not enabled now
SSCC plan to not use a new column family

So it is good to remove this extra, no use column family


traflm:
There is no way to remove the CF as far as I know.
It is harmless, the extra CF never contains value, so no HFile will be generated. 
It is an extra Store for HBase, I am not quite clear its impact yet.
But since it is never used, so I remove it here.

traflm:
jenkins, extra tests

PR #1551, opened 2018-05-07
[TRAFODION-3056] Change UPDATE STATISTICS to tolerate absence of RMS

DaveBirdsall:
When RMS is absent, UPDATE STATISTICS either fails with an internal error 9200 or cores.

This set of changes adds code to UPDATE STATISTICS so that it will ignore any 2024 warnings returned by the Executor due to RMS being absent.


zellerh:
+1
Looks good to me.

robertamarton:
Looks good.

PR #1553, opened 2018-05-08
[TRAFODION-1861] Run Trafodion on CentOS 7

traflm:
Solve the link issue when using GCC 4.8 on CentOS 7
So Trafodion can start on CentOS 7 with this change
It also depends on https://github.com/apache/trafodion/pull/1552 to make the code build success


selvaganesang:
Will the RH7 build be enabled on daily basis to ensure that new code change doesn't introduce build and run-time regressions with RH7?

I had made an attempt to build on RH7 earlier via #1260  , but the new code caused build failures again. So, it is good idea to introduce daily builds on RH7

traflm:
Yes, Selva, I created a JIRA to track the new code change, there is another PR to fix all current issues. But to enable new jenkins for RH7, I assume we need a working build first.
I plan to discuss this on mail list after these two changes merged.
Then before the jenkins setup, I plan to monitor the changes and tried on RH7 until Jenkins setup.

core/sqf/src/seabed/src/Makefile
```diff
@@ -55,6 +55,16 @@ LIBSBUTIL	= $(LIBEXPDIR)/libsbutil.so
 LIBSQSTATESB	= $(LIBEXPDIR)/libsqstatesb.so
 # neet -lrt for clock_gettime
 LIBSX	       += -lrt
+#
+# the order of these objects have to be swapped based on the GCC version
+MS_MPITMSG_O	= $(OUTDIR)/mpitmsg.o \
+		  $(OUTDIR)/ms.o 
+ifeq ($(GCC_MAJOR_VERSION_GT_3),1)
+ifeq ($(GCC_MINOR_VERSION_GT_4),1)
```
DaveBirdsall: I'm not sure this is what you want. For example, if there is ever a gcc version 5.1, the first "ifeq" will be true but not the second.

traflm: You are right Dave, so it should be GCC_MAJOR_VERSION_EQ_4 and GCC_MINOR_VERSION_GT_4 
We don't know how GCC 5.1 behave now, so let only check GCC 4.4 vs. GCC 4.8 at present. Let me change the logic more accurate
That is, we only change the link order for GCC 4.x , where x>4

core/sqf/src/seabed/src/Makefile
```diff
@@ -55,6 +55,16 @@ LIBSBUTIL	= $(LIBEXPDIR)/libsbutil.so
 LIBSQSTATESB	= $(LIBEXPDIR)/libsqstatesb.so
 # neet -lrt for clock_gettime
 LIBSX	       += -lrt
+#
+# the order of these objects have to be swapped based on the GCC version
+MS_MPITMSG_O	= $(OUTDIR)/mpitmsg.o \
+		  $(OUTDIR)/ms.o 
+ifeq ($(GCC_MAJOR_VERSION_EQ_4),1)
+ifeq ($(GCC_MINOR_VERSION_GT_4),1)
+MS_MPITMSG_O	= $(OUTDIR)/ms.o \
```
selvaganesang: If any release above 4.4 should use this order of .sos this change won't work either. You may consider change it as follows
By default, the order should be ms.o and mpitmsg.o, 
Change the order when it is less than 4.4.

traflm:
Thanks Selva, @DaveBirdsall do you still have concerns? I later need to check each gcc versions, the latest is GCC 8.1
That will take a lot of time. So I suggest we first support gcc 4.8, which CentOS/RH 7 , that latest CentOS7.5 is still using gcc 4.8.5, not newer GCC, so I think this change will serve Trafodion for a long time.

PR #1558, opened 2018-05-09
[TRAFODION-3066] Fix bug that causes 2053 warning on deeply nested subqueries

DaveBirdsall:
When a query has at least three levels of subquery, and when the subqueries are transformed from semi-joins to inner-joins, we may see an Optimizer assertion failure warning (2053) when preparing the query. This bug has been there for a long time but is more likely now due to the changes in https://github.com/apache/trafodion/pull/1530.

The problem is that when transforming a semi-join to an inner-join + group by, the new GroupByAgg node was getting its GroupAttributes as a copy of the child node. If the child node happened to also be a Join (as would happen with nested subqueries), the numJoinedTables_ field for the GroupByAgg's GroupAttributes would remain 2 or more. If during optimization, the GroupByOnJoinRule fired, pushing the GroupByAgg down, we end up with another Join expression in the same group. But having numJoinedTables_ being 2 or more in that Group would allow the LeftShiftJoinRule to fire, with the result that we might get multiple GroupBy criteria in the same join backbone. The Analysis.cpp module does not expect this expansion of the join back bone, and raises the assertion, which results in the 2053 warning.

The fix is that when Join::transformSemiJoin creates the new GroupByAgg node, it must reset numJoinedTables_ in its new GroupAttributes object to 1.

### DIFF
--- a/core/sql/optimizer/GroupAttr.h
+++ b/core/sql/optimizer/GroupAttr.h
@@ -812,8 +812,21 @@ private:
 
   // values used to avoid bushy join trees
   Int32          numBaseTables_;    // # of base tables involved in this subtree
-  Int32          numJoinedTables_;  // # of tables in join backbone
+  Int32          numJoinedTables_;  // # of tables in join backbone (see note below)
   Int32          numTMUDFs_;        // # of table-mapping UDFs in this subtree
+
+  // Note: For numJoinedTables_, we calculate this when we create the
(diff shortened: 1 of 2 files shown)


DaveBirdsall:
Note: I will be running full regressions overnight tonight on a workstation for this change. Please do not merge this until I report back on that.

DaveBirdsall:
FYI: Full regressions on a workstation passed for this change overnight last night. So this is OK to merge.

PR #1563, opened 2018-05-12
[TRAFODION-3015] retrieve a value from numeric type  get no result if…

traflm:
… using xx='value'

When the column is defined as NUMERIC(19,0), underlying data type will use FLOAT to save the number. 
So an implicit cast will generate wrong result when user input is string literal.
For example:
create table t3015 (a1 NUMERIC(19,0) );
insert into t3015 values(27380468);
select * from t3015 where a1='27380468';

will have wrong output.

The fix is to report error, if the precision of NUMERIC is greater than 18, and try to cast STRING into it.


traflm:
I will run full regression and fix all issues. But it will be very appreciated one can review the basic ideas. Thanks in advance!

PR #1564, opened 2018-05-13
[TRAFODION-3015] retrieve a value from numeric type get no result

traflm:
When the column is defined as NUMERIC(19,0), underlying data type will use FLOAT to save the number.
So an implicit cast will generate wrong result when user input is string literal.
For example:
create table t3015 (a1 NUMERIC(19,0) );
insert into t3015 values(27380468);
select * from t3015 where a1='27380468';

will have wrong output.

The fix is to report error, if the precision of NUMERIC is greater than 18, and try to cast STRING into it.

### DIFF
--- a/core/sql/optimizer/BindItemExpr.cpp
+++ b/core/sql/optimizer/BindItemExpr.cpp
@@ -2249,11 +2249,17 @@ static ItemExpr * ItemExpr_handleIncompatibleComparison(
     //3. Comparing date to numeric
     //4. Comparing interval to numeric
     //6. Comparing date to char of form: DD-MON-YYYY
-
+    NAString LiteralNumeric("NUMERIC");
     //check for numeric to character comparison
     if ((type1.getTypeQualifier() == NA_CHARACTER_TYPE) &&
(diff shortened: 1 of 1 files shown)


DaveBirdsall:
I don't believe this is the correct fix. I notice that if I do a "showplan select * from t3015 where a1=27380468" (without quotes around the number), the expression generated does not involve floats at all, rather it sticks to BIGNUM data types. I suspect that is what we should do in the '27380468' quoted string case as well: We should convert the string literal to a BIGNUM type then do BIGNUM comparison. @anoopsharma00, what do you think?

traflm:
You are right Dave, I didn't realize there is a BIGNUM data type. Let me check it. And very appreciated to hear from Anoop's advice.
Decimal will use string as storage, I was always thinking NUMERIC will use 8 bytes to store the value. So precision 19 is out of scope of 64bit. We may already save NUMERIC (19) as BIGNUM, I need to read more into the code.

traflm:
This is a wrong fix. Close it.

PR #1569, opened 2018-05-19
[TRAFODION-3075] Installer needs to support RH7 / Centos7

traflm:


traflm:
I will merge this since the RH7 jenkins build need this change.

PR #1570, opened 2018-05-19
[TRAFODION-3015] query no result when comparing string with BIGNUM

traflm:
When the column is BIGNUM, and the predicate is comparing to a string. Trafodion cast the string into FLOAT and the convert FLOAT into BIGNUM, 
There is an issue when convert FLOAT into BIGNUM that didn't considering the rounding:

so 2.73804679999999968E+007 will become 27380467 instead of 27380468


traflm:
jenkins, extra tests

traflm:
double check to run extra test

PR #1576, opened 2018-05-23
[TRAFODION-3083] Fix compile problem on CentOS7

DaveBirdsall:
Add -lpthread to list of libraries used for link.


zellerh:
+1
Looks good to me.

PR #1581, opened 2018-05-28
[TRAFODION-2901] Self-reference Holloween problem fix

Guhaiyan:
For query " insert into to t1 select seqnum(seq1, next) from t1;", there is no SORT as left child of TSJ, and it is a self-referencing updates Halloween problem.  In NestedJoin::genWriteOpLeftChildSortReq(), child(0) producing no outputs for this query, which means that there is no column to sort on. So we  solve this by having the source for Halloween insert produce at least one output column always.

### DIFF
--- a/core/sql/optimizer/RelExpr.cpp
+++ b/core/sql/optimizer/RelExpr.cpp
@@ -13342,7 +13342,9 @@ void GenericUpdate::pushdownCoveredExpr(const ValueIdSet &outputExpr,
 				predicatesOnParent,
 				&localExprs);
 
-  if (avoidHalloween() && child(0) && child(0)->getGroupAttr())
+  if (avoidHalloween() && child(0) &&
+      child(0)->getOperatorType() == REL_SCAN &&
+      child(0)->getGroupAttr())
     {
       ValueIdSet cur_output = child(0)->getGroupAttr()->getCharacteristicOutputs();
       if (cur_output.isEmpty())
--- a/core/sql/optimizer/RelExpr.cpp
+++ b/core/sql/optimizer/RelExpr.cpp
@@ -13342,12 +13342,17 @@ void GenericUpdate::pushdownCoveredExpr(const ValueIdSet &outputExpr,
 				predicatesOnParent,
 				&localExprs);
 
+/*to fix jira 18-20180111-2901  
+ *For query " insert into to t1 select seqnum(seq1, next) from t1;", there is no SORT as left child of TSJ, and it 
+ *is a self-referencing updates Halloween problem. In NestedJoin::genWriteOpLeftChildSortReq(), child(0)
+ *producing no outputs for this query, which means that there is no column to sort on. So we solve this by 
+ *having the source for Halloween insert produce at least one output column always.
+ * */
   if (avoidHalloween() && child(0) &&
       child(0)->getOperatorType() == REL_SCAN &&
       child(0)->getGroupAttr())
     {
-      ValueIdSet cur_output = child(0)->getGroupAttr()->getCharacteristicOutputs();
-      if (cur_output.isEmpty())
+      if (child(0)->getGroupAttr()->getCharacteristicOutputs().isEmpty())
         {
           ValueId exprId;
           ValueId atLeastOne;
@@ -13364,8 +13369,7 @@ void GenericUpdate::pushdownCoveredExpr(const ValueIdSet &outputExpr,
                   break;
                 }
             }
-         cur_output = child(0)->getGroupAttr()->getCharacteristicOutputs();
-         if (cur_output.isEmpty())
+         if (child(0)->getGroupAttr()->getCharacteristicOutputs().isEmpty())
            {
              child(0)->getGroupAttr()->addCharacteristicOutputs(atLeastOne);
            }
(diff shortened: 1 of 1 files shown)


traflm:
jenkins, add user

traflm:
@svarnau please help, I am not an admin

svarnau:
jenkins, add user

svarnau:
Hmm, I thought it should have worked for you, Ming (current whitelist users).

core/sql/optimizer/RelExpr.cpp
```diff
@@ -13341,6 +13345,33 @@ void GenericUpdate::pushdownCoveredExpr(const ValueIdSet &outputExpr,
...
+            {
+              atLeastOne = exprId;
+              if (!(exprId.getItemExpr()->doesExprEvaluateToConstant(FALSE, TRUE)))
+                {
+                  child(0)->getGroupAttr()->addCharacteristicOutputs(exprId);
+                  break;
+                }
+            }
+         cur_output = child(0)->getGroupAttr()->getCharacteristicOutputs();
+         if (cur_output.isEmpty())
```
DaveBirdsall: I don't understand this code path. If cur_output is empty, that means that the "for" loop did not find a ValueId to add to it. In particular. "atLeastOne" has its initial value, which I think is the null ValueId, 0. This "if" would cause us to add the null ValueId to child(0)'s characteristic outputs, which does not sound correct to me.

zellerh:
Could you explain (in the code or in the JIRA) why this fix solves the reported problem? The fix looks ok to me, but it would be good to understand what the underlying problem is. If we have to avoid the Halloween problem, do we need to retrieve the clustering key of the source table? If so, maybe that should be added to localExprs.

core/sql/optimizer/RelExpr.cpp
Guhaiyan: If cur_output is empty, it means that in RelExpr::pushdownCoveredExpr,  child(0).getPtr()->getGroupAttr()->computeCharacteristicIO set requiredOutputs_ to be empty, that is there is no output for Scan. So I go through the original_output and pick a ValueId which does NOT EvaluateToConst  and add it to the characteristic output of Scan. Since the original_output is not empty, so we can always find one ValueId to be added to Scan's output . After "for" loop, atLeastOne is the last ValueId in the original_output, not its initial value(0), so I do not think the "if" would cause adding an null ValueId in Scan's output.

DaveBirdsall: Ah, thank you! I misread the code. But is it possible that original_output itself is empty? If so, the "for" loop would not be entered and exprId would remain 0. Do we need a check for that?

Guhaiyan:
jenkins,retest

Guhaiyan:
The test error BatchTest.testBatchUpdateOverflow seems not be caused by this commit , instead it looks much more like the test env problem. would you please check for that?

svarnau:
No, it causes mxosrvr to core. That is not environment problem.
http://traf-testlogs.esgyn.com/PullReq/1581/2715/jdbc_test-cdh-rh6/corefiles.log

sandhyasun:
Looks like the test has passed now. If you can reply to @zellerh about his general question on how this solves the original problem, then we should be able to merge this in.

Guhaiyan:
@svarnau I updated my first comment about 2 weeks ago to answer @zellerh 's question. But maybe I updated it in the wrong place so that you did not notice that. So I paste my explanations below.

For query " insert into to t1 select seqnum(seq1, next) from t1;", there is no SORT as left child of TSJ, and it is a self-referencing updates Halloween problem. In NestedJoin::genWriteOpLeftChildSortReq(), child(0) producing no outputs for this query, which means that there is no column to sort on. So we solve this by having the source for Halloween insert produce at least one output column always.

zellerh:
You are right, I did not see your earlier reply with the explanation. Please add this to the code, because I think it helps to understand the fix. I looked at method NestedJoin::genWriteOpLeftChildSortReq(), and that code has an if statement marked "fix for soln 10-090826-4155". I assume that we get into this if statement and that we want to make sure that this if statement will produce an output.

It looks like your fix is only needed when the underlying table has a SYSKEY? You could add that check, but just adding the comment will help tremendously, since people can then find the place where the generated output is needed.

core/sql/optimizer/RelExpr.cpp
```diff
@@ -13341,6 +13341,36 @@ void GenericUpdate::pushdownCoveredExpr(const ValueIdSet &outputExpr,
 				newExternalInputs,
 				predicatesOnParent,
 				&localExprs);
+
+  if (avoidHalloween() && child(0) &&
+      child(0)->getOperatorType() == REL_SCAN &&
+      child(0)->getGroupAttr())
+    {
+      ValueIdSet cur_output = child(0)->getGroupAttr()->getCharacteristicOutputs();
```
zellerh: This works, but is is not very efficient to create a copy of a ValueIdSet just to check whether it is empty. You could just check

```
if (child(0)->getGroupAttr()->getCharacteristicOutputs().isEmpty())
```
in the line below.

core/sql/optimizer/RelExpr.cpp
```diff
@@ -13341,6 +13341,36 @@ void GenericUpdate::pushdownCoveredExpr(const ValueIdSet &outputExpr,
...
+               output_source.advance(exprId))
+            {
+              atLeastOne = exprId;
+              if (!(exprId.getItemExpr()->doesExprEvaluateToConstant(FALSE, TRUE)))
+                {
+                  child(0)->getGroupAttr()->addCharacteristicOutputs(exprId);
+                  break;
+                }
+            }
+         cur_output = child(0)->getGroupAttr()->getCharacteristicOutputs();
```
zellerh: Same comment as above, on line 13349.

zellerh:
+1
Thanks for adding the comment.

PR #1584, opened 2018-05-29
[TRAFODION-3088] Fix data conversion issue with INTERVAL literals

DaveBirdsall:
ValueIdList::convertToTextKey (optimizer/ValueDesc.cpp) in some cases was encountering INTERVAL literal values containing the text "INTERVAL ... < qualifier >". The code expects this text to be there in some code paths but not all. With this change, I've taken care of the remaining code paths. So we should no longer see data conversion errors in the cases noted in the JIRA.

### DIFF
--- a/core/sql/optimizer/ValueDesc.cpp
+++ b/core/sql/optimizer/ValueDesc.cpp
@@ -6382,7 +6382,32 @@ void ValueIdList::convertToTextKey(const ValueIdList& keyList, NAString& result)
 	 {
            short vLen = val.length();
 
-	   if ((constType->getTypeQualifier()  == NA_NUMERIC_TYPE) &&
+	   if (constType->getTypeQualifier() == NA_INTERVAL_TYPE)
+	     {
+	       // In some code paths, the text may have "INTERVAL 'xxx' <qualifier>"
+	       // junk around it so we have to strip that off. (Example: An equality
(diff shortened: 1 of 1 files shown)


DaveBirdsall:
I am running a full regressions run on this change on a workstation overnight. Please do not merge until I report back with those results.

core/sql/optimizer/ValueDesc.cpp
```diff
@@ -6382,7 +6382,32 @@ void ValueIdList::convertToTextKey(const ValueIdList& keyList, NAString& result)
...
+	       // that example.)
+	       Lng32 start = val.index("'");
+	       Lng32 minus = val.index("-");
+	       if (start > 0)
+	         {
+	           Lng32 end = val.index("'", start+1);
+	           if (end > 0)
+	             {
+	               val = val(start+1, (end-start-1));
+	               if (minus > 0)
```
zellerh: Should this also check for minus < start? Could we encounter year-month intervals with a '-' inside the literal, such as interval '1-1' year to month?

DaveBirdsall: Indeed, you are correct! Good catch. Will fix.

zellerh:
+1
Looks good to me.

PR #1585, opened 2018-05-30
[TRAFODION 3091] Unnamed Dynamic parameter doesn't work in WITH state…

andyyangcn:
…ment

### DIFF
--- a/core/sql/optimizer/ItemExpr.cpp
+++ b/core/sql/optimizer/ItemExpr.cpp
@@ -3886,23 +3886,7 @@ ItemExpr * Parameter::copyTopNode(ItemExpr *derivedNode, CollHeap* outHeap)
 
 ItemExpr * DynamicParam::copyTopNode(ItemExpr *derivedNode, CollHeap* outHeap)
 {
-  ItemExpr *result;
-
-  if (derivedNode == NULL) {
-    result = new (outHeap) DynamicParam(paramName_, indicatorName_, outHeap);
-    ((DynamicParam *) result)->setRowsetSize(rowsetSize_);
-    ((DynamicParam *) result)->setRowsetInfo(rowsetInfo_);
-    ((DynamicParam *) result)->setParamHeading(heading_);
-    ((DynamicParam *) result)->setParamTablename(tablename_);
-    // we remember our original dynamic parameter because we
-    // must use their valueid at dynamicparam::codegen time
-    ((DynamicParam *) result)->setOriginal(this);
-  }
-
-  else
-    result = derivedNode;
-
-  return Parameter::copyTopNode(result, outHeap);
+  return this;
 }
 
 const NAType * DynamicParam::pushDownType(NAType& desiredType,
--- a/core/sql/optimizer/BindItemExpr.cpp
+++ b/core/sql/optimizer/BindItemExpr.cpp
@@ -8622,7 +8622,12 @@ ItemExpr *CurrentTimestampRunning::bindNode(BindWA *bindWA)
 ItemExpr *Parameter::bindNode(BindWA *bindWA)
 {
   if (nodeIsBound())
+  {
+    OperatorTypeEnum opTyp = getOperatorType();
+    // All user inputs are treated as outer references in the current scope.
+    bindWA->getCurrentScope()->addOuterRef(getValueId());
     return getValueId().getItemExpr();
+  }
   if (bindWA->getCurrentScope()->context()->inTDFunction())
   {
     //Paramaters and outer references are not supported with rank function.


core/sql/optimizer/ItemExpr.cpp
```diff
@@ -3886,23 +3886,7 @@ ItemExpr * Parameter::copyTopNode(ItemExpr *derivedNode, CollHeap* outHeap)
...
-    // we remember our original dynamic parameter because we
-    // must use their valueid at dynamicparam::codegen time
-    ((DynamicParam *) result)->setOriginal(this);
-  }
-
-  else
-    result = derivedNode;
-
-  return Parameter::copyTopNode(result, outHeap);
+  return this;
```
DaveBirdsall: I do not believe this is correct. The copyTopNode method is used to make a copy of a given node, sometimes on a different heap than the original. (For example, we might be copying a bound query tree from the statement heap which is destroyed at the end of statement compilation to the query cache heap which lives until the end of the process.)

DaveBirdsall: @zellerh, can you comment on copyTopNode? Is my understanding correct?

zellerh: +1

Andy and I discussed this and I like the fix. There are certain ItemExprs like parameters and BaseColumns that should never be copied. We may have others, like bound aggregate functions, constants, cuts that also should not be copied. The copyTopNode and copyTree methods were initially used in very restricted situations, but we now use them in many more places and some issues creep in. I think this fix is an improvement over the existing solution. Someday we should probably fix BaseColumn::copyTopNode in the same way, I doubt that the current implementation works correctly.

My definition of copyTopNode is that it should create a new copy of a tree that can be modified and bound _where it is semantically meaningful_, but things like parameters and base columns refer to and represent (with their value id) existing objects that cannot be copied. At least, they cannot be copied as a side-effect of copyTopNode().

Examples: In this particular query, we want to copy an unbound tree as part of the WITH clause handling. IMHO, sharing the parameter between the two copies is the right thing to do.

Another example would be making a copy of an unbound aggregate in such a context (WITH clause parsing), which is probably ok. Making a copy of a bound aggregate in a later phase would be incorrect, however, because that aggregate would not be in the aggregate expression of any groupby.

A similar situation exists with the copy of a BaseColumn node. Its new ValueId would not appear in the column list of any TableDescriptor, so it would not be meaningful. Copying the unbound form of a BaseColumn, a ColReference, is ok.

Sorry for the long answer. We probably need to do more work in this area and define more clearly what we mean by copying trees.

DaveBirdsall: Thanks, @zellerh. I take it then that when we copy trees from the Statement heap to the Context heap we use something different than copyTopNode? Otherwise it's not clear to me how dangling pointers are avoided.

zellerh: Good question, and I'm not sure I know the answer to that. It probably is not possible to take trees that contain things like parameters, bound aggregate nodes and BaseColumn nodes and make a copy that is valid.

Another thing I should have mentioned: I tried to remember what my original design goal was when implementing copyTopNode. Now I think I remember - I said it will just make a copy of an object with all data members, regardless of whether that is meaningful or not. It would be the responsibility of the caller to fix up things that are invalid in a copy. I don't like that design anymore, given that we use copyTopNode() and copyTree() in so many places.

Something to be aware of is that these methods aren't generally usable, one needs to be careful when calling them. Here are some cases that should generally work:

- Copying the before pattern in an optimizer rule (note that this does not copy the "cut" nodes, nor does it copy any ItemExpr nodes)
- Copying an unbound RelExpr and ItemExpr tree in the parser (e.g. for WITH clause processing)
- Copying ItemExprs that can be evaluated at compile time, consisting only of constants and ItemExpr operators, with no aggregates, parameters or other special ItemExprs in them like current_time, sequence functions, etc.
- I'm sure there are more we could add to this list

DaveBirdsall: I did a quick experiment. I put a breakpoint at DynamicParam::copyTopNode in sqlci, then did some prepares using dynamic parameters. I wanted to see if this method was called to copy a tree into the query cache. I didn't get any hits. So my tentative conclusion is that this method isn't used for such copies, or perhaps plans with dynamic parameters are not cacheable (which would surprise me). In either case though the change looks safe. So I'm OK with it.

DaveBirdsall:
Tests failed due to a (hopefully) transient failure in the Phoenix tests. Will rerun them.

DaveBirdsall:
jenkins, retest

core/sql/optimizer/ItemExpr.cpp
zellerh: As far as I understand, the query cache uses a cache key that is a string and it stores the generated object as a binary. We should not use copyTopNode to populate the query cache.

andyyangcn:
Hi Dave & Hans, thanks a lot for your questions and answers. Learnt a lot from you discussion.

PR #1586, opened 2018-05-30
[TRAFODION-2946] Support to provide a locking mechanism for LOB insert/update operations

sandhyasun:
This is an interim solution to provide a locking mechanism for LOB operations. See JIRA for a document on the issue and the details.

### DIFF
--- a/core/sql/cli/Context.cpp
+++ b/core/sql/cli/Context.cpp
@@ -3163,6 +3163,59 @@ Lng32 ContextCli::setSecInvalidKeys(
 
 }
 
+Int32 ContextCli::checkLobLock(char *inLobLockId, NABoolean *found)
+{
+  Int32 retcode = 0;
+  *found = FALSE;
+  CliGlobals *cliGlobals = getCliGlobals();
+  StatsGlobals *statsGlobals = GetCliGlobals()->getStatsGlobals();
+  if (cliGlobals->getStatsGlobals() == NULL)
+  {
+    (diagsArea_) << DgSqlCode(-EXE_RTS_NOT_STARTED);
+    return diagsArea_.mainSQLCODE();
+  }
+  statsGlobals->checkLobLock(cliGlobals,inLobLockId);
+  if (inLobLockId != NULL)
+    *found = TRUE;
+  return retcode;
+}
+Lng32 ContextCli::setLobLock(
+     /* IN */    char *lobLockId // objID+column number
+                             )
+{
+  CliGlobals *cliGlobals = getCliGlobals();
+  if (cliGlobals->getStatsGlobals() == NULL)
+  {
+    (diagsArea_) << DgSqlCode(-EXE_RTS_NOT_STARTED);
+    return diagsArea_.mainSQLCODE();
+  }
+  ComDiagsArea *tempDiagsArea = &diagsArea_;
+  tempDiagsArea->clear();
+ 
+  IpcServer *ssmpServer = ssmpManager_->getSsmpServer(exHeap(),
+                                 cliGlobals->myNodeName(), 
+                                 cliGlobals->myCpu(), tempDiagsArea);
+  if (ssmpServer == NULL)
+    return diagsArea_.mainSQLCODE();
+
+  SsmpClientMsgStream *ssmpMsgStream  = new (cliGlobals->getIpcHeap())
+        SsmpClientMsgStream((NAHeap *)cliGlobals->getIpcHeap(), 
+                            ssmpManager_, tempDiagsArea);
+  ssmpMsgStream->addRecipient(ssmpServer->getControlConnection());
+  LobLockRequest *llMsg = 
+    new (cliGlobals->getIpcHeap()) LobLockRequest(
+                                      cliGlobals->getIpcHeap(), 
+                                      lobLockId);
+  *ssmpMsgStream << *llMsg;
+  // Call send with no timeout.  
+  ssmpMsgStream->send(); 
+  // I/O is now complete.  
+  llMsg->decrRefCount();
+  cliGlobals->getEnvironment()->deleteCompletedMessages();
+  ssmpManager_->cleanupDeletedSsmpServers();
+  return diagsArea_.mainSQLCODE();
+
+}
 ExStatisticsArea *ContextCli::getMergedStats(
             /* IN */    short statsReqType,
             /* IN */    char *statsReqStr,
--- a/core/sql/cli/Context.cpp
+++ b/core/sql/cli/Context.cpp
@@ -3184,17 +3184,50 @@ Lng32 ContextCli::setLobLock(
                              )
 {
   CliGlobals *cliGlobals = getCliGlobals();
+  NABoolean releasingLock = FALSE;
   if (cliGlobals->getStatsGlobals() == NULL)
   {
     (diagsArea_) << DgSqlCode(-EXE_RTS_NOT_STARTED);
     return diagsArea_.mainSQLCODE();
   }
   ComDiagsArea *tempDiagsArea = &diagsArea_;
+  IpcServer *ssmpServer = NULL;
   tempDiagsArea->clear();
- 
-  IpcServer *ssmpServer = ssmpManager_->getSsmpServer(exHeap(),
-                                 cliGlobals->myNodeName(), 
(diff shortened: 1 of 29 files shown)


core/sql/executor/ExExeUtil.h
```diff
@@ -3083,6 +3083,7 @@ class ExExeUtilLobUpdateTcb : public ExExeUtilTcb
   ExLobStats lobStats_;
   char statusString_[200];
   fstream indata_;
+  char lobLockId_[12];
```
DaveBirdsall: Should "12" be "LOB_LOCK_ID_SIZE"?

core/sql/executor/ExExeUtilLoad.cpp
```diff
@@ -3459,6 +3458,25 @@ short ExExeUtilLobUpdateTcb::work()
...
             Int32 outHandleLen;
             Int64 requestTag = 0;
+            if (lobTdb().lobLocking())
+              {
+                ExpLOBoper::genLobLockId(uid,lobNum,lobLockId_);
+                NABoolean found = FALSE;
+                retcode = SQL_EXEC_CheckLobLock(lobLockId_ , &found);
+                if (! retcode && !found) 
+                  {    
+                    retcode = SQL_EXEC_SetLobLock(lobLockId_);
```
DaveBirdsall: Might this call fail? It looks like we ignore the retcode in this code path.

core/sql/cli/Context.cpp
```diff
@@ -3163,6 +3163,59 @@ Lng32 ContextCli::setSecInvalidKeys(
...
+        SsmpClientMsgStream((NAHeap *)cliGlobals->getIpcHeap(), 
+                            ssmpManager_, tempDiagsArea);
+  ssmpMsgStream->addRecipient(ssmpServer->getControlConnection());
+  LobLockRequest *llMsg = 
+    new (cliGlobals->getIpcHeap()) LobLockRequest(
+                                      cliGlobals->getIpcHeap(), 
+                                      lobLockId);
+  *ssmpMsgStream << *llMsg;
+  // Call send with no timeout.  
+  ssmpMsgStream->send();
```
DaveBirdsall: Is this a "datagram" protocol? That is, do we wait for some acknowledgement from the SSMP that our message was processed? Or do we just continue on? Do we assume that we have the lock once send() completes?

core/sql/generator/GenPreCode.cpp
```diff
@@ -4520,7 +4520,9 @@ RelExpr * GenericUpdate::preCodeGen(Generator * generator,
     {
       oltOptInfo().setOltOpt(FALSE);
       generator->oltOptInfo()->setOltOpt(FALSE);
-      generator->setAqrEnabled(FALSE);
+      //enabling AQR to take care of the lock conflict error 8558 that
+      // should be retried.
+      //      generator->setAqrEnabled(FALSE);
```
DaveBirdsall: How would AQR work for an INSERT/SELECT of one table into another where a LOB column is being copied?

sandhyasun:
Answers to general questions:
All these are addressed by the RMS infrastructure itself. Query invalidation uses this mechanism too to set security invalidation keys. This is a similar model. 

Are Lob Locks local to a node? (I didn't see logic to propagate locks to all nodes in the cluster.) If so, how does a client know which node to talk to for a given Lob?
     They are propagated. We use the mxsscp processes running on every node to propagate the lock. The executor sends a message to the mxssmp on the local node. In turn the mxssmp propagates the same lock to all mxsscps and the lock will get set in the shared segment. (see ssmpipc.cpp change ::actOnLobReq which sends the message to all sscps on each node  and sscpipc.cpp : processLobReq which sets the flag in the shared segment)  So to check the lock , the CLI can simply check if the lock is set in it's own nodes shared segment. (SQL_EXEC_CheckLobLock) 
Is Lob lock acquisition synchronous?
What happens if two clients on the same node ask for a lock on the same Lob at the same time? Are we guaranteed that one of them will get the lock and the other won't?

Yes the RMS mechanism to set and send the message to mxssmp and mxsscps and setting in the shared segment is done under a semaphore. 

It wasn't clear to me when locks are released. Do we wait until end of transaction to release the locks?
They are released after every I/O - see ExpLOB.cpp   and the call SQL_EXEC_ReleaseLobLock.

core/sql/executor/ExExeUtil.h
sandhyasun: Yes. Changing that.

core/sql/executor/ExExeUtilLoad.cpp
sandhyasun: Will add the error condition for non zero retcode too.

core/sql/cli/Context.cpp
sandhyasun: Yes that's how it's "supposed" to work and we are indeed replying on all this working well jsut like we do for Query Invalidation.  More testing and error and crash scenarios may need to be addressed in a future checkin.

core/sql/generator/GenPreCode.cpp
sandhyasun: The while insert would get retried again. The insert/select operation of lob columns treats the input to the  target table as a char string and does it in chunks. If there is lock clash, the transaction gets aborted and we retry.

DaveBirdsall:
So, suppose two clients attempt to lock the same LOB, and the clients are on different nodes. Suppose they do so at the same time. Each checks its local RMS segment and finds no lock. So each sends a message to the SSMP locking the LOB. I assume at that point, each client thinks it has the lock. Now the SSMPs will begin to propagate. And when each receives the other's lock request, will the conflict be detected? Will the transaction be doomed at that point?

sandhyasun:
I assume since the check also has to get the semaphore,  only one of them wins. See StatsGlobals::checkLobLock.

DaveBirdsall:
But the semaphore is local to a node, right? In this example, the clients are on different nodes.

core/sql/runtimestats/sscpipc.cpp
```diff
@@ -907,3 +910,53 @@ void SscpNewIncomingConnectionStream::processSecInvReq()
...
+  ex_assert( !moreObjects(), "unknown object follows LobLockRequest.");
+  SscpGlobals *sscpGlobals = getSscpGlobals();
+  StatsGlobals *statsGlobals = sscpGlobals->getStatsGlobals();
+  int error = statsGlobals->getStatsSemaphore(sscpGlobals->getSemId(),
+                  sscpGlobals->myPin());
+  statsHeap = statsGlobals->getStatsHeap();
+  char *ll = new (statsHeap) char [LOB_LOCK_ID_SIZE];
+  memcpy(ll,request->getLobLockId(),LOB_LOCK_ID_SIZE+1);
+  SyncHashQueue *lobLockList = statsGlobals->getLobLocks();
+  if (ll[0] == '+') // If it's a positive value, we are supposed to insert it.
```
sandhyasun: To address the scenario where 2 clients on different nodes check for a lock and being the process of setting it, instead of simply setting/inserting into the HashQueue, we should do an extra check for existence here before setting . If it's already there, some other process on another node may have propagated the same lockLockId first . So we should send a diagsArea back here to indicate the set  for this client/process failed.

selvaganesang: I would suggest the following to avoid this race condition.

1)  Obtain a node id based on the lock key. It  could be as simple as hash of the lock key % number of nodes in the cluster
2)  Send a message to SSMP to obtain a lock if it is not locked already.  Let SSMP send the lock info to other nodes via SSCP
3)  Let SQL process check if it is already locked via the shared segment, If the lock exists already , poll till the lock is released or pre-determined timeout expires, When the lock is released send a message to SSMP pertaining to lock key to lock,
 4) If this request returns if it is locked already, continue to poll locally to check the lock is released as in 3 to take care of two waiters waking up at the same time to obtain lock.

However this would cause all SQL processes to have opens to all SSMPs in the cluster.
Also, this concept of locking via RMS can overwhelm the RMS infrastructure. So, we should have mechanism to disable it if needed.

sandhyasun: Made all the needed changes to support the synchronization

PR #1588, opened 2018-05-31
[TRAFODION-3096] clip the varchar type data from mxosrvr to jdbct4

partnation:
When fetching varchar type data from SQL engine, mxosrvr will apply for enough memory to ensure that the data can be stored, regardless of the actual varchar length.Then send such a large piece of data to jdbct 4.This results in unnecessary network overhead and JDBC memory usage.
So I added a function to clip the resulting data. The data to be sent to JDBCT4 in mxosrvr is trimmed to reduce network overhead and memory consumption.This function cuts out the redundant parts of varchar type data, and is very useful for the case where varchar is very long but actually occupies very little.
I added a switch to jdbct4 code to turn this function on and off according to the situation.
I tested my code with phx

### DIFF
--- a/core/conn/jdbcT4/src/main/java/org/trafodion/jdbc/t4/CONNECTION_CONTEXT_def.java
+++ b/core/conn/jdbcT4/src/main/java/org/trafodion/jdbc/t4/CONNECTION_CONTEXT_def.java
@@ -35,6 +35,7 @@ class CONNECTION_CONTEXT_def {
 	short autoCommit;
 	short queryTimeoutSec;
 	short idleTimeoutSec;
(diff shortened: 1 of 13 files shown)


DaveBirdsall:
jenkins, add user

DaveBirdsall:
Thanks for your contribution. Could you supply a JIRA for this change?

DaveBirdsall:
jenkins, retest

sandhyasun:
Looks good to me. @arvind-narain,   if you can give a +1, I'll merge this one.

sandhyasun:
Merging this  now.

PR #1592, opened 2018-06-02
TRAFODION-3025 support order by with limit N in subquery

traflm:
ORDER BY in subquery is ignored by Trafodion, when in the subquery, there is a limit N, it is good to keep the order by:

select * from t ( select * from t order by c limit 20);

for example


traflm:
jenkins, extra tests

traflm:
test failure not related. I will merge this.

PR #1593, opened 2018-06-04
TRAFODION-3086 Traf support for DDL operations on Hive objects

anoopsharma00:
-- Support for TRAFODION-3086. Details in document attached to jira.

Other changes:
-- support for "if not exists", "if exists" clause for create/drop view

-- Support for:  truncate T, truncate table T.
  -- same as purgedata

-- showddl <tab>, detail

-- unregister hive schema <sch>
  -- will unregister all objects in specified schema

-- Support for "if not registered", "if registered" clauss for
   register/unregister command.


anoopsharma00:
jenkins retest

PR #1594, opened 2018-06-06
[TRAFODION 3091] revert the change provisionally for test case issue.

andyyangcn:


zellerh:
+1
Looks good to me.
Sorry for encouraging you to do this fix, which is causing trouble (not entirely unexpected). I hope you can eventually solve the problems and restore your fix.

PR #1597, opened 2018-06-09
Trafodion 3084

gdgy:


sandhyasun:
@anoopsharma00  Does this look ok ? Looks ok to me and can merge with your +1.

anoopsharma00:
Were testcases added to regressions to test this feature?
Would this work for hbase "_ROW_", "_CELL_", "_MAP_" formats? They all have HBASE as catalog name.
Would 'drop hbase table' command remove added comments? 
Would unregister command remove comments? It will remove the added objectuid for that hbase table.
It will also be better to use 'if not registered' instead of 'if not exists' construct. They both are supported.

gdgy:
@anoopsharma00 thank you for your review, I have modiy and test my code, but I have some questions.
when I register a HBase native table use MAP format ,error message is the object dose not exist
>>register hbase table hbase."_CELL_"."t06194";

--- SQL operation complete.
>>register hbase table hbase."_ROW_"."t06194";

--- SQL operation complete.
>>register hbase table hbase."_MAP_"."t06194";

*** ERROR[3251] REGISTER operation could not be completed. Reason: Specified object t06194 does not exist.

Is my syntax is worng? Thank you!

traflm:
jenkins, retest

core/sql/sqlcomp/CmpSeabaseDDLcommentOn.cpp
```diff
@@ -289,6 +289,31 @@ void  CmpSeabaseDDL::doSeabaseCommentOn(StmtDDLCommentOn   *commentOnNode,
...
+      }
+      else
+      {
+          //HBASE erternal table already register in create time no need to register again
+          NAString ddl = "REGISTER INTERNAL HBASE TABLE IF NOT REGISTERED ";
+          ddl.append(extObjName.data());
+
+          Lng32 cliRC = cliInterface.executeImmediate(ddl.data());
+          // error code -8102 means the object already register
+          if (cliRC < 0 && cliRC != -8102)
```
anoopsharma00: Since the ddl issued is 'register ...if not registered', an 8102 error will
not be returned if object already exists.
This check should not be done and all errors should be returned.

core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
```diff
@@ -10558,7 +10558,12 @@ short CmpSeabaseDDL::unregisterNativeTable
      objectNamePart.data(),
      objType
      );
-  
+
+  // drop comment text
+  retcode = deleteFromTextTable(&cliInterface,
+                                objUID,
+                                ComTextType::COM_OBJECT_COMMENT_TEXT,
+                                0);
```
anoopsharma00: if retcode is < 0, it should return -1 indicating an error.

PR #1604, opened 2018-06-12
Trafodion 2884

zcorrea:
Implements: https://issues.apache.org/jira/browse/TRAFODION-2884


core/sqf/sqenvcom.sh
```diff
@@ -673,17 +673,31 @@ export SQ_LUNMGR_VERBOSITY=1
...
+#   NAME-SERVER - to disable process replication and enable the name-server
+#
+# Uncomment the next environment variable
+#export SQ_NAMESERVER_ENABLED=1
+if [[ "$SQ_NAMESERVER_ENABLED" == "1" ]]; then
+  export NS_COMM_PORT=23370
+  export NS_SYNC_PORT=23360
+  export NS_M2N_COMM_PORT=23350
+  export MON2MON_COMM_PORT=23340
+fi
```
svarnau: PORT numbers should not be hard-coded here. I think most default ports are coded elsewhere, but these need to at leaset be coded to allow over-ride, such as:
export NS_COMM_PORT=${NS_COMM_PORT:-23370}

zcorrea: Is the '-' part of the syntax?

svarnau: Yes

zcorrea: I'll fix this. Thanks!

sureshsubbiah:
+1

PR #1606, opened 2018-06-13
[TRAFODION-3105] DatabaseMetaData.getColumns should return 24 columns

mashengchen:
odbc & jdbc have two rules, odbc return 18 columns while jdbc needs 24 columns, so there use two case to separate.
also we don't support java.sql.types.REF, so value of the new added columns are just null.


core/sqf/sqenvcom.sh
```diff
@@ -40,12 +40,12 @@ export TRAFODION_VER="${TRAFODION_VER_MAJOR}.${TRAFODION_VER_MINOR}.${TRAFODION_
...
 ##############################################################
 # Trafodion authentication:
-#    Set TRAFODION_ENABLE_AUTHENTICATION to YES to enable
+#    Set TRAFODION_ENABLE_AUTHENTICATION=NO
 # authentication in the Trafodion environment.
 # LDAP configuration must also be setup--see
 # $TRAF_HOME/sql/scripts/traf_authentication_config for details.
 ##############################################################
-export TRAFODION_ENABLE_AUTHENTICATION=${TRAFODION_ENABLE_AUTHENTICATION:-NO}
+export TRAFODION_ENABLE_AUTHENTICATION=NO
```
venkat1m: Was this an accidental commit for sqenvcom.sh where you disabled authentication to test your other changes?

export TRAFODION_ENABLE_AUTHENTICATION=${TRAFODION_ENABLE_AUTHENTICATION:-NO}  is what we need

mashengchen: sorry, wrong for git add

sureshsubbiah:
+1. Please consider adding some sample output to the JIRA.

PR #1607, opened 2018-06-13
[TRAFODION-2668] fix nid been accidentally deleted

mashengchen:
in previous [TRAFODION-2668] , there deleted nid accidentally, this result when mxosrvr restart, there will do a whole sqlcheck instead of node check


sandhyasun:
+1 Merging...

PR #1608, opened 2018-06-13
 jira TRAFODION-1782 Auto assignment of NOT NULL attr to primary key cols

anoopsharma00:
details in jira.


core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
```diff
@@ -8047,14 +8149,36 @@ void CmpSeabaseDDL::alterSeabaseTableAddPKeyConstraint(
...
-      return;
+      // remove NATable so current definition could be loaded
+      ActiveSchemaDB()->getNATableDB()->removeNATable
+        (cn,
+         ComQiScope::REMOVE_FROM_ALL_USERS, 
+         COM_BASE_TABLE_OBJECT, 
+         alterAddConstraint->ddlXns(), FALSE);
+      
+      // copy tempTable data into newly created table
+      str_sprintf(buf, "insert with no rollback into %s select * from %s",
```
zellerh: A general question, this also happens in the existing code: When we make a copy of a table in a DDL operation, do we need to make sure that nobody writes data into the table? If there are inserts or updates, what would happen? Could the DDL operation fail with a conflict, or would we potentially lose that data?

zellerh:
Thanks, I wanted this feature for many years, now you made it happen :-)

PR #1610, opened 2018-06-15
[TRAFODION-3109] comment can not transfer single quote normally in co…

zlei929:
When comment use a quote in the string, we need to use single quote to transfer the quote.


zellerh:
+1
Looks good to me.

PR #1613, opened 2018-06-19
TRAFODION-3111 CentOS 7 build error

traflm:
Some new code checkin broke the build on CentOS 7

I will continously monitor the CentOS 7 build until the jenkins setup.
CentOS 7 jenkins test now failed on some UDR regression, it is hard to find out root cause yet, still need some time.

### DIFF
--- a/core/sqf/monitor/linux/cluster.cxx
+++ b/core/sqf/monitor/linux/cluster.cxx
@@ -42,6 +42,7 @@ using namespace std;
 #include <sys/resource.h>
 #include <errno.h>
 #include <limits.h>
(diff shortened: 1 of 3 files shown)


traflm:
jenkins, retest

traflm:
@svarnau could you help to review this change?

I am still not finding the root cause of CentOS 7 regression failure in UDR, this change is not related, but it is required to make a CentOS 7 build.

I need to study how C++ TMUDF works, and try to figure out the regression issue, then we can setup the Jenkins test, and no longer need this kind of fixes.

PR #1619, opened 2018-06-25
[TRAFODION-3101] enhance get command to show rows get

traflm:
At field, engineers issue get tables to list all tables, and want to count how many tables got, it will be good to show that info by 'get' command itself.
The current implementation is to show that count as part of get command output, just as header info.

### DIFF
--- a/core/sql/executor/ExExeUtilGet.cpp
+++ b/core/sql/executor/ExExeUtilGet.cpp
@@ -3346,14 +3346,13 @@ short ExExeUtilGetMetadataInfoTcb::work()
 
 	case DONE_:
 	  {
-            if (NOT getMItdb().noHeader())
+            if (NOT getMItdb().noHeader() && getReturnRowCount() > 0)
             {
               short rc = 0;
               char returnMsg[256];
               memset(returnMsg, 0, 256);
-              sprintf(returnMsg, "\n=======================\nSummary info of get:\nGet %d rows.", getReturnRowCount());
+              sprintf(returnMsg, "\n=======================\n %d row(s) returned", getReturnRowCount());
               moveRowToUpQueue(returnMsg, strlen(returnMsg), &rc);
-              moveRowToUpQueue(" ");
             }
 
 	    retcode = handleDone();
@@ -3907,14 +3906,13 @@ short ExExeUtilGetMetadataInfoComplexTcb::work()
 
 	case DONE_:
 	  {
-            if (NOT getMItdb().noHeader())
+            if (NOT getMItdb().noHeader() && getReturnRowCount() > 0)
             {
               short rc = 0;
               char returnMsg[256];
               memset(returnMsg, 0, 256);
-              sprintf(returnMsg, "\n=======================\nSummary info of get:\nGet %d rows.", getReturnRowCount());
+              sprintf(returnMsg, "\n=======================\n %d row(s) returned", getReturnRowCount());
               moveRowToUpQueue(returnMsg, strlen(returnMsg), &rc);
-              moveRowToUpQueue(" ");
             }
 	    retcode = handleDone();
 	    if (retcode == 1)
--- a/core/sql/executor/ExExeUtilGet.cpp
+++ b/core/sql/executor/ExExeUtilGet.cpp
@@ -1851,6 +1851,7 @@ short ExExeUtilGetMetadataInfoTcb::work()
 	    headingReturned_ = FALSE;
 
 	    numOutputEntries_ = 1;
+            returnRowCount_ = 0 ;
 
 	    objectUid_[0] = 0;
 	  }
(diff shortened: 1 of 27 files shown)


core/sql/executor/ExExeUtilGet.cpp
```diff
@@ -949,6 +949,7 @@ short ExExeUtilGetMetadataInfoTcb::displayHeading()
 {
   if (getMItdb().noHeader())
     {
+printf("LMDBG noheader is true\n");
```
traflm: sorry, I will remove this...

core/sql/regress/compGeneral/EXPECTED023
```diff
@@ -23,6 +23,10 @@ SB_HISTOGRAMS
 SB_HISTOGRAM_INTERVALS
 SB_PERSISTENT_SAMPLES
 
+=======================
```
venkat1m: Would it be better to change the GET output like SELECT where we say "n row(s) returned" where n >=0

anoopsharma00:
select * from (get tables in schema sch) x(a);

will return the count of number of tables returned .

core/sql/regress/compGeneral/EXPECTED023
traflm: I agree, this looks much better!

traflm: Let me change it step by step.

anoopsharma00:
What was the decision on what the final row count will be displayed as? Is it 'n rows selected' as Venkat pointed out?

Does this change work for all GET commands including complex GET stmts which internally call get work method multiple times? method ExExeUtilGetMetadataInfoComplexTcb::work

It seems that count is kept in a local var returnRowCount. Would it retain the final count if that work tcb returns in cases like up queue is full, etc.

Are all regressions run before delivery and not just checktests?

core/sql/regress/compGeneral/EXPECTED023
zellerh: Agree with Venkat. The GET statement should behave like a SELECT as much as possible. Ideally, we would someday allow it in places where we allow select:

```
select *
from (get users union all get schemas) T(o)
where o like 'finance%'
```

I know, that's science fiction for now, but it would be good to make sure this will some day be possible.

anoopsharma00: This is not that much of science fiction. 
get commands can be used as derived tables within selects.

For ex: one can do the following to get this result:
>>create table if not exists "finance" (a int);

--- SQL operation complete.
>>select left(b, 10) from
+>((select * from (get tables in schema sch) x(a))
+> union all
+> (select * from (get users) x(a))) z(b)
+>where b like 'finance%';

(EXPR)                                  
----------------------------------------

finance                                 

--- 1 row(s) selected.
>>

zellerh: Very cool, thanks!

traflm:
jenkins, extra tests

zellerh:
@traflm, what will happen when I do the following, if I have two tables:

select count(*) from (get tables) t;

Will that display 2 or something else, because we now return additional rows in get tables?

traflm:
jenkins, extra tests

traflm:
@zellerh That query works fine will NOT count the extra message rows.

>>set schema sch;

--- SQL operation complete.
>>get tables;

Tables in Schema TRAFODION.SCH
==============================

SB_HISTOGRAMS
SB_HISTOGRAM_INTERVALS
SB_PERSISTENT_SAMPLES
T1

=======================
Summary info of get:
Get 4 rows.


--- SQL operation complete.
>>

>>>>select count(*) from (get tables) t(a);
(EXPR)
--------------------

                   4

--- 1 row(s) selected.

anoopsharma00:
when a GET statement is used in a subquery, then parser automatically sets noHeader flag in the get operator. That causes header and footer rows to not be returned. 
It behaves the same as "get tables, no header".

traflm:
Hi, Anoop, not ALL get command work. As you mentioned, some GET command will invoke GET recursively, in that case, it doesn't work, I still cannot figure out the reason. But you remind me that it is another work() method.
And, yes, the local var is not good

traflm:
jenkins, extra tests

traflm:
jenkins, retest

traflm:
@anoopsharma00 , @venkat1m , I have refactor the code. Could you help to take a review again?
Thanks,
Ming

traflm:
@anoopsharma00 , @venkat1m @zellerh 
Let me explain this PR: it is pure for user friendly change. The rows returned message is part of the 'header' message, so if no header option spefied, that message will not display as well. So you can think it as an extension to the header display information.
This will be useful at field. For example, after running a script to create huge amount of tables, it will be good to validate the number of tables created successfully. This may not look like a common operation, but this is a true use case. I am in that project and at field. We are required to write DDL in a very short time to create 30+ tables and external tables and indexes. Run a script, and want to check how many tables created, and found this 'issue', so it is something sometimes really helpful.

zellerh:
+1
Looks good to me.

PR #1622, opened 2018-06-26
[TRAFODION-3118] Improve on parts of EXTRACT

kakaxi3019:


core/sql/parser/sqlparser.y
```diff
@@ -500,6 +501,7 @@ static void enableMakeQuotedStringISO88591Mechanism()
 %token <tokval> TOK_CARDINALITY
 %token <tokval> TOK_CASE
 %token <tokval> TOK_CAST
+%token <tokval> TOK_CENTURY
```
DaveBirdsall: The new tokens, TOK_CENTURY, TOK_DECADE, etc., need to be added to the nonreserved_word production, otherwise they will be taken as reserved words (notwithstanding the NONRESTOKEN_ flag in ParKeyWords.cpp). Please also add them to test core/TEST037, which tests that non-reserved words really are non-reserved.

core/sql/regress/seabase/TEST030
```diff
@@ -98,6 +98,11 @@ select HOUR(interval '5:2:15:36.33' day to second(2)) from (values(1)) as t(a);
 select MINUTE(interval '5:13:25:2.12' day to second(2)) from (values(1)) as t(a);
 select extract (year from INTERVAL '97-02' YEAR TO MONTH) from (values (1)) as t(a);
 select interval '8' year / 4 from dual;
+select extract(week from date '2005-01-01') from (values(1)) as t(a);
+select extract(epoch from timestamp '2000-12-30 20:38:40.12') from (values(1)) as t(a);
+select extract(dow from timestamp '2018-06-21 20:38:40') from (values(1)) as t(a);
+select extract(doy from timestamp '2018-06-21 20:38:40') from (values(1)) as t(a);
+select extract(wom from timestamp '2018-06-21 20:38:40') from (values(1)) as t(a);
```
DaveBirdsall: Consider adding tests for the other options, e.g. CENTURY, DECADE

core/sql/optimizer/SynthType.cpp
```diff
@@ -4416,6 +4416,22 @@ const NAType *Extract::synthesizeType()
...
   }
+  if ( type != NA_DATETIME_TYPE )
+    {
+      enum rec_datetime_field eField = getExtractField();
+      NAString sErr;
+      if ( REC_DATE_WEEK == eField
+          || REC_DATE_DOW == eField
+          || REC_DATE_DOY == eField
+          || REC_DATE_WOM == eField
+          || REC_DATE_CENTURY == eField)
```
DaveBirdsall: Should DECADE be in this list too?

core/sql/optimizer/SynthType.cpp
```diff
@@ -4470,15 +4489,31 @@ const NAType *Extract::synthesizeType()
...
     scale += dti.getFractionPrecision();
   }
+  if (getExtractField() == REC_DATE_EPOCH)
+    {
+      prec  += dti.getFractionPrecision();
+      scale += dti.getFractionPrecision();
+    }
+  NABoolean bNegValue = FALSE;
+  if ( getExtractField() >= REC_DATE_CENTURY && extractStartField <= REC_DATE_WOM )
+    bNegValue = TRUE;
```
DaveBirdsall: I'm confused. From this code, it looks like all of the new options can yield negative values. Yet, earlier, for example, QUARTER is limited to a precision of 1. I could imagine QUARTER being negative if it were with respect to the first quarter of year 1 AD. But since the precision is 1, I'm guessing that QUARTER is always with respect to the current year, so it only takes the values 1, 2, 3 or 4. I'm wondering what DECADE, DOM, DOY and WOM are; what are they relative to? And can they be negative?

core/sql/parser/sqlparser.y
kakaxi3019: add TOK_CENTURY, TOK_DECADE, TOK_DOW,  TOK_DOY ,TOK_EPOCH, TOK_WOM to nonreserved_word 
TOK_WEEK, TOK_QUARTER already exists in nonreserved_func_word

core/sql/regress/seabase/TEST030
kakaxi3019: add century,decade,quarter tests

core/sql/optimizer/SynthType.cpp
kakaxi3019: DECADE,EPOCH,QUARTER should not be in this list, because DECADE support interval.
exp.
EXTRACT(DECADE FROM INTERVAL '9' YEAR - INTERVAL '99' YEAR);

core/sql/optimizer/SynthType.cpp
kakaxi3019: Thanks Dave, there are some problems here.
options DECADE,QUARTER,EPOCH can be negative values

PR #1623, opened 2018-06-27
TRAFODION-3099

gdgy:


core/sql/cli/CliExpExchange.cpp
```diff
@@ -3804,7 +3893,10 @@ InputOutputExpr::inputValues(atp_struct *atp,
           if (!source) {
 	    continue;
 	  }
-
+          if(isOdbc)
```
zellerh: I wonder whether it is correct that JDBC passses a string of the format _iso88591'ABCD' here. I would have expected that trafci should have already converted this to an unquoted string. You can try the following with your fix and see what you get:

trafci
execute xx using '_iso88591''ABCD''';

Will this insert ABCD into the table or _iso88591'ABCD'? If it inserts ABCD, then I think the real bug is in trafci, not in this method, InputOutputExpr::inputValues().

core/sql/cli/CliExpExchange.cpp
```diff
@@ -3383,6 +3383,94 @@ InputOutputExpr::inputRowwiseRowsetValues(atp_struct *atp,
...
+                        break;
+                    }
+                }
+                Lng32 valBeg = perfix_end+1;
+                if (!source[valBeg])
+                {
+                    valBeg += 1;
+                }
+                memcpy(&source[perfix_beg], &source[valBeg], valEnd-valBeg+1);
+                memcpy(&source[valEnd-valBeg+1+perfix_beg], &source[valEnd+3], valBeg+2);
```
zellerh: Wouldn't this copy uninitialized memory beyond the end of the value? This also shows a problem, that you would need to shorten the string. Another indicator that maybe the bug is in trafci and not here (see comment below).

core/sql/cli/CliExpExchange.cpp
```diff
@@ -1603,7 +1603,7 @@ InputOutputExpr::outputValues(atp_struct *atp,
 	    sourceType = REC_DECIMAL_LSE;
 	  }
 	  
-	  // 5/18/98: added to handle SJIS encoding charset case.
+      // 5/18/98: added to handle SJIS encoding charset case.
```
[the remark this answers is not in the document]
gdgy: ok， I will set my  editor  each tab to 8 spaces.

core/sql/cli/CliExpExchange.cpp
gdgy: I have test if we not remove the prefix the prefix will insert into the table.
tim 20180628093450

zellerh: Maybe trafci does not require quotes around string literals in the "using" clause. Also, I don't think we should allow this kind of syntax in the "using" clause. We are specifying the value of a parameter that already has a data type. Specifying a character set at this time does not make sense, that syntax is meant for SQL queries, not for variables of parameters.

Note that the "using" clause does not accept SQL syntax, only values. So, execute xx using 3+4 would probably insert '3+4' into the database, not '7'.

DaveBirdsall:
I think there is a conceptual issue with this fix and with the original JIRA. The datatype of the dynamic parameter is fixed by the compiler at compile time, so it cannot depend on an expression supplied at run time. There are mechanisms at the CLI level that allow run time manipulation; it may be that ODBC/JDBC allow access to those. (Trafci does not.) But those are limited to changing the datatype only; they do not involve lexical analysis or parsing of the expression. Indeed, given a string 'ABCD' in an EXECUTE USING statement, trafci automatically converts from the character set of the SQL statement to the character set of the column, making the N'xxxx' and _iso88591'xxxx' syntaxes unnecessary.

DaveBirdsall:
Following up on that remark: Perhaps the best way to resolve this particular JIRA is with a documentation change. We can simply document this behavior in the SQL Reference Manual.

PR #1626, opened 2018-06-29
[TRAFODION-3126] Refactored HDFS client implementation should also su…

selvaganesang:
…pport Alluxio file system

Alluxio doesn't support direct ByteBuffer access. Circumvented
this problem by using non-direct ByteBuffer to read hdfs files
when it belongs to Alluxio file system.

No need to change the default setting of USE_LIBHDFS for Alluxio to work.

### DIFF
--- a/core/sql/src/main/java/org/trafodion/sql/HDFSClient.java
+++ b/core/sql/src/main/java/org/trafodion/sql/HDFSClient.java
@@ -87,6 +87,7 @@ public class HDFSClient
    private static ExecutorService executorService_ = null;
    private static FileSystem defaultFs_ = null;
    private static CompressionCodecFactory codecFactory_ = null;
+   private static boolean alluxioNotInstalled_ = false;
    private FileSystem fs_ = null;
    private int bufNo_;
    private int rangeNo_;
@@ -95,6 +96,7 @@ public class HDFSClient
    private OutputStream outStream_;
    private String filename_;
    private ByteBuffer buf_;
+   private ByteBuffer savedBuf_;
    private byte[] bufArray_;
    private int bufLen_;
    private int bufOffset_ = 0;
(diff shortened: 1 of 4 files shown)


selvaganesang:
@zellerh If you can, please review this PR

core/sql/src/main/java/org/trafodion/sql/HDFSClient.java
```diff
@@ -142,21 +154,32 @@
...
+         } 
+         else  {
+            // alluxio doesn't support direct ByteBuffer reads
+            // Hence, create a non-direct ByteBuffer, read into
+            // byteArray backing up this ByteBuffer and 
+            // then copy the data read to direct ByteBuffer for the 
+            // native layer to process the data
+            if ((! alluxioNotInstalled_) && fs_ instanceof alluxio.hadoop.FileSystem) {
+               savedBuf_ = buf_;
+               buf_ = ByteBuffer.allocate(savedBuf_.capacity());
```
zellerh: Do we do this allocation every time we call the call() method? If so, could we avoid that and do this only once? The same applies to line 163 above.

selvaganesang:
You will get ClassNotFoundException or ClassDefNotFoundError during instanceof

core/sql/src/main/java/org/trafodion/sql/HDFSClient.java
```diff
@@ -126,6 +128,16 @@
...
       }
+      try {
+         boolean alluxioFs = defaultFs_ instanceof alluxio.hadoop.FileSystem;
+      }
+      catch (Throwable rte)
+      {
+         // Ignore the exception. It is not needed for alluxio to be installed
+         // for the methods of this class to work if 
+         // alluxio filesystem is NOT required
+         alluxioNotInstalled_ = true;
```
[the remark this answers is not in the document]
selvaganesang: You would get stack trace in the JIRA

core/sql/src/main/java/org/trafodion/sql/HDFSClient.java
selvaganesang: HdfsClient object is constructed for every range or every HdfsScanBuf size within a range in HdfsScan.  The call method is called only once. So, it is ok to allocate.

However, It is possible to optimize on this, if you push the detection of file system before HdfsClient object is constructed to HdfsScan, if the allocation becomes a bottleneck.

zellerh: Thanks for the explanation!

PR #1632, opened 2018-07-06
[TRAFODION-3127] tochar enhancement

kakaxi3019:


core/sql/exp/exp_datetime.cpp
```diff
@@ -3294,6 +3338,39 @@ convertMonthToStr(Lng32 value, char *&result, UInt32 width)
...
+    "FRI",
+    "SAT"
+  };
+
+  if (bAbbreviation)
+    strcpy(result, dayofweek_abb[value-1]);
+  else
+    strcpy(result, dayofweek[value-1]);
+  // Update result pointer to point to end of string.
+  result += width;
```
DaveBirdsall: Why width instead of strlen(result)? (This is probably OK; I'm just curious.)

DaveBirdsall:
The manuals should be updated. Perhaps you could write a separate JIRA for that purpose? @liuyu000 might be interested.

kakaxi3019:
Hi, Dave.Thanks for reminding. I have wrote a JIRA TRAFODION-3127.

core/sql/exp/exp_datetime.cpp
kakaxi3019: Yes. The string length is always 3 or 9, so i assign the value for width. 
For example 
strlen("SUNDAY")=6, but strlen("WEDNESDAY")=9. 
so the length will always be 9.
But here "SUNDAY   "(with 3 whitespace), so your suggestion is good, thank you!

liuyu000:
@DaveBirdsall Thanks for your remiding, I'll update the SQL Reference Manual 😁

PR #1633, opened 2018-07-09
[TRAFODION-2743] Windows ODBC LOB support

Weixin-Xu:
The operation is following JDBC T4

### DIFF
--- a/win-odbc64/Common/TransportBase.h
+++ b/win-odbc64/Common/TransportBase.h
@@ -134,6 +134,8 @@ enum SRVR_API {
 	SRVR_API_SQLFASTEXECDIRECT,					//OK WMS
 	SRVR_API_SQLFASTFETCH_PERF,					//OK WMS
 	SRVR_API_GETSEGMENTS,						//OK WMS
+    SRVR_API_EXTRACTLOB,                        //OK LOB
+    SRVR_API_UPDATELOB,                         //OK LOB
 	SRVR_API_LASTAPI								//Add new APIs before this
 
 };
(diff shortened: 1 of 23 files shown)


win-odbc64/Common/TransportBase.h
```diff
@@ -134,6 +134,8 @@ enum SRVR_API {
 	SRVR_API_SQLFASTEXECDIRECT,					//OK WMS
 	SRVR_API_SQLFASTFETCH_PERF,					//OK WMS
 	SRVR_API_GETSEGMENTS,						//OK WMS
+    SRVR_API_EXTRACTLOB,                        //OK LOB
+    SRVR_API_UPDATELOB,                         //OK LOB
```
DaveBirdsall: You may want to check the tab settings in your editor; the indentation seems to frequently be wrong. Best practice is to store spaces instead of tabs.

Weixin-Xu: The above use tab while I use 4 spaces instead of tab.

DaveBirdsall: My guess is that the original code uses tabs (which is bad) and the tab setting in github assumes 8 spaces per tab. Consider changing your editor to use 8 spaces per tab.

win-odbc64/odbcclient/drvr35/lob.cpp
```diff
@@ -54,8 +54,12 @@ namespace ODBC {
         CStmt *pStatement = (CStmt *)hstmt;
         IDL_long extractLen = length;
 
+        IDL_string lobHandleTemp = (IDL_string)malloc(lobHandleLen + 1);
```
DaveBirdsall: Is there a place where this string will be deleted? If not, is this a memory leak?

PR #1636, opened 2018-07-10
[TRAFODION-3133] limit n support in subquery

Guhaiyan:
Just as the title shows, it is to support limit n in subquery.


DaveBirdsall:
What is the intent behind this change? Is it to just accept the syntax?

If it is intended to be the full feature, there should be some tests added that demonstrate that the feature works.

PR #1637, opened 2018-07-10
[TRAFODION-3131] Fix some reserved and non-reserved word issues

DaveBirdsall:

1. Fixes some issues with reserved words not being recognized as reserved words.
2. Updates core/TEST037 to include test cases for some recently added functions (CRC32, MD5, SHA, SHA1 and SHA2)
3. Updates the expected results for core/TEST037.

Note: I'm aware that the comments in core/TEST037 are not perfectly aligned with the expected results. (But they are much more aligned than they were.) Will leave further improvements to follow-up work.

Note: I considered refactoring common/ComResWords.cpp and the "prettify SQL text" functions in common/NAString.cpp, moving them to the parser and changing them to use the ParKeyWords table in parser/ParKeyWords.cpp. I wanted to do this to simplify development, consolidating three keyword tables into one. I started down this path and discovered that this was far from easy; moreover I concluded that it wouldn't actually simplify development much. So I fell back to the position of merely adding thorough documentation to parser/ParKeyWords.cpp, giving instructions on how to keep these tables in sync.


zellerh:
Do you know whether the old behavior of TEST037 just masked errors and looked like success or whether the statements using reserved words really succeeded? Having too many reserved words is one of the biggest complaints of new Trafodion users, so increasing the number of reserved words would not be good.

I think @anoopsharma00 made some changes a while ago that moved many keywords from the reserved to the non-reserved category.

DaveBirdsall:
The old behavior of TEST037 was an artifact of turning off reserved word checking in some code paths. Notice in the old expected results, the comments say, "expect error 3128" but no error occurs. (See the code change to common/NAString.cpp. The old code for function IsSqlReservedWord formerly returned FALSE. I've changed it to call the ComResWords methods instead. Actually there is a derived product that also calls the ComResWords methods here, so I believe this is correct.)

zellerh:
+1
Thanks for the explanation. Now you got me thinking what that "derived product" might be :-)

PR #1639, opened 2018-07-11
[TRAFODION-3138] Part 1 of removing static SQL code from ustats + one bug fix

DaveBirdsall:
This pull request contains two changes:

1. It removes obsolete static SQL code from two ustat modules (hs_read.cpp and hs_update.cpp). It also removes obsolete metadata version checking code. Thanks to @anoopsharma00 for these changes.

2. It fixes a bug in incremental UPDATE STATISTICS. If one has a table name containing a dot (.), incremental UPDATE STATISTICS fails. For example:

```
>>update statistics for table "A.B" on existing columns incremental where a > 128;

*** ERROR[15005] Unmatched quote in input (unterminated string): 
select count(*) from B" where  a > 128;

>>
```

This bug has been fixed (hs_globals.cpp). Thanks to @anoopsharma00 also for this change.

### DIFF
--- a/core/sql/ustat/hs_globals.cpp
+++ b/core/sql/ustat/hs_globals.cpp
@@ -5328,13 +5328,16 @@ Lng32 HSGlobalsClass::validateIUSWhereClause()
 {
   Lng32 retcode = 0;
 
+  // use QualifiedName constructor to correctly handle delimited names.
+  QualifiedName qualTableName(user_table->data(), 1);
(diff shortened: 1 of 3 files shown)


DaveBirdsall:
Note: There is also static SQL code remaining in ustat/hs_cli.cpp and ustat/hs_globals.cpp. This code will be removed in a later set of changes.

PR #1646, opened 2018-07-16
[TRAFODION-3145] Make "hbase_options" as default setting and use cqd …

traflm:
…to change the default setting

As we know, for performance thinking, we always need to add below syntax in “create table” syntax. And I know in latest version, we have made “ATTRIBUTES ALIGNED FORMAT” as default, do we have a plan that make HBASE_OPTIONS(including ENCODING、COMPRESSION、MEMESTORE) as the default setting when creating table? I think this will be easier for end users?
ATTRIBUTES ALIGNED FORMAT
HBASE_OPTIONS
(
DATA_BLOCK_ENCODING = 'FAST_DIFF',
COMPRESSION = 'SNAPPY',
MEMSTORE_FLUSH_SIZE = '1073741824'
)
;

Trafodion has
cqd HBASE_DATA_BLOCK_ENCODING_OPTION ‘FAST_DIFF’;
cqd HBASE_COMPRESSION_OPTION ‘SNAPPY’;

Add 
cqd HBASE_MEMSTORE_FLUSH_SIZE_OPTION '1073741824' ;

So one can use these values without explicitly specify them when create table.


core/sql/sqlcomp/CmpSeabaseDDLcommon.cpp
```diff
@@ -10991,7 +10997,20 @@ CmpSeabaseDDL::setupHbaseOptions(ElemDDLHbaseOptions * hbaseOptionsClause,
...
+    {
+      hbaseOption = new(STMTHEAP) HbaseCreateOption("MEMSTORE_FLUSH_SIZE", 
+                                                    flushSize.data());
+      hbaseCreateOptions.insert(hbaseOption);
 
+      if (ActiveSchemaDB()->getDefaults().userDefault
+          (HBASE_MEMSTORE_FLUSH_SIZE_OPTION) == TRUE)
+        {
+          numHbaseOptions += 1;
+          sprintf(optionStr, "MEMSTORE_FLUSH_SIZE='%s'|", flushSize.data());
```
zellerh: I would recommend using snprintf instead of sprintf, to avoid the possibility of a buffer overrun (probably not likely here, but just in case).

```
snprintf(optionStr, sizeof(optionStr), ...)
```

zellerh:
+1
Looks good to me.

selvaganesang:
I presume that  there is no need for  Trafodion SQL engine to parse the Hbase options because the engine doesn't do any specific action based on these options.  If so,  Hbase create options could have been implemented like pass thru.  If there is a problem with hbase create options pass through string, hbase would return an error. Trafodion can just return an error and with DDL transactions enabled automatically, the transaction is rolled back to ensure the metadata consistency. This way, there is no need to update our code for any new options.

core/sql/sqlcomp/CmpSeabaseDDLcommon.cpp
traflm: I agree, I can change this into a more robust coding style, using snprintf.

traflm:
@selvaganesang , this is only for user friendly. By default, Trafodion will NOT take any assumption of what HBase options to use. All these CQD by default is null.
At field, when user want to create a table, in most cases, they have to explicitly specify three HBASE_OPTIONS. We saw them in nearly 100% use cases. 
It is hard for end user, or even experienced supporting engineers to remember these details. And since it is Trafodion extension, many times, we have to change existing tools to add these options syntax.
With these three CQDs, it will help in above use cases. And it doensn't change default behavior, so user use it with clear understanding of what they want the table with hbase options.
This is a pure user-friendly improvements.

selvaganesang:
@traflm My suggestion is to have a single CQD that would let users to specify the default hbase options. Then Trafodion uses it in create table,  Trafodion could validate the CQD at the time of issuing the CQD by creating a dummy table with these hbase options.

traflm:
@selvaganesang  A single CQD is a very good idea, I can work on it later, that CQD will have highest priority and when it is set, all other hbase option CQD will be ingored and as you mentioned, we just pass it to hbase.
Let me do it in next iteration ^_^

selvaganesang:
Thanks @traflm for volunteering to implement the suggestion.

PR #1647, opened 2018-07-16
[Trafodion-3049] Numeric for ODBC need more effcient and correct

Weixin-Xu:
When the target DataType is like Char or Varchar, the function need skip.
And I copy the function before the value used.


DaveBirdsall:
This pull request has conflicts that need to be resolved before we can go further with it.

PR #1651, opened 2018-07-18
[TRAFODION-3150] fix to_char & extract bugs

kakaxi3019:


zellerh:
-1 - for now.

This would make incompatible changes to Trafodion and I think we should discuss first what the best solution is. Please see my comments in the JIRA.

kakaxi3019:
Thanks Hans!. I have left my comments in the JIRA.

zellerh:
Hi @kakaxi3019, thank you for looking at this issue some more. I also put another comment into the JIRA. My recommendation would be to start a discussion on the user@trafodion.apache.org list to see what other people think.

PR #1652, opened 2018-07-18
[TRAFODION-3153] Fix issue with CREATE EXTERNAL TABLE

DaveBirdsall:
Change the parser production for CREATE EXTERNAL TABLE to point to the proper token for the optional IF NOT EXISTS clause.


robertamarton:
+1 looks good to me

PR #1660, opened 2018-07-23
[TRAFODION-3158] Make EncodedValue error handling more robust

DaveBirdsall:
This pull request makes two code changes to function EncodedValue::constructorFunction.

1. Formerly, the function would check for the existence of an error in CmpCommon::diags() and remove it if present. This wasn't sufficient though as the test case in the JIRA generated *two* errors in CmpCommon::diags(); the code as it was removed only one of them. The result was that the Normalizer retried the query, and the warnings concerning the bad histogram values were lost.

2. In my initial attempt to fix this problem, I put an unguarded call to CmpCommon::diags()->mark() to capture the state of CmpCommon::diags() on entry to EncodedValue::constructorFunction. This turned out to cause cores in the mxssmp process at sqstart time, because in that process it happened that the EncodedValue constructor for a global object was called before the global CmpCommon::diags() was initialized. This could have happened in any process; it's just the luck of the draw in what order the linker decides to invoke global constructors. So to fix that problem, I was forced to add a flag so that code paths involving global objects do not try to dereference CmpCommon::diags(). This change is to my mind ugly; I added extensive comments explaining the situation. Some significant refactoring will be required to produce a more elegant solution.

### DIFF
--- a/core/sql/optimizer/EncodedValue.cpp
+++ b/core/sql/optimizer/EncodedValue.cpp
@@ -648,8 +648,62 @@ EncodedValue::EncodedValue (ItemExpr *expr,
 void
 EncodedValue::constructorFunction (const NAWchar * theValue,
                                    const NAColumnArray &columns, 
+                                   NABoolean okToReportErrors,
                                    ConstValue* cvPtrs[])
 {
+  // Some notes about error reporting for this function:
+  //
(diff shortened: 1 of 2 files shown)


DaveBirdsall:
Note: I ran full regressions on this change last night on a workstation. They pass.

PR #1663, opened 2018-07-24
[TRAFODION-3152] Fix bug in UPDATE STATISTICS CREATE SAMPLE

DaveBirdsall:
If the "_HIVESTATS_" schema didn't already exist, UPDATE STATISTICS ... CREATE SAMPLE and UPDATE STATISTICS ... REMOVE SAMPLE would fail with weird errors.

With this change, they now will succeed. The "_HIVESTATS_" schema and associated histogram tables will be created automatically.

### DIFF
--- a/core/sql/ustat/hs_globals.cpp
+++ b/core/sql/ustat/hs_globals.cpp
@@ -3186,38 +3186,13 @@ HSGlobalsClass::~HSGlobalsClass()
   DeleteHSContext(contID_);
 }
 
+
 // -----------------------------------------------------------------------
-//
+// Initialize stats schema for Hive or native HBase tables if needed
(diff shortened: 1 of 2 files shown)


DaveBirdsall:
An explanation of the change for reviewers: I took the code that creates the Hive stats (and native HBase stats) schema and the histograms tables out of HSGlobalsClass::Initialize and put them in a new method, HSGlobalsClass::InitializeStatsSchema. The former method now calls the latter. I also added a call to the latter to function managePersistentSamples; this function is executed for UPDATE STATISTICS ... CREATE / REMOVE SAMPLE.

zellerh:
+1
Looks good to me.

PR #1664, opened 2018-07-25
[TRAFODION-2884] Multiple fixes with Name Server enabled logic.

zcorrea:
- More than one Name Server is now supported
- Node failures handled with with/without Name Server in node
- Long process names are now the default and support clusters larger that 1000 nodes
- In a real cluster, the monitor startup now uses the AGENT mode logic which in which monitor
  processes join the cluster through the master/lead monitor process.

Note: The Name Server functionality is not currently enabled. To enable for testing purposes uncomment the SQ_NAMESERVER_ENABLED=1 environment variable in 'sqenvcom.sh'


DaveBirdsall:
Is there a JIRA associated with this? Could you put it in the subject line? Thanks.

core/sqf/monitor/linux/process.cxx
```diff
@@ -4775,46 +4818,85 @@ void CProcessContainer::Bcast (struct message_def *msg)
...
+                    name[i+4] = 'A'+cnt[i];
+                }
             }
         }
+        name[8] = '\0';
+    }
+    else
+    {
+        // We are skipping 'A', 'I', 'O', and 'U' to distinguish between zero
+        // and one digits, and for political correctness in generated names
```
DaveBirdsall: No English four-letter words with just E as the vowel?

carolp222: We could get closer to the workarounds that teenagers use:  V for U, zero for O.  That gives you a few more.  But for me, this seems good enough to start with.

PR #1671, opened 2018-07-31
[TRAFODION-3710] odb don't report error message.

SuJinpei:
**root cause:** SQLGetDiagRec expected an error message buffer length parameter which type is signed short, but odb defined a macro LINE_CHUNK as 51200 and this macro will be pass to SQLGetDiagRec as error message buffer length parameter,  51200 is a negative value for signed short, so no error message returned.
**solution:** define a new macro ERR_MSG_LEN with a small value for SQLGetDiagRec function.

### DIFF
--- a/core/conn/odb/src/odb.c
+++ b/core/conn/odb/src/odb.c
@@ -40,6 +40,7 @@ char *odbauth = "Trafodion Dev <trafodion-development@lists.launchpad.net>";
 #define MAX_ARGS    11          /* Max arguments for interactive mode */
 #define ARG_LENGTH  128         /* Max argument length in interactive mode */
 #define LINE_CHUNK  51200       /* size of memory chunks allocated to store lines */
+#define ERR_MSG_LEN 512         /* size of error message buffer */
 #define MAX_VNLEN   32          /* Max variable name length */
 #define MAX_PK_COLS 16          /* Max number of PK elements */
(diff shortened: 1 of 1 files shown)


core/conn/odb/src/odb.c
```diff
@@ -40,6 +40,7 @@ char *odbauth = "Trafodion Dev <trafodion-development@lists.launchpad.net>";
 #define MAX_ARGS    11          /* Max arguments for interactive mode */
 #define ARG_LENGTH  128         /* Max argument length in interactive mode */
 #define LINE_CHUNK  51200       /* size of memory chunks allocated to store lines */
+#define ERR_MSG_LEN 512         /* size of error message buffer */
```
xiaozhongwang: The value of short is [32768, 32767], why do you define it so small?
I am worried that this value is too small and will cause a buffer overflow.

SuJinpei: This value came from the driver implementer @Weixin-Xu 's advise. By the way, the small buffer length won't cause the buffer overflow issue, because the driver will protect this by truncating the error message.

SuJinpei:
test failed was caused by maven build issue, some dependencies can't be download. So, Jenkins retest.

core/conn/odb/src/odb.c
xiaozhongwang: +1 if you have think about this.

PR #1672, opened 2018-07-31
[TRAFODION-3164] Restart mxosrvrs on-demand

zellerh:
This code adds a zookeeper value to signal to mxosrvr processes
when it is time to exit. This allows to phase out processes when
things in the environment change. The process will wait until it
is idle before exiting.


arvind-narain:
+1 Looks good - thank you very much for this cool way of restarting mxosrvrs

traflm:
Thanks Hans, this is something I always want to do. 
At some special environment, mxosrvr will run into issue, we want to periodically kill it and restart it. But there is no such way at present.
This will help the stability of the system.

I will merge it .

zellerh:
Hi @CoderSong2015, yes, this could allow us to change code in the mxosrvr and get all the mxosrvrs to restart, without having to stop and start the DCS service. So, users would see no interruption of service at all.

PR #1673, opened 2018-08-01
[TRAFODION-3108] Correct the description of Certification widget

Weixin-Xu:
From Atanu's suggestion, the new description can to be better understood.


DaveBirdsall:
I merged this change this morning but for reasons unknown Jenkins does not mark this PR as "Merged". I checked in a fresh instance and verified that this change is in. So the problem is fixed. I think the only thing that needs to be done now is for this pull request to be marked closed. @Weixin-Xu, would you close this?

PR #1674, opened 2018-08-01
[TRAFODION-3171] Refactor Hive sequence file reading to use the new i…

selvaganesang:
…mplementation

### DIFF
--- a/core/sql/executor/ExHdfsScan.cpp
+++ b/core/sql/executor/ExHdfsScan.cpp
@@ -130,8 +130,6 @@ ExHdfsScanTcb::ExHdfsScanTcb(
   Space * space = (glob ? glob->getSpace() : 0);
   CollHeap * heap = (glob ? glob->getDefaultHeap() : 0);
   useLibhdfsScan_ = hdfsScanTdb.getUseLibhdfsScan();
-  if (isSequenceFile())
-     useLibhdfsScan_ = TRUE;
   lobGlob_ = NULL;
   hdfsScanBufMaxSize_ = hdfsScanTdb.hdfsBufSize_;
   headRoom_ = (Int32)hdfsScanTdb.rangeTailIOSize_;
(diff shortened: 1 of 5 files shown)


core/sql/executor/ExHdfsScan.cpp
```diff
@@ -571,6 +569,7 @@ ExWorkProcRetcode ExHdfsScanTcb::work()
              hdfsScan_ = HdfsScan::newInstance((NAHeap *)getHeap(), hdfsScanBuf_, hdfsScanBufMaxSize_, 
                             hdfsScanTdb().hdfsIoByteArraySizeInKB_, 
                             &hdfsFileInfoListAsArray_, beginRangeNum_, numRanges_, hdfsScanTdb().rangeTailIOSize_, 
+                            isSequenceFile(), hdfsScanTdb().recordDelimiter_,
```
sureshsubbiah: If we did not pass in the recordDelimiter_ previously, do we know if previous code worked correctly for text format, when the record delimiter was something other than /n ? This PR does not seem change anything for text format reads, so if there was issue previously, it might still exist.

core/sql/src/main/java/org/trafodion/sql/HDFSClient.java
```diff
@@ -240,7 +256,48 @@ int compressedFileRead(int readLenRemain) throws IOException
...
+          {
+              eof = true;
+              break;
+          }
+          byteArray = ((Text)value_).getBytes();
+          readLen = ((Text)value_).getLength();
+          if (readLen <= lenRemain) {
+                            
+              buf_.put(byteArray, 0, readLen);
+              buf_.put(recDelimiter_);
```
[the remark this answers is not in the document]
selvaganesang: Yes. I need to consider the extra byte added for delimiter. Good catch, Will fix it.

core/sql/executor/ExHdfsScan.cpp
selvaganesang: There is no need pass the record delimiter for text format because it is a raw read of text formatted table. It should have the record delimiter as per the hive metadata.

sureshsubbiah: Thank you for explaining

sureshsubbiah:
+!

PR #1676, opened 2018-08-02
TRAFODION-3165 HDFS storage option not work well if table use namespace

traflm:
If the table has namespace, the path of that table in HDFS should not in /hbase/data/default, but in /hbase/data/<namespace>/


traflm:
thanks Dave

PR #1677, opened 2018-08-02
TRAFODION-3146 Support ANSI OVERLAY function

anoopsharma00:
OVERLAY modifies a source string by replacing a given
substring of the string, which is specified by a given numeric starting
position and a given numeric length, with a replacement string).

When the length of the substring is zero, nothing is removed
from the source string and the string returned by the function is the
result of inserting the replacement string into the source string at the
starting position.

STUFF is syntactic variation of OVERLAY.

Example:
  overlay ('source original string' placing 'modified ' from 8 for 9)
  stuff ('source original string', 8, 9, 'modified ')
    will return:

  'source modified string'

  overlay ('source original string' placing 'modified ' from 8 for 0)
    will return:

  'source modified original string'


DaveBirdsall:
When I read your description of the semantics, I was surprised that the starting position is 0-based. (The starting position for SUBSTRING, in contrast, is 1-based.) But looking at ANSI SQL 2003, I believe you have it right; OVERLAY does use a 0-based starting position.

core/sql/exp/exp_function.cpp
```diff
@@ -786,13 +786,18 @@ ExpRaiseErrorFunction::ExpRaiseErrorFunction (Attributes **attr,
...
-    tableName_((char *)tableName)
+  theSQLCODE_(sqlCode),
+  constraintName_((char *)constraintName),
+  tableName_((char *)tableName)
 {
-    setRaiseError(raiseError);
+  setRaiseError(raiseError);
+
+
+  strncpy(optionalStr_, optionalStr, MAX_OPTIONAL_STR_LEN);
```
DaveBirdsall: This will abend if "optionalStr" is null

core/sql/bin/SqlciErrors.txt
```diff
@@ -1535,6 +1535,7 @@ $1~String1 --------------------------------
 8142 ZZZZZ 99999 ADVANCED INFRM LOGONLY An error was artificially injected, to test error handling. Testpoint $0~string0, Value $1~string1.
 8143 ZZZZZ 99999 ADVANCED INFRM LOGONLY The requested operation stopped $0~Int0 server processes.
 8144 ZZZZZ 99999 ADVANCED MAJOR DIALOUT Corruption is detected in table $0~string0$1~string1.
+8145 ZZZZZ 99999 BEGINNER MAJOR DBADMIN This statement is not supported or incorrect options were specified. Reason: $0~string0
```
DaveBirdsall: Could you add this to the Messages Guide too?

core/sql/exp/exp_function.cpp
anoopsharma00: thanks, good catch. this showed up during regr failures as well.
It is fixed.

anoopsharma00:
start position is 1-based and not 0-based. What was the semantic description
that suggested that it is 0-based? 

An error will be returned if a value of <= 0 is used.

>>select overlay('abc' placing 'xy' from 0 for 2) from dual;

*** ERROR[8145] This statement is not supported or incorrect options were specified. Reason: Start position of replacement string specified in function OVERLAY cannot be less than or equal to zero.

--- 0 row(s) selected.
>>

core/sql/regress/charsets/TEST315
```diff
@@ -315,6 +315,26 @@ select converttohex(_ucs2 x'515a515b515c515d') from dual;
 select char_length(reverse(_ucs2 x'515a515b515c515d')) from dual;
 select converttohex(reverse(_ucs2 x'515a515b515c515d')) from dual;
 
+-- tests for OVERLAY function
+select overlay('w1234567xyz' placing 'abcde' from 3)  from dual;
+select overlay('w1234567xyz' placing 'abcde' from 3 for 4)  from dual;
```
arvind-narain: Do we need a test case to cover the example in the description  where numeric length is 0?

> select overlay('w1234567xyz' placing 'abcde' from 3 for 0)  from dual;

PR #1679, opened 2018-08-05
[TRAFODION-3163] Add script to do STFS health check

traflm:
It is hard to test the STFS status if the cluster has many nodes and the STFS configured with many directories. This script will help to do basic checking.

### DIFF
--- a/core/sqf/src/tm/tools/dtmci.cpp
+++ b/core/sqf/src/tm/tools/dtmci.cpp
@@ -1810,7 +1810,7 @@ int main(int argc, char *argv[])
            {
               const char* v = ms_getenv_str((const char*)lp_nextcmd);
               if(v == NULL)
-                cout <<"*** Enviorement Var not exist" << endl;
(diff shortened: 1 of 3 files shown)


core/sqf/src/tm/tools/dtmci.cpp
```diff
@@ -1801,6 +1801,20 @@ int main(int argc, char *argv[])
...
+        else if (!strcmp(lp_nextcmd, "showenv"))
+        {
+           get_cmd(lp_inputstr, lp_nextcmd);
+           if (lp_nextcmd[0] == '\0')
+             cout <<"TODO: show all env var\n" << endl;
+           else
+           {
+              const char* v = ms_getenv_str((const char*)lp_nextcmd);
+              if(v == NULL)
+                cout <<"*** Enviorement Var not exist" << endl;
```
DaveBirdsall: Typo. "Environment"

traflm: thanks Dave, I will change this typo

PR #1680, opened 2018-08-06
[TRAFODION-3173] enhance ExeCliInterface fetchAllRows to save the dat…

traflm:
…a type

### DIFF
--- a/core/sql/executor/ExExeUtilCli.cpp
+++ b/core/sql/executor/ExExeUtilCli.cpp
@@ -1322,7 +1322,6 @@ short ExeCliInterface::fetchAllRows(Queue * &infoList,
 	    if (nullTerminate)
 	      r[len] = 0;
 	  }
-//comehere add type
 	  oi->insert(j, r, len, type);
 	}
 	
--- a/core/sql/executor/ExExeUtilCli.cpp
+++ b/core/sql/executor/ExExeUtilCli.cpp
@@ -64,6 +64,13 @@ void OutputInfo::insert(Lng32 index, char * data, Lng32 len)
   len_[index] = len;
 }
 
+void OutputInfo::insert(Lng32 index, char * data, Lng32 len, Lng32 type, Lng32 *indOffset , Lng32 *varOffset )
+{
+  data_[index] = data;
+  len_[index] = len;
+  type_[index] = type;
+}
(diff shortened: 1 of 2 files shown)


selvaganesang:
I am not not clear on the intent of this change. Is it for debugging purpose. I didn't see the additional parameters like indOffet and varOffset are not being populated. The new type parameter is also not being used.

traflm:
Hi, Selva,
The indOffset and varOffset I can remove them, I just need them in the next change to this interface.
At present, all the user of ExeCliInterface is using it to retrieve small amount of data and in types of character string. So there is no current need for the type and null indicator.
I am trying to use this interface as part of the Oracle Connect By implementation, where I need to retrieve a row from a table and must know the SQL data type for each column and null value.
The null value is not urgent, so I leave it for the next change.

traflm:
The 'Connect By' implementation I am doing is to call a SQL stmt, which is constructed on the fly
select  * from basetable where parentCol = val
in the work() method. 
And the parentCol can be any data type. So the val must consider to add quote sign or not (string or numeric value for example)

After this change merged, I will send out the 'Connect By' code for review, then you will see how this is used.

traflm:
Hi, Selva,
I need this change for another task, since you don't have objection, I will merge this change.

zellerh:
+1
Sounds good to me to merge. There is a limit of 100 output columns. Won't that be a problem with a CONNECT BY feature?

traflm:
yes, Hans, That is a limitation, I noticed it as well, I am planning to enhance that next. But postpone it since it is only required by me at this phase.
For CONNECT BY , it only use the fetchAllRows in very special cases, and I will add protection code: it is used in this case: save the children for a given parent, I hope the fan out will not bigger than 100 in most use cases, but it is highly possible, however, If the user don't use _SYS_CONNECT_IS_LEAF, I will never invoke this interface. I can also disable SYS_CONNECT_IS_LEAF at present.
And in the next , I am planning to simulate the fetchAllRow by adding a new method, which not save all results in an array, but in a Queue. 
That needs a lot of coding, I plan to do it next. Since it is only used by SYS_CONNECT_IS_LEAF function, which I hope is not very frequently used, so postpone it until the end user really require that.

PR #1685, opened 2018-08-07
[TRAFODION-3180] At times establishing a JDBC/ODBC connection takes o…

selvaganesang:
…bservably long time

Analysis revealed that the mxosrvr process in connecting state was attempting to open the
ssmp process on the node for a non-unique query as part of establishing connection.
The ssmp process has many ports in CLOSE_WAIT state. It looks like the client happens
to hit on a port that is in CLOSE_WAIT state.  The port transitions to ESTABLISHED state
after some time. Hence the connection was taking a longer time.

The mxssmp process keeps the port in CLOSE_WAIT because the socket wasn't closed on the
server side when client exits gracefully as well as abruptly. The seabed layer in
Trafodion doesn't handle  more than one open to a process in a correct way. I have changed
the IPC infrastructure in SQL to ensure that the ssmp process is opened only once
in mxosrvr process.

The API msg_get_phandle opens the process with the given name to obtain the handle. This API
is now replaced with XFILENAME_TO_PROCESSHANDLE_

Also added code to possibly fix the regression failure hive/TEST006

### DIFF
--- a/core/sql/common/IpcGuardian.cpp
+++ b/core/sql/common/IpcGuardian.cpp
@@ -212,14 +212,10 @@ NABoolean GuaProcessHandle::compare(const GuaProcessHandle &other) const
 
 NABoolean GuaProcessHandle::fromAscii(const char *ascii)
 {
- SB_Phandle_Type *tempPhandle;
 
- tempPhandle = get_phandle_with_retry((char *)ascii);
-
- if (!tempPhandle)
+  int retcode = get_phandle_with_retry((char *)ascii, &phandle_);
+  if (retcode != FEOK)
     return FALSE;
-
- memcpy ((void *)&phandle_, (void *)tempPhandle, sizeof(SB_Phandle_Type));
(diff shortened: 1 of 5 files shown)


core/sql/common/IpcGuardian.cpp
```diff
@@ -4135,13 +4131,9 @@ void IpcGuardianServer::useProcess(ComDiagsArea **diags,
...
-    short gprc = 0;
-    procHandle = get_phandle_with_retry(tmpProcessName, &gprc);
-    if (procHandle != NULL)
-      rc = 0;
-    else
-      rc = gprc;
-    if ((rc != 0) || (procHandle == NULL))
+    int gprc = 0;
+    gprc = get_phandle_with_retry(tmpProcessName, &procHandle);
+    if (rc != FEOK)
```
arvind-narain: Should rc be set first or use gprc ?

PR #1688, opened 2018-08-10
[TRAFODION-3034] Support Oracle Hierarchy Query (Connect By)

traflm:
This PR is the first patch to support Oracle Hierarchy query feature (CONNECT BY).
In this PR, the feature is implemented as a new SQL utility, It is standalone, rather clear isolated with all other SQL functions, so the impact is minimal.
In the long run, we should finish the ANSI recursive feature (recursive WITH) , and at that time, this feature can be considered to refactor to use that infrastructure.
This is just the first phase of this feature.

The basic logic is simple: the utility will run a query to get all start values (specified by the START WITH clause), then it will construct queries to search for children of the root, and loop until no children can be found.

Oracle has 3 pseudo columns, to support the ISLEAF and CONNECT_BY_PATH, the utility will have to run a query for each parent, it will be rather slow. If a query doesn't have those two pseudo columns required, the utility will run in batch mode, for each iteration, get all children in one query. That will be much faster.

One can check the executor/TEST021 for how this feature works first.

This will be a long review process, there must be many places to be modified and enhanced, thank you all for help in advance.


core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -2505,7 +2505,62 @@ RelExpr *RelExpr::bindSelf(BindWA *bindWA)
...
+    {
+       if (!getConnectBy().isEmpty() && getConnectBy().containsSubquery())
+       {
+         *CmpCommon::diags() << DgSqlCode(-4337);
+         bindWA->setErrStatus();
+         return this;
+       }
+    }  
+   }
+#endif
```
traflm: oops, I will remove these dead code.

traflm:
I need a big rework of this patch. Please hold until I finish.
The current implemention takes assumption there is only one table in the from_clause, which is not correct.

core/sql/optimizer/ValueDesc.cpp
```diff
@@ -4555,7 +4555,7 @@ void ValueIdSet::unparse(NAString &result,
 
   NAString connectorText;
 
-  if ((form == MVINFO_FORMAT) || (form == QUERY_FORMAT))
+  if ((form == MVINFO_FORMAT) || (form == QUERY_FORMAT) || (form == CONNECT_BY_FORMAT))
```
sureshsubbiah: How is the CONNECT_BY_FORMAT different from QUERY or MVINFO formats? The answer might be in other parts of code, its just that I cannot tell. Thank you for explaining

core/sql/optimizer/RelExpr.h
```diff
@@ -525,7 +540,7 @@ class RelExpr : public ExprNode
   // QSTUFF
 
   // --------------------------------------------------------------------
-  // normalizeNode() performs predicate pushdown and also ensures
+  // normalizeNode() performs predicate pushdown and also ensuresL
```
sureshsubbiah: typo

core/sql/executor/ExExeUtil.h
```diff
@@ -4138,6 +4138,7 @@ class ExExeUtilLobInfoTableTcb : public ExExeUtilTcb
     DONE_
   };
   Step step_;
+  char * data_;
```
sureshsubbiah: Why do we need a change in a Lob class?

core/sql/executor/ExExeUtil.h
```diff
@@ -4172,6 +4173,48 @@ class ExExeUtilLobInfoPrivateState : public ex_tcb_private_state
 protected:
 };
 
+class connectByStackItem
```
sureshsubbiah: I think it is unusual in the executor to have classes that do not have a base class. How about memory management for this class. I might understand more as I see how these classes are used.

core/sql/optimizer/ValueDesc.cpp
traflm: CONNECT_BY_FORMAT has only one difference from QUERY_FORMAT, that the column name only has the last part, not qualified. 
The reason is: the column name will be used in the work() method of CONNECT_BY, where alias of the table name is lost. But we can make sure, there is only one table in that query in the work() method, so only the column name is required.

PR #1690, opened 2018-08-12
[TRAFODION-1881] A better way to solve TRAFODION-1858

traflm:
In my previous fix to TRAFODION-1858, I add a new method to check if a column is a system column by comparing the column name with "_SALT_", "_DIVISION_" etc, that is not a good way.
This is a new implementation, it use more robust way to tell if a column is a system column.

To check the RI, the predicate only need to check the user specified Primary Column.


traflm:
thank you Dave for the review

PR #1691, opened 2018-08-12
[TRAFODION-3182] cleanup monitor log content

traflm:
At field, when we want to do trafodion trouble shooting. Open mon.xxx.log file, in many times, it is flood with message related STDIN redirection entries.
These log entries are never real issues, and it is very hard to understand, so this PR is to move these log entries into DEBUG level. 
So only when an expert want to debug very complicated issue and need to turn-on the DEBUG log/trace, these info will be printed out.

I am trying to make log file more useful, this is the first attempt.


selvaganesang:
Any idea what is the root cause of the stdin redirection log entries?

traflm:
Just a theory.
It was a timing issue. One process exit(crash) and before monitor can do proper reaction, but it is for process std IO, which is not a fatal issue. These log info never are used for any real trouble shooting, but just bring confusion.

traflm:
I will merge it, those log entries never been used as valuable troubleshooting clues, just confusion. Nearly no one knows what they mean. And there are too many of them in the field.

PR #1692, opened 2018-08-13
[TRAFODION-3171] Refactor Hive sequence file reading to use the new i…

selvaganesang:
…mplementation

Ensured that split of non-compressed sequence file works. In case of compressed
sequence filee, an exception is thrown when the file is split.


selvaganesang:
jenkins, retest

PR #1693, opened 2018-08-13
[TRAFODION-3180] At times establishing a JDBC/ODBC connection takes o…

selvaganesang:
…bservably long time

When the sock stream is in use, socket close is kept pending in a queue till it can
be closed. But, the sockets from this queue is never closed. Added code
to close the sockets from this queue at the time of creating the socket stream next
time.


selvaganesang:
@narendragoyal , can you please review this change.

PR #1696, opened 2018-08-14
[TRAFODION-3110] Refactor LOB access to use the new implementation of…

selvaganesang:
… HdfsClient

LOB: Extract lobtofile() to a hdfs file returns 8442 error

Implemented the missing code to support this functionality via the new
implementation.


selvaganesang:
@sandhyasun Can you please review.

core/sql/exp/ExpLOBaccess.cpp
```diff
@@ -2455,16 +2455,59 @@ Ex_Lob_Error ExLob::readDataToHdfsFile(char *tgtFileName,  Int64 offset, Int64 s
...
+        overwrite = FALSE; 
+     if ((LobTgtFileFlags)fileflags == Lob_Append_Or_Error)
+        append = TRUE; 
+     tgtHdfsClient = HdfsClient::newInstance(getLobGlobalHeap(), NULL, hdfsClientRetcode);
+     ex_assert(hdfsClientRetcode == HDFS_CLIENT_OK, "Internal error: HdfsClient::newInstance returned an error");
+     if (tgtHdfsClient->hdfsCreate(tgtFileName, overwrite, append, FALSE) !=  HDFS_CLIENT_OK)
+        return LOB_TARGET_FILE_OPEN_ERROR; 
+     Int32 bytesRead; 
+     Int32 bytesWritten;
+     while (remainLen > 0)
```
sandhyasun: In the original code there is code to handle multiple chunks.  That seems to be missing  in this case ? Note that when there are multiple chunks in the LOB data file, they could be a t different offsets - they won't be contiguous.

core/sql/exp/ExpLOBaccess.cpp
```diff
@@ -2455,16 +2455,59 @@ Ex_Lob_Error ExLob::readDataToHdfsFile(char *tgtFileName,  Int64 offset, Int64 s
...
+     while (remainLen > 0) 
+     {
+        if (remainLen > lobMaxChunkMemLen)
+           readLen = lobMaxChunkMemLen;
+        else
+           readLen = remainLen;
+        if (lobData == NULL)
+           lobData = new (lobGlobalHeap_) char[readLen];
+        bytesRead = hdfsClient_->hdfsRead(pos, lobData, readLen, hdfsClientRetcode);
+        if (hdfsClientRetcode == HDFS_CLIENT_OK)
```
sandhyasun: Needs some error handling if retcode is  HDFS_CLIENT_OK ?

selvaganesang: The error handling is done at line 2501

core/sql/exp/ExpLOBaccess.cpp
selvaganesang: Yes. I will make the change

sandhyasun:
+1 looks good.

PR #1698, opened 2018-08-19
[TRAFODION-2968] update SQL manual for MySQL functions unix_timestamp, uuid, sleep etc

traflm:
add documentation for some newly added MySQL functions. For comparability enhancement of Trafodion.

### DIFF
--- a/docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
+++ b/docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
@@ -444,7 +444,7 @@ the expressions have NULL values, the function returns a NULL value.
 | <<nvl_function,NVL Function>>                                   | Returns the value of the first operand unless it is NULL, in which case it returns the value of the second operand.
 | <<user_function,USER Function>>                                 | Returns either the database user name of the current user who invoked the function or the database user name
 associated with the specified user ID number.
-| <<sys_guid_function,SYS_GUID Function>>                         | Returns a globally unique identifier. (Oracle extention)
+| <<sys_guid_function,SYS_GUID Function>>                         | Returns a globally unique identifier, similar to Oracle's SYS_GUID function.
 | <<uuid_function,UUID Function>>                                 | Returns a globally unique identifier.
 | <<uuid_short_function,UUID_SHORT Function>>                     | Returns a “short” universal identifier as a 64-bit unsigned integer.
 | <<sleep_function,SLEEP Function>>                               | Sleep for a number of seconds
@@ -11303,8 +11303,9 @@ A UUID is designed as a number that is globally unique in space and time.
 Two calls to UUID() are expected to generate two different values, even if 
 these calls are performed on two separate devices not connected to each other.
 
-NOTE: But UUID() can only be used in SELECT list, or very simple expressions to get a UUID.
-Do not use it in WHERE clause, the behavior is non-deterministic.
+NOTE: UUID() behavior is non-deterministic. Therefore it is best to use it 
+only in SELECT lists. Using it in a WHERE clause, for example, may result in i
+non-deterministic predicate evaluation.
 
 
 ```
@@ -11335,8 +11336,9 @@ ID
 == UUID_SHORT Function
 Returns a “short” universal identifier as a 64-bit unsigned integer.
 
-NOTE: But UUID_SHORT() can only be used in SELECT list, or very simple expressions to get a UUID.
-Do not use it in WHERE clause, the behavior is non-deterministic.
+NOTE: UUID_SHORT() behavior is non-deterministic. Therefore it is best to use it 
+only in SELECT lists. Using it in a WHERE clause, for example, may result in i
+non-deterministic predicate evaluation.
 
 
 ```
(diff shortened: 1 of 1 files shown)


docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
```diff
@@ -442,7 +444,10 @@ the expressions have NULL values, the function returns a NULL value.
 | <<nvl_function,NVL Function>>                                   | Returns the value of the first operand unless it is NULL, in which case it returns the value of the second operand.
 | <<user_function,USER Function>>                                 | Returns either the database user name of the current user who invoked the function or the database user name
 associated with the specified user ID number.
-| <<sys_guid_function,SYS_GUID Function>>                         | Returns a globally unique identifier.
+| <<sys_guid_function,SYS_GUID Function>>                         | Returns a globally unique identifier. (Oracle extention)
```
DaveBirdsall: Typo: extention should be extension. Suggested wordsmith: "Returns a globally unique identifier, similar to Oracle's SYS_GUID function."

docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
```diff
@@ -442,7 +444,10 @@ the expressions have NULL values, the function returns a NULL value.
...
+| <<uuid_function,UUID Function>>                                 | Returns a globally unique identifier.
```
DaveBirdsall: How does this contrast to SYS_GUID? Is UUID standard? Or is it like some other vendor?

docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
```diff
@@ -11290,6 +11295,99 @@ ID
...
 
+<<<
+[[uuid_function]]
+== UUID Function
+Returns a Universal Unique Identifier (UUID) generated according to RFC 4122. 
+A UUID is designed as a number that is globally unique in space and time. 
+Two calls to UUID() are expected to generate two different values, even if 
+these calls are performed on two separate devices not connected to each other.
+
+NOTE: But UUID() can only be used in SELECT list, or very simple expressions to get a UUID.
```
DaveBirdsall: Do we prevent it from being used in the WHERE clause? If not, suggested word-smith: "NOTE: UUID() behavior is non-deterministic. Therefore it is best to use it only in SELECT lists. Using it in a WHERE clause, for example, may result in non-deterministic predicate evaluation."

docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
```diff
@@ -11290,6 +11295,99 @@ ID
 --- 1 row(s) selected.
 ```
 
+<<<
```
DaveBirdsall: Should there be a SYS_GUID example somewhere also?

docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
traflm: UUID is MySQL function. I don't think ANSI define standard of SQL functions.  In SQL Server, it is called newId(). So let me add comments like SYS_GUID, by adding "similar to MySQL UUID function"?

docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
traflm: Hi, Dave, we already have a SYS_GUID example

docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
traflm: We don't prevent it from being used in the WHERE clause. It can be used in some special case. It just we don't know where optimizer will choose to evaluate this function. If in ROOT, it will be one value for per statement, if it pushdown to SCAN, it will be one value for each row.
So I will update the words as you suggested.

docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
traflm: thank you, I will update the words

docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
DaveBirdsall: That's fine.

docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
DaveBirdsall: OK, thanks.

docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
DaveBirdsall: OK, thanks.

PR #1699, opened 2018-08-19
[TRAFODION-2404] enhance the regexpr performance

traflm:
Previously, each time evaluate a row, it needs to invoke the regcomp(), by this change, only when the pattern is changed, invoke the regcomp()

### DIFF
--- a/core/sql/exp/exp_like.cpp
+++ b/core/sql/exp/exp_like.cpp
@@ -498,11 +498,10 @@ ex_expr::exp_return_type ExRegexpClauseChar::eval(char *op_data[],
   NABoolean matchFlag = true;
   Lng32 len1 = getOperand(1)->getLength(op_data[-MAX_OPERANDS+1]);
   Lng32 len2 = getOperand(2)->getLength(op_data[-MAX_OPERANDS+2]);
-  regex_t reg;
+  char * pattern;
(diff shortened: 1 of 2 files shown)


core/sql/exp/exp_like.cpp
```diff
@@ -513,11 +512,16 @@ ex_expr::exp_return_type ExRegexpClauseChar::eval(char *op_data[],
 
   str_cpy_all(pattern, op_data[2], len2);
   str_cpy_all(srcStr, op_data[1], len1);
-
-  z = regcomp(&reg, pattern, cflags);
+  if(rpattern_ != pattern)
```
DaveBirdsall: This is a pointer comparison rather than a content comparison. Since rpattern_ is a data member and pattern is a stack variable, I expect this comparison will always mismatch, so (rpattern_ != pattern) will always be true. I am guessing that you intended to compare the content instead?

traflm: Hi, Dave, rpattern_ is NAString, so I think this comparison is correct? My intent is to compare the content.
Let me check the NAString further.

traflm:
@DaveBirdsall NAString overload the == operator, so the comparison is correct.
rpattern_ is not a char* but a NAString type

PR #1700, opened 2018-08-20
[TRAFODION-3187] increase the ref count

wenjun-zhu:
The original code lacks the ref count increment, compared the correct code, so add it now.


traflm:
jenkins, add user

PR #1704, opened 2018-08-25
[TRAFODION-2952] large amount of data will cause error in sequence ge…

traflm:
…nerating

Currently, the sequence next range value is NOT protected by DTM, but a retry logic. 
If multiple processes try to use the same sequence, when the cache is used, all of them try to get the next range and go into race. So one may get a range that another get. The logic is any one get the next range will check the updated timestamp, if it is itself, then it is good for him. Otherwise, the range is got by someone else. So the code wait and retry.

For the use case described in JIRA TRAFODION-2952, there are 48 partitioned table and want to do a load of 4 billion rows. It is high concurrency, and many times conflict. 
The that large load, it is very frequent to fail, since the retry time is hardcoded to 10.

This patch introduce a CQD to control the retry number. It is set to a higher number by default, so in most cases, users don't need to modify this. But if still run into issue, one can change the CQD.

### DIFF
--- a/core/sql/cli/Cli.cpp
+++ b/core/sql/cli/Cli.cpp
@@ -10529,9 +10529,9 @@ static Lng32 SeqGenCliInterfaceUpdAndValidateMulti(
       cliRC = cqdCliInterface->holdAndSetCQD("traf_no_dtm_xn", "ON");
     }
 
-  Lng32 numTries = 0;
+  Lng32 numTries = 0, maxRetryNum = sga->getSGRetryNum();
   NABoolean isOk = FALSE;
-  while ((NOT isOk) && (numTries < 10))
+  while ((NOT isOk) && (numTries < maxRetryNum))
     {
       if (startLocalXn)
         {
@@ -10599,7 +10599,10 @@ static Lng32 SeqGenCliInterfaceUpdAndValidateMulti(
       
       numTries++;
       
-      DELAY(100 + numTries*25);
+      if( 100 + numTries*25 < 1000)   //MAX is 1 second
+          DELAY(100 + numTries*25);
+      else
+          DELAY(1000);
     }
 
     // could not update it after 10 tries. Return error.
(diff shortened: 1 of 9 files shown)


core/sql/sqlcomp/DefaultConstants.h
```diff
@@ -2946,6 +2946,12 @@ enum DefaultConstants
   // this is used to change cache size of sequence numbers for a session.
   // It overwrites the cache size that was specified during sequence creation.
   TRAF_SEQUENCE_CACHE_SIZE,
+ 
+  // this is used to set the retry time if two concurrent update of sequence
+  // conflict, and how many times will retry
+  // by default it is 250, when you saw error 1583, you can try to increase
```
DaveBirdsall: In nadefaults.cpp, it looks like the default is 100, not 250, so the comment is wrong.

core/sql/cli/Cli.cpp
```diff
@@ -10599,7 +10599,10 @@ static Lng32 SeqGenCliInterfaceUpdAndValidateMulti(
       
       numTries++;
       
-      DELAY(100 + numTries*25);
+      if( 100 + numTries*25 < 1000)   //MAX is 1 second
+          DELAY(100 + numTries*25);
+      else
+          DELAY(1000);
```
DaveBirdsall: You might want to consider adding some small random amount to this maximum delay. Otherwise you might get a bunch of processes in lock step, all retrying at the same time.

selvaganesang:
It is possible that I may be misreading the code. It looks like the run-time retry value doesn't seem to come from the cqd settings. I could trace that it flows to ExFunctionSequenceValue, but not sure how it influences the run-time value in the cli layer.

core/sql/sqlcomp/DefaultConstants.h
traflm: thanks Dave, yes, I will change this

core/sql/cli/Cli.cpp
traflm: Yes, this is a good suggestion

traflm:
At ExFunctionSequenceValue::eval(), it will invoke seqValGen->setRetryNum(getRetryNum());
In the execution time, the retry number will be initilaized:
   maxRetryNum = sga->getSGRetryNum();
I have tested this, it is very simple to reproduce it. But I will double check it.

PR #1705, opened 2018-08-27
[TRAFODION-3190] expression involving NULL should be treated as FALSE, not TRUE.

wenjun-zhu:


core/sql/exp/exp_comp.cpp
```diff
@@ -116,9 +116,9 @@ ex_expr::exp_return_type ex_comp_clause::processNulls(char *op_data[],
...
        if (getOperand(i)->getNullFlag() && (!op_data[i])) // missing value 
    	{
-   	  // move null value to result.
-   	  *(Lng32 *)op_data[2 * MAX_OPERANDS] = -1;
-   	  return ex_expr::EXPR_NULL;
+            // move null value to result.
+            *(Lng32 *)op_data[2 * MAX_OPERANDS] = 0;
+            return ex_expr::EXPR_NULL;
    	}
      }
```
anoopsharma00: this change is not correct.

3-valued boolean logic requires that comparison of null to another value during intermediate operations should make the result null and not false.
The place where null becomes false is at the very end of that expression.

As an example, with the changed logic, this expression will return incorrect result.
  select * from t where not (a = 1);
if a is null, the where predicate will become true which is incorrect.

With the original 3-valued result, the where pred will become false and not return any result.

anoopsharma00:
checktests have also failed due to this change.

Since this is a fundamental change, make sure that all regressions are run and pass
before a change in this area is merged.

wenjun-zhu:
jenkins, retest

wenjun-zhu:
jenkins, retest

anoopsharma00:
jenkins,retest will not fix incorrect code changes.
Make sure you read review comments and make appropriate changes.

wenjun-zhu:
OK, I will do more investigation.

PR #1708, opened 2018-08-30
Fix for [TRAFODION-3192] UNION ALL up and down queue size default value too small, or not dynamically adjusted

sandhyasun:
Fix for queue sizes for Union, Transpose and Sequence operators.


sandhyasun:
jenkins retest

sandhyasun:
jenkins retest

selvaganesang:
+1 Looks good.  HBASE_VSBB_ROWSET_SIZE is set to 1024 by default. It is not clear how the default queue size of 2048 would be utilized by the operators when a non-queue related CQD restricts the number of entries to 1024 in the  upsert operation.

sandhyasun:
The 2048 size is the maximum it can grow to - it won't be set to 2048 anymore.
Also there is this calculation for the upsert case specifically in GenRelUpdate.cpp  

  if ((getInsertType() == Insert::VSBB_INSERT_USER || // covers upsert
       getInsertType() == Insert::UPSERT_LOAD)  && // covers upsert using load and bulk load
      generator->oltOptInfo()->multipleRowsReturned())
  {
    downqueuelength = getDefault(HBASE_ROWSET_VSBB_SIZE);
    queue_index dq = 1;
    queue_index bits = downqueuelength;
    while (bits && dq < downqueuelength) {
        bits = bits  >> 1;
        dq = dq << 1;
    }
    downqueuelength = dq;
  }

In addition, in the Unpack operator, we further limit queue lengths based on rowlength otherwise the rowset becomes too large to fit in memory. So there is a lot of manipulation of queue lengths throughout the code .

PR #1709, opened 2018-09-08
[TRAFODION-3200] correct label and version

PierreSmits:


tools/docker/build-base-docker.sh
```diff
@@ -19,7 +19,7 @@ set -e -x -u
 
 BASE_SCRIPT_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 
-export DOCKER_ENV_VERSION="0.1"
+export DOCKER_ENV_VERSION="2.3"
 export BASE_IMAGE_NAME="trafodion/base:${DOCKER_ENV_VERSION}"
```
svarnau: Do you want to put this in the release2.3 branch instead of master?

PierreSmits: For sure the changes need to wind up in the 2.3 branch. 

However, having them also in master will show there what latest stable release should be worked with when doing a deployment (implementation) in an adopters environment

DaveBirdsall:
Thanks for this change. I will merge it now.

PR #1710, opened 2018-09-09
[TRAFODION-3193]update cqd reference manual about the new CQD

traflm:
Add a new CQD which end user will use, so update the manual


traflm:
jenkins, retest

traflm:
jenkins, retest

PR #1715, opened 2018-09-19
[TRAFODION-3209] The variable of won't be initialized before a new statement

Weixin-Xu:
Windows ODBC won't initialize the variable before a new statement while Linux ODBC have this operation.


Weixin-Xu:
jenkins, retest

PR #1719, opened 2018-09-27
[TRAFODION-3215] length return 0 when the source and target charset is same

Weixin-Xu:


selvaganesang:
+1.  

1. Ensure that surrounding code conventions like indentation are followed for easy reading.
2. It will be good to create JIRA with symptom description rather than solution description. This would help consumers to pick up this PR if they encounter a similar problem. Otherwise, the consumers need to guess the PR to be picked up when they encounter a problem.

Weixin-Xu:
Actully, I use 4 blank and total 8 black but context use 2 tabs, so it seems different indentation.

DaveBirdsall:
Yeah, I've run into these indentation problems too. Best practice is to use blanks, not tabs, in new functions and methods. When I add lines within an existing method, I try to copy the white space from surrounding lines so if it has tabs it gets them and if it has blanks, it gets them instead. That tends to keep the appearance more uniform in the code. 

It is a bother, though; perhaps someday we'll do a blanket replacement of all tabs with spaces and prevent new tabs from coming in.

PR #1721, opened 2018-09-28
[TRAFODION-3216]UDR libraries need to be stored in metadata table as a blob

sandhyasun:
Udr libraries that used to be on local directories will now be stored as Blobs n the Libraries metadata table. 
There are changes in compile time and runtime components to make this work. See JIRA for details and attached document within JIRA for change details. 

https://issues.apache.org/jira/browse/TRAFODION-3216

This change also involves a change to a metadata table so it requires an upgrade. The version of metadata has been bumped from 2.1.0 to 2.1.1


sandhyasun:
Would like the reviewers to comment on the changes . PLEASE DO NOT MERGE this PR. Would like to ensure we are ready for an upgrade before merging this PR.

sandhyasun:
@zellerh @robertamarton  Can you please review ?

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -563,32 +795,332 @@ void CmpSeabaseDDL::dropSeabaseLibrary(StmtDDLDropLibrary * dropLibraryNode,
...
+    {
+      *CmpCommon::diags() << DgSqlCode(-1382)
+                        << DgString0(libFileName);
+      processReturn();
+      return;
+      
+    }
+  char buf[2048]; // filename max length is 512. Additional bytes for long
+  // library names.
+  str_sprintf(buf, "update %s.\"%s\".%s set library_filename = '%s' , library_storage = filetolob('%s') where library_uid = %ld",
```
zellerh: This brings up a question: It looks like we overwrite the LIBRARIES row when we alter a library. That means that if the old library is not already cached, it will no longer be available. Initially, I thought that we might keep the old version available in the metadata, but maybe that's too complicated. Do you think we could run into an error in the following situation:

```
Process a: prepare s from call sp1();
Process b: alter library lib1 .... ; -- lib1 is the library for sp1
Process a: execute s;
```

Now, if we assume that the jar file for lib1 is not yet in the cache, executing s will try to extract it, but it won't find it in the metadata anymore - assuming that updating the table will invalidate the LOB handle we used for the previous version?

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -1925,3 +2730,568 @@ short CmpSeabaseDDL::createSeabaseLibmgrCPPLib(ExeCliInterface * cliInterface)
...
+            libNameNoPath = libFileName(lastSlash+1, libFileName.length()-lastSlash-1);
+          str_sprintf(ubuf," update %s.\"%s\".%s set library_filename = '%s', library_storage = filetolob('%s') where library_uid = %ld",
+                      getSystemCatalog(),SEABASE_MD_SCHEMA, lti.newName,
+                      libNameNoPath.data(),libName,libuid
+                      );
+          cliRC = cliInterface->executeImmediate(ubuf);
+          if (cliRC < 0)
+            {
+              if (failedLibraries.length() ==0)
+                failedLibraries += "Libraries Upgrade failed for :";
```
zellerh: I assume this is just a warning and we continue using the old flavor of the library?

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -1925,3 +2730,568 @@ short CmpSeabaseDDL::createSeabaseLibmgrCPPLib(ExeCliInterface * cliInterface)
...
+      if (cliRC < 0)
+        {
+          cliInterface->retrieveSQLDiagnostics(CmpCommon::diags());
+          return -1;
+        }
+
+      //update contents into  the new library table so the libname doesn't contain the path and the 
+      //lib contents get loaded into the blob column.
+      userLibsQ->position();
+      for (size_t i = 0; i < userLibsQ->numEntries(); i++)
```
zellerh: This says it loops over userLibsQ, assuming that these are user libraries, but I don't see where it would exclude our system libraries for the ValidateRoutine SPJ and the predefined UDFs.

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -1925,3 +2730,568 @@ short CmpSeabaseDDL::createSeabaseLibmgrCPPLib(ExeCliInterface * cliInterface)
...
+        default:
+          return -1;
+        }
+    } // while
+
+  return 0;
+
+}
+
+short CmpSeabaseDDL::createLibraries(ExeCliInterface * cliInterface)
```
zellerh: This name confused me somewhat. As far as I understand it now, this creates the new ```"_MD_".LIBRARIES``` table, but not any libraries.

core/sql/bin/SqlciErrors.txt
```diff
@@ -37,7 +37,7 @@
 1035 ZZZZZ 99999 BEGINNER MAJOR DBADMIN Catalog $0~CatalogName already exists.
 1036 ZZZZZ 99999 BEGINNER MAJOR DBADMIN Authorization ID $0~String0 cannot grant to authorization ID $1~String1 because it could create a circular dependency.
 1037 ZZZZZ 99999 BEGINNER MAJOR DBADMIN Revoke failed because of a dependent grant between authorization ID $0~string0 and authorization ID $1~string1.
-1038 ZZZZZ 99999 BEGINNER MAJOR DBADMIN --- unused ---
+1038 ZZZZZ 99999 BEGINNER MAJOR DBADMIN Unable to upgrade library management
```
robertamarton: Don't forget to update asciidocs for this and the next message change.

core/sql/comexe/ComTdbUdr.h
```diff
@@ -322,8 +325,12 @@ class ComTdbUdr : public ComTdb
 
   Int32 javaDebugPort_;                                   // 236-239
   Int32 javaDebugTimeout_;                                // 240-243
+  Int64 libraryRedefTime_;                                // 244-251
+  NABasicPtr libraryBlobHandle_;                          // 252-259
+  NABasicPtr librarySchName_;                             // 260-267
+  Int32 libraryVersion_;                                  // 268-271
   // Make sure class size is a multiple of 8
-  char fillerComTdbUdr2_[28];                             // 244-271
+  char fillerComTdbUdr2_[24];                             // 272-295
```
robertamarton: Just a question to understand the code a bit better.  Are there rules on how big a filler of a structure is?  The new field size adds up to 28 which would take up the entire filler.  So a 24 extra bytes were added to allow for extra expansion.  Why was 24 chosen?

core/sql/common/ComMisc.cpp
```diff
@@ -357,3 +358,119 @@ NABoolean ComTrafReservedColName(
 
   return FALSE;
 }
+
+
+Int32  ComGenerateUdrCachedLibName(NAString libname,Int64 redeftime, NAString schemaName, NAString userid, NAString &cachedLibName, NAString &cachedLibPath)
```
robertamarton: Just to understand this function.  It creates the structure of directories (if not already created) and returns  the cached library name and cached library path.   If the user name is passed, then that username is used in the path.  If the user name is not passed, then that portion of the directory is set to DB__ROOT.   According the the header file, the username can also be public.  For the cached library name, we spilt the passed in name into two parts, the library name and the suffix (.so/.jar) and insert the redefinition timestamp in between.

core/sql/common/ComMisc.cpp
```diff
@@ -357,3 +358,119 @@ NABoolean ComTrafReservedColName(
...
+{
+  NAString libPrefix, libSuffix;
+  struct stat statbuf;
+  NAString redefTimeString = Int64ToNAString(redeftime);
+  size_t lastDot = libname.last('.');
+  if (lastDot != NA_NPOS)
```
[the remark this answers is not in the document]
robertamarton: I was able to create a library that contains special characters: create library "name/sih.char%" file '/mnt/rmarton/gitws/traf-plus';  So an embedded period is allowed.  I am wondering if, at least until we allow more special characters in names,  library names should follow rules of other identifiers and return error 1422 if the specified name contains characters not allow in hbase.   (isValidHbaseName).

core/sql/common/ComMisc.cpp
```diff
@@ -357,3 +358,119 @@ NABoolean ComTrafReservedColName(
...
+ 
+  //when isolated user support is added   
+  if (userid.length()!=0)       
+    {
+
+      cachedLibPath = getenv("TRAF_HOME") ;
+      cachedLibPath += "/udr";
+      if ( stat(cachedLibPath, &statbuf) != 0)
+         {
+           if (mkdir(cachedLibPath,S_IRWXU|S_IRWXG|S_IRWXO))
```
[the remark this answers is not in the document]
robertamarton: Or just pass in a username.  It looks like an empty name is passed.  also, instead of DB__ROOT you could use something like "SYSTEM".

core/sql/common/ComMisc.cpp
```diff
@@ -357,3 +358,119 @@ NABoolean ComTrafReservedColName(
...
+              return -1;
+            }
+               
+        }
+      cachedLibPath += "/" + schemaName;
+      if ( stat(cachedLibPath, &statbuf) != 0)
+         {
+           if (mkdir(cachedLibPath,S_IRWXU|S_IRWXG|S_IRWXO))
+             {
+               return -1;
```
robertamarton: Is it possible to retrieve the actually error on why mkdir failed?  If so, we could return this and make error 4316 more meaningful.  Could return an additional parameter that returns error details, for example:  Error occurred while extracting library to local cache : unable to mkdir <directory name>, error xx

core/sql/common/ComMisc.cpp
```diff
@@ -357,3 +358,119 @@ NABoolean ComTrafReservedColName(
...
+               
+         }
+     
+    }
+      
+  
+  
+  cachedLibName += libPrefix + "_" ;
+  cachedLibName += redefTimeString;
+  cachedLibName += libSuffix ;
```
robertamarton: If the passed in name does not contain a suffix (lastDot is NA_NPOS)? Should that we an error - illformed name or internal error?

core/sql/executor/ExExeUtilLoad.cpp
```diff
@@ -3288,6 +3288,45 @@ short ExExeUtilLobExtractTcb::work()
   return 0;
 }
 
+
+short ExExeUtilLobExtractLibrary(ExeCliInterface *cliInterface,char *libHandle, char *cachedLibName,ComDiagsArea *toDiags)
+{
+  char buf[1000];
```
robertamarton: I assume that cachedLibName won't be greater than 1000 bytes.  You could:  char buf[strlen(cachedLibName) + strlen(libHandle) + 200] so you have a better indication of the size.

core/sql/executor/ExExeUtilLoad.cpp
```diff
@@ -3288,6 +3288,45 @@ short ExExeUtilLobExtractTcb::work()
...
+  Int32 cliRC =0;
+  str_sprintf(buf, "extract lobtofile(LOB '%s','%s');",libHandle,cachedLibName);
+               
+
+  cliRC = cliInterface->fetchRowsPrologue(buf, TRUE/*no exec*/);
+  if (cliRC < 0)
+    {
+      cliInterface->retrieveSQLDiagnostics(toDiags);
```
robertamarton: Now that I have a better understanding of how diags areas work, is toDiags always guaranteed to be not NULL?  That is, should you call allocAndRetrieveSQLDiagnostics?  The reason I mentioned this is that I am in the process of fixing an issue in ExExeUtilGet where the diags pointer is NULL and the code does a retrieveSQLDiags instead of allocAndRetrieve ..

core/sql/executor/ExExeUtilLoad.cpp
```diff
@@ -3288,6 +3288,45 @@ short ExExeUtilLobExtractTcb::work()
...
+  cliRC = cliInterface->clearExecFetchClose(NULL, 0);
+  if (cliRC < 0)
+    {
+      cliInterface->retrieveSQLDiagnostics(toDiags);
+      //Ignore error if the target file exists. This could be because the cached
+      // file already got created by another process at the same time. So we can ignore
+      // the error and use the already cached file.
+      ComCondition *cond = NULL;
+      Int32 entryNumber;
+      cond = toDiags->findCondition(-EXE_ERROR_FROM_LOB_INTERFACE, &entryNumber);
```
robertamarton: Could you get more than one error in diags when the target file exists?  Instead of just removing the error in question, you could mark the diags area before the call and then rewind if the target exists.

core/sql/executor/ExUdr.cpp
```diff
@@ -631,7 +632,10 @@ Int32 ExUdrTcb::fixup()
 
   UdrDebug1("[BEGIN TCB FIXUP] %p", this);
   setUdrTcbState(FIXUP);
-    
+  // CliGlobals *cliGlobals = getGlobals()->castToExExeStmtGlobals()->
+  // castToExMasterStmtGlobals()->getCliGlobals();
+  CliGlobals *cliGlobals = GetCliGlobals();
```
robertamarton: nit: remove commented lines?

core/sql/executor/ExUdr.cpp
```diff
@@ -792,9 +796,57 @@ Int32 ExUdrTcb::fixup()
...
+      if ((myTdb().getLibraryRedefTime() != -1) && (myTdb().getLibraryRedefTime() != 0))
+        {
+          // Cache library locally. 
+          NAString dummyUser;
+          NAString libOrJarName;
+        
+          if (myTdb().getLanguage() == COM_LANGUAGE_JAVA)
+            libOrJarName = myTdb().getPathName();
+          else
+            libOrJarName = myTdb().getContainerName();
```
robertamarton: Are there any special considerations for SPSQL jars?

core/sql/executor/ExUdr.cpp
```diff
@@ -792,9 +796,57 @@ Int32 ExUdrTcb::fixup()
...
+          if(ComGenerateUdrCachedLibName(libOrJarName.data(),
+                                         myTdb().getLibraryRedefTime(),
+                                         myTdb().getLibrarySchName(),
+                                         dummyUser,
+                                         cachedLibName, cachedLibPath))
+            {
+              NAString cachedFullName = cachedLibPath+"/"+cachedLibName;
+                  *getOrCreateStmtDiags() <<  DgSqlCode(-4316)
+                                << DgString0(( char *)cachedFullName.data());;
```
robertamarton: As briefly mentioned this earlier.  One thing that has been discussed is to have better error messages that would make debugging customer problems easier.  In this case, instead of just saying unable to extract library, you could mention that is was because we were unable to create the directory structure and include details returned from mkdir.

core/sql/generator/GenUdr.cpp
```diff
@@ -1567,7 +1579,10 @@ IsolatedScalarUDF::codeGen(Generator *generator)
                            NULL);              // TMUDF only
   
   if (effectiveMetadata.getRoutineID() > 0)
-    generator->objectUids().insert(effectiveMetadata.getRoutineID());
+    {
+      generator->objectUids().insert(effectiveMetadata.getRoutineID());
+      generator->objectUids().insert(metadata.getLibObjUID());
```
robertamarton: Did you want ... insert (effectiveMetadata.getLibObjUID())?

core/sql/generator/Generator.cpp
```diff
@@ -2352,6 +2352,14 @@ TrafDesc *Generator::createVirtualRoutineDesc(
...
    routine_desc->routineDesc()->librarySqlName = new GENHEAP(space) char[strlen(routineInfo->library_sqlname)+1];
    strcpy(routine_desc->routineDesc()->librarySqlName, routineInfo->library_sqlname);
+   routine_desc->routineDesc()->libRedefTime = routineInfo->lib_redef_time;
+   routine_desc->routineDesc()->libBlobHandle = routineInfo->lib_blob_handle;
+
+   routine_desc->routineDesc()->libVersion = routineInfo->library_version;
+   routine_desc->routineDesc()->libObjUID = routineInfo->lib_obj_uid;
+   //routine_desc->routineDesc()->libSchName = new GENHEAP(space) char[strlen(routineInfo->lib_sch_name)+1];
+   //strcpy(routine_desc->routineDesc()->libSchName ,routineInfo->lib_sch_name);
+   routine_desc->routineDesc()->libSchName = routineInfo->lib_sch_name;
```
robertamarton: Remove comments?  Do you need a deep copy for lib_sch_name?

core/sql/optimizer/UdfDllInteraction.cpp
```diff
@@ -136,6 +137,64 @@ NABoolean TMUDFDllInteraction::describeParamsAndMaxOutputs(
...
+    {
+      // Cache library locally. 
+      NAString dummyUser;
+      NAString libOrJarName;
+      NAString cachedLibName,cachedLibPath;  
+      if (routine->getLanguage() == COM_LANGUAGE_JAVA)
+        libOrJarName = routine->getExternalPath();
+      else
+        libOrJarName = routine->getContainerName();
+      if(ComGenerateUdrCachedLibName(libOrJarName.data(),
```
robertamarton: This  code looks a lot like ExUdr.cpp.  Is it similar enough to create a subprocedure that both can call?

core/sql/regress/tools/rgrCleanup
```diff
@@ -0,0 +1,4 @@
+#temp solution to cleanup udr cached files
+#can be removed when cleanup mechanism for cached files in implemented
+echo "rm -rf $TRAF_HOME/udr/$MY_UDR_CACHE_LIBDIR/* 2>$NULL"
+rm -rf $TRAF_HOME/udr/$MY_UDR_CACHE_LIBDIR/* 2>$NULL
```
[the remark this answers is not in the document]
robertamarton: If the location of TRAF_HOME changes then we may want to clean up the previous location.  Perhaps we could include this step somehow when upgrading to a new version.   I see you added it to runregr, let's talk I may have some ideas...

core/sql/sqlcat/TrafDDLdesc.cpp
```diff
@@ -603,7 +603,7 @@ Lng32 TrafRoutineDesc::unpack(void * base, void * reallocator)
   librarySqlName = (librarySqlName ? (char*)((char*)base - (Long)librarySqlName) : NULL);
   libraryFileName = (libraryFileName ? (char*)((char*)base - (Long)libraryFileName) : NULL);
   signature = (signature ? (char*)((char*)base - (Long)signature) : NULL);
-
+  libBlobHandle = (libBlobHandle ? (char*)((char*)base - (Long)libBlobHandle) : NULL);
```
robertamarton: Should we also be including the schema name?

core/sql/sqlcomp/CmpDDLCatErrorCodes.h
```diff
@@ -69,7 +69,7 @@ enum CatErrorCode { CAT_FIRST_ERROR = 1000
                   , CAT_CATALOG_ALREADY_EXISTS                    = 1035
                   , CAT_CIRCULAR_PRIVS                            = 1036
                   , CAT_DEPENDENT_PRIV_EXISTS                     = 1037
-                  // unused                                       = 1038
+                  , CAT_UNABLE_TO_UPGRADE                         = 1037
```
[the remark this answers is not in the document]
robertamarton: I assume you want this to be 1038?

core/sql/sqlcomp/CmpSeabaseDDLinitraf.cpp
```diff
@@ -692,7 +675,62 @@ short CmpSeabaseDDL::initTrafMD(CmpDDLwithStatusInfo *dws)
...
+                    {
+                      setValuesInDWS(dws, IT_STEP_FAILED,
+                                     "Update MDSPJ : Failed", 0, TRUE,
+                                     FALSE, TRUE, TRUE);                      
+
+                      return 0;
+                    }
+                  setValuesInDWS(dws, IT_CREATE_PRIVMGR_REPOS,
+                                 "Create Libraries Tables: Completed", 0, TRUE,
+                                 FALSE, TRUE, TRUE);
```
robertamarton: Note to myself, since libraries are upgraded before privmgr, I may be able to make privmgr upgrade faster since it doesn't have to deal with library privs - to check

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -389,12 +388,245 @@ void CmpSeabaseDDL::createSeabaseLibrary(
...
               createLibraryNode->getVersion());
+    
+ 
+  Lng32 cliRC = cliInterface.executeImmediate(query);
+
+  NADELETEBASIC(query, STMTHEAP);
+  if (cliRC < 0)
+    {
+      cliInterface.retrieveSQLDiagnostics(CmpCommon::diags());
+      processReturn();
```
robertamarton: dallocEHI?

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -389,12 +388,245 @@ void CmpSeabaseDDL::createSeabaseLibrary(
...
+  
+  return 0;
+  NADELETEBASIC(query, STMTHEAP);
+ 
+}
+
+void CmpSeabaseDDL::createSeabaseLibrary2(
+				      StmtDDLCreateLibrary * createLibraryNode,
+				      NAString &currCatName, 
+                                      NAString &currSchName)
```
robertamarton: There is a lot of duplicate code between createSeabaseLibrary and createSeabaseLibrary2.  In a future release, maybe these can be refactored.

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -563,32 +795,332 @@ void CmpSeabaseDDL::dropSeabaseLibrary(StmtDDLDropLibrary * dropLibraryNode,
...
+      processReturn();
+      return;
+    }
+
+  if (!isDDLOperationAuthorized(SQLOperation::DROP_LIBRARY,
+                                objectOwnerID,
+                                schemaOwnerID))
+  {
+     *CmpCommon::diags() << DgSqlCode(-CAT_NOT_AUTHORIZED);
+     processReturn ();
```
robertamarton: deallocEHI?  Probably in existing dropSeabaseLibrary2 code also.

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -697,6 +1229,26 @@ void  CmpSeabaseDDL::alterSeabaseLibrary(StmtDDLAlterLibrary  *alterLibraryNode,
...
+  Int64 libUID = 0;
+  short retcode = 0;
+      if (stat(cachedLibName, &statbuf) != 0)
+        {
+          retcode =  ExExeUtilLobExtractLibrary(cliInterface, libHandle, cachedLibName, 
+                                                CmpCommon::diags());
+          if (retcode < 0)
+            {
+              *CmpCommon::diags() <<  DgSqlCode(-4316)
+                                  << DgString0(cachedLibName);
```
robertamarton: Again, if we could return why unable to extract the library - useful for debugging issues.

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -804,13 +1362,15 @@ void CmpSeabaseDDL::createSeabaseRoutine(
...
+  
+  // read the library  name from the LIBRARIES metadata table
   char * buf = new(STMTHEAP) char[200];
-  str_sprintf(buf, "select library_filename from %s.\"%s\".%s"
+ 
+  str_sprintf(buf, "select library_filename, library_storage from %s.\"%s\".%s"
               " where library_uid = %ld for read uncommitted access",
               getSystemCatalog(), SEABASE_MD_SCHEMA, SEABASE_LIBRARIES, libUID);
-
+
```
robertamarton: Would we ever get here if library_storage did not exist in the LIBRARIES table?

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -839,21 +1399,27 @@ void CmpSeabaseDDL::createSeabaseRoutine(
...
+  cliInterface.getPtrAndLen(2, ptr, len);
+  str_cpy_all(libBlobHandle, ptr, len);
+  libBlobHandle[len] = '\0'; 
+    
+  
+  NAString extPath(externalPath);
+  size_t lastDot = extPath.last('.');
+     
+  
+  if (lastDot != NA_NPOS)
```
robertamarton: Here is another place that would potentially fail if the library name contained an embedded period.  Should an internal error be generated if there is no period in the name?

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -995,180 +1561,293 @@ void CmpSeabaseDDL::createSeabaseRoutine(
...
+
+      if( libBlobHandle[0] != '\0' )
+        {
+          NAString dummyUser;
+          NAString cachedLibName, cachedLibPath;
+          
+          if(ComGenerateUdrCachedLibName(extPath,redefTime,libSchNamePart,dummyUser, cachedLibName, cachedLibPath))
+            {
+              *CmpCommon::diags() << DgSqlCode(-1231)
+                                  << DgString0(extRoutineName);
```
robertamarton: There is a literal for this: CAT_UNABLE_TO_CREATE_ROUTINE.  This is another error that could fail for various reasons and it could be improved to return more details.

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -995,180 +1561,293 @@ void CmpSeabaseDDL::createSeabaseRoutine(
...
+                  containerName = externalPrefix(separatorPos+1,
+                                                 externalPrefix.length()-separatorPos-1);
+                  externalPrefix.remove(separatorPos,
+                                        externalPrefix.length()-separatorPos);
+                }
+              else
+                {
+                  // assume the entire string is a local name
+                  containerName = externalPrefix;
+                  externalPrefix = ".";
```
robertamarton: This adds a period to the externalPrefix, however, if a slash exists, the period is not included.  Is this intentional?

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -995,180 +1561,293 @@ void CmpSeabaseDDL::createSeabaseRoutine(
...
+               extLibraryName.data(),
+               &routineHandle,
+               CmpCommon::diags()) != LME_ROUTINE_VALIDATED)
+        {
+          if (routineHandle != NullCliRoutineHandle)
+            cliInterface.putRoutine(routineHandle,
+                                    CmpCommon::diags());
+
+          CMPASSERT(CmpCommon::diags()->mainSQLCODE() < 0);
+          processReturn();
```
robertamarton: deallocEHI?

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -1749,7 +2431,124 @@ short CmpSeabaseDDL::grantLibmgrPrivs(ExeCliInterface *cliInterface)
...
+           getSystemCatalog(),SEABASE_MD_SCHEMA, SEABASE_LIBRARIES, dllLocation.data(),
+           getSystemCatalog(),SEABASE_MD_SCHEMA, SEABASE_OBJECTS, SEABASE_LIBMGR_LIBRARY_CPP);
+  CMPASSERT(stmtSize < sizeof(queryBuf));
+
+  cliRC = cliInterface->executeImmediate(queryBuf);
+  if (cliRC < 0)
+    {
+      cliInterface->retrieveSQLDiagnostics(CmpCommon::diags());
+      return -1;
+    }
```
robertamarton: Is it possible to remove the library path updates for system libraries?

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -1925,3 +2730,568 @@ short CmpSeabaseDDL::createSeabaseLibmgrCPPLib(ExeCliInterface * cliInterface)
...
+        
+            return 0;
+          }
+          break;
+
+        case 2:
+          {
+            // drop old libraries
+            if (dropLibraries(cliInterface, TRUE/*old */))
+              return -3;  // error, but no recovery needed
```
robertamarton: unable to drop old libraries, it is okay - that is because it will just use up space but not cause any difficulties?

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -1925,3 +2730,568 @@ short CmpSeabaseDDL::createSeabaseLibmgrCPPLib(ExeCliInterface * cliInterface)
...
+          break;
+
+        case 8:
+          {
+            // copy old contents into new 
+           
+            if (copyOldLibrariesToNew(cliInterface))
+              {
+                mdui->setMsg(" Copy Old Libraries failed ! Drop  and recreate the following :   ");
+                //return -1;  // error, need to drop new libraries then undo rename
```
robertamarton: Should this be commented out?

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -1925,3 +2730,568 @@ short CmpSeabaseDDL::createSeabaseLibmgrCPPLib(ExeCliInterface * cliInterface)
...
+short CmpSeabaseDDL::createLibraries(ExeCliInterface * cliInterface)
+{
+ Lng32 cliRC = 0;
+
+  char queryBuf[20000];
+
+  NABoolean xnWasStartedHere = FALSE;
+
+
+  for (Int32 i = 0; i < sizeof(allLibrariesUpgradeInfo)/sizeof(MDUpgradeInfo); i++)
```
robertamarton: Since we are not adding any libraries, this code won't be exercised?

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -1925,3 +2730,568 @@ short CmpSeabaseDDL::createSeabaseLibmgrCPPLib(ExeCliInterface * cliInterface)
...
+  return errcode;
+}
+
+
+short CmpSeabaseDDL::alterRenameLibraries(ExeCliInterface * cliInterface,
+                                          NABoolean newToOld)
+{
+ Lng32 cliRC = 0;
+
+  char queryBuf[10000];
```
robertamarton: 10000 seems a little big for an alter library statement.

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
```diff
@@ -1925,3 +2730,568 @@ short CmpSeabaseDDL::createSeabaseLibmgrCPPLib(ExeCliInterface * cliInterface)
...
+      if ((! lti.newName) || (! lti.oldName) || (NOT lti.upgradeNeeded))
+        continue;
+
+      if (newToOld)
+        str_sprintf(queryBuf, "alter table %s.\"%s\".%s rename to %s ; ",
+                    getSystemCatalog(), SEABASE_MD_SCHEMA, lti.newName, lti.oldName);
+      else
+        str_sprintf(queryBuf, "alter table %s.\"%s\".%s rename to %s ; ",
+                    getSystemCatalog(), SEABASE_MD_SCHEMA, lti.oldName, lti.newName);
+
```
robertamarton: Outside of your changes but does a rename table retain any views created on table?  How about any privileges granted to the table?

core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
```diff
@@ -5619,8 +5620,80 @@ void CmpSeabaseDDL::alterSeabaseTableAddColumn(
       processReturn();
       return;
      }
-  char * col_name = new(STMTHEAP) char[colName.length() + 1];
+  */
+  char *col_name = new(STMTHEAP) char[colName.length() + 1];
   strcpy(col_name, (char*)colName.data());
+ if ((datatype == REC_BLOB) ||
```
robertamarton: Now supporting alter table add column <blob>, require a documentation change?

core/sql/sqlcomp/CmpSeabaseDDLtable.cpp
```diff
@@ -5619,8 +5620,80 @@ void CmpSeabaseDDL::alterSeabaseTableAddColumn(
...
+     memset(lobHdfsServer,0,256);
+     strncpy(lobHdfsServer,CmpCommon::getDefaultString(LOB_HDFS_SERVER),sizeof(lobHdfsServer)-1);
+     Int32 lobHdfsPort = (Lng32)CmpCommon::getDefaultNumeric(LOB_HDFS_PORT);
+      Int32 rc = sendAllControls(FALSE, FALSE, TRUE);
+      
+      Int64 objUID = getObjectUID(&cliInterface,
+                                  catalogNamePart.data(), schemaNamePart.data(), 
+                                  objectNamePart.data(),
+                                  COM_BASE_TABLE_OBJECT_LIT);
+
```
robertamarton: Is it possible that an invalid objectUID is returned?  For example, an error occurred reading OBJECT table.  What would happen if objectUID = -1?

core/sql/sqlcomp/CmpSeabaseDDLupgrade.cpp
```diff
@@ -314,7 +326,7 @@ short CmpSeabaseMDupgrade::executeSeabaseMDupgrade(CmpDDLwithStatusInfo *mdui,
   // now as they are stateless
   CmpSeabaseUpgradeRepository upgradeRepository;
   CmpSeabaseUpgradePrivMgr upgradePrivMgr;
-
+  CmpSeabaseUpgradeLibraries upgradeLibraries;
```
robertamarton: Since we create repository, libraries, and then privmgr, perhaps upgradeLibraries should come before upgradePrivMgr?

core/sql/sqlcomp/CmpSeabaseDDLupgrade.cpp
```diff
@@ -1637,6 +1660,43 @@ short CmpSeabaseMDupgrade::executeSeabaseMDupgrade(CmpDDLwithStatusInfo *mdui,
...
+                mdui->setStep(UPGRADE_PRIV_MGR);
+                mdui->setSubstep(0);
+                break;
+              }
+
+            if (xnInProgress(&cliInterface))
+              {
+                *CmpCommon::diags() << DgSqlCode(-20123);
+                
+                mdui->setStep(UPGRADE_FAILED_RESTORE_OLD_REPOS);
```
robertamarton: Should this message mention failure during upgrade library step?

core/sql/sqlcomp/CmpSeabaseDDLupgrade.cpp
```diff
@@ -1657,7 +1717,7 @@ short CmpSeabaseMDupgrade::executeSeabaseMDupgrade(CmpDDLwithStatusInfo *mdui,
 		    {
 		      *CmpCommon::diags() << DgSqlCode(-20123);
 
-		      mdui->setStep(UPGRADE_FAILED_RESTORE_OLD_REPOS);
+		      mdui->setStep(UPGRADE_FAILED_RESTORE_OLD_LIBRARIES);
```
robertamarton: We have already completed the upgrade library step, should this step (and others that follow) indicate we were in the update version step?  In a subsequent release, we may not be upgrading libraries but something else.

core/sql/sqlcomp/CmpSeabaseDDLupgrade.cpp
```diff
@@ -1739,7 +1799,31 @@ short CmpSeabaseMDupgrade::executeSeabaseMDupgrade(CmpDDLwithStatusInfo *mdui,
...
+	  {
+            if (upgradeLibraries.needsUpgrade(this))
+              {
+                if (upgradeLibraries.doDrops(&cliInterface,mdui,this))
+                  {
+                    // no status message in this case so no return
+                    cliInterface.clearGlobalDiags();
+                    mdui->setStep(OLD_REPOS_DROP);
+                    mdui->setSubstep(0);
+                    mdui->setEndStep(TRUE);
```
robertamarton: In this step, we fail to drop the old libraries.  Do we just ignore any errors?

core/sql/sqlcomp/CmpSeabaseDDLupgrade.cpp
```diff
@@ -1901,7 +1985,50 @@ short CmpSeabaseMDupgrade::executeSeabaseMDupgrade(CmpDDLwithStatusInfo *mdui,
...
+                      }
+                  }
 
+                if (upgradeLibraries.doUndo(&cliInterface,mdui,this))
+                  {
+                    // ignore errors; no status message so just continue on
+                    cliInterface.clearGlobalDiags();
+                    mdui->setStep(OLD_REPOS_DROP);
+                    mdui->setSubstep(0);
+                    mdui->setEndStep(TRUE);
```
robertamarton: again, if we can't undo, what happens?

robertamarton:
Very nice job.

core/sqf/sqenvcom.sh
```diff
@@ -200,6 +207,7 @@ export DTM_COMMON_JAR=trafodion-dtm-cdh-${TRAFODION_VER}.jar
 export SQL_JAR=trafodion-sql-cdh-${TRAFODION_VER}.jar
 export UTIL_JAR=trafodion-utility-${TRAFODION_VER}.jar
 export JDBCT4_JAR=jdbcT4-${TRAFODION_VER}.jar
+export MY_UDR_CACHE_LIBDIR=cached_libs
```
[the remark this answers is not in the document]
sandhyasun: Ok I can change both and remove the MY_

core/sql/sqlcomp/CmpDDLCatErrorCodes.h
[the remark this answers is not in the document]
sandhyasun: Leftover from the temp upgrade we had done earlier. Don't need this error anymore...

core/sql/common/ComMisc.cpp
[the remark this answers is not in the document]
sandhyasun: Kept it separate due to the permission issue - for Isolated User support  I assumed we needed stricter access - i.e read/write/execute for user only and no one else.

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
sandhyasun: I'v eadded QI support for alter library. So when an alter library is done, the objUid of the library is added to the list of invalidationKeys. So the prepare by process 'a' will get invalidated . So we avoid the issue of a stale library being executed .

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
sandhyasun: Included the system libraries in this upgrade. So they will also have an entry in the blob volumn. There is one exception to this - validateRoutine that would show up as a "failed" upgrade but it doesn't prevent the upgrade operation.

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
sandhyasun: Yes this is creating the LIBRARIES table only.

core/sql/comexe/ComTdbUdr.h
sandhyasun: Just has to be a multiple of 8. These were more relevant when we had  versioning support. It's not the case now but we try to keep some filler anyway.

core/sql/sqlcomp/CmpSeabaseDDLupgrade.cpp
sandhyasun: We ar ein the phase of rewinding the upgrade. Unless the "needsUpgrade" flag is set, the restore steps wont do anything. So we are ok. We just need to start with last step and rewind but some of the in between steps may not do any rewinding.

core/sql/sqlcomp/CmpSeabaseDDLupgrade.cpp
sandhyasun: Yes all the drop steps ignore errors.

core/sql/sqlcomp/CmpSeabaseDDLupgrade.cpp
sandhyasun: This is just like what happens with repository upgrade undo.

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
zellerh: Thanks for the explanation.

core/sql/sqlcomp/CmpSeabaseDDLroutine.cpp
zellerh: Couldn't there still be a short window where one query is executing and another process upgrades the library? If so, that's probably ok, mostly wondering whether I understand this correctly.

sandhyasun: Maybe a small window where the alter of the library has finished and another process has started executing but  not yet extracted the library to cache.  Not sure what we can do about that. If the library is already cached it'll just use the old copy.  I think we'll need to simulate this and address it separately.. If needed we may need to maintain the old version of the library.

core/sql/common/ComMisc.cpp
sandhyasun: Good idea - added errno to the error message

sandhyasun:
Made some fixes based on review comments.

PR #1722, opened 2018-09-29
[trafodion-3211]refactor the usage of setDiagsArea()

wenjun-zhu:
Two points:
- append a letter `x` to the function name setDiagsArea()
- add a new function shareDiagsArea(), which increase the refCount of the source ComDiagsArea.

This PR serves as a code base for further discussion. And the appendix letter `x` would be removed before merging.


DaveBirdsall:
Looks like compGeneral/TEST006 might be coring.

Might want to run full regressions on a workstation for this change.

sandhyasun:
I don't see any comments in the JIRA on the change description. So trying to summarize here with a question .... Is the change to simply change all places (in ex)queu class and ExAtp class) where we do a setDiagsArea(da)  followed by da->incrRefCount()  to be done with one descriptive method - shareDiagsArea(da) ?

PR #1724, opened 2018-10-09
[TRAFODION-3220] Executor GUI

zellerh:
Adding the ability to display query execution with the "display"
command in Trafodion (sqlci only, requires an X Windows display,
just as the compiler GUI does).


core/sql/SqlCompilerDebugger/ExeQueueDisplay.h
```diff
@@ -33,22 +18,43 @@
 // under the License.
 //
 // @@@ END COPYRIGHT @@@
- *
- *
- *****************************************************************************
- */
+#ifndef EXEQUEUEDISPLAY_H
```
zellerh: I have no clue why git says I renamed ComSqlexedbg.h to ExeQueueDisplay.h. In reality, I deleted file ComSqlexedbg.h and added a new file ExeQueueDisplay.h. Please ignore the the ComSqlexedbg.h file, it doesn't have any meaningful content.

sureshsubbiah:
+1. This is a wonderful change Hans. Can't wait to try it out.

selvaganesang:
Is this feature available in debug mode build only?

zellerh:
@selvaganesang, yes, the executor GUI is available in debug builds only, for performance reasons. The compiler GUI should work for both debug and release builds. This is controlled by the executor Makefile.

PR #1727, opened 2018-10-16
[TRAFODION-3222] Windows odbc build fail

Weixin-Xu:
1. header file "cconnect.h" missed in "sqltocconv.h"
2. In README.txt, vs2013 is recommended to build windows odbc but now only the version is set vs2015 or higher.
3. update windows odbc and odb version to 2.3


hegdean:
Changes are  ok. Please update the installer  file with the correct release number

Weixin-Xu:
I have updated the version and test pass

sbroeder:
Changes look good, but the copyright year needs to be updated to 2018.

PR #1728, opened 2018-10-16
[TRAFODION-2626] Make logs directory location configurable

svarnau:
Makes path to log directory independent of TRAF_HOME. The default location
is still $TRAF_HOME/logs. The DCS and REST log directories are now
sub-directories of the main TRAF_LOG directory.

Changes were pretty straight-forward, except the monitor stdout_* files.  Someone familiar with foundation should have a look.

Tested in local_hadoop environment, with TRAF_LOG set to alternate path and normal logs directory blocked.

### DIFF
--- a/core/sql/sqludr/SqlUdrPredefLogReader.cpp
+++ b/core/sql/sqludr/SqlUdrPredefLogReader.cpp
@@ -533,7 +533,7 @@ void ReadCppEventsUDFInterface::processData(UDRInvocationInfo &info,
 
     logrootdir = getenv("TRAF_LOG");
     if (strlen(logrootdir) > 1000)
-	throw UDRException(38001, "TRAF_HOME is longer than 1000 characters");
+	throw UDRException(38001, "TRAF_LOG is longer than 1000 characters");
     std::string logDirName(logrootdir);
 
     switch (logLocationIndex) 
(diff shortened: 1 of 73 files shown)


core/sql/sqludr/SqlUdrPredefLogReader.cpp
```diff
@@ -530,37 +530,28 @@ void ReadCppEventsUDFInterface::processData(UDRInvocationInfo &info,
   {
     char* logrootdir = NULL;
     char* confrootdir = NULL;
+
+    logrootdir = getenv("TRAF_LOG");
+    if (strlen(logrootdir) > 1000)
+	throw UDRException(38001, "TRAF_HOME is longer than 1000 characters");
```
DaveBirdsall: Should say, TRAF_LOG is longer than 1000 characters

core/sql/sqludr/SqlUdrPredefLogReader.cpp
```diff
@@ -530,37 +530,28 @@ void ReadCppEventsUDFInterface::processData(UDRInvocationInfo &info,
...
+      logDirName += "/dcs";
       break ;
     case 2:
-      logrootdir = getenv("REST_INSTALL_DIR");
-      if (!logrootdir)
-	throw UDRException(38001, "REST_INSTALL_DIR not set");
-      else if (strlen(logrootdir) > 1000)
-	throw UDRException(38001, "REST_INSTALL_DIR is longer than 1000 characters");
-      break ;
+      logDirName += "/rest";
```
DaveBirdsall: You need a "break;" statement after this one, otherwise you fall through to the default case and raise an exception. This is probably why test udr/TEST002 failed in the regression test run.

PR #1736, opened 2018-11-02
[TRAFODION-3226] SQLGetTypeInfo returns incorrect precision of int

Weixin-Xu:
SQLGetTypeInfo(hstmt, SQL_INTEGER);
When the returned type is INTEGER and INTEGER SIGNED, the column width value is expected to be 10, but it is actually 9.


DaveBirdsall:
@Weixin-Xu, do you want to try committing this yourself (now that you have committer privileges) or would you like me to commit it for you?

Weixin-Xu:
@DaveBirdsall Actually I do not have the committer privileges. If you can, please commit it for me.
Thanks.

PR #1739, opened 2018-11-07
[TRAFODION-3230] The number of error message is less than the failed input rows

Weixin-Xu:
the parameter of SQLGetDiagRec need to be initialized

### DIFF
--- a/core/conn/odb/src/odb.c
+++ b/core/conn/odb/src/odb.c
@@ -10087,6 +10087,7 @@ static int Oloadbuff(int eid)
                 lcr += (unsigned long)etab[eid].Oresl;
             break;
         default:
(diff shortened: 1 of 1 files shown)


core/conn/odb/src/odb.c
```diff
@@ -10183,6 +10183,7 @@ static int Oloadbuff(int eid)
                 }
                 Oi++;
             }
+            Oi = 1; //the record number need to be initialize for next loop
```
DaveBirdsall: This is fine, but it might be a bit clearer to the reader if the initialization was at the beginning of the loop, for example at line 10090.

Weixin-Xu: Thanks, I have updated the change.

PR #1741, opened 2018-11-12
[TRAFODION-3190] expression involving NULL should be treated as FALSE, not TRUE.

wenjun-zhu:


core/sql/exp/exp_bool.cpp
```diff
@@ -96,7 +96,7 @@ ex_expr::exp_return_type ex_branch_clause::eval(char *op_data[],
   switch (getOperType())
     {
     case ITM_AND:
-      if (*(Lng32 *)op_data[1] == 0)
+      if (*(Lng32 *)op_data[1] == 0 || *(Lng32 *)op_data[1] == -1)  // null treated as false
```
DaveBirdsall: I don't believe this is correct. Consider the query, "select a from t1x where not(b = 0 and c = 0)". When B and C are both null, both equal predicates evaluate to null, and the AND evaluates to null. The NOT then also evaluates to null. The WHERE clause should treat the result of the NOT as false. But with this fix, the result of the AND will be false, making the NOT true. There needs to be three cases here for ITM_AND: If the first operand is false, then the AND is false. If the first operand is true, then the result is the second operand. If the first operand is null, then if the second operand is false, the result is false otherwise the result is null. Similar logic needs to be added to the ITM_OR case.

anoopsharma00:
Dave is correct. Thanks for pointing it out.

Branch clause need to evaluate the second leg if the first leg is 'not false' which means
it is 'true or null'. With this change, it will branch out by treating null as false.

Similar comments were added to an earlier PR 1705 for this same jira
and those comments apply here as well.

For the original issue where CASE WHEN stmt was not proceeding to ELSE leg, a change
need to be added so boolean result is evaluated to at the very end of  THEN clause.
At that time, NULL can be treated as FALSE and control to move to ELSE part.
This is same as a WHERE predicate where only at the end of predicate evaluation can
one determine if the result is true or false/null.

PR #1746, opened 2018-11-15
[TRAFODION-3234] Refactor hive meta calls to be less resource intensive to support hive partitions if needed	

selvaganesang:
Refactored the interaction with hive metastore to use the efficient APIs to obtain the necessary
information to construct the hive table descriptors. This refactoring is expected to reduce the
memory requirements, efficient and high performing even when there are thousands of partitions.

Expiration Timestamp concept to detect the change in hive definition or hive file timestamp is now removed. Data modification check was unnecessarily done in the scan/insert operator. This has been removed. Redefine timestamp is now updated when the hive statistics is obtained.

Prior to this change

Timestamp of the directory is obtained many times resulting in increased path length at the time of compilation and as well as in execution.
The expiration timestamp concept of hive meta data information for both definition and statistics results in non-deterministic behavior in some tests when the system is loaded in Jenkins.
The current concept:

During the compilation of the query:
a)	When the hive table is not found in NATable cache,
The table and partition descriptors are obtained
Currently hive table statistics is obtained via libhdfs calls including the listing of files in partitions and highest directory time stamp of the partition directories.
This is recorded as redefinition timestamp in the descriptor.
b)	When the hive table is found in NATable cache,
The table definition is validated always to determine if there is any change in the definition of the hive table.
The highest directory time stamp is compared with redefinition timestamp to determine if there are any new files added or removed.
If so, this table along with statistics is removed from NATable cache and steps in a) are executed again. If not, the NATable info is good enough to be used in query plans.

During the execution of the query:
a)	Just before execute, similarity check is done to verify if there is any change in the directory timestamp. If so, the query will be AQR-ed just as it was done before.
b)	AQR is expected to remove the query from the query cache and re-prepare the query.
c)	This step is always done irrespective of the query is prepared recently for execDirect or picked up from query cache or a user prepared query.


core/sql/executor/HiveClient_JNI.cpp
```diff
@@ -42,17 +43,16 @@ static const char* const hvcErrorEnumStr[] =
  ,"Java exception in close()."
```
DaveBirdsall: Should there also be exception text for ,HVC_ERROR_INIT_EXCEPTION?

core/sql/executor/HiveClient_JNI.cpp
```diff
@@ -42,17 +43,16 @@ static const char* const hvcErrorEnumStr[] =
...
  ,"Preparing parameters for getRedefTime()."
  ,"Java exception in getRedefTime()."
  ,"Java exception in getAllSchemas()."
  ,"Preparing parameters for getAllTables()."
  ,"Java exception in getAllTables()."
  ,"Preparing parameters for executeHiveSQL()."
  ,"Java exception in executeHiveSQL()."
+ ,"Preparing parameters for getHiveTableInfo()."
+ ,"Java exception in getHiveTableInfo()."
+ ,"Error in getHiveTableInfoDetails()."
```
DaveBirdsall: Should there also be error text for HVC_ERROR_POPULATE_SDS_ERROR? More generally, the entries in this array of strings do not match in number those of the enum HVC_RetCode in HiveClient_JNI.h.

core/sql/executor/HiveClient_JNI.cpp
```diff
@@ -68,6 +68,27 @@ char* HiveClient_JNI::getErrorText(HVC_RetCode errEnum)
     return (char*)hvcErrorEnumStr[errEnum-HVC_FIRST-1];
 }
 
+//////////////////////////////////////////////////////////////////////////////
+// 
+//////////////////////////////////////////////////////////////////////////////
+HiveClient_JNI* HiveClient_JNI::newInstance(NAHeap *heap, HVC_RetCode &retCode)
+{
+   QRLogger::log(CAT_SQL_HDFS, LL_DEBUG, "HiveClient_JNI::newInstance() called.");
```
DaveBirdsall: Just curious: Why do we use CAT_SQL_HDFS in this QRLogger::log call, but elsewhere in this module we use CAT_SQL_HBASE? (Perhaps this use is correct and the others are all wrong?)

core/sql/optimizer/hiveHook.h
```diff
@@ -140,31 +145,42 @@ struct hive_sd_desc
...
+        location_ = strduph(loc, heap_);
+     else
+        location_ = NULL;
+     inputFormat_ = strduph(ift, heap_);
+     outputFormat_= strduph(of, heap_);
+     nullFormat_ = (nf ? strduph(nf, heap_) : NULL);
+     if (pColVals)
+       partitionColValues_ = strduph(pColVals, heap_);
+     else
+       partitionColValues_ = NULL;
```
DaveBirdsall: The code is correct. It's interesting that the styles are mixed though: For nf, we use a ternary operator and for loc and pColVals we use if/then/else.

core/sql/sqlcomp/CmpDescribe.cpp
```diff
@@ -2355,10 +2355,134 @@ short CmpDescribeHiveTable (
       NAString colString(buf);
       Int32 j = i;
       outputColumnLine(space, colString, j);
-    }
+  }
 
   outputShortLine(space, "  )");
 
+  // show hive table partitions and buckets defination.
```
DaveBirdsall: Nit. Typo: defination (should be "definition")

core/sql/sqlcomp/CmpDescribe.cpp
```diff
@@ -2355,10 +2355,134 @@ short CmpDescribeHiveTable (
...
+  // this default schema name is what the Hive default schema is called in SeaHive
```
DaveBirdsall: What's SeaHive?

core/sql/sqlcomp/CmpDescribe.cpp
```diff
@@ -2355,10 +2355,134 @@ short CmpDescribeHiveTable (
...
+  HiveMetaData* md = bindWA.getSchemaDB()->getNATableDB()->getHiveMetaDB();
+  NAString defSchema = ActiveSchemaDB()->getDefaults().getValue(HIVE_DEFAULT_SCHEMA);
+  defSchema.toUpper();
+  struct hive_tbl_desc* htbl;
```
DaveBirdsall: This variable doesn't seem to be used anywhere. Can we delete it?

core/sql/executor/HiveClient_JNI.cpp
selvaganesang: Yes. Fixed it

core/sql/executor/HiveClient_JNI.cpp
selvaganesang: Yes. Corrected and  ensured that it matches

core/sql/sqlcomp/CmpDescribe.cpp
selvaganesang: Fixed

core/sql/sqlcomp/CmpDescribe.cpp
selvaganesang: Changed the comment line entirely

core/sql/sqlcomp/CmpDescribe.cpp
selvaganesang: Deleted

core/sql/executor/HiveClient_JNI.cpp
selvaganesang: Let me try to get rid of logging unless it makes sense

selvaganesang:
@dave and @sureshsubbiah  Can you please review the commit 3 and 4. I have fixed some more issues along with taking care of your review comments

PR #1747, opened 2018-11-15
[TRAFODION-3235] add div and trim function for map feature, improve bad file print function.

SuJinpei:
Some customer asks for this feature and these feature has been tested.

### DIFF
--- a/core/conn/odb/src/odb.c
+++ b/core/conn/odb/src/odb.c
@@ -6249,7 +6249,7 @@ static void Oload(int eid)
         char op;                /* 1=substr, 2=dconv, 3=tconv, 4=tsconv, 5=replace,
                                    6=toupper, 7=tolower, 8=firstup, 9=csubstr, 10=translit,
                                    11=comp, 12=comp3, 13=zoned, 14=emptyasconst, 15=emptyasempty,
-                                   16=div */
+                                   16=div, 17 trim */
         char **el;              /* dataset element array pointer */
         unsigned int prec;      /* COMP3/ZONED Precision */
         unsigned int scale;     /* COMP3/ZONED Scale */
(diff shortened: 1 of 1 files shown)


core/conn/odb/src/odb.c
```diff
@@ -6246,7 +6248,8 @@ static void Oload(int eid)
              trt[16];           /* translit to array */         
         char op;                /* 1=substr, 2=dconv, 3=tconv, 4=tsconv, 5=replace,
                                    6=toupper, 7=tolower, 8=firstup, 9=csubstr, 10=translit,
-                                   11=comp, 12=comp3, 13=zoned, 14=emptyasconst, 15=emptyasempty */
+                                   11=comp, 12=comp3, 13=zoned, 14=emptyasconst, 15=emptyasempty,
+                                   16=div */
```
DaveBirdsall: Should the comments also include 17=trim?

core/conn/odb/src/odb.c
```diff
@@ -10076,6 +10114,7 @@ static int Oloadbuff(int eid)
         clock_gettime(CLOCK_MONOTONIC, &tsp1);
 #endif
         Or = SQLExecute(thps[tid].Os) ;         /* Execute INSERT (load/copy) or tgt command */
+        SQLLEN tLastRow = -1;           /* this is special solution for China Union Pay, print only first error message for state 22003 */
```
DaveBirdsall: Probably best not to put customer names into the code. Consider changing the comment to just "/* print only first error message for state 22003 */"

core/conn/odb/src/odb.c
```diff
@@ -14685,6 +14743,56 @@ static void addGlobalPointer(void *ptr)
...
+            else return 0;
+            break;
+        case 7:
+            if (!isdigit(str[i])) return 0;
+            break;
+        default:
+            return 0;
+        }
+    }
+    if (s == 3 || s == 7 || s == 5) return 1;
```
DaveBirdsall: I am wondering if s == 4 should be allowed here too. Consider the string "1234."

core/conn/odb/src/odb.c
SuJinpei: Thank you, I will fix this.

core/conn/odb/src/odb.c
SuJinpei: Thank you, I will fix this.

core/conn/odb/src/odb.c
SuJinpei: Yes, this is also a valid numeric value.

PR #1749, opened 2018-11-24
[TRAFODION-3236] allocate revcive buffer by using dynamic array sizing instead of allocate from heap dynamically

Guo-Heng:


DaveBirdsall:
Not sure why the regression tests are failing. The symptom is very weird. On the chance that it is environmental, I'll rerun the tests.

DaveBirdsall:
jenkins, retest

DaveBirdsall:
The regression tests failed again with similar results. @Guo-Heng, could you run the checktest2 suite in your environment and see if it also fails for you?

Guo-Heng:
I discussed it with a few colleagues including Kenny. The result of the discussion is that we agree that the chances of a buffer being too long are very small, so we don't have to do this.

PR #1750, opened 2018-11-28
[TRAFODION-3238] ODBC can not connect to server when the local language is gbk

Weixin-Xu:
1. Set your current language used in non-unicode programs to GBK.
2. Create user name with chinese.
3. Test odbc.

### DIFF
--- a/win-odbc64/security_dll/native/source/secpwd.cpp
+++ b/win-odbc64/security_dll/native/source/secpwd.cpp
@@ -175,6 +175,10 @@ SecPwd::SecPwd(const char *dir, const char* fileName,
 {
     char* dir_from_env = NULL;
     char  certDir[MAX_SQL_IDENTIFIER_LEN + 1];
+    char* serverNameGBKToUtf8 = NULL;
+    LCID lcid = GetSystemDefaultLCID();
+    int  translen = 0;
+    char transError[128] = { 0 };
 
     if (serverName == NULL)
         throw SecurityException (INPUT_PARAMETER_IS_NULL, " - serverName.");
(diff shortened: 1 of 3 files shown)


win-odbc64/security_dll/native/source/secpwd.cpp
```diff
@@ -217,8 +221,25 @@ SecPwd::SecPwd(const char *dir, const char* fileName,
     if(stat(certDir,&st) != 0)
         throw SecurityException(DIR_NOTFOUND, (char *)certDir);
 
-    certFile = buildName(certDir, fileName, serverName, CER);
-    activeCertFile = buildName(certDir, activeFileName, serverName, ACTIVE_CER);
+    if (lcid == 0x804) // if local charset is not utf8
```
DaveBirdsall: Looking at https://msdn.microsoft.com/en-us/library/cc233965.aspx, it appears 0x804 = language tag zh-CN, with default sort order. Do we know that all other locales use UTF-8? (Perhaps the comment on this line should be, "if locale ID is zh-CN")

win-odbc64/security_dll/native/source/secpwd.cpp
```diff
@@ -217,8 +221,25 @@ SecPwd::SecPwd(const char *dir, const char* fileName,
...
+    {
+        serverNameGBKToUtf8 = (char *)malloc(MAX_SQL_IDENTIFIER_LEN + 1);
```
DaveBirdsall: You use malloc to allocate this storage but delete to free it. This behavior is undefined in general. It would  be better to use free() to delete it, or to use new instead of malloc.

Weixin-Xu: Thanks, I will modify it.

win-odbc64/security_dll/native/source/secpwd.cpp
Weixin-Xu: I will change "is not utf8" to "is gbk". Do you know how many character sets we need?

DaveBirdsall: No... I suppose we'll let the marketplace tell us.

PR #1753, opened 2018-11-30
[TRAFODION-3241]support both jdk1.7 and jdk1.8

kevinxu021:
Previously, it doesnot support jdk1.8 because of site. Here use profile as a solution. By the way, to accelerate to the compilation speed, add if-else in Makefile that if you want to re-compile, please do make clean first. With my experiences, most of the time if the error is not in DCS/JDBC, you might not want to recompile this module.

### DIFF
--- a/core/conn/jdbcT4/Makefile
+++ b/core/conn/jdbcT4/Makefile
@@ -27,8 +27,8 @@ all: build_all
 
 build_all: LICENSE NOTICE 
 	@if [ ! -f target/*.jar ]; then \
-		echo "$(MAVEN) package -DskipTests"; \
-		set -o pipefail && $(MAVEN) package -DskipTests | tee build_jdbct4.log | grep --line-buffered -E -e '^\[[^WId]' -e '^\[INFO\] B[Uu][Ii][Ll][Dd]' -e 'to compile'; \
+		echo "$(MAVEN) install -DskipTests"; \
+		set -o pipefail && $(MAVEN) install -DskipTests | tee build_jdbct4.log | grep --line-buffered -E -e '^\[[^WId]' -e '^\[INFO\] B[Uu][Ii][Ll][Dd]' -e 'to compile'; \
 		cp target/jdbcT4-${TRAFODION_VER}.jar ${TRAF_HOME}/export/lib; \
 		mkdir -p ../clients; \
 		mv target/jdbcT4-${TRAFODION_VER}.zip ../clients; \
--- a/core/conn/jdbcT4/Makefile
+++ b/core/conn/jdbcT4/Makefile
@@ -26,17 +26,20 @@ include ../../macros.gmk #top level
 all: build_all
 
 build_all: LICENSE NOTICE 
-	echo "$(MAVEN) package -DskipTests"
-	set -o pipefail && $(MAVEN) package -DskipTests | tee build_jdbct4.log | grep --line-buffered -E -e '^\[[^WId]' -e '^\[INFO\] B[Uu][Ii][Ll][Dd]' -e 'to compile'
-	cp target/jdbcT4-${TRAFODION_VER}.jar ${TRAF_HOME}/export/lib
-	mkdir -p ../clients
-	mv target/jdbcT4-${TRAFODION_VER}.zip ../clients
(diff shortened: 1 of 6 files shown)


DaveBirdsall:
@svarnau, @selvaganesang, @narendragoyal, do any of you have concerns on this change?

core/conn/jdbcT4/Makefile
```diff
@@ -26,17 +26,20 @@ include ../../macros.gmk #top level
...
 
 build_all: LICENSE NOTICE 
-	echo "$(MAVEN) package -DskipTests"
-	set -o pipefail && $(MAVEN) package -DskipTests | tee build_jdbct4.log | grep --line-buffered -E -e '^\[[^WId]' -e '^\[INFO\] B[Uu][Ii][Ll][Dd]' -e 'to compile'
-	cp target/jdbcT4-${TRAFODION_VER}.jar ${TRAF_HOME}/export/lib
-	mkdir -p ../clients
-	mv target/jdbcT4-${TRAFODION_VER}.zip ../clients
-	#ln -sf ${TRAF_HOME}/export/lib/jdbcT4-${TRAFODION_VER}.jar ${TRAF_HOME}/export/lib/jdbcT4.jar
-	`cd ${TRAF_HOME}/export/lib;ln -sf jdbcT4-${TRAFODION_VER}.jar jdbcT4.jar`
+	@if [ ! -f target/*.jar ]; then \
```
narendragoyal: Instead of target/*.jar, should it check for the exact file that we are trying to build. What if TRAFODION_VER changes or there is some other JAR already in the 'target' directory.

core/conn/jdbcT4/Makefile
```diff
@@ -26,17 +26,20 @@ include ../../macros.gmk #top level
...
+		cp target/jdbcT4-${TRAFODION_VER}.jar ${TRAF_HOME}/export/lib; \
+		mkdir -p ../clients; \
+		mv target/jdbcT4-${TRAFODION_VER}.zip ../clients; \
+		cd ${TRAF_HOME}/export/lib;ln -sf jdbcT4-${TRAFODION_VER}.jar jdbcT4.jar; \
+	fi
 
 clean:
-	-$(MAVEN) clean | grep ERROR
-	$(RM) build_jdbct4.log ${TRAF_HOME}/export/lib/jdbcT4*.jar
+	@if [ -f target/*.jar ]; then \
```
narendragoyal: Just my opinion :) : I think the 'clean' target should not be conditional on the existence of the target/*JAR, it should execute the commands and perhaps the 'maven clean' will cleanup the intermediate class files also (even when a target/*.jar does not exist).

core/conn/jdbcT4/Makefile
svarnau: If *.jar expands to more than one file, won't that generate a shell syntax error?

kevinxu021: @narendragoyal @svarnau  sure, it should be an exactly name.

core/conn/jdbcT4/Makefile
kevinxu021: The classes are in folder "target". if this folder does not exist, 'mvn clean' actually will do nothing. So here yes it should b -d target.

selvaganesang:
@kevinxu021 Are we compiling the sources to be compatible with JDK 1.7 or JDK 1.8?  I am asking because I couldn't decipher from the changes and I am not familiar with maven rules. Also, I see the hard coded versions 1.7 is being changed to use property such as source.version and target.version

kevinxu021:
@selvaganesang it uses "<jdk>[1.8,)</jdk>" to activate the property  "<javadoc.opts>-Xdoclint:none</javadoc.opts>", and with jdk1.8 you will got error while "mvn site". this is a preparation for our project running on jdk1.8. If this changes made, you can compile this project both on jdk1.7 and jdk1.8. So it's '-Xdoclint:none' stop JDK1.8 compatibility.

kevinxu021:
@selvaganesang Sorry for confused you. Here supporting JDK1.7 and JDK1.8 means compilation. Before the changes, those projects cannot be compiled on JDK1.8 actually.  Then if we upgrade JDK from JDK1.7(current version) to JDK1.8, you will see some error on compilation time. This fixes are for that, to keep the code being compiled successfully on JDK1.7 and JDK1.8.

kevinxu021:
Closed because of conflict.

PR #1757, opened 2018-12-07
[TRAFODION-3246] add slf4j in jdbc

mashengchen:
add a new jar which can use slf4j to print log

### DIFF
--- a/licenses/lic-clients-bin
+++ b/licenses/lic-clients-bin
@@ -32,3 +32,30 @@ MIT like license. (http://site.icu-project.org/)
   this Software without prior written authorization of the copyright holder.
 
 ===============================================================================
+The binary distribution of Apache Trafodion JDBC driver bundle
+SLF4J (Simple Logging Facade for Java) software which is available under the
+MIT/X11 License.  http://www.slf4j.org
+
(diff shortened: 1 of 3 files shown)


DaveBirdsall:
@arvind-narain, @hegdean , @svarnau: Any comments on this one?

svarnau:
This seems to be bundling in slf4j with JDBC driver.  If that is the case, then the slf4j license, which is already in licenses/lic-server-bin needs to also be added to licenses/lic-clients-bin.

mashengchen:
thank you for your help Steve

DaveBirdsall:
@svarnau, does this look good to you now?

licenses/lic-clients-bin
```diff
@@ -32,3 +32,30 @@ MIT like license. (http://site.icu-project.org/)
   this Software without prior written authorization of the copyright holder.
 
 ===============================================================================
+The binary distribution of Apache Trafodion DCS and REST components bundle
```
svarnau: here we should say JDBC driver instead of "DCS and REST components"

DaveBirdsall:
@svarnau, does this look good to you now?

PR #1760, opened 2018-12-11
[TRAFODION-3247] support customer define client charset

mashengchen:
1. change charset map in InterfaceUtilities.java to Canonical Charset Name according to jdk define.

2. if user specify clientCharset, then it will change that charset when display string value


DaveBirdsall:
Is there a JIRA for this change?

mashengchen:
@DaveBirdsall sorry for miss JIRA id, add it to the title now.

DaveBirdsall:
@arvind-narain, do you have any comments?

PR #1763, opened 2018-12-15
[TRAFODION-3249] Update Syntax Descriptions and Add Examples for *CREATE/ALTER/DROP SEQUENCE Statement* in *Trafodion SQL Reference Manual*

liuyu000:


liuyu000:
@DaveBirdsall Thanks for your review :-D

PR #1766, opened 2018-12-21
[TRAFODION-3234] Refactor hive meta calls

selvaganesang:
Hive similarity check was detecting if there is a change in the hive
directory timestamp only. It has been changed to detect the DDL changes too.

Cleaned up code further and removed all the unwanted code.


selvaganesang:
@sureshsubbiah Can you please review this PR

sureshsubbiah:
+1. Looks good @selvaganesang, Thank you for including me in the review.

PR #1770, opened 2019-01-15
Do not make again before MAKE CLEAN

kevinxu021:
check exactly jar name for each Makefile

Do mvn install in project Makefile


kevinxu021:
duplicated

PR #1771, opened 2019-01-15
[TRAFODION-3242]Do not make again before MAKE CLEAN

kevinxu021:
Ref to PR-1755


kevinxu021:
Please help to merge it. i don't want to have next round of conflicting. Thanks!

svarnau:
+0  Personally, I'm not a fan of this change.  It breaks makefile protocol and expected behavior, in that it forces removal of files in order to re-build them instead of detecting they are out of date with sources.

I understand that our makefiles are not in particularly good shape of re-building things unnecessarily, so that is why I'm not a -1.

This is an ugly work-around rather than fixing makefiles.  But I agree that it is an expedient work-around for some parts of long re-build times.

kevinxu021:
Hi Steve, Do you have some better idea? I was trying to find one instead of the current one but failed. Makefile may be not a good choice for MAVEN project. Maybe, i need to put those scripts into another shell script file and Makefile call the new script file, agree? Whatever, any good suggestion would be appreciated and i only want to solve the issue experienced every time almost.

kevinxu021:
I have a new idea for this and will create a new pull request for it.

svarnau:
Hi Kevin,
You may be right that this is a fundamental problem of mixing makefiles and maven into single build project.  

My concern with this change is that it only checks for existence of a file rather than checking it against the dependencies to see if it needs to be re-built.  Ideally, makefiles should be written so that if none of the source files have changed, make does not rebuild anything. But if a source file changes, then the right things do get re-built without needing to do make clean. 

Right now, we have the problem of way too much re-building that does not need to be done.  With your change, we have less of that problem, but the developer needs to know what to remove/clean if they do need to re-build something.  That may be a better problem to have.  It just goes against the ideal makefile methods.

Example:

Trafci_jar:
	mvn clean install -DskipTests

This rule always runs becuase "Trafci_jar" is not a real file, instead make is asking maven to make sure it is up to date, but maven is told not to trust existing objects by doing clean first. (Ugh).  You are changing this to:

Trafci_jar:
	@if [ ! -f target/trafciInstaller.jar ] ; then \
		mvn install -DskipTests ; \
	fi

Which is not running maven at all regardless of dependencies if the jar exists.  The simpler way to write that rule is:

target/trafci.jar:
          mvn install -DskipTests

(Replacing all references of "Trafci_jar" in the makefile with the full jar file name.)  Even better would be something like this:

target/trafci.jar: src/main/java/org/trafodion/ci/*
           mvn install -DskipTests

Which would cause a rebuild if any sources in that directory changed.  Other dependencies may also be needed, and nice to have.  But at least use the simpler makefile syntax instead of shell script in the rule.

In this particular case, the rules are even more complicated because there are two different pom files in the same directory that probably interfere with each other.  So fixing this is not easy.

kevinxu021:
Thanks Steve,  I know what you were saying. I made a change for  core/conn/jdbcT4/Makefile. Could you please take a look again and how to deal with CLEAN? I tried it and it works well. If you agree with it, i will change the rest of them. Let me have a version for 2 pom files with one directory later.

kevinxu021:
@saruna I don't have a better idea for CLEAN target. Please take a look the latest changes for CLEAN as well. I was using shell script.

svarnau:
@kevinxu021 Yes, I like these changes much better!  I agree that is the way to go.

I don't know why the clean target is a concern.  That should not be used so often, right?  Why not just call mvn clean for each pom file in the makefile clean target?  It should not take so long, right?

kevinxu021:
@svarnau it will take you around 6s for each project. Trafodion has more than 10 project( some project has more than one pom.xml which might be in different target folder now and in the future), that means it will save more than 60s. Trafodion was complained for compilation, that was why i have to do something for it.

kevinxu021:
@svarnau I have done for all the changes. Please take a look. Thanks!

kevinxu021:
@svarnau Error from Git, please help, thanks!

Please upgrade your git client.
GitHub.com no longer supports git over dumb-http: https://github.com/blog/809-git-dumb-http-transport-to-be-turned-off-in-90-days

svarnau:
> @svarnau Error from Git, please help, thanks!
> 
> Please upgrade your git client.
> GitHub.com no longer supports git over dumb-http: https://github.com/blog/809-git-dumb-http-transport-to-be-turned-off-in-90-days

You should not use https protocol with github. (Sorry if wiki misled.)  Use git@github.com:apache/trafodion.

svarnau:
test failure indicates some problem with TRX files.

kevinxu021:
@svarnau Git issue confused me. Finally, I found the issue is lost dtm jar. So far, the changes has been done. Please review and merge. Thanks!

PR #1772, opened 2019-01-15
[TRAFODION-3256] Fix error path in CLI similarity check code

DaveBirdsall:
Add missing "return ERROR" (and an NADELETEBASIC to avoid memory leak) in method Statement::doHiveTableSimCheck.


DaveBirdsall:
@selvaganesang, could you review this please?

PR #1777, opened 2019-01-23
[TRAFODION-3262] Insert float to numeric(19,0) lost precision for some case.

SuJinpei:
**root cause:**  when the target numeric type's precision is bigger than 18, the source float or double data will first be converted to char string which will call the double_to_char function, then transfer to the server. original double_to_char only keep '**precision**' significant digits. In this case, the caller provides the precision with 6 for float datatype, so last 2 digits' precision lost.
**solution:** now, the '**precision**' parameter means the number of digits after the decimal point which is the caller really meant to do. To prevent precision lost,  the conversion will fail if the caller doesn't provide enough space to receive the result.
**verify:**  no new issue was found by running coast to verify the fix.

### DIFF
--- a/docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
+++ b/docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
@@ -525,7 +525,7 @@ ACOS (COS (0.3491))
 
 <<<
 [[add_months_function]]
-=== ADD_MONTHS Function
+== ADD_MONTHS Function
 
 The ADD_MONTHS function adds the integer number of months specified by
 _int_expr_ to _datetime_expr_ and normalizes the result. ADD_MONTHS is a {project-name} SQL
@@ -1930,7 +1930,7 @@ CEILING (2.25)
 
 <<<
 [[char_function]]
-=== CHAR Function
+== CHAR Function
 
 The CHAR function returns the character that has the specified code
 value, which must be of exact numeric with scale 0.
@@ -2409,7 +2409,7 @@ COS (0.3491)
 
 <<<
 [[cosh_function]]
-=== COSH Function
+== COSH Function
 
 The COSH function returns the hyperbolic cosine of a numeric value
 expression, where the expression is an angle expressed in radians.
@@ -5629,7 +5629,7 @@ See <<ucase_function,UCASE Function>>.
 
 <<<
 [[left_function]]
-=== LEFT Function
+== LEFT Function
 
 The LEFT function returns the leftmost specified number of characters
 from a character expression. Every character, including multi-byte
@@ -5786,7 +5786,7 @@ LOG (2.0)
 
 <<<
 [[log10_function]]
-=== LOG10 Function
+== LOG10 Function
 
 The LOG10 function returns the base 10 logarithm of a numeric value
 expression.
@@ -6810,7 +6810,7 @@ FROM mining.seqfcn SEQUENCE BY ts;
 
 <<<
 [[nullif_function]]
-=== NULLIF Function
+== NULLIF Function
 
 The NULLIF function compares the value of two expressions. Both
 expressions must be of comparable types. The return value is NULL when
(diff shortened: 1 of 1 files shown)


SuJinpei:
Hi @Weixin-Xu, @arvind-narain, do you have some comments on it?

win-odbc64/odbcclient/drvr35/drvrglobal.cpp
```diff
@@ -868,47 +870,48 @@ bool use_gcvt(double number, char* string, short size)
...
-	return rc;
+    char format[16] = { '\0' };
+    size_t actualLen = 0;
+    bool need_reprint = false;
+
+    // make sure any precision of possible double value can be format to the buf.
+    char buf[MAX_DOUBLE_TO_CHAR_LEN] = { '\0' };
+
+    // precission should be limit to a reasonable range.
+    if ((precision < 0) || (precision >(DBL_MANT_DIG - DBL_MIN_EXP)))
```
Weixin-Xu: DBL_MIN_EXP means min binary exponent. The number here is decimal.

win-odbc64/odbcclient/drvr35/drvrglobal.cpp
```diff
@@ -868,47 +870,48 @@ bool use_gcvt(double number, char* string, short size)
...
+        return false;
+
+    if (sprintf(format, "%%.%dlf", precision) < 0)
+        return false;
+
+    // use scientific notation if data is small or the result is too long.
+    if (fabs(number) < 1.0 || (actualLen = sprintf(buf, format, number)) >= size) {
```
Weixin-Xu: You can use some other function to find out whether the number has integer part rather than comparsion between the absolute value of the number and 1.0 .

win-odbc64/odbcclient/drvr35/drvrglobal.cpp
```diff
@@ -868,47 +870,48 @@ bool use_gcvt(double number, char* string, short size)
...
+    bool need_reprint = false;
+
+    // make sure any precision of possible double value can be format to the buf.
+    char buf[MAX_DOUBLE_TO_CHAR_LEN] = { '\0' };
+
+    // precission should be limit to a reasonable range.
+    if ((precision < 0) || (precision >(DBL_MANT_DIG - DBL_MIN_EXP)))
+        return false;
+
+    if (sprintf(format, "%%.%dlf", precision) < 0)
```
Weixin-Xu: Why do you return false since you have confirmed the precision is reasonable ?

win-odbc64/odbcclient/drvr35/drvrglobal.cpp
```diff
@@ -868,47 +870,48 @@ bool use_gcvt(double number, char* string, short size)
...
+        return false;
+
+    // use scientific notation if data is small or the result is too long.
+    if (fabs(number) < 1.0 || (actualLen = sprintf(buf, format, number)) >= size) {
+        if (sprintf(format, "%%.%dle", precision) < 0)
+            return false;
+        need_reprint = true;
+    }
+
+    if (need_reprint) {
```
Weixin-Xu: Is this flag necessary ?

win-odbc64/odbcclient/drvr35/drvrglobal.cpp
SuJinpei: Yes, DBL_MIN_EXP is min binary exponent, but we actually should use this value to compute the max precision.  You can refer to https://en.wikipedia.org/wiki/Double-precision_floating-point_format.  The example is: 
2^(-1)  -> precision: 1
2^(-2) -> precision:2
...
2^(-52) -> precision:52
So the fraction part will generate DB_MANT_DIG  precision.
The fraction part should multiply 2^(DBL_MIN_EXP) which will generate another DBL_MIN_EXP precision,  so the total max precision is DBL_MANT_DIG - DBL_MIN_EXP.

Weixin-Xu: The number is decimal.
The result is like "123" not "‭1111011‬".
You can not find a double number larger than DBL_MANT_DIG - DBL_MIN_10_EXP.

PR #1779, opened 2019-01-23
[TRAFODION-3241]support maven compilation with both jdk1.7 and jdk1.8

kevinxu021:
Ref to PR-1753, closed because of conflict.
Optimize DCS Makefile.
Support maven compilation with both jdk1.7 and jdk1.8


svarnau:
I like the idea, but don't know enough about maven/java.
If someone compiles with 1.8, does that mean it won't work with 1.7 runtime?

PR #1780, opened 2019-01-24
[TRAFODION-3263]Changes to disable LOB locking and refactor IUD operations

sandhyasun:
The order of operations with this change is as follows :

1.Call JNI hdfs Write API to write the lob data to hdfs. 
2.Use return data offset from JNI hdfswrite API in 1. as the offset to store in the LOB descriptor tables. 
3.If there are multiple chunks to write, do it in a loop and append to the first chunk. This way each chunk can be anywhere in hdfs and not necessarily continguous. But we are guaranteed that whatever we wrote will be stored in our internalLOB descriptor files.
4. If any failure or TM erro occurs while writing to the LOB descriptor tables,  transaction gets rolled back and the chunk of hdfs data written becomes "dead data". It doesn't harm the next operation. 
5. GC check is now done before an update or insert. Earlier it was done as part of the ::allocateDesc operation to get the EOD of the file.


sandhyasun:
@selvaganesang  Can you please review these changes ?

sandhyasun:
Thanks Selva.  Lets keep the JIRA open so hdfsWriteImmediate can return a status if the FileStatus.size just before the "write" and FileStatus.size just after the write don't differe by exactly btesWritten. If a error status can be written, we can retry the hdfsWriteImmediate upto 3 times and if not return an error to the callerand fail the insert/update operation.

PR #1781, opened 2019-01-25
[TRAFODION-2626] Change pathes of sqconf, ms.env, sqcert, udr

svarnau:
This changeset covers the remaining instances of hard-coded paths based
on TRAF_HOME for dynamic files that should be in specific paths such as
TRAF_CONF or TRAF_VAR.

All files referenced via TRAF_HOME should be read-only software, not written
during run-time.

I was able to test this premise by setting TRAF_CONF, _LOG, _VAR, and rundir variables to alternates paths and then change the entire core/sqf tree as read-only (except local_hadoop). In this environment of read-only TRAF_HOME, I bring up instance and run all SQL regressions.


svarnau:
Looking for some reviewers for this change.  The changes are mostly mechanical path changes across many files, but given the effects on locations of auth file, sqconfig, ms.env, udr, I'm thinking these folks may be interested:  @robertamarton @zcorrea @sandhyasun @zellerh

zellerh:
This is a great idea and the parts I reviewed (mostly UDR-related) look good to me.

sandhyasun:
+1   Thanks for making this useful change across so many files.

PR #1782, opened 2019-01-28
[TRAFODION-3264] eliminate unneeded function call in Lob-related functions

wenjun-zhu:
in file core/sql/executor/ExExeUtilLoad.cpp, lines like:
```
	    ex_expr::exp_return_type exprRetCode =	
	      lobTdb().inputExpr_->eval(NULL, 	
					workAtp_);
```
call function eval(), with first parameter NULL.

Then code in ex_expr::evalClauses() accesses NULL pointer, so the server aborts.

As @sandhyasun said, this invocation is unneeded, so remove these code.

### DIFF
--- a/core/sql/executor/ExExeUtilLoad.cpp
+++ b/core/sql/executor/ExExeUtilLoad.cpp
@@ -2653,15 +2653,6 @@ short ExExeUtilLobExtractTcb::work()
 	  
 	case GET_NO_CHILD_HANDLE_:
 	  {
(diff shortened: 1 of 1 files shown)


sandhyasun:
Did you run all the extract commands and  "update lob" commands and ensure this expression eval can be eliminated ? Did executor/TEST130 pass ?

sandhyasun:
We can merge this in if we get the results on executor/TEST130 and the JDBC lob tests. Please update this PR  with the results.

PR #1783, opened 2019-01-29
[TRAFODION-3259] Length of varchar exceeds 65536 insert data will fail

zhenxingh:
The precision_ of Attributes class is defined as Int16, this patch
expanded it to 32bit by adding a precision_high_ to hold the higher
16 bits of the precision.


svarnau:
jenkins, okay to test

core/sql/exp/exp_attrs.h
```diff
@@ -759,9 +763,10 @@ class SimpleType : public Attributes
   Int16 getScale()            {return scale_;}
   UInt16 getScaleAsUI()       {return (UInt16)scale_;}
 
-  Int32 getPrecision()        {return (UInt16)precision_;} // Treat as UInt16 in case 0x8000 or bigger
+  Int32 getPrecision()
+  {return (UInt16)precision_ | ((UInt32)precision_high_ << 16);}
   inline void setPrecision(Int32 precision)
-                              {precision_ = (Int16)precision;}
+  {precision_ = (Int16)precision; precision_high_ = (Int16)((UInt32)precision >> 16);}
```
selvaganesang: I would think it should be ok to change the precision to Int32 instead of keeping two different variables precision_ and precision_high_. Currently Trafodion doesn't support versioning. From this perspective, it is ok to change these structures which could be serialized across processes.

zhenxingh: Good to know that, thanks!

zlei929:
This issue may be similar to the M7785 and i have fixed it already.

zhenxingh:
> This issue may be similar to the M7785 and i have fixed it already.

Yes, the same issue, but we also need to fix it for trafodion.

PR #1784, opened 2019-01-30
[TRAFODION-3268] [TRAFODION-3269] [TRAFODION-3269] Library manager fixes and LOB alter fix

sandhyasun:
Fixes for library management 
1. A non existent library would not return an error during create due to incorrect error/diagnostics handling
2. Added a retry during extract of libraries to try to avoid random "file too short" errors.
3. Made a fix to LOB alter code to avoid creating LOB_MD table when one already exists.


sandhyasun:
@robertamarton  Can you please review these ?

core/sql/cli/Cli.cpp
```diff
@@ -9517,20 +9517,37 @@ Lng32 SQLCLI_LOBddlInterface
...
-	
-	cliRC = cliInterface->executeImmediate(query);	
-	currContext.resetSqlParserFlags(0x1);
+        // set parserflags to allow ghost table
+        currContext.setSqlParserFlags(0x1);
 	
-	if (cliRC < 0)
-	    goto error_return;
+        cliRC = cliInterface->executeImmediate(query);	
+        currContext.resetSqlParserFlags(0x1);
```
robertamarton: Can someone set this parserflag before calling the method and now it is reset?

core/sql/cli/Cli.cpp
```diff
@@ -9517,20 +9517,37 @@ Lng32 SQLCLI_LOBddlInterface
...
 	
+        if (cliRC < 0)
+          {
+            ComDiagsArea *tempDiags = ComDiagsArea::allocate(currContext.exHeap());
```
robertamarton: Do you have to deallocate tempDiags?

core/sql/executor/ExUdr.cpp
```diff
@@ -809,20 +809,17 @@ Int32 ExUdrTcb::fixup()
...
+             
                char errString[200];
                NAString errNAString;
-               sprintf(errString , "Error %d creating directory :",err); 
-               errNAString = errString;
-               errNAString += cachedFullName;
-             
-              
+               sprintf(errString , "Error %d creating directory %s :",err, cachedLibPath.data()); 
+               errNAString = errString;
```
robertamarton: Can cachedLibPath be longer than 200 characters?  Will errString size be long enough?

core/sql/optimizer/UdfDllInteraction.cpp
```diff
@@ -158,12 +158,10 @@ NABoolean TMUDFDllInteraction::describeParamsAndMaxOutputs(
                                   dummyUser,
                                      cachedLibName, cachedLibPath))
         {
-           NAString cachedFullName = cachedLibPath+"/"+cachedLibName;
            char errString[200];
            NAString errNAString;
-           sprintf(errString , "Error %d creating directory :",errno); 
+           sprintf(errString , "Error %d creating directory : %s", err, cachedLibPath.data());
```
robertamarton: If cachedLibPath > 200 bytes, it will overflow errString.

core/sql/cli/Cli.cpp
sandhyasun: It's set just before the executeImmediate and reset right after.

core/sql/cli/Cli.cpp
sandhyasun: decrRefCount will deallocate the diags if refcount goes down to 0.  So inn this case we are good- it will get dealllocated since refCount is equal to 1.

core/sql/executor/ExUdr.cpp
sandhyasun: Thanks for pointing this out. Fixing in both places this is called.

PR #1786, opened 2019-01-30
[TRAFODION-3260] SSMP may wait 3 seconds before handling requests

zhenxingh:


svarnau:
jenkins, add user

core/sql/runtimestats/ssmpipc.cpp
```diff
@@ -540,7 +540,7 @@ static Int64 SikGcInterval = -1;
 void SsmpGlobals::work()
 {
 
-  getIpcEnv()->getAllConnections()->waitOnAll(getStatsMergeTimeout());
+  getIpcEnv()->getAllConnections()->waitOnAll(0);
```
selvaganesang: We would want ssmp to wake up every 3 seconds if there are no other requests to it and do some cleanup tasks. So, it is not clear why do you want to set this to 0.

zhenxingh: I have move the wait out to the main loop, so there is no need to wait here.

zhenxingh: The wait in the main loop will wake up every 3 seconds, so here we only need to check if there is any work to do, and no need to wait.

selvaganesang: I got it now. My understanding is that we wait for openers or messages from openers in XWAIT. If there are open messages, it is consumed at cc->wait(0) and if there are messages from openers it is consumed here. But we wake up every 3 seconds to do bookkeeping.

zhenxingh: Yes, that's right

core/sql/bin/ex_ssmp_main.cpp
```diff
@@ -246,8 +247,12 @@ void runServer(Int32 argc, char **argv)
     }
   }
 */
-    // wait for system messages only until ssmp starts receiving msgs.
-    cc->wait(300);
+    // Wait for messages, but we need ssmp to wake up periodically to
+    // perform garbage collection.
+    short mask = XWAIT(LREQ | LDONE, ssmpGlobals->getStatsMergeTimeout());
+    if (mask & LREQ) {
+      cc->wait(0);
```
selvaganesang: This change looks good except that it is not encapsulated.  XWAIT, LREQ and LDONE concepts should have been part of the IPC infrastructure and not expected to be used in the callers of IPC infrastructure.  So, @narendragoyal and I tried the following  and it seems to achieve the same effect. 

Add the following 2 lines in lieu of this code
if (cc->getConnection() == NULL)
   cc->wait(ssmpGlobals->getStatsMergeTimeout());

And undo the change at ssmpGlobals::work().

Prior to Trafodion there used to at least one opener to ssmp always. Hence this issue was not observed earlier.  It is good catch

selvaganesang: Please consider modifying to encapsulate as suggested or some other way

zhenxingh: There is a subtle difference, the waitOnAll() function called in ssmpGlobals::work() does not wait on all connections simultaneously, it actually only waits on the first connection with an exponentially increased timeout (3 seconds maximum), and then cycle through all other connections without waiting. This can still lead to a unnecessary wait if there are data arrival on connections other than the first.

selvaganesang: WaitOnAll contains a set of connections : Control Connection + all other opener connections. Control connection can receive system messages like open, close etc, while other connection receive  messages sent by the openers. Though waitOnAll waits exponentially till the given timeout, as soon as any system message or messages from the opener is received it will break the  wait to respond to the message.

zhenxingh: It's OK to merge the PR as is now, we can do the encapsulation later, we may also need to do some refactoring and cleanup of the code, which may be risky and need more QA testing.

PR #1788, opened 2019-02-04
[TRAFODION-3270] Handle incompatible types in key expressions

DaveBirdsall:
If we have a join predicate of incompatible types and nested join is chosen, we get an assertion failure in the Generator in debug builds.

This changes fixes that, by adding logic to handle incompatible types when generating key expressions.


DaveBirdsall:
Do not merge. A few regression tests failed and the failure needs to be investigated.

DaveBirdsall:
The rework takes care of the case of nullable keys. Without the rework, tests such as core/TEST005 would fail when using an is null key predicate on an index. @robertamarton, could you review this a second time? Thanks.

PR #1792, opened 2019-02-10
Sean traf 2.4

sbroeder:


svarnau:
This pull request reverted earlier changes in the hbase-trx Makefile. I think that was incorrect. The change should hae been made in the new format.

PR #1802, opened 2019-02-20
[TRAFODION-3272] Change installer settings: TRAF_LOG, TRAF_CONF, TRAF_VAR

svarnau:
Changes python-installer and ambari-installer.

Installers set TRAF_CONF to /etc/trafodion/conf
Installers allow user to specify TRAF_VAR, but defaults to /var/lib/trafodion
Installers allow user to specify TRAF_LOG, but defaults to /var/log/trafodion

Changes DCS conf directory from DCS_INSTALL_DIR/conf to TRAF_CONF/dcs
Changes REST conf directory from REST_INSTALL_DIR/conf to TRAF_CONF/rest


svarnau:
jenkins, retest -- automation change to make parent dir when installing regressions

svarnau:
@mkby  Could you take a look at this trafodion change?

install/python-installer/db_uninstall.py
```diff
@@ -63,26 +63,21 @@ def main():
...
         config_file = options.cfgfile
         p = ParseInI(config_file, 'dbconfigs')
         cfgs = p.load()
+        traf_var = cfgs['traf_var']
         node_list = cfgs['node_list']
-    # user input
-    else:
+      # user input
+      else:
+        traf_conf = '/var/lib/trafodion'
```
mkby: I think `traf_conf` here should be `traf_var`

install/python-installer/scripts/traf_ldap.py
```diff
@@ -34,8 +34,9 @@ def run():
 
     db_root_user = dbcfgs['db_root_user']
     traf_home = os.environ['TRAF_HOME']
+    traf_conf = os.environ['TRAF_CONF']
     sqenv_file = traf_home + '/sqenvcom.sh'
-    traf_auth_config = '%s/sql/scripts/.traf_authentication_config' % traf_home
+    traf_auth_config = '%s/.traf_authentication_config' % traf_conf
```
mkby: Should we remove the `.` from the ldap config file? Otherwise it will be hided in $TRAF_CONF dir if not provide `ls -a` command. I'm curious why we hide this config file in the original design, maybe it's time to change it.

install/python-installer/scripts/traf_user.py
```diff
@@ -103,7 +105,9 @@ def run():
     nodes = dbcfgs['node_list'].split(',')
     trafodion_config = """
 export TRAF_HOME="%s"
-export TRAF_VAR=$TRAF_HOME/tmp
+export TRAF_VAR="%s"
+export TRAF_CONF="%s"
+export TRAF_LOG="%s"
 export MY_SQROOT=$TRAF_HOME # for compatibility
```
mkby: I think it's time to remove My_SQROOT, please confirm.

mkby:
Looks good! I like the idea to put all configs in $TRAF_CONF

install/python-installer/db_uninstall.py
svarnau: Good catch. I will fix it.

install/python-installer/scripts/traf_ldap.py
svarnau: Yes. Unrelated to this change, but I agree that it is strange we name the file that way.

install/python-installer/scripts/traf_user.py
svarnau: Yes. Again unrelated to this change, but it makes sense to me that we no longer need that compatibility setting.  That suggestion is localized, so I can include removing it in this change.

PR #1810, opened 2019-03-06
add a scripts to make trafodion build easily and Visualized

CoderSong2015:
There are several unpleasant places when compiling the trafodion(not make package):
1. We need to redirect the make information manually,
   otherwise the information will flood the screen
2. The build information mixed together so that we couldn't locate the build error quickly
3. We need to make all again when we fix a building error of one component.

This scripts could build trafodion component based.
And it will write the compiling information into build_log file.
Also you can continue to compile the remaining components which havn't been compiled by commented components which have finished compiled.

### DIFF
--- a/build_easily.sh
+++ b/build_easily.sh
@@ -38,9 +38,7 @@ function process(){
     make_path=$4
     printf "[%-30s]........................[\e[33m%s\e[0m]\r" "building $component_name" "waiting"
     cd $make_path
-    #echo $log_file
-    #ret=$($cmd > "$log_dir/$log_file" 2>&1;echo $?)
-    ret=0
+    ret=$($cmd > "$log_dir/$log_file" 2>&1;echo $?)
     if [ 0 -eq $ret ]; then
         printf "[%-30s]...................................[\e[32m%s\e[0m]\r\n" "build $component_name" "ok"
     else
--- a/build_easily.sh
+++ b/build_easily.sh
@@ -1,0 +1,67 @@
+#!/bin/sh
+log_dir=$TRAF_HOME/../../"build_logs"
+
+function process(){
+    component_name=$1
+    cmd=$2
+    log_file=$3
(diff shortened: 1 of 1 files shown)


mkby:
Super cool! Can we also add 'make package'?

CoderSong2015:
> Super cool! Can we also add 'make package'?

Yes,. It is better to do it in next Pr.

build_easily.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/sh
```
xiaozhongwang: need the copyright header here

build_easily.sh
```diff
@@ -0,0 +1,67 @@
...
+fi
+
+if [ ! -d $log_dir/"foundation" ];then
+    mkdir $log_dir/"foundation"
+fi
+
+#export LC_ALL=en_US.UTF-8
+cd $TRAF_HOME/../
+./bldenvchk.sh;
+export SQ_BUILD_TYPE=debug
```
xiaozhongwang: you can use debug as default, and developer can change it via parameter.

ryzuo:
Finally! This looks great!

weiqxu:
It's an amazing work.

build_easily.sh
CoderSong2015: Hi, this srcipt now can olny be used in debug env.

CoderSong2015:
Can anyone make a review？

CoderSong2015:
> I think building via this script will be slower, since it is not doing things with as much parallelism as make would. It also has potential to get out of synch with makefiles / dependencies. So I would not document it as the primary build mechanism, but it seems fine as an alternative for those that want to use it.

@svarnau The script just call the make of the trafodion components so that it will not build slower than executing make directly .Besides, the dependency between components make that impossible to build component in parallel.

PR #1815, opened 2019-03-18
[TRAFODION-3281] Fix syntax error introduced in previous change

svarnau:
Earlier changes for this jira introduced syntax error in script
with spaces, where tabs are expected.


svarnau:
Merging without further review, since this problem is already impacting developers.

PR #1820, opened 2019-04-02
[TRAFODION-3280] Reduce path length in Trafodion for improved performance and scalability

selvaganesang:
This PR fixes the following issues to improve performance.
1. When the session is dropped for the default context, we were unnecessarily dropping HBase connections.
2. There were 2 HBase connections per mxosrvr - one for the default context and another for repository context. Now it has been changed to one HBase connection per process.
3. cli_globals::currContext() is now optimized for default context.
4. Changes to ensure that logging related objects are created and the corresponding functions are called only when the logging Level is set higher than the recording level,


core/sql/cli/Globals.h
```diff
@@ -89,6 +89,7 @@ class ExProcessStats;
 class CliGlobals;
 class CLISemaphore;
 class HBaseClient_JNI;
+class MonarchClient_JNI;
```
DaveBirdsall: Was this line intended?

core/sql/cli/Globals.h
```diff
@@ -510,6 +523,9 @@ inline
   LmLanguageManagerC *langManC_;
   LmLanguageManagerJava *langManJava_;
   NABoolean espProcess_;
+  HBaseClient_JNI *hbaseClientJNI_;
+  HBaseClient_JNI *bigtableClientJNI_;
+  MonarchClient_JNI *monarchClientJNI_;
```
DaveBirdsall: Not clear to me what bigtableClientJNI_ and monarchClientJNI_ are intended for.

selvaganesang: I will remove it in my next PR unless there is a need for another commit

core/sql/cli/Globals.h
selvaganesang: No. I will remove it in my next PR.

selvaganesang:
@DaveBirdsall I have pushed in a commit to take care of your comments. Can you please resolve the conversations and approve this PR

PR #1821, opened 2019-04-02
[TRAFODION-3291] Fix core when multi-column stats are done on lots of columns

DaveBirdsall:
The problem given in the JIRA is that creating a multi-column histogram on 500 columns results in a tdm_arkcmp core.

There are two changes in this change set:

1. Fixes the core by addressing a buffer overrun problem. When we log messages, we use a fixed size buffer for the message text. When a column list is long enough, we were overrunning this buffer, causing the core. The fix truncates the column list text to avoid these overruns.

2. Adds a limit on the number of columns in a multi-column histogram. As a practical matter, they are not useful on large column sets. Too, the request could conceivably be a syntax mistake by the user; it may be the user intended to specify a list of columns to create single column histograms upon. The limit is set at 10 (we already have such a limit when automatically generating multi-column histograms for key prefixes). It can be overridden using CQD USTAT_MULTI_COLUMN_LIMIT.


core/sql/ustat/hs_globals.cpp
```diff
@@ -994,7 +994,8 @@ void HSGlobalsClass::formGroupSets()
       {
          if (LM->LogNeeded())
          {
-            sprintf(LM->msg, "\tMC: GROUP (%s) has state DONT_TRY, is skipped", mgroup_set->colNames->data());
+            sprintf(LM->msg, "\tMC: GROUP (%s) has state DONT_TRY, is skipped", 
+              LM->truncate(mgroup_set->colNames->data(),sizeof(LM->msg-200)));
```
nonstop-qfchen: I wonder if there is a slightly better way to do this so that the total msg length is exactly 2000. 

One way to do so would be to compute the the skeleton string length L (such as strlen("\MC: GROUP () has state DONT_TRY, is skipped"), and then set the truncation length to 2000 - L. 

When L is available, we can supply it to sprint() as follows. See https://stackoverflow.com/questions/2239519/is-there-a-way-to-specify-how-many-characters-of-a-string-to-print-out-using-pri.

printf ("Here are the first %d chars: %.*s\n", L, L, "A string that is more than 8 chars");

core/sql/ustat/hs_globals.cpp
```diff
@@ -1004,7 +1005,8 @@ void HSGlobalsClass::formGroupSets()
 
          if (LM->LogNeeded())
          {
-            sprintf(LM->msg, "\tMC: GROUP (%s) is a HEAD GROUP ", mgroup_set->colNames->data());
+            sprintf(LM->msg, "\tMC: GROUP (%s) is a HEAD GROUP ", 
+              LM->truncate(mgroup_set->colNames->data(),sizeof(LM->msg-200)));
             LM->Log(LM->msg);
```
nonstop-qfchen: same comment as before.

core/sql/ustat/hs_globals.cpp
```diff
@@ -1050,14 +1052,15 @@ void HSGlobalsClass::formGroupSets()
               myNextNeighbpr = myNextNeighbpr->mcis_next;
 
               sprintf(LM->msg, "\tMC: GROUP (%s) is neighbor of HEAD GROUP (%s)", 
-                      myNextNeighbpr->colNames->data(), mgroup_set->colNames->data());
+                      LM->truncate(myNextNeighbpr->colNames->data(),sizeof(LM->msg-1100)), 
+                      LM->truncate(mgroup_set->colNames->data(),sizeof(LM->msg-1100),1));
               LM->Log(LM->msg);
```
nonstop-qfchen: same comment as before.

core/sql/ustat/hs_globals.cpp
nonstop-qfchen: It may also be a good idea to use snprintf().

DaveBirdsall: This is complicated in general. For example, some of the log messages have multiple parameters substituted in the sprintf. I would need to do several sprintfs, each one calculating the L from the previous substitution. There are many examples of this. Too, where two strings are involved that may need truncation, I would need additional logic to evenly apportion the available space to them. 

My feeling is the effort does not justify the benefit.

I did consider using snprintf. The disadvantage there is that it would truncate the end of the message. That might leave out interesting and vital information that today follows the column name list. Instead, I truncated the column name list and visually checked in all cases that I truncated it enough that the end of the message fits within the sprintf buffer.

I don't claim this is ideal. A better approach would be to completely rewrite the HSLogMan methods to use a C++ streaming style into, say, an NAString, eliminating the need for sprintf. I could avoid truncation completely, having HSLogMan fold the message into multiple lines if the underlying logger has any line length limitations. However, the code changes would be truly massive; logging is ubiquitous throughout the UPDATE STATISTICS code. Again, I did not feel that the effort was justified.

DaveBirdsall: Actually, let me think about the HSLogMan refactoring. I could provide new methods alongside the old, and then convert only the problem code paths to use the new methods. I will consider that.

DaveBirdsall:
Do *not* merge. I am reworking this change.

DaveBirdsall:
@nonstop-qfchen and @sureshsubbiah, could you take another look?

Thank you, @nonstop-qfchen, for the suggestions.

In two places, where sprintf is being used to populate a local buffer, I took your suggestion and used %.Ns. This is much simpler than the approach I previously used.

Everywhere else, I added new methods to HSLogMan to use an NAString buffer instead of a fixed length buffer for accumulating log message text. The buffer is populated using C++-style operator<<. That buffer is cleared when the accumulated text is written to the log. I checked the QRLogger and underlying CommonLogger code and they appear to support unlimited message size. So, the new methods in HSLogMan now also support this. This eliminates the problematic use of sprintf.

For now, the old logging methods and new logging methods will live side-by-side; over time use of the old methods can be replaced with the new. I believe I've changed all the places that were needed for long multi-column lists.

core/sql/ustat/hs_log.h
```diff
@@ -277,5 +283,8 @@ class HSLogMan
          timeval prevTime_;
          char title_[MAX_TIMING_EVENTS][1000];
          static THREAD_P HSLogMan* instance_;      /* 1 and only 1 instance           */
+
+         // intended as a long-term replacement for msg + Log()
+         NAString logStreamBuffer_;
```
sureshsubbiah: I like this change very much. Using an NAString seems better than using fixed length buffers. I do wonder if this NAString needs to be allocated on a partcular heap. I think the default, is to use system heap. Since this class is a singleton maybe the context heap is a better place for it, allowing heap usage to traced using executor counters. I am not sure if the heap change is necessary or even beneficial when considering all aspects such as multiple Contexts being present in situations such as T2 and repository threads. Thread safety of this variable when we have multiple T2 connections doing update stats on different tables from the same process may also be an interesting test. These are minor concerns. The approach and change are a resounding +1 for me.

DaveBirdsall: Thanks for the review. I'll explore putting logStreamBuffer_ on an NAHeap in a later pull request. One bit of good news is that this code runs in the tdm_arkcmp process, so we won't be affecting global heap in mxosrvr.

PR #1824, opened 2019-04-04
[TRAFODION-3280] Reduce path length in Trafodion for improved performance and scalability 

selvaganesang:
Removed/avoided the following APIs shown as hotspots when Trafodion is run with OLTP kind of queries.

java.net.Socket.setSoTimeout
org.trafodion.jdbc.t4.T4LoggingUtilities.log
java.net.Socket.getLocalSocketAddress
java.lang.String.split
java.lang.String.replaceAll
java.lang.Class.getMethod
org.trafodion.jdbc.t4.TrafT4ResultSet.findColumn

Now socket.read waits for the network timeout. It can be configured via a property networkTimeout or call to connection.setNetworkTimeout. If the networkTimeout is not configured, a default value of 10 secs is assumed.

The caller is expected to set the timeout depending upon the operation being performed like loginTimeout, queryTimeout. Socket.read will wake up for every networkTimeout expiration and check if the timeout for the operation hasn't exceeded. When it exceeds, SocketTimeoutException is thrown.

Changed the T4Statement and T4Restult classes to have "has a" relationship with T4Connection for better understanding of the code. Earlier, it had "is a" relationship with T4Connection.

JDBC driver was parsing the sql string passed to it by the application to determine the type of SQL statement and to check if the sql string has parameters. It doesn't parse the sql string anymore
and passes the sql string to SQL engine via mxosrvr. In case of PreparedStatement, the JDBC driver will
set the sqlStmtType based on the query type returned. In case of Statement, the JDBC driver will set the
sqlStmtType as SQL_TYPE_UNKNOWN.

Now, the PreparedStatement is always prepared for setting array values for parameter for all
statement types such as IUD and select. SQL engine will report an error if the array values are
not supported for any statement type.

Changed Hashtable to HashMap where possible with the assumption that the JDBC connection
shouldn't be used from different threads. TrafT4Connection object needs be made thread-safe.

Optimized org.trafodion.jdbc.t4.TrafT4ResultSet.findColumn by using HashMap of columnName to
columnIndex. This should avoid excessive calls to java.lang.String.equalsIgnoreCase.

Removed excessive calls to java.lang.Class.getMethod from TrafT4ResultSet.

When the statement type is TYPE_UNKNOWN, mxosrvr was executing the extra code
to load controls, GetHashInfo etc. GetHashInfo was corrupting the sqlString by
truncating spaces in between the literals. Removed these code from mxosrvr because it is specific to QueryStatsServer as part of NeoView and WMS.

There are many places in our code getenv is called repeatedly. Cleaned up code so that
repeated getenv calls are avoided during prepare and execute operations.


core/conn/jdbcT4/src/main/java/org/trafodion/jdbc/t4/InterfaceStatement.java
```diff
@@ -1228,20 +1111,13 @@ void prepare(String sql, int queryTimeout, TrafT4PreparedStatement pstmt) throws
 //			txId = Bytes.createIntBytes(0, false);
```
hegdean: can we delete the commented line

core/conn/odbc/src/odbc/nsksrvrcore/sqlinterface.cpp
```diff
@@ -5492,10 +5492,12 @@ SQLRETURN SRVR::PREPARE2withRowsets(SRVR_STMT_HDL* pSrvrStmt)
 	HANDLE_ERROR(retcode, sqlWarning);
 
 	if (pSrvrStmt->maxRowsetSize > 1 
+/*
```
hegdean: remove the comment section

core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
```diff
@@ -1209,8 +1209,8 @@ odbc_SQLSvc_Prepare2withRowsets_sme_(
 		if (pSrvrStmt->maxRowsetSize == ROWSET_NOT_DEFINED) pSrvrStmt->maxRowsetSize = DEFAULT_ROWSET_SIZE;
 
 // this part is for NAR (not Atomic Rowset Recovery)
-		if (pSrvrStmt->maxRowsetSize > 1
-            && (pSrvrStmt->sqlStmtType == TYPE_INSERT_PARAM))
+		if (pSrvrStmt->maxRowsetSize > 1)
+//            && (pSrvrStmt->sqlStmtType == TYPE_INSERT_PARAM))
```
hegdean: same here. delete commented lines

PR #1828, opened 2019-04-10
[TRAFODION-3296] Fix handling of [first n] + ORDER BY subqueries

DaveBirdsall:
Queries containing IN-subqueries involving both [first n] and ORDER BY were giving wrong results. The ORDER BY essentially was being ignored. This has now been fixed. The [first n] is now evaluated with respect to the ORDER BY clause.

For [any n] + ORDER BY (or equivalently, LIMIT n + ORDER BY), the ORDER BY continues to be ignored, since [any n] does not specify which rows should be returned and therefore ordering has no importance.

The combination [last n] + ORDER BY in an IN-subquery is now forbidden, and results in a 4484 error. The reason is that [last n] is typically used as a performance tool when used at the top level of a query, but that purpose makes no sense in the context of a subquery.

This particular set of changes is gated by a new CQD, COMP_BOOL_114. When set to 'ON' (which is the default), the fix executes. When set to 'OFF', the fix logic is bypassed. This is a short-term measure to allow testing of before-and-after behavior.


core/sql/regress/core/TEST002
```diff
@@ -613,8 +623,44 @@ select
...
+
+prepare xx from select val from t002x1 where val in (select [first 1] val from t002x2 order by val);
+
+-- query plan should be a sequence plan, not a firstn plan; the ORDER BY causes a rewrite
+explain options 'f' xx;
+
+-- should return one row, -1865644273
+execute xx;
+
+prepare xx from select val from t1 where val in (select [any 1] val from t2 order by val);
```
DaveBirdsall: The table name here should be t002x1, not t1. (Found this when I reran this test on a workstation after cleaning my instance.) Similarly for t2.

core/sql/regress/core/TEST002
```diff
@@ -613,8 +623,44 @@ select
...
+
+-- query plan should be a firstn plan
+explain options 'f' xx;
+
+-- should fail with error 4484
+prepare xx from select val from t002x1 where val in (select [last 1] val from t002x2 order by val);
+
+-- limit n has same semantics as [any n]
+prepare xx from select val from t1 where val in (select val from t2 order by val limit 1);
```
DaveBirdsall: Another example where t1 should be t002x1, and t2 should be t002x2

DaveBirdsall:
FYI: I ran full regressions overnight on these changes. Aside from core/TEST002 (which failed because of my editing mistake), all tests passed. I've since pushed a fix to core/TEST002.

core/sql/optimizer/BindRelExpr.cpp
```diff
@@ -6807,6 +6846,100 @@ RelExpr *RelRoot::bindNode(BindWA *bindWA)
...
+  ColRefName * colRefName = new (bindWA->wHeap()) ColRefName(rowNumberColumnName, bindWA->wHeap());
+  RenameCol * rename = new (bindWA->wHeap())RenameCol(rowNumberOverOrderBy, colRefName);
+
+  // add it to select list of the current RelRoot
+  compExprTree_ = new (bindWA->wHeap()) ItemList(compExprTree_,rename);
+
+  // put a RelSequence node on top of the query node, and make it the child of this
+  RelSequence * sequenceNode = new (bindWA->wHeap()) RelSequence(query,NULL);
+  sequenceNode->setHasOlapFunctions(TRUE);
+  this->child(0) = sequenceNode;
```
sandhyasun: Does the sequence node sends a GET_ALL or GET_N to it's child ?

sandhyasun: Also what happens if there are ESPs How does the plan look ?

DaveBirdsall: Looking at the code in executor/ExSequence.cpp, it looks like it sends GET_ALL. It is at the sequence node that the data is ordered and the OLAP function is calculated. We have to do a GET_ALL in order to get the ordering.

DaveBirdsall: Re: ESPs: I will research and find out. If I recall correctly, sequence plans were not easy to parallelize. But that recollection is many years old. It may have improved since.

sandhyasun: > Looking at the code in executor/ExSequence.cpp, it looks like it sends GET_ALL. It is at the sequence node that the data is ordered and the OLAP function is calculated. We have to do a GET_ALL in order to get the ordering.
Ok so if that's the case, then with this first N in subquery, we won't be getting any performance  , correct ? It will work correctly but wondering if the user will expect it to run  faster  than if they did not provide the first N ?

DaveBirdsall: Hmmmm... The presence of an ORDER BY implies that all rows must be read and sorted. There is no escaping that. Even in firstn plans (where [first n] + ORDER BY is specified at the top level query) we do that. There may be an optimization possibility of doing Top N sort rather than a full sort. But in any case, all rows must be read.

DaveBirdsall: Re: ESPs (again): I just did an example with two tables of 2 million rows each, and forced ESP parallelism. The sequence operator is done serially, in one ESP. So my recollections were correct, we do not parallelize OLAP operators today.

DaveBirdsall: I should add that in my experiment, all other parts of the plan were parallelized. So for example, the scans underneath the sequence operator were done in parallel, and each was sorted in parallel, with the results funneled into the (serialized) sequence operator.

sandhyasun:
+1  Change looks good.  Performance is a more minor issue - it's good that this query actually works correctly now.

DaveBirdsall:
Thank you, @sandhyasun, for your review.

PR #1831, opened 2019-04-18
[TRAFODION-3300] Fix overflow issues with extreme big nums + ROUND

DaveBirdsall:

1. Implements the ROUND function for BIG NUM data types (e.g. NUMERIC(n,m) where n > 18) in the expression evaluator. Formerly, the binder would rewrite ROUND as a rather complicated series of other functions, which unfortunately cause overflows in extreme cases.
2. Adds extensive white box tests to the regression library for the new ROUND implementation.
3. Adds error message 4054 to cover the case where a user specifies ROUND(A,m) where m is larger than the maximum precision allowed for the data type. Formerly, we would give an incorrect error in this case.
4. Add method NAType::display to make it easier for developers to view NAType information when debugging the compiler.
5. As a test aid, I left the old code in. If CQD COMP_BOOL_15 is set to 'OFF', we use the old code. If it is 'ON' we use the new code. The default is 'ON'. It is intended that a future JIRA will remove this CQD and remove the old code, once the new code has sufficient exposure.


core/sql/generator/GenPreCode.cpp
```diff
@@ -10033,17 +10033,56 @@ ItemExpr * MathFunc::preCodeGen(Generator * generator)
...
+              if (!typ1.isInteger() || 
+                  ((typ1.getFSDatatype() != REC_BIN8_SIGNED) &&
+                   (typ1.getFSDatatype() != REC_BIN8_UNSIGNED) && 
+                   (typ1.getFSDatatype() != REC_BIN16_SIGNED) &&
+                   (typ1.getFSDatatype() != REC_BIN16_UNSIGNED) &&
+                   (typ1.getFSDatatype() != REC_BIN32_SIGNED) &&
+                   (typ1.getFSDatatype() != REC_BIN32_UNSIGNED) &&
+                   (typ1.getFSDatatype() != REC_BIN64_SIGNED) &&
+                   (typ1.getFSDatatype() != REC_BIN64_UNSIGNED) ) )
+                {
```
anoopsharma00: Could use isBinaryNumeric method from dfs2rec class  unless these
specific checks are needed

core/sql/regress/executor/TEST101
```diff
@@ -214,6 +214,97 @@ control query default BIGNUM_IO 'ON';
...
+-- negative test: round using a non-integer decimal column
+select a,round(b,f) from t101big;
+
+-- negative test: round using rounding values > max big num precision
+select a,round(b,129) from t101big;
+
+-- but -129 works fine
+select a,round(b,-129) from t101big;
+
+drop table t101big;
```
anoopsharma00: good to see extensive tests. These would probably catch more issues than visual code review.

core/sql/generator/GenPreCode.cpp
DaveBirdsall: Didn't know about this method. Just took a look at it. It includes REC_BPINT_UNSIGNED, which I had not added run-time support for. I'll add it and make this change.

DaveBirdsall: Done! Thanks for the suggestion.

PR #1837, opened 2019-04-25
[TRAFODION-3306] Add the TRUNC Function in the Trafodion SQL Reference Manual

liuyu000:
@DaveBirdsall Could you please help review? Many thanks.

### DIFF
--- a/docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
+++ b/docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
@@ -368,9 +368,9 @@ It is *reverse* of the CEIL function.
 
 | <<round_function, ROUND Function>> 
 
-|`ROUND (_numeric-expression_, _num_)` 
+|`ROUND (_numeric-expression_, _decimal-place_)` 
 
-|Rounds a `_numeric-expression_` on `_num_` decimal positions.
+|Rounds a `_numeric-expression_` on `_decimal-place_` decimal positions.
 
 This is the most complicated one of the four functions. For more information, see <<round_function,ROUND Function>>.
 
@@ -379,11 +379,11 @@ This is the most complicated one of the four functions. For more information, se
 |<<trunc_function, TRUNC Function>>  +
 (TRUNCATE)
 
-|`TRUNC (_numeric-expression_, _num_)` +
+|`TRUNC (_numeric-expression_, _decimal-place_)` +
 
-`TRUNCATE (_numeric-expression_, _num_)`
+`TRUNCATE (_numeric-expression_, _decimal-place_)`
 
-|Truncates a `_numeric-expression_` on `_num_` decimal positions without rounding.
+|Truncates a `_numeric-expression_` on `_decimal-place_` decimal positions without rounding.
 
 |The TRUNC function is similar to the ROUND function in terms of *reducing the number of decimal places*. +
 
@@ -5993,8 +5993,8 @@ The FLOOR function rounds down a `_numeric-expression_` to the largest integer l
 
 `FLOOR (_numeric-expression_)`
 
-[[syntax_description_of_floor]]
-=== Syntax Description of FLOOR
+[[syntax_descriptions_of_floor]]
+=== Syntax Descriptions of FLOOR
 
 `_numeric-expression_`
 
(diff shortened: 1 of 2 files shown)


docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
```diff
@@ -12460,6 +12460,122 @@ CHAR(20). This expression uses the TRIM function to return the value
 TRIM(first_name) || ' ' || TRIM (last_name)
 ```
 
+[[trunc_function]]
+== TRUNC Function
+
+The TRUNC (TRUNCATE) function truncates `_numeric-expression_` on an optional `_decimal-place_` decimal position(s), that is, it rounds down `_numeric-expression_` to the nearest (or equal) integer closer to zero *without rounding*.
```
DaveBirdsall: The second argument makes this description awkward. If the second argument were zero, then we are rounding down to the nearest integer closer to zero. If the second argument is positive. we aren't necessarily rounding to an integer. TRUNC(1.23,1) gives 1.2, and 1.2 isn't an integer. And I'm not sure what happens when the second argument is negative. Maybe TRUNC(123,-1) gives 120?

If this is true, one way to describe this is TRUNC is like ROUND, except it always rounds toward zero.

Here is a possible wordsmith (please do a bit of experimentation to see if this is right);

The TRUNC (TRUNCATE) function truncates '_numeric-expression_'. If the second argument is not supplied, or is supplied and is zero, this rounds the '_numeric-expression_' to the nearest integer closer to zero. If the second argument is supplied and is positive, say, n, it rounds the '_numeric_expression_' to the nearest multiple of 10^-n that is closer to zero. In these two cases, decimal places to the right of the decimal point are effectively deleted. If the second argument is supplied and is negative, say, -n, it rounds '_numeric_expression_' to the nearest multiple of 10^n. Any decimal places to the right of the decimal point are effectively deleted.

docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
```diff
@@ -12460,6 +12460,122 @@ CHAR(20). This expression uses the TRIM function to return the value
...
+For more information, see <<numeric_value_expressions,Numeric Value Expressions>>.
+
+* `_decimal-place_`
+
++
+is an integer specifying the precision of the truncation.
+
+NOTE: `_numeric-expression_` and `_decimal-place_` cannot be NULL.
+
+NOTE: Currently, `_decimal-place_` can be zero or a positive integer, while *negative integer* is *not supported*.
```
DaveBirdsall: Ah, this makes it easier. We don't have to worry about the negative case for the second argument.

docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
```diff
@@ -12460,6 +12460,122 @@ CHAR(20). This expression uses the TRIM function to return the value
...
+  &#9702; `_decimal-place_` is not specified 
+
+| *TRUNC* function = *CEIL* function
+
+|===
+
+* The TRUNC function is similar to the ROUND function in terms of *reducing the number of decimal places*. 
+
++
+However, the TRUNC function *does not perform any rounding* as the ROUND function does.
```
DaveBirdsall: I would say this differently. I would say that TRUNC rounds toward zero. While ROUND rounds to the nearest integer.

liuyu000: @DaveBirdsall Thank you, I've incorporated your comment.

docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
liuyu000: For most databases, TRUNC(123,-1) gives 120, while Trafodion doesn't support negative _decimal-place_.

docs/sql_reference/src/asciidoc/_chapters/sql_functions_and_expressions.adoc
liuyu000: For most databases, TRUNC(123,-1) gives 120, while Trafodion doesn't support negative decimal-place.
Your suggestions are correct, I've incorporated them. Thanks again.

PR #1839, opened 2019-05-10
TRAFODION-3280] Reduce path length in Trafodion for improved performance and scalability

selvaganesang:
Improvements in DTM register region concept.

1. TransactionalTableClient is retired. ttable in RMInterface is now HTable. SSCC transaction type might need some rework.
2. registerTransaction method is now refactored to reduce pathlength.
3. All the regions during scan are registered upfront at the time of scan instead of trying to figure out if new regions are accessed while the rows are processed.
4. Removed unnecessary synchronization of RMInterface methods.
5. Changed idTm to be one per process.
6. Changes in SQL layer to bypass the java layer when start row id is less than the end row id.
7. RegionList is now collected per table on the client side using HashMap instead of Set. This
allows client to register to TM the region only if it not already registered within a transaction.

Added to code to enable java debugging of TM process via the variable JVM_DEBUG_PORT in ms.env.
If JVM_DEBUG_PORT=xx000, then the debug port is set to xx * 1000 + <tm_pid> % 1000.


selvaganesang:
jenkins, retest

selvaganesang:
jenkins, retest

selvaganesang:
jenkins, retest

PR #1840, opened 2019-05-15
[TRAFODION-3308] Improve error reporting when a process cannot be started

DaveBirdsall:
This pull request contains improvements to error reporting when a process server cannot be started. The following changes are made:

1. Logic in the compiler (optimizer/NodeMap.cpp) to compute node names has been corrected. Thanks to @sandhyasun for this change.
2. The correct node name is now reported in error messages 2012 and 2013 (common/IpcGuardian.cpp). Formerly, a hard-coded value of "NSK" was reported.
3. Redundant 2013 and 2002 error messages are no longer generated (common/IpcGuardian.cpp and cli/ExSqlComp.cpp).
4. Interpretive text has been added to error 2012 to give a more human-understandable reason behind certain common error codes (common/IpcGuardian.cpp and bin/SqlciErrors.txt).
5. The nodeName_ member in IpcGuardianServer is now computed at run time, and only if needed (common/IpcGuardian.cpp). (This happens during error reporting only.) This is done because the compile time value may be different in the case of ESPs due to node down conditions. Former logic that put "NSK" in this field has been removed.
6. The function ComRtGetOSClusterName is an obsolete function that in predecessor products returned the cluster name. On Trafodion, it just returns the string "NSK". I removed much but not all of the logic that calls this function. This resulted in the removal of some dead code (executor/ExCancel.cpp, executor/ExExeUtilCommon.cpp).
7. The method IpcServerClass::getProcessName required the node name in predecessor products but does not in Trafodion. I removed the node name and node name length parameters from this method and associated code (common/Ipc.cpp). I was motivated to do this because the old calling code sometimes took strlen(nodeName_) to get the node name length, but due to the changes above nodeName_ is now null in some cases.
8. The Messages Guide entries for messages 2012 and 2013 has been updated.

### DIFF
--- a/core/sql/cli/ExSqlComp.cpp
+++ b/core/sql/cli/ExSqlComp.cpp
@@ -172,7 +172,8 @@ ExSqlComp::ReturnStatus ExSqlComp::createServer()
   //  
   if (ret == ERROR)
   {
-	error(arkcmpErrorServer);
+	if ((!diagArea_->contains(-2013)) && (!diagArea_->contains(-2012))) // avoid generating redundant error
+	  error(arkcmpErrorServer);
 	if (getenv("DEBUG_SERVER"))
 	  MessageBox(NULL, "ExSqlComp:createServer", "error ", MB_OK|MB_ICONINFORMATION);
   }
@@ -911,7 +912,7 @@ ExSqlComp::ExSqlComp(void* ex_environment,
                      short version,
                      char *nodeName ,
                      IpcEnvironment *env):
-isShared_(FALSE), lastContext_(NULL), resendingControls_(FALSE)
+isShared_(FALSE), lastContext_(NULL), resendingControls_(FALSE), nodeName_(NULL)
 {
   h_ = h;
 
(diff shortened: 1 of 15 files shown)


DaveBirdsall:
jenkins, retest

DaveBirdsall:
jenkins, retest

DaveBirdsall:
jenkins, retest

core/sql/executor/ExUdrServer.cpp
```diff
@@ -298,7 +287,7 @@ ExUdrServer::ExUdrServerStatus ExUdrServer::start(ComDiagsArea **diags,
     ipcServer_ =
       udrServerClass_->allocateServerProcess(diags,
                                              diagsHeap,
-                                             nodeName,
+                                             NULL,
```
sandhyasun: Would the udr server still be colocated with the ESP ? It needs to know the nodename ot start it right ?

DaveBirdsall:
jenkins, retest

core/sql/cli/ExSqlComp.cpp
```diff
@@ -172,7 +172,8 @@ ExSqlComp::ReturnStatus ExSqlComp::createServer()
   //  
   if (ret == ERROR)
   {
-	error(arkcmpErrorServer);
+	if ((!diagArea_->contains(-2013)) && (!diagArea_->contains(-2012))) // avoid generating redundant error
+	  error(arkcmpErrorServer);
```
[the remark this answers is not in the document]
DaveBirdsall: Actually, the error that we suppress here is error 2002, which simply says, "internal error". See the JIRA for an example.

core/sql/executor/ExUdrServer.cpp
DaveBirdsall: Yes. This is determined by the next parameter, "collocatedCPU".

core/sql/cli/ExSqlComp.cpp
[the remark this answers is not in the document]
sandhyasun: Ok . Yes I see... misread the if condition.

core/sql/executor/ExUdrServer.cpp
sandhyasun: Ok

svarnau:
jenkins, retest -- changed one jenkins option, checking to see if it makes a difference

svarnau:
jenkins, retest -- previous change worked, but installer ran into a new problem, added new work-around

svarnau:
jenkins, retest

PR #1842, opened 2019-05-22
[TRAFODION-3310] Commit transaction needs to be complete before returning to the caller.

selvaganesang:
The no-waited commit is not an option that is valid in Trafodion. SQL needs to ensure that commit completes before it gives control to the application. Hence removed all the code that are related to nowait commit as per review.


selvaganesang:
jenkins, retest

svarnau:
jenkins, retest -- jenkins server has been restarted

selvaganesang:
> +1 Looks good to me. Looks like the primary change is to remove the predecessor concept of "DP2 transaction", and also to allow the CLI to not wait for transaction commit if the environment variable DTM_EARLYCOMMITREPLY is defined and non-zero. Is that correct?

Yes. It is no good to use the passed in parameter to commitTransaction  method to determine if the commit needs to be waited or not. It is important that commit completes before caller gets the control to ensure data consistency

selvaganesang:
jenkins, rettest

selvaganesang:
jenkins, retest

PR #1843, opened 2019-05-23
[TRAFODION-3311] Trafodion to start transactions for select statements with FOR UPDATE

selvaganesang:
A CQD BEGIN_TRANSACTION_FOR_SELECT is introduced. This CQD can take a value 1,2 or any numeric value. Default is 1.

1 - means mark the statement as transaction required if it is SELECT ... FOR UPDATE
2 - means mark the statement as transaction required if it is SELECT statement with
    isolation other than SKIP CONFLICT ACCESS or READ UNCOMMITTED ACCESS.
Any other value - SELECT statement is not marked to begin transaction.

Trafodion always change the isolation level to READ COMMITTED ACCESS when it is
READ UNCOMMITTED ACCESS.


selvaganesang:
@anoopsharma00 Can you please review this change

svarnau:
jenkins, retest -- trying new improved jenkins

PR #1844, opened 2019-06-07
[TRAFODION-2311] Avoid logging error 8605 and 8609 during SQLEndTrans if no trans active

arvind-narain:
Frequently , during performance test runs like OE and other regression tests, following errors
are observed in master logs:

ERROR[8605] Committing a transaction which has not started.
ERROR[8609] Waited rollback performed without starting a transaction.

MXOSRVR does return success to the client for these errors but since commit/rollback is executed, the error gets logged.

Some of these can be avoided if mxosrvr implements the following ODBC standard and skips executing the commit/rollback statement.

"For drivers and data sources that support transactions, calling SQLEndTran with either SQL_COMMIT or SQL_ROLLBACK when no transaction
is active returns SQL_SUCCESS (indicating that there is no work to be committed or rolled back) and has no effect on the data source."

### DIFF
--- a/core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
+++ b/core/conn/odbc/src/odbc/nsksrvrcore/srvrothers.cpp
@@ -3234,15 +3234,26 @@ odbc_SQLSvc_EndTransaction_sme_(
 	SRVRTRACE_ENTER(FILE_SME+5);
 
 	char stmtLabel[MAX_STMT_LABEL_LEN+1];
-	Int32 rc = SQL_SUCCESS;
 	SRVR_STMT_HDL	*pSrvrStmt = NULL;
+	bool isTransPending = (WSQL_EXEC_Xact(SQLTRANS_STATUS, 0) == 0);
+	Int32 rc = SQL_SUCCESS;
+
+	exception_->exception_nr = 0;
+	sqlWarning->_buffer = NULL;
+	sqlWarning->_length = 0;
 
 	switch (transactionOpt) {
 	case SQL_COMMIT:
-		pSrvrStmt = getSrvrStmt("STMT_COMMIT_1", FALSE);
+		if (isTransPending)
+			pSrvrStmt = getSrvrStmt("STMT_COMMIT_1", FALSE);
+		else
+			return;
 		break;
 	case SQL_ROLLBACK:
-		pSrvrStmt = getSrvrStmt("STMT_ROLLBACK_1", FALSE);
+		if (isTransPending)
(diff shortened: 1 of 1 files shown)


sbroeder:
I have some reservations about this change because sometimes this warning message serves as an indicator for some other error that has caused a transaction to be aborted unilaterally.  When this happens, a user transaction can turn into autocommit SQL statements and the business transaction unit may not be applied atomically.  If the warning message is removed there will be no indication this has happened.

selvaganesang:
@arvind-narain  Can you please confirm what is the status returned by SQL_EXEC_Xact when the transaction is unilaterally aborted. If it returns that there is a still active transaction, then this change should be good. You can test it via SPJ that seems to unilaterally abort the transaction when an error is encountered.

arvind-narain:
Thanks @sbroeder and @selvaganesang. Sorry for the delay. Will check SQL_EXEC_Xact and update. This change is mainly needed since some packages written for other databases may expect a transaction to be started even for a select. If the code is ported over then such application fills up the logs on every commit done after a select. As stated above ODBC standards do state that SQL_SUCCESS should be returned on a SQLEndTrans if there are no transactions pending. If numerous 8605 and 8609 messages are present in the logs then troubleshooting other issues becomes difficult.

selvaganesang:
The ODBC specification says the following when SQLEndTran API is called 

For drivers and data sources that support transactions, calling SQLEndTran with either SQL_COMMIT or SQL_ROLLBACK when no transaction is active returns SQL_SUCCESS (indicating that there is no work to be committed or rolled back) and has no effect on the data source.

When a driver is in autocommit mode, the Driver Manager does not call SQLEndTran in the driver. SQLEndTran always returns SQL_SUCCESS regardless of whether it is called with a CompletionType of SQL_COMMIT or SQL_ROLLBACK.

Hence this change can be merged.

xiaozhongwang:
Very good feature for compatible.

On 星期四, 六月 27, 2019 08:04 CST, Selvaganesan Govindarajan <notifications@github.com> 撰寫:
 
The ODBC specification says the following when SQLEndTran API is called
For drivers and data sources that support transactions, calling SQLEndTran with either SQL_COMMIT or SQL_ROLLBACK when no transaction is active returns SQL_SUCCESS (indicating that there is no work to be committed or rolled back) and has no effect on the data source.
When a driver is in autocommit mode, the Driver Manager does not call SQLEndTran in the driver. SQLEndTran always returns SQL_SUCCESS regardless of whether it is called with a CompletionType of SQL_COMMIT or SQL_ROLLBACK.
Hence this change can be merged.

PR #1845, opened 2019-07-01
[TRAFODION-3263] and other misc fixes for lob locking and refactoring 

sandhyasun:
These fixes are cherry picked from another branch and all were reviewed earlier.
1. Fixes a core that was caused since ExLOB::update was missing a chunk of code very similar to insert . This code path is taken when updating a NULL entry. Added a testcase for it in TEST130.
2. Regressions on Jenkins have been getting an InterruptException. Made a change to call "close" only once instead of twice. Also added a retry to take care of a small window that can cause issues when there are multiple LOB updates/inserts going on parallelly.
3. SQL_EXEC_SwitchContext was not allowing switching back to default context. In the case of Udr ResultSets we make an exception with this change.
In general all work will be done in the connections that are set up. So setContext in Udr ResultSet will still not allow switching to the default context. Only resetContext() will allow it. This is a day-one bug.


core/sql/cli/Cli.cpp
```diff
@@ -8554,17 +8558,34 @@ Lng32 SQLCLI_LOBcliInterface
...
+	    goto error_return;
+        char *inputValues = new (currContext.exHeap()) char[sizeof(Int32) + strlen(lobDescHandleName) + sizeof(Int64)*2];
+        *(Int32 *)inputValues= strlen(lobDescHandleName);
+        memcpy(&inputValues[sizeof(Int32)],lobDescHandleName,strlen(lobDescHandleName));
+       
+        inputValues[sizeof(Int32) + strlen(lobDescHandleName)] = descPartnKey;
+        inputValues[sizeof(Int32) + strlen(lobDescHandleName)+sizeof(Int64)] = (dataLen ? *dataLen : 0);
+        inputValuesLen = sizeof(Int32) + strlen(lobDescHandleName)+sizeof(Int64)*2;
+        cliRC = cliInterface->clearExecFetchCloseOpt((char *)inputValues,inputValuesLen,(char*)&descSyskey,&len, &rowsAffected);
+        */
```
DaveBirdsall: Why was this block of commented-out code added?

core/sql/cli/SQLCLIdev.h
```diff
@@ -37,7 +37,20 @@
...
+#if defined(_THROW_EXCEPTIONS)
+     if (cliWillThrow())
+       {
+         throw;
+       }
+#endif
+   }
+
+   RecordError(NULL, retcode);
+   return retcode;*/
```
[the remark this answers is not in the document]
sandhyasun: Good catch. It's this way int he source branch too where the merge was done from. Removing in both.

core/sql/cli/Cli.cpp
```diff
@@ -8554,17 +8558,34 @@ Lng32 SQLCLI_LOBcliInterface
     case LOB_CLI_INSERT:
       {
 	// insert into lob descriptor handle table
-	str_sprintf(query, "select syskey from (insert into table(ghost table %s) values (%ld, 1, %ld)) x",
-		    lobDescHandleName, descPartnKey, (dataLen ? *dataLen : 0));
+        str_sprintf(query, "select syskey from (insert into table(ghost table %s) values (%ld, 1, %ld)) x",
+                    lobDescHandleName, descPartnKey, (dataLen ? *dataLen : 0));
+                /*str_sprintf(query, "select syskey from (insert into table(ghost table cast(? as varchar(1024)) values (cast(? as largeint), 1, cast(? as largeint))) x");*/
+
```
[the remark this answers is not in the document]
sandhyasun: Removed the commented out code.

core/sql/cli/Cli.cpp
sandhyasun: This was from the source branch when a new method of saving prepared statements was attemtped. Removing it since it's being reworked.

core/sql/exp/ExpLOB.cpp
```diff
@@ -6,7 +6,9 @@
 // distributed with this work for additional information
 // regarding copyright ownership.  The ASF licenses this file
 // to you under the Apache License, Version 2.0 (the
-// "License"); you may not use this file except in compliance
+// "License
+
+"); you may not use this file except in compliance
```
DaveBirdsall: Will this compile? Looks like a stray line feed got added to the code.

sandhyasun: It compiled since it was  encompassed in a larger comment but correcting it anyway.

PR #1846, opened 2019-07-02
[TRAFODION-3314] Avoid generating redundant DDL in OSIM for unique constraints

DaveBirdsall:
The problem: When a table has a unique constraint, OSIM generates both the ALTER TABLE statement to create the constraint and the CREATE INDEX statement for the system-generated index into CREATE_TABLES_DDL.txt. When we then do OSIM LOAD, the CREATE INDEX is executed, then the ALTER TABLE. The ALTER TABLE fails because the index already exists.

The fix: Don't write the CREATE INDEX to the CREATE_TABLES_DDL.txt file.

### DIFF
--- a/core/sql/optimizer/OptimizerSimulator.cpp
+++ b/core/sql/optimizer/OptimizerSimulator.cpp
@@ -541,31 +541,43 @@ void OptimizerSimulator::dumpDDLs(const QualifiedName & qualifiedName)
                                    <<"."<< qualifiedName.getSchemaName() 
                                    << ";" << endl;
         
+        // skippingSystemGeneratedIndex is set to TRUE to avoid generating redundant
+        // DDL for system-generated indexes
+        NABoolean skippingSystemGeneratedIndex = FALSE;
         outQueue->position();//rewind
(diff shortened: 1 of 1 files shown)


core/sql/optimizer/OptimizerSimulator.cpp
```diff
@@ -541,31 +541,43 @@ void OptimizerSimulator::dumpDDLs(const QualifiedName & qualifiedName)
...
+                }
+
+                //output ddl  
+                (*createTable) << ptr << endl;
             }
 
-            //output ddl    
-            (*createTable) << ptr << endl;
+            if (skippingSystemGeneratedIndex && (strcmp(ptr,";") == 0)) // at end of DDL to be skipped?
+              skippingSystemGeneratedIndex = FALSE;
```
DaveBirdsall: Note to reviewers: The code from lines 556 through 576 in the new file is merely re-indented lines 548 through 568 from the old file, with an "if (!skippingSystemGenerated) { ... }" around it. Unfortunately the comparison tool doesn't render this difference nicely.

PR #1848, opened 2019-07-09
[TRAFODION-3316] Three fixes to UPDATE STATISTICS

DaveBirdsall:
Three fixes to UPDATE STATISTICS:

1. Add support for Hive TIMESTAMP data type. The only change needed is to tolerate a precision of 9 on a TIMESTAMP.

2. When creating a sample table on a small Hive table, the SAMPLING_RATIO recorded in SB_PERSISTENT_SAMPLES is incorrect. This happened because the estimated size for the table was off by two orders of magnitude. We calculated the sampling ratio from the number of rows returned in the sample and this estimated size. There was logic to adjust the estimated size of the table as well from the statistics of the sample SELECT, however that logic was stubbed out because it didn't work. That logic has been removed -- the design premise was flawed anyway because it does not take into account the possibility that a key predicate will be used in incremental UPDATE STATISTICS. Instead, we now re-estimate the rows in the original table using the user-specified sampling ratio and the number of rows in the sample.

Note: As part of this change, a parameter to HSFuncExecQuery is nearly obsolete. I did not replace it however because this function is called in over a hundred places in the code. Will leave that for a later clean-up.

3. There was one place (extractTblName and its caller) where we were using a const char * pointer into an NAString object built on the stack that goes out of scope. This causes occasional failures. This has been fixed.

### DIFF
--- a/core/sql/ustat/hs_cli.cpp
+++ b/core/sql/ustat/hs_cli.cpp
@@ -176,6 +176,9 @@ Lng32 HSExecDirect( SQLSTMT_ID * stmt
 //          inactivateErrorCatcher = TRUE if the caller already has an
 //                HSErrorCatcher object active (that is, the caller wants
 //                to capture diagnostics itself).
+// Note: srcTabRowCount is an obsolete parameter. Its only function now
+// is to control whether plan information should be kept. Later we can
+// clean this out, replacing it with an NABoolean.
 // -----------------------------------------------------------------------
 Lng32 HSFuncExecQuery( const char *dml
                     , short sqlcode
@@ -194,7 +197,7 @@ Lng32 HSFuncExecQuery( const char *dml
   HSErrorCatcher errorCatcher(retcode, sqlcode, errorToken, TRUE,
                               inactivateErrorCatcher);
   retcode = HSFuncExecQueryBody(dml,sqlcode,rowsAffected,errorToken,
-                                srcTabRowCount,tabDef,errorToIgnore,checkMdam);
+                                srcTabRowCount != NULL,tabDef,errorToIgnore,checkMdam);
   HSHandleError(retcode);
   return retcode;
 }
@@ -209,7 +212,7 @@ Lng32 HSFuncExecQueryBody( const char *dml
                     , short sqlcode
                     , Int64 *rowsAffected
                     , const char *errorToken
-                    , Int64 *srcTabRowCount
+                    , NABoolean printPlan
                     , const HSTableDef *tabDef
                     , short errorToIgnore
                     , NABoolean checkMdam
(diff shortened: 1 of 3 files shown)


core/sql/ustat/hs_cli.cpp
```diff
@@ -194,7 +197,7 @@ Lng32 HSFuncExecQuery( const char *dml
   HSErrorCatcher errorCatcher(retcode, sqlcode, errorToken, TRUE,
                               inactivateErrorCatcher);
   retcode = HSFuncExecQueryBody(dml,sqlcode,rowsAffected,errorToken,
-                                srcTabRowCount,tabDef,errorToIgnore,checkMdam);
+                                srcTabRowCount != NULL,tabDef,errorToIgnore,checkMdam);
   HSHandleError(retcode);
```
nonstop-qfchen: We could use !srcTabRowCount here.

core/sql/ustat/hs_globals.cpp
```diff
@@ -4307,11 +4302,12 @@ Lng32 HSSample::make(NABoolean rowCountIsEstimate, // input
     // 2 cqds allow user control of these retries.
     Int32 centiSecs = getDefaultAsLong(USTAT_RETRY_DELAY);
     Int32 limit = getDefaultAsLong(USTAT_RETRY_LIMIT);
+    Int64 printPlan = 1;
     if (limit < 1 || centiSecs < 1) // user does not want any retry
```
nonstop-qfchen: Can we use NABoolean here through out?  It is a little bit more efficient than Int64 when pass to a function.

core/sql/ustat/hs_globals.cpp
```diff
@@ -4392,6 +4388,12 @@ Lng32 HSSample::make(NABoolean rowCountIsEstimate, // input
         LM->Log(LM->msg);
       }
 
+ 
+    double sampleRatio = samplePercent / 100;
```
nonstop-qfchen: Just wonder if samplePercent can be 0. If so, we may have a core when dividing by sampleRatio.

core/sql/ustat/hs_globals.cpp
DaveBirdsall: Yes, I wanted to do this. Unfortunately, there are over 100 calls to HSFuncExecQuery, so I chose not to do that now. (It also complicates the merge up into derivative products, as there are even more calls to HSFuncExecQuery there.)

core/sql/ustat/hs_globals.cpp
DaveBirdsall: I tried this, actually. We get "infinity" for the floating point number. That's why I added the "isnormal" test. We also error out with a 9207 error if the sample table is empty.

PR #1854, opened 2019-08-15
[TRAFODION-3318] Changed process management rules for DTM process:

zcorrea:
- DTM is now a 'monitor primitive persistent' process
           o When persistent process retries exceed the persist time window,
             the node is brought down
           o Improves HA behavior in that DTM process death does no longer
             kills all processes in its node
         - Removed TmSync logic and DTM dependency on it
         - Removed SoftNodeDown/Up logic and protocol triggered by DTM process
           abnormal termination
         - This change required enabling monitor AGENT mode in Python installations
         - Moved obsolete test files to monitor/test-legacy directory
         - Removed obsolete code files in monitor/linux and monitor/test directories


zcorrea:
Jenkins, retest

zcorrea:
Jenkins, retest

PR #1855, opened 2019-09-09
[TRAFODION-3325] Enable index access for non-VEG equality join predicates

DaveBirdsall:
The problem is that sometimes join equality predicates are not converted into VEGs, and therefore do not get pushed down to their children. That prevents Scan::addIndexInfo from considering indexes when there are column references in the join predicate. That can result in missed opportunities to use efficient index plans and instead doing full table scans + hash joins.

The fix is to look for equality predicates that didn't get pushed down, and if we find one and it has a child that is covered, we create an IS NOT NULL predicate on that child and push that down. That allows Scan::addIndexInfo to notice that an index might be useful. Later, if the join is converted to a TSJ, the equality predicate will be pushed down and the Optimizer can make use of the index.

An example of a join equality predicate that is not converted to a VEG is a join predicate involving a tuple list.

I added a CQD, COMP_BOOL_194, which if set to 'OFF' turns off this fix. The default is 'ON'.

When unit testing this fix, I noticed a bug in some of the optimizer debugging code that caused a core. I fixed that too (optimizer/opt.cpp).

### DIFF
--- a/core/sql/optimizer/RelExpr.cpp
+++ b/core/sql/optimizer/RelExpr.cpp
@@ -1330,14 +1330,14 @@ void RelExpr::pushdownCoveredExpr(const ValueIdSet & outputExpr,
       // -----------------------------------------------------------------
       ValueIdSet referencedInputs[MAX_REL_ARITY];
       // -----------------------------------------------------------------
-      // Allocate a ValueIdSet to contain the ValueIds of the roots of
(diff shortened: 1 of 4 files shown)


DaveBirdsall:
*** Do not merge ***  Presently analyzing regression failures.

DaveBirdsall:
*** DO NOT MERGE *** I will be closing this pull request and replacing this fix shortly.

PR #1856, opened 2019-09-12
[TRAFODION-3325] Pass down index hints for non-VEG equality predicates

DaveBirdsall:
The problem is that sometimes join equality predicates are not converted into VEGs, and therefore do not get pushed down to their children. That prevents Scan::addIndexInfo from considering indexes when there are column references in the join predicate. That can result in missed opportunities to use efficient index plans and instead doing full table scans + hash joins.

My first attempt at fixing this problem was to generate an IS NOT NULL predicate for the children of such equality predicates, then attempt to push those down. This didn't work well: If the column was not nullable, logic in HBaseTable.java would try to optimize it out resulting in a null filter which resulted in an exception. Too, it also resulted in redundant predicates of the form \<constant\> = \<constant\> in explain plans for some queries.

So, this fix takes a different approach. We add a member to RelExpr, possibleIndexColumns_, to contain the covered children of non-VEG equality predicates. We push down such children. At Scan::addIndexInfo, we consider such ValueIds when computing the set of indexes to try. That is, this fix creates an alternative mechanism to push down information to Scan::addIndexInfo, without generating possibly redundant predicates.

An example of a join equality predicate that is not converted to a VEG is a join predicate involving a tuple list.

I added a CQD, COMP_BOOL_194, which if set to 'OFF' turns off this fix. The default is 'ON'.

When unit testing this fix, I noticed a bug in some of the optimizer debugging code that caused a core. I fixed that too (optimizer/opt.cpp).

### DIFF
--- a/core/sql/optimizer/RelExpr.cpp
+++ b/core/sql/optimizer/RelExpr.cpp
@@ -811,6 +811,9 @@ RelExpr * RelExpr::copyTopNode(RelExpr *derivedNode,CollHeap* outHeap)
   if (selection_ != NULL)
     result->selection_ = selection_->copyTree(outHeap)->castToItemExpr();
 
+  // copy possible index column hints for non-VEG equality predicates
+  result->possibleIndexColumns_ = possibleIndexColumns_;
+
   // -- Triggers
   // Copy the inlining information and access sets.
   result->getInliningInfo().merge(&getInliningInfo());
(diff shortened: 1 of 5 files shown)


core/sql/optimizer/RelExpr.cpp
```diff
@@ -1434,6 +1449,44 @@ void RelExpr::pushdownCoveredExpr(const ValueIdSet & outputExpr,
...
+            {
+              for (iter = firstChild; iter < lastChild; iter++)
+                {
+                  for (CollIndex i = 0; i < ie->getArity(); i++)
+                    {
+                      // If the child is covered, it might contain a column reference
+                      // that is useful for index selection. Push that down.
+                      ItemExpr * ieChildi = ie->child(i);
+                      if ((CmpCommon::getDefault(COMP_BOOL_194) == DF_ON) &&
+                          (coveredSubExprNotUsed[iter].contains(ieChildi->getValueId())))
```
nonstop-qfchen: It may be more efficient to move the check on CB_194 out of the topmost FOR loop.

core/sql/optimizer/RelExpr.cpp
```diff
@@ -1434,6 +1449,44 @@ void RelExpr::pushdownCoveredExpr(const ValueIdSet & outputExpr,
       computeValuesReqdForPredicates(predicatesOnParent,
                                      exprToEvalOnParent);
```
nonstop-qfchen: I wonder if we subtract those push-down non-VEG equality predicates from the predicates to be evaluated on the parent.

core/sql/optimizer/RelExpr.cpp
DaveBirdsall: Will do. Thanks.

core/sql/optimizer/RelExpr.cpp
DaveBirdsall: We don't today, and I have not changed this. As far as I can tell, this does not cause any issues.

PR #1857, opened 2019-10-01
[TRAFODION-3327] LOB support in T2 driver

selvaganesang:


core/conn/jdbc_type2/native/SrvrOthers.cpp
```diff
@@ -1529,3 +1529,127 @@ odbc_SQLSvc_CloseUsingLabel_sme_(
...
+
+    if ((QryLobExtractSrvrStmt = createSrvrStmt(dialogueId, "MXOSRVR_EXTRACTLOB", &sqlcode, 
+			NULL, 0, 0, TYPE_UNKNOWN, false,false)) == NULL) 
+    {
+       exception_->exception_nr = odbc_SQLSvc_ExtractLob_SQLInvalidhandle_exn_;
+       return;
+    }
+    switch (extractLobAPI) {
+    case 0:
+        snprintf(lobExtractQuery, sizeof(lobExtractQuery), "EXTRACT LOBLENGTH(LOB'%s') LOCATION %llu", lobHandle, (Int64)&lobLength);
```
DaveBirdsall: Is there a maximum size on strlen(lobHandle)? If not, then long lobHandle values (on the order of 980 characters or so) will cause truncation of lobExtractQuery, resulting in a SQL error. One way to work around this is to use a dynamic array size when declaring the lobExtractQuery array, so we can guarantee enough space.

selvaganesang:
Thanks @DaveBirdsall . Yes. Lob support is still not functional.  I will be making another round of changes with more testing.

PR #1858, opened 2019-10-01
[TRAFODION-3328] Code cleanup in NADefaults

selvaganesang:
Avoid reading VERSIONS and DEFAULTS table after it is done once in a process.

This required following cleanup
1. cqd MODE_SEABASE is removed.
2. Removed the support to read CQD from a file.
3. NADefaults constructor now reads the defaults table once and
   saves off the relevant information such as currentDefaults, currentTokens,
   provenances and currentFloats to be reused by the other NADefaults objects.
4. T2 driver doesn't use the defaultContext to compile any query. Hence, it is not
   possible to use the heap from CmpContext to save off the defaults value.
   The code has been changed the heap from CliGlobals to save the current defaults.
5. Fixed code to use static literals as it is from the defaultsDefaults to
   avoid invalid memory in the saved defaults
6. Changes to optimize initialization of currentTokens_ in NADefaults
7. CQD * RESET RESET
   CQD * RESET
   Now resets to the defaults when CQD * RESET RESET was issued.
8. Also optimized initCurrentsFromDefaultsTable further.


DaveBirdsall:
jenkins, retest

core/sql/generator/GenRelUpdate.cpp
```diff
@@ -779,6 +779,7 @@ short HbaseDelete::codeGen(Generator * generator)
 
   NABoolean isAlignedFormat = getTableDesc()->getNATable()->isAlignedFormat(getIndexDesc());
   NABoolean isHbaseMapFormat = getTableDesc()->getNATable()->isHbaseMapTable();
+  NABoolean isTrafMDTable = getTableDesc()->getNATable()->isSeabaseMDTable();
```
DaveBirdsall: Not sure why this change is here. We don't seem to use isTrafMDTable anywhere?

selvaganesang:
@DaveBirdsall It looks like this change exposed some timing issue with mxosrvr code. I am still attempting to figure out this issue

core/sql/generator/GenRelUpdate.cpp
```diff
@@ -2449,6 +2450,7 @@ short HbaseInsert::codeGen(Generator *generator)
   }
 
   const NATable *naTable = getTableDesc()->getNATable();
+  NABoolean isTrafMDTable = getTableDesc()->getNATable()->isSeabaseMDTable();
```
DaveBirdsall: Same observation here

core/sql/sqlcomp/nadefaults.cpp
```diff
@@ -3629,11 +3622,12 @@ void NADefaults::initCurrentDefaultsWithDefaultDefaults()
   }
 // End: Temporary workaround for SQL build regressions to pass
 
-  // Cache all the default keywords up front,
   // leaving other non-keyword token to be cached on demand.
```
DaveBirdsall: Should this comment line (line 3625 in the new file) be deleted also? It doesn't make sense by itself.

selvaganesang:
jenkins, retest

PR #1859, opened 2019-10-02
[TRAFODION-3329] Code cleanup in Type 2 driver

selvaganesang:
Remove unused static module related code from T2 driver.

Cleanup of T2 code exposed an issue with cleanup of volatile schema
at the time of drop session and t2 disconnect.


core/sql/sqlmxevents/logmxevent_traf.h
```diff
@@ -189,4 +190,7 @@ class SQLMXLoggingArea
 																									 
 };
 
+void logAnMXEventForError( ComCondition & condition, SQLMXLoggingArea::ExperienceLevel emsEventEL);
```
DaveBirdsall: Is there a function body somewhere that goes with this? I don't see it in this pull request.

selvaganesang: The implementation was already available at cli/CliExtern.cpp. The corresponding include file will be sqlcli.h that contains the interface to be used by the clients to Trafodion SQL engine. Hence, I added this declaration in logmxevent_traf.h

PR #1863, opened 2019-12-02
TRAFODION-3332 Component log filenames that are generated contain node id instead of node name

hegdean:


hegdean:
jenkins retest

PR #1867, opened 2020-01-29
[TRAFODION-3328] Code cleanup in Trafodion

selvaganesang:
Retired the HDFS text/sequence file scan using lib hdfs.

Fix for udr/TEST102 failure

### DIFF
--- a/core/sql/sqlcomp/nadefaults.cpp
+++ b/core/sql/sqlcomp/nadefaults.cpp
@@ -3623,6 +3623,7 @@ void NADefaults::initCurrentDefaultsWithDefaultDefaults()
   }
 // End: Temporary workaround for SQL build regressions to pass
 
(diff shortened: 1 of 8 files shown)


core/sql/sqlcomp/nadefaults.cpp
```diff
@@ -5092,12 +5093,8 @@ enum DefaultConstants NADefaults::validateAndInsert(const char *attrName,
           else
             Reset_SqlParser_Flags(IN_MODE_SPECIAL_4);
 	}
-      break;
-
-
-	      insert(UPD_SAVEPOINT_ON_ERROR, "OFF", errOrWarn);
+        break;
```
DaveBirdsall: Nice catch. Must be from a past merge error?

PR #1870, opened 2020-02-18
[TRAFODION-3335] DDL takes a longer time when there are many snapshots

selvaganesang:
Avoided use of the HBaseAdmin.listSnapshots() that causes unnecessary overhead whereever it is possible

HBaseAdmin.listSnapshots will continue to be called when CQD TRAF_TABLE_SNAPSHOT_SCAN is set to 'LATEST'.

Removed latest snapshot name being part of table descriptor and NATable
  Added code to get the latest snapshot at the time of code generation
  It was done unconditionally at times and this was causing performance problems during DDL
     when there are many snapshots in the cluster.


selvaganesang:
@sureshsubbiah Can you please review this PR

